PR #32, opened 2016-02-10
Add function mapping for timestamp operators and sum aggregate function

omegaga:
Add function mapping in timestamp operators, so that peloton type system can recognize these functions from Postgres's plan tree.

Check `src/postgres/backend/utils/fmgrtab.cpp` to see complete function mapping.


jarulraj:
Merged it :+1:

PR #57, opened 2016-03-12
add support for CASE

omegaga:
Handle two forms of CASE:
1. ```CASE expression
    WHEN value THEN result
    [WHEN ...]
    [ELSE result]
END```
2. ```CASE WHEN condition THEN result
     [WHEN ...]
     [ELSE result]
END```
and add corresponding regression test.


apavlo:
Do we have test cases for this?

omegaga:
Still writing unit tests and asking Joy for some specific problem.

PR #60, opened 2016-03-13
add support for four operators

normanyahq:
1. CASE_EXPR
2. UNARY_MINUS
3. COALESCE
4. NULLIF

A SQL test script covering above four is also included.


normanyahq:
I've just installed the latest version of protobuf.

It seems that NewCallBack should be called by:
google::protobuf::**internal::**NewCallback(&Call); instead of  

what is written in master branch now:
google::protobuf::NewCallback(&Call);

anyway, to make it compile in current environment. I commit with the second version.

PR #61, opened 2016-03-13
Networking (tcp socket + libevent + serialization + rpc)

Michael-Tieying-Zhang:
Network stuff for peloton nodes. The nodes use raw tcp socket and libevent to send and recv data. TCP stream control is also supported. I add header for each message to represent the length and when the receiver recvs the data, it separates the messages using the header. Livevent is used for asysc calback. I have merged with the latest master. This PR also contains:
--new thread pool
--supporting network with thread pool
--adding protobuf controller
--unit test for thread pool and network

I also did simple performance test:

client sends 5000rpc with x threads, server can process n rpc/s with x threads:
client:3 threads server: 3 threads
speed:-----------------------------------------------rpc: 13275.9/s----------
client:5 threads server:5 threads
TRACE ------------------------------------------------rpc: 23302.9/s----------
client: 10 threads server:10 threads
----------------------------------------------------------rpc: 33142.8/s----------
client: 20 threads server:20 threads
----------------------------------------------------------rpc: 28575.8/s----------


jarulraj:
Code review comments : Please don't use "using std::xyz". Before sending a PR, compile without --enable-debug flag. I had to remove some unused variables because they were only used within an assert statement.

PR #63, opened 2016-03-15
replace stx-btree with cpp-btree.

yingjunwu:


yingjunwu:
a bug is detected, will fix it before commit.

PR #68, opened 2016-03-21
re-implement mvcc.

yingjunwu:
The design follows that proposed by Hekaton team. It is optimistic, and does not allow cascading abort.


yingjunwu:
TODO: address phantom problem. update wiki.

jarulraj:
Yingjun -- This looks good. But it is a major refactoring.
- Can you add more comments in the txn manager and the executors ? 
- Also, please delete any unused code. 
- Uncomment the logging stuff in the executors and move them to an appropriate place within DExecute.

yingjunwu:
Sure. We should no longer use Hyrise's design. It may raise costly overhead in index operations, e.g., insertion and deletion. I will also rewrite the wiki.

PR #80, opened 2016-03-24
Fix memory leak in case expression

omegaga:


omegaga:
I noticed that due to this memory leak, my previous commit failed `make check` on local machines. More specifically, it failed to pass valgrind check. However, it passed tests on Travis CI. Does Travis CI check memory leak?

yingjunwu:
please remember to set a pointer to nullptr once you've deleted the object it points to.

omegaga:
I think it's safe not to set to nullptr in the destructor, since the pointer itself is discarded anyway.

omegaga:
This stackoverflow post discussed this problem in detail: http://stackoverflow.com/questions/3060006/is-it-worth-setting-pointers-to-null-in-a-destructor

yingjunwu:
http://blogs.microsoft.com/cybertrust/2012/04/24/guarding-against-re-use-of-stale-object-references/
this is the rule that makes the code safe and the rule that we should always follow. or use smart pointer whenever possible.

normanyahq:
ahh...My mistake, sorry. I should have used shared_ptr

omegaga:
Hmm, you're right. In this expression, if `default_result` shares the same object with a `clause`, then if the object is not set as nullptr in `clause`'s destructor, it will be double freed since `CaseExpression`'s destructor will think the object has not been deleted.

omegaga:
Anyway, using shared_ptr is always a better idea. However, it looks like the CaseExpr transformation logic is rollbacked or deleted for some reasons. I will do the basic set-as-nullptr for now.

Issue #21, opened 2016-02-08
Isolation Levels


jarulraj:
We use a multi-version concurrency control protocol. Right now, we only support the default snapshot isolation level.

1. We need to identify anomalies and avoid them in our protocol.

2. We need to support higher isolation levels, like the *serializable* (isolation level)[https://en.wikipedia.org/wiki/Isolation_%28database_systems%29#Isolation_levels].

3. We need a testing suite for validating different isolation levels.

jarulraj:
cmu-db/peloton/pull/82 adds basic support for testing isolation levels.

PR #84, opened 2016-03-24
Refactor transaction test

ranxian:
* Use {Insert, Delete, Update}Executor to execute transaction operations.
* Add test for transaction abort
* Refactor transaction test helpers to `transaction_tests_util.cpp`

#21


yingjunwu:
have bug. we should talk before commit.

PR #89, opened 2016-03-26
fix all kinds of bugs in current mvocc implementation. support bw-tree. refactor. add more test cases.

yingjunwu:
@MattPerron @eric-haibin-lin you may like to have a look at it.


ranxian:
@yingjunwu CI has failed?

yingjunwu:
@ranxian bw-tree has bugs. I'll remove it.

PR #95, opened 2016-03-28
Dev practices

jarulraj:


jarulraj:
This PR adds support for additional code validators. One such validator prevents people from using `std::cout` and `printf`. Instead, they should only use the `LOG_INFO` and similar methods.

PR #96, opened 2016-03-28
fix potential bugs in current mvocc and mv2pl implementation. refactor. add more test cases for concurrency control.

yingjunwu:


src/backend/executor/index_scan_executor.cpp
```diff
@@ -179,13 +179,21 @@ bool IndexScanExecutor::ExecIndexLookup() {
...
+            return res;
+          }
         } else {
           expression::ContainerTuple<storage::TileGroup> tuple(tile_group.get(), tuple_id);
           auto eval = predicate_->Evaluate(&tuple, nullptr, executor_context_).IsTrue();
           if (eval == true) {
             visible_tuples[tile_group_id].push_back(tuple_id);
-            transaction_manager.PerformRead(tile_group_id, tuple_id);
+            auto res = transaction_manager.PerformRead(tile_group_id, tuple_id);
+            if(!res){
```
ranxian: @jessesleeping Is returning false enough? Do we need to set a result tile?

yingjunwu: you mean SetOutput? I think it is unnecessary to do that in this case, as the transaction fails, and no output should be returned.

PR #100, opened 2016-03-30
make the logging work.

yingjunwu:
the logging team has finished implementing the basic logging systems.
@MattPerron @eric-haibin-lin @abhishekjoshi2 You may like to describe how to make the logging system work (how to configure it?) in this pr.


MattPerron:
We probably want to wait until this is more thoroughly tested before merging into cmu-db/master. There are still some parts of the code that I do not have high confidence in. Can we have until early next week to complete this?

jarulraj:
Sure Matt. Here are some suggestions : 

- Avoid using free(), instead use smart pointers
- Add more unit tests 
- Please update the wiki on the logging section based on the updated design. You can remove figures that are no longer valid
- Check if recovery works in the entire system. Like use pg_ctl to stop and restart the system and see if all the data is recovered. For this, you do not need to write unit tests. Test it using psql.

jarulraj:
Yingjun: 
- Please update the concurrency control wiki page with the new design. 
- We need to add a paragraph on top of each txn manager to describe its design choices in the source code.
- Integrate logging with the transaction managers.

yingjunwu:
ok. let me fix the compilation error before merging the pr.

eric-haibin-lin:
@jarulraj  Thanks for the suggestions. We'll work on these.
We don't have the permission to edit the wiki - how do we make an update?

jarulraj:
@eric-haibin-lin I have sent an invite to join the Peloton team.

yingjunwu:
@eric-haibin-lin @MattPerron @abhishekjoshi2 Hi, could you please help figure out why the test fails?

jarulraj:
It is failing due to a validator check. Somewhere in the code, `free` is being used.

MattPerron:
We found where we are using it and @eric-haibin-lin is working in getting it fixed now.

Issue #99, opened 2016-03-30
CleanPlan is not working properly


shlee0605:
```
create table test1(a int);
create table test2(a int);
insert into test1 values (1);
insert into test2 values (2);

select test1.a + 1 from test1,test2;
```

If this sequence of queries are executed, the peloton fails with segmentation fault.
Our team looked into the stack trace and found that the following line causes the error.
If we comment out the "delete root", it no longer causes segmentation but it may possibly lead
the memory leak. 

https://github.com/cmu-db/peloton/blob/master/src/backend/bridge/dml/mapper/mapper.cpp#L195

I include the stack trace.
```
#0  0x0000000000000085 in ?? ()
#1  0x00007ffff6d7d776 in std::default_delete<peloton::catalog::Schema const>::operator() (this=0x7fffdc0bef10, __ptr=0x7fffdc0bee60) at /usr/include/c++/4.8/bits/unique_ptr.h:67
#2  0x00007ffff6d7d511 in std::unique_ptr<peloton::catalog::Schema const, std::default_delete<peloton::catalog::Schema const> >::~unique_ptr (this=0x7fffdc0bef10, __in_chrg=<optimized out>) at /usr/include/c++/4.8/bits/unique_ptr.h:184
#3  0x00007ffff6d7d816 in peloton::planner::ProjectionPlan::~ProjectionPlan (this=0x7fffdc0beee0, __in_chrg=<optimized out>) at ../../src/backend/planner/projection_plan.h:29
#4  0x00007ffff6d7d862 in peloton::planner::ProjectionPlan::~ProjectionPlan (this=0x7fffdc0beee0, __in_chrg=<optimized out>) at ../../src/backend/planner/projection_plan.h:29
#5  0x00007ffff6d79b65 in peloton::bridge::PlanTransformer::CleanPlan (root=0x7fffdc0beee0) at ../../src/backend/bridge/dml/mapper/mapper.cpp:194
#6  0x00007ffff6d7b7eb in std::_Sp_counted_deleter<peloton::planner::AbstractPlan const*, void (*)(peloton::planner::AbstractPlan const*), std::allocator<int>, (__gnu_cxx::_Lock_policy)2>::_M_dispose (this=0x7fffdc0d3870) at /usr/include/c++/4.8/bits/shared_ptr_base.h:347
[...]

PR #102, opened 2016-04-01
Fix bug in duplicate deletion due to usage of raw pointers

omegaga:
Fix bug in first statement in #99


pmenon:
So I ran into this problem a while back and I don't think this is the right approach.

1. Why are the mappers not returning unique pointers?  What else has a handle to these plan nodes to require them to be shared?  Only the root of the plan tree is shared between the cache and the executing thread.
2. We should be explicit when using unique pointers.  If a class takes ownership of an object, the constructor must accept an r-value reference to the object (std::unique_ptr<Object> &&r) so that the caller **has** to issue a std::move(r).  This makes it very clear that the caller is relinquishing ownership to the class.

omegaga:
That makes sense. The plan nodes are shared in rare cases now but they shouldn't. Actually one of these bugs is that our code share same schema in multiple plan nodes, which can be avoided by using unique_ptr. I will refactor to use unique_ptr.

omegaga:
I will finish the refactoring and create a new pull request.

PR #104, opened 2016-04-02
finished implementing different kinds of concurrency-control protocols.

yingjunwu:
currently, we support multiple kinds of concurrency-control protocols:

1. pessimistic (see pessimistic_txn_manager.h/.cpp);
2. optimistic (see optimistic_txn_manager.h/.cpp);
3. pessimistic + eager write (see eager_write_txn_manager.h/.cpp);
4. optimistic + speculative read (see speculative_read_txn_manager.h/.cpp);
5. timestamp ordering (see ts_order_txn_manager.h/.cpp);
6. serializable snapshot isolation (see ssi_txn_manager.h/.cpp).

There may still contain bugs in the current implementation, but we will not make major changes to the program skeleton in recent days.

@jarulraj let me add more test cases before merging this pr.


jessesleeping:
Add a conditional insert API in the index.h base class. This API is design for insert of primary index in our MVCC implementation.

virtual bool ConditionalInsertEntry(const storage::Tuple *key,
                                      const ItemPointer &location,
                                      std::function<bool(const storage::Tuple *, const ItemPointer &)> predicate) = 0;

What it should do is:
1) Retrieval all key-value pairs that have the given key
2) Apply the predicate on these key-value pairs. If any of these pairs satisfy the predicate, return false.
3) Else, insert the given key-value pair into the index and return true.
4) step 2) and step 3) should be atomic.

And the predicate should be, under our MVCC scenario, a dirty read visibility check.

By doing this, we can handle tuple insertion/update in primary/unique indexes.

PR #107, opened 2016-04-03
Add copy functions to expressions.

amaliujia:
In order to support intra-query parallelism, peloton plan tree has to be copied. During copy, the difficult but necessary part is to copy expression classes. 

In this commit, AbstractExpression is added a Copy interface, like following:

`  virtual AbstractExpression *Copy() const = 0;`

Each child class of AbstractExpression  implements this interface, and do recursively copy in a clean manner.

In expression_test, the old tests are extended, to test if copied expressions are correct.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -211,3 +211,5 @@ local.properties
 
 # Code Recommenders
 .recommenders/
+
+CMakeLists.txt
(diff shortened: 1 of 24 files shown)


jarulraj:
Nice job.

PR #110, opened 2016-04-04
Add copy functions to planner.

amaliujia:
In this PR I added copy functions to all planners.

This merge built on the top of RP #107 

Basically, I add following interface to AbstractPlan:
`virtual AbstractPlan *Copy() const = 0;`

And I added copy functions to all plans.

Two things need to mention:

1. In the future, all new plans are forced to implement Copy function, but it is ok since the guy writing that plan should know how to copy that, and since I have provided all necessary copy helper functions in Expressions, Schema, ProjectInfo classes.

2. I didn't add tests to test this part of code. The reason is directly doing unit tests is trivial and not pratical. Since plans are simple, and I will build intra-query parallelism on top of copies, I will cover these functions in my tests. If I catch any bugs, I will change them.


pmenon:
This is good work.  However, the interface should be returning `unique_ptr` instead of raw pointers.  It might be too late to revert this merge, so I'll create a bug for @amaliujia to fix.  Additionally, since this code is specific to the parallelism team, why was it merged into master?  It may conflict with some of the other teams are doing ...

jarulraj:
Yeah, I agree that we should be using unique pointers here as well. I decided to merge this in because I think that these functions might be needed by other teams as well.

amaliujia:
I am wondering why should use smart pointer here? From my point of view, it is the caller who take responsible of managing the return pointer. 

Do you think use smart pointer is safer?

pmenon:
By returning `unique_ptr`, the interface is clear that Copy() does not retain a reference to what is returned.  While comments are great, having the compiler verify and enforce your assumption is even better.

apavlo:
@pmenon is correct.

@amaliujia can you fix this and send a new pull request? 

Nice work otherwise.

amaliujia:
I will fix this one.

Issue #116, opened 2016-04-06
Refactor CASE expression type


pmenon:
Right now, CASE expressions are handled between `CaseExpression` and `OperatorCaseWhenExpression`, where the former relies on the latter throwing exceptions to determine whether when clauses are false.  This is unnecessarily complex.

One solution is to have a single `CaseExpression` class that accepts a list of pairs of expressions as clauses.  The first component of the pair is the "when" condition (either boolean or a value whose type is comparable against the initial expression if one is provided), and the second component is the "then" expression.  An initial idea is something like:

```c++
typedef std::pair<AbstractExpression*, AbstractExpression*> WhenClause;

class CaseExpression {
 public:
  CaseExpression(AbstractExpression* initial, 
                             std::vector<WhenClause> &clauses,
                             AbstractExpression* default);
};
```

Evaluation of a case expression is as simple as iterating over the clauses, evaluating the first component of the pair, and if true, returning the evaluation of the second component of the pair.

apavlo:
@pmenon's assessment and suggested fix is correct.

normanyahq:
Got it. I will take a look.

PR #119, opened 2016-04-07
Refactored Case

normanyahq:
Mentioned in #116 

Also, refactored nullif, coalesce, and unary_minus.


src/backend/expression/case_expression.h
```diff
@@ -9,69 +9,74 @@
...
-
+#include <utility>
 #include "backend/expression/abstract_expression.h"
 #include "backend/expression/expression_util.h"
 
 namespace peloton {
 namespace expression {
 
+// the first part is condition, and the second part is expression
+typedef std::pair<AbstractExpression *, AbstractExpression *> WhenClause;
```
pmenon: Can you move this `typedef` into `CaseExpression`

src/backend/expression/case_expression.h
```diff
@@ -9,69 +9,74 @@
...
+
 class CaseExpression : public AbstractExpression {
  public:
-  CaseExpression(ValueType vt, const std::vector<AbstractExpression *> &clauses,
-                 AbstractExpression *default_result)
+  CaseExpression(ValueType vt, const std::vector<WhenClause *> &clauses,
+                 WhenClause *default_clause)
```
pmenon: With the modification to the typedef above, use `std::move` to move the vector parameter's contents.  Also pass in an r-value `unique_ptr&&` for the default clause.

src/backend/expression/case_expression.h
```diff
@@ -9,69 +9,74 @@
...
       : AbstractExpression(EXPRESSION_TYPE_OPERATOR_CASE_EXPR),
         clauses(clauses),
-        default_result(default_result),
+        default_clause(default_clause),
         case_type(vt) {}
 
   ~CaseExpression() {
```
pmenon: You won't need a destructor with the changes above.

pmenon:
Thanks for the quick turnaround! 

It looks good overall, but I've added a few comments to address.  Try to use `unique_ptr` whenever you see a `new` anywhere in the code.  Apply this technique to any code you're modifying (in this case, CASE, COALESCE and NULLIF expressions too).

Try to make the above changes, and fix the build issue that's come up

normanyahq:
OK... I forgot to modify unit test.
btw, I can't make check on my machine with gcc-5.2.1

normanyahq:
Done. I will add unit tests for nullif, coalesce later.

src/backend/expression/case_expression.h
```diff
@@ -18,60 +20,55 @@ namespace expression {
 
 class CaseExpression : public AbstractExpression {
  public:
-  CaseExpression(ValueType vt, const std::vector<AbstractExpression *> &clauses,
-                 AbstractExpression *default_result)
+  // the first part is condition, and the second part is expression
+  typedef std::unique_ptr<AbstractExpression> AbstractExprPtr;
+  typedef std::pair<AbstractExprPtr, std::unique_ptr<AbstractExpression>>
+      WhenClause;
```
pmenon: Be consistent with your typedef usage.

src/backend/expression/case_expression.h
```diff
@@ -18,60 +20,55 @@ namespace expression {
...
-      delete clause;
-      clause = nullptr;
-    }
-    if (default_result != nullptr) {
-      delete default_result;
-      default_result = nullptr;
+        default_clause(std::move(default_clause)),
+        case_type(vt) {
+    for (auto &clause : t_clauses) {
+      clauses.push_back(std::move(clause));
```
pmenon: You should be able to construct `clauses` as `std::move(t_clauses)`

pmenon:
I've added a few minor edits to make.  Also, since we no longer need `OperatorCaseWhenExpression`, can you remove that?

src/backend/expression/case_expression.h
```diff
@@ -9,7 +9,9 @@
 // Copyright (c) 2015-16, Carnegie Mellon University Database Group
 //
 //===----------------------------------------------------------------------===//
-
```
pmenon: Missing an include guard here.  Just pop in `#pragma once`

src/backend/expression/case_expression.h
```diff
@@ -18,60 +20,55 @@ namespace expression {
...
   }
 
  private:
   // Case expression clauses
-  std::vector<AbstractExpression *> clauses;
+  std::vector<WhenClausePtr> clauses;
 
   // Fallback case result expression
-  AbstractExpression *default_result;
+  WhenClausePtr default_clause;
```
pmenon: Use the member variable naming convention by putting an underscore as a suffix. Do `clauses_` as an example.

pmenon:
I've added a few minor edits to make. Also, since we no longer need `OperatorCaseWhenExpression`, can you remove that?

src/backend/expression/case_expression.h
```diff
@@ -9,6 +9,7 @@
 // Copyright (c) 2015-16, Carnegie Mellon University Database Group
 //
 //===----------------------------------------------------------------------===//
+#pragma once
```
pmenon: Just add some spacing between the comment, the guard and the first include.

pmenon:
This is awesome. Just one final formatting comment and we can merge this in.

normanyahq:
Done. Thank you for all the suggestions. I learned a lot.

PR #123, opened 2016-04-08
add boost thread pool

yingjunwu:
Currently, every time a query session is created, a thread is spawned and detached. however, this is quite inefficient when processing massive number of transactions. We may use a thread pool that can reuse threads efficiently. Due to the lack of efficient thread pool, I added boost thread pool in this pr for future usage.

### DIFF
--- a/src/backend/common/worker_thread_pool.h
+++ b/src/backend/common/worker_thread_pool.h
@@ -18,42 +18,50 @@
 #include <boost/function.hpp>
 
 namespace peloton {
-  // a wrapper for boost worker thread pool.
-  class WorkerThreadPool{
-  public:
-    WorkerThreadPool() : pool_size_(0), work_(io_service_){}
-    WorkerThreadPool(const size_t &pool_size) : pool_size_(pool_size), work_(io_service_){}
+// a wrapper for boost worker thread pool.
+class WorkerThreadPool {
+ public:
+  WorkerThreadPool() : pool_size_(0), work_(io_service_) {}
+  WorkerThreadPool(const size_t &pool_size)
+      : pool_size_(pool_size), work_(io_service_) {}
 
-    ~WorkerThreadPool(){
-      io_service_.stop();
-      thread_pool_.join_all();
+  ~WorkerThreadPool() {
+    io_service_.stop();
+    thread_pool_.join_all();
+  }
+
+  void InstantiatePool(const size_t &pool_size) {
+    pool_size_ = pool_size;
+    assert(pool_size_ != 0);
+    for (size_t i = 0; i < pool_size_; ++i) {
+      // add thread to thread pool.
+      thread_pool_.create_thread(
+          boost::bind(&boost::asio::io_service::run, &io_service_));
     }
+  }
 
-    void InitiatePool(const size_t &pool_size){
-      pool_size_ = pool_size;
-      assert(pool_size_ != 0);
-      for (size_t i = 0; i < pool_size_; ++i){
-        thread_pool_.create_thread(boost::bind(&boost::asio::io_service::run, &io_service_));
-      }
-    }
+  // submit task to thread pool.
+  // it accepts a function and a set of function parameters as parameters.
+  template <typename FunctionType, typename... ParamTypes>
+  void SubmitTask(FunctionType &&func, const ParamTypes &&... params) {
+    // add task to thread pool.
+    io_service_.post(std::bind(func, params...));
+  }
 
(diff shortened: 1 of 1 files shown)


domino-succ:
https://github.com/cmu-db/peloton/blob/master/src/backend/common/thread_manager.h

domino-succ:
There are two kinds of thread-pool in the current code can be used: std:thread and pthread:https://github.com/cmu-db/peloton/blob/master/src/backend/common/thread_manager.h

yingjunwu:
@domino-succ I am aware of this thread manager. But boost thread pool is highly optimized, especially compared with self-implemented thread pools.

domino-succ:
Optimized for ? Boost is a wrapper around pthreads, similar to the std::thread.

apavlo:
We should have a single-thread pool for workers.

Make sure you guys coordinate with @amaliujia on this. They have been working on multi-threaded queries.

amaliujia:
I don't think multi-threaded queries have high performance requirements on thread pool, since single thread has heavy workload so overhead of context switch or thread scheduling should not matter that much.

But it is hard to tell. I can test different thread pools at the end when parallel implementation is fixed and use the data to see if thread pools matter.

yingjunwu:
Thread pool is not a bottleneck for OLAP processing, and it is probably not a bottleneck for processing OLTP benchmarks with only a small number of connections (i.e., number of terminals in oltpbench). However, when processing workloads with hundreds of connections, relying on OS to schedule hundreds of threads may not be a good idea, and it may also bring additional overhead to transaction processing, garbage collection, as well as database logging. That's why we may need to rewrite the multithreading model in Peloton.

pmenon:
My two cents: use Intel's TBB. It outperforms Boost and pulls in fewer
dependencies IIRC.  The less we need Boost, the better.

apavlo:
Doesn't TBB require us to build a dependency graph for each query plan? @jarulraj  originally wanted to use TBB over a year ago and I talked him out of it. I don't remember the details of our conversation but I remember convincing myself that it was a bad (beyond that it was GPL).

We don't need to come up with the perfect solution right now. @yingjunwu's just needs a simple worker pool with a configurable number of threads.

yingjunwu:
Actually, our ultimate goal (if possible) is to schedule concurrent threads at database level (i.e., use thread pool) instead of at OS level. Current implementation (https://github.com/cmu-db/peloton/blob/master/src/backend/common/thread_manager.h) adopts conditional variables and mutex to synchronize threads. However, such synchronization primitives are very expensive when processing OLTP transactions with dozens of cores. While there are many open-source thread pool implementations, after discussed with @jarulraj, we think that boost thread pool might be a good choice. In fact, we have also implemented our own thread pool with advanced synchronization mechanisms applied, but at current stage, we cannot guarantee the efficiency of our implementation, especially compared with industrial-level libraries, such as boost.

btw, please consider removing mutex if possible, as the incurred context switch is extremely costly when processing short-running OLTP transactions.

TBB is a good choice, but I am not sure whether we can use it (any license issue??). Using TBB can make our life much easier as it supports all kinds of concurrent data structures.

eric-haibin-lin:
@MattPerron you may want to take a look

src/backend/common/worker_thread_pool.h
```diff
@@ -0,0 +1,59 @@
...
+  private:
+    WorkerThreadPool(const WorkerThreadPool &);
+    WorkerThreadPool& operator=(const WorkerThreadPool &);
+
+  public:
+    
+  private:
+    size_t pool_size_;
+    boost::thread_group thread_pool_;
+    boost::asio::io_service io_service_;
```
jarulraj: Can you add more info on boost stuff ?

src/backend/common/worker_thread_pool.h
```diff
@@ -0,0 +1,59 @@
...
+  public:
+    WorkerThreadPool() : pool_size_(0), work_(io_service_){}
+    WorkerThreadPool(const size_t &pool_size) : pool_size_(pool_size), work_(io_service_){}
+
+    ~WorkerThreadPool(){
+      io_service_.stop();
+      thread_pool_.join_all();
+    }
+
+    void InitiatePool(const size_t &pool_size){
```
jarulraj: `InstantiatePool`

yingjunwu:
@jarulraj I have added comments on thread pool implementation.

Let me know if further changes are required. Thanks.

apavlo:
This is an important decision. Let's discuss it at Tuesday's meeting.

PR #130, opened 2016-04-09
Sending query plan between peloton nodes

Michael-Tieying-Zhang:
This PR mainly provide the function of sending query plan between peloton nodes. The serialization/deserialization for Postgres's nested structure takes a lot of time. The details include:

1.AbstractPlan and SeqScanPlan serialization/deserialization
2.Expression serialization/deserialization
3.Value serialization/deserialization
4.Postgres TupleDesc serialization/deserialization
5.TupleResult serialization/deserialization
6.Refacor protobuf related code
7.Unit test


src/backend/bridge/dml/executor/plan_executor.h
```diff
@@ -38,10 +38,20 @@ class PlanExecutor {
   static void PrintPlan(const planner::AbstractPlan *plan,
                         std::string prefix = "");
 
+/* TODO: Delete this mothod
```
jarulraj: Delete this method.

src/backend/bridge/dml/executor/plan_executor.h
```diff
@@ -38,10 +38,20 @@ class PlanExecutor {
...
 
+  /*
+   * Use std::vector<Value> as params to make it more elegant for networking
+   * Before ExecutePlan, a node first receives value list, so we should pass
+   * value list directly rather than passing Postgres's ParamListInfo
+   */
+  static peloton_status ExecutePlan(const planner::AbstractPlan *plan,
+                                    const std::vector<Value> &params,
+                                    TupleDesc m_tuple_desc);
  private:
```
jarulraj: Remove `private`

src/backend/common/serializer.h
```diff
@@ -512,7 +512,7 @@ class CopySerializeOutput : public SerializeOutput {
  public:
   // Start with something sizeable so we avoid a ton of initial
   // allocations.
-  static const int INITIAL_SIZE = 8388608;
+  static const int INITIAL_SIZE = 1024*4; // the previous value is 8388608 which is too big for network
```
jarulraj: Add a note saying 4 KB or something like that.

jarulraj: You could add another constructor that takes in the `initial_size` as a parameter and use it in the networking stuff, and leave the default `8388608` as it is.

src/backend/executor/index_scan_executor.cpp
```diff
@@ -130,6 +130,22 @@ bool IndexScanExecutor::DExecute() {
 bool IndexScanExecutor::ExecIndexLookup() {
   assert(!done_);
 
+  /*
+   * If query is IN+Subquery, the values should be set using context params
+   * for now, the flag params_exec_ is set to 1 only in nestloop join
+   * in this case, the params is the results of the outer plan
+   * We can add more cases in future
+   */
+  //TODO: we probably need to add more for other cases in the future
```
jarulraj: I think we should chat about this during the group meeting. I think that we can do this stuff in the transformation module and leave the executors unchanged.

src/backend/networking/abstract_service.proto
```diff
@@ -524,4 +546,5 @@ service AbstractPelotonService {
     rpc Heartbeat(HeartbeatRequest) returns (HeartbeatResponse);
```
jarulraj: Add comments describing the purpose of each RPC type.

src/backend/networking/expression.proto
```diff
@@ -0,0 +1,39 @@
...
+  VALUE_TYPE_ARRAY = 99;
+}
+
+message AbstractExpression {
+    // The left node pointer
+    optional AbstractExpression left = 1;
+    // The right node pointer
+    optional AbstractExpression right = 2;
+    // parameter
+    optional bool b_parameter = 3;
```
jarulraj: What is `b_parameter` ? Use a more descriptive name.

src/backend/networking/expression.proto
```diff
@@ -0,0 +1,39 @@
...
+    // value size
+    optional int32 value_size = 4;
+    // inBytes
+    optional bool in_bytes = 5;
```
jarulraj: Try to give a more useful description instead of repeating the variable name.

src/backend/networking/peloton_service.cpp
```diff
@@ -267,12 +273,31 @@ void PelotonService::Heartbeat(::google::protobuf::RpcController* controller,
       LOG_ERROR("No response: site is null");
     }
 
+
+    /*
```
jarulraj: Do not use /* .. */ style comments. Use just // comments even for multi-line comments. Later, if we need to comment something quickly for temporary hacking, then we can use /* .. */ style comments.

src/backend/networking/peloton_service.cpp
```diff
@@ -305,5 +330,94 @@ void PelotonService::TimeSync(::google::protobuf::RpcController* controller,
...
+            case PLAN_NODE_TYPE_SEQSCAN: {
+                LOG_TRACE("SEQSCAN revieved");
+                std::string plan = request->plan();
+                ReferenceSerializeInputBE input(plan.c_str(), plan.size());
+                std::shared_ptr<peloton::planner::SeqScanPlan> ss_plan =
+                        std::make_shared<peloton::planner::SeqScanPlan>();
+                ss_plan->DeserializeFrom(input);
+
+                //peloton_status status =
+                        peloton::bridge::PlanExecutor::ExecutePlan(ss_plan.get(),
```
jarulraj: Indentation.

src/backend/networking/queryplan.proto
```diff
@@ -0,0 +1,55 @@
...
+// NameData should be transformed 
+message FormAttributeMsg {
+    required uint64 attrelid = 1;
+    // The original is NameData structure which contains 64 bytes char array
+    optional bytes attname = 2;
+    optional uint64 atttypid = 3;
+    optional int32 attstattarget = 4;
+    // The original is int16 which should be cast for conversion
+    optional int32 attlen = 5;
+    // The original is int16 which should be cast for conversion
```
jarulraj: Indentation.

src/backend/networking/expression.proto
domino-succ: Pls ignore expression.proto. It is not used in this version, I will delete it.

src/backend/networking/rpc_utils.cpp
```diff
@@ -0,0 +1,252 @@
...
+    if (tuple_desc_msg.has_constr()) {
+        AttrDefault* attrdef = (AttrDefault*)malloc(sizeof(AttrDefault));
+        ConstrCheck* constrch = (ConstrCheck*)malloc(sizeof(ConstrCheck));
+        TupleConstr* tuple_constr = (TupleConstr*)malloc(sizeof(TupleConstr));
+
+        std::string adbinss = tuple_desc_msg.constr().adbin();
+        std::string ccbinss = tuple_desc_msg.constr().ccbin();
+        std::string ccnamess = tuple_desc_msg.constr().ccname();
+
+        char* adbin = (char*)malloc( adbinss.size() + 1 );
```
jarulraj: Why are you using malloc ? Use `new` and smart pointers.

src/backend/planner/seq_scan_plan.cpp
```diff
@@ -0,0 +1,241 @@
...
+              break;
+            }
+          }
+    }
+
+    return true;
+}
+/**
+ *
+ * SeqScanPlan is serialized as:
```
jarulraj: These kind of figures will be very helpful later.

src/backend/networking/rpc_utils.cpp
domino-succ: That's a good suggestion. I also plan to use smart pointer here. But these raw pointers are inside tupleDesc which is a Postgres structure. We can't use std::move to transfer the ownership to these raw pointers. If we just use .get() to pass the raw pointer, we will get dangling pointers after the function call. 

Do you have any idea about that?

jarulraj:
Can you resolve the merge conflicts ?

Michael-Tieying-Zhang:
Sure. I am going to merge it again. For the problem of "malloc" for postgres structure （TupleDesc, TupleConstr ConstrCheck AttrDefault）, I figure out a solution.  We could add constructor and deconstructor for TupleDesc. In constructor, we use "new" for  TupleConstr ConstrCheck AttrDefault, and in deconstructor， we use "delete". 

But we will change Posgres's code. How do you think? Or anyone else have ideas about that?

src/backend/bridge/dml/executor/plan_executor.cpp
```diff
@@ -181,6 +193,15 @@ executor::ExecutorContext *BuildExecutorContext(ParamListInfoData *param_list,
 }
 
 /**
+ * @brief Build Executor Context
+ */
+executor::ExecutorContext *BuildExecutorContext(const std::vector<Value> &params,
+                                                concurrency::Transaction *txn) {
+  return new executor::ExecutorContext(
+      txn, params);
+}
```
pmenon: Fix this formatting.

src/backend/bridge/dml/executor/plan_executor.cpp
```diff
@@ -31,6 +32,12 @@ namespace bridge {
 executor::ExecutorContext *BuildExecutorContext(ParamListInfoData *param_list,
                                                 concurrency::Transaction *txn);
 
+/*
+ * Added for network invoking efficiently
+ */
+executor::ExecutorContext *BuildExecutorContext(const std::vector<Value> &params,
```
pmenon: Make this comment more meaningful.

src/backend/executor/abstract_executor.h
```diff
@@ -63,13 +63,13 @@ class AbstractExecutor {
...
-  // Set the context
-  void SetContext(Value value, uint32_t flag) {
-    executor_context_->SetParams(value);
-    executor_context_->SetParamsExec(flag);
-  }
+  // set the context 
+  void SetContext(Value value, ParamsExecFlag flag) {
+ 	  executor_context_->SetParams(value);
+ 	  executor_context_->SetParamsExecFlag(flag);
+   }
```
pmenon: Formatting.

pmenon:
From the looks of it, I think we can handle the subquery logic during plan transformation (i.e., when transforming Postgres plans to Peloton plans).  If so, we won't need to bring in Postgres stuctures in to Peloton plans, which should solve the whole malloc/palloc issue.

Michael-Tieying-Zhang:
@pmenon the malloc/palloc issue is not related to "transforming plan". This is the execution code of peloton: PlanExecutor::ExecutePlan(mapped_plan.get(), param_values, tuple_desc);
When we execute a plan, we need three stuff: mapped_plan, param_values, and tuple_desc. Mapped_plan is already transformed through peloton PlanTransform. But tuple_desc is the original postgres structure, and peloton also uses postgres's functions to parse tuple_desc inside ExecutePlan.

pmenon:
Micheal and I discussed the issue. The solution is to have a separate plan executor method outside the bridge that accepts a Peloton query plan and the parameter list, and returns a list of logical tiles (even better, an iterator over the result tiles). The existing plan executor method should invoke the new method to iterate over the output tuples, and convert them into Postgres tuples.  Two benefits here:

1. The networking code doesn't need to worry about serializing Postgres structures.
2. We only exchange Peloton tuples over the network, isolating us from Postgres. This allows clients to manipulate the results of a remote call without conversion to/from Peloton/Postgres all the time.

Michael-Tieying-Zhang:
Cool 




tieyingz@cs.cmu.edu

From: Prashanth Menon
Date: 2016-04-11 15:25
To: cmu-db/peloton
CC: Michael
Subject: Re: [cmu-db/peloton] Sending query plan between peloton nodes (#130)
Micheal and I discussed the issue. The solution is to have a separate plan executor method outside the bridge that accepts a Peloton query plan and the parameter list, and returns a list of logical tiles (even better, an iterator over the result tiles). The existing plan executor method should invoke the new method to iterate over the output tuples, and convert them into Postgres tuples. Two benefits here:
The networking code doesn't need to worry about serializing Postgres structures. 
We only exchange Peloton tuples over the network, isolating us from Postgres. This allows clients to manipulate the results of a remote call without conversion to/from Peloton/Postgres all the time.

apavlo:
@pmenon's suggestions LGTM.

jarulraj:
@Michael-Tieying-Zhang  can you remove the code you have added to support subqueries from this PR and fix the merge conflicts ? Then, we can merge just the networking stuff.

Michael-Tieying-Zhang:
All the tests pass on my local machine. Why one of them fails?

Michael-Tieying-Zhang:
Done.

Michael-Tieying-Zhang:
For the execution results, I didn't send LogicalTile. Because LogicalTile doesn't contain the real tuple, when the client recvs LogicalTile, the only stuff the client can do is to send it back which means "I want the tuples in this LogicalTile".

So I materialize the LogicalTile into a physical tile and serializes the physical tile and send it back. There might be several LogicalTile, the server may keep a handler, and the client can requests the next LogicalTile. The current code doesn't include the handler，and we may implement that in the next step. I also checked the current execution method, it also gets all the tuples/tiles and sends them back (https://github.com/cmu-db/peloton/blob/master/src/backend/bridge/dml/executor/plan_executor.cpp#L92), which may need to be refatored later.

jarulraj:
Ok, sounds good. Can you another merge ?

Michael-Tieying-Zhang:
Done

Issue #124, opened 2016-04-08
Move message_queue to common.


amaliujia:
Can @jarulraj move message_queue (https://github.com/cmu-db/peloton/blob/master/src/backend/networking/message_queue.h) to common and rename it?

Such producer-consumer queue is necessary for parallelism and I am using my version too.  Let's create one for everyone who need it.

Can you also add `void Push(const vector<T>& item)` for convenience?

jarulraj:
Feel free to do this and send a PR.

yingjunwu:
I don't find it is necessary to add function like `void Push(const vector<T>& item)`. The semantic is somewhat confusing. Do we need to push a set of items to the queue *atomically*? If yes, then you'd better directly use `queue<std::vector<some_type>>`.

Feel free to comment on my PR #134.

PR #134, opened 2016-04-10
add lockfree queues

yingjunwu:
@amaliujia I have ported two lockfree queues to Peloton. I think they can meet your need in #124.


amaliujia:
That's fantastic! Thanks.

MattPerron:
@eric-haibin-lin Can we use this?

Issue #143, opened 2016-04-14
Prepared Statements not working/supported?


larryxiao:
Tested on 6a88a585ca6a215b34dbac9e0e3221278ed28be1 (Apr 12) and 71d7a687c6d474f5ff95dbf8272f08b88b03932f (Mar 13).

```
CREATE TABLE test ( key VARCHAR(200) PRIMARY KEY, value VARCHAR(2048), flag smallint, size smallint );

PREPARE GET (text) AS SELECT key, flag, size, value FROM TEST WHERE key = $1;
EXECUTE GET('1');
```

```
postgres=# PREPARE GET (text) AS SELECT key, flag, size, value FROM TEST WHERE key = $1;
PREPARE
postgres=# EXECUTE GET('1');
 key | flag | size | value 
-----+------+------+-------
(0 rows)

postgres=# EXECUTE GET('1');
server closed the connection unexpectedly
        This probably means the server terminated abnormally
        before or while processing the request.
The connection to the server was lost. Attempting reset: Failed.
```

```
10:36:54,405 [../../src/backend/bridge/dml/mapper/dml_utils.cpp:845:CopyExprState] TRACE - ExprState tag : 400 , Expr tag : 303 
10:36:54,405 [../../src/backend/bridge/dml/mapper/dml_utils.cpp:845:CopyExprState] TRACE - ExprState tag : 400 , Expr tag : 303 
10:36:54,405 [../../src/backend/bridge/dml/mapper/dml_utils.cpp:845:CopyExprState] TRACE - ExprState tag : 400 , Expr tag : 303 
10:36:54,405 [../../src/backend/bridge/dml/mapper/dml_utils.cpp:845:CopyExprState] TRACE - ExprState tag : 400 , Expr tag : 303 
10:36:54,405 [../../src/backend/bridge/dml/mapper/mapper_index_scan.cpp:75:TransformIndexScan] INFO  - Index scan on test using oid 20342, index name: test_pkey
10:36:54,405 [../../src/backend/bridge/dml/mapper/mapper_index_scan.cpp:79:TransformIndexScan] TRACE - Scan order: 1
10:36:54,405 [../../src/backend/bridge/dml/mapper/mapper_index_scan.cpp:86:TransformIndexScan] TRACE - num of scan keys = 1, num of runtime key = 0
10:36:54,405 [../../src/backend/bridge/dml/tuple/tuple_transformer.cpp:144:GetValue] TRACE - len = 1 , text = "1"
[...]

PR #149, opened 2016-04-22
fix bug in transforming empty strings in scan key

omegaga:
This PR mainly fixes the bug in issue #143.

As mentioned in that post, Postgres uses generic plan after several rounds of execution. In a generic plan, a parameter will be filled with a `Param` node rather than a `Const` node, and will be replaced by runtime_key in executor. However, Peloton does not handle some data types correctly (like strings and lists) when transforming the `Param` node.

This PR first fixes the bug in handling text, varchar, bpchar. What we do here is to simply transform it to an empty Value, since it will be replaced by runtime key anyway.

Things are more complicated for arrays. An `ArrayExpr` is passed as `runtime_key`, and Peloton does not support transforming `ArrayExpr` now. <del>Since this is sort of independent issue, I will send a separate PR when I finish it.</del> I added a `TransformArrayExpr` to transform `ArrayExpr` to `VectorExpression` in our VoltDB-like expression system.

### DIFF
--- a/src/backend/bridge/dml/expr/expr_transformer.cpp
+++ b/src/backend/bridge/dml/expr/expr_transformer.cpp
@@ -76,6 +76,10 @@ expression::AbstractExpression *ExprTransformer::TransformExpr(
       peloton_expr = TransformScalarArrayOp(expr_state);
       break;
 
+    case T_ArrayExpr:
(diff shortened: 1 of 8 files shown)


omegaga:
UPDATE: bugs in transforming arrays have been fixed.

PR #150, opened 2016-04-22
refactor and enhancement

yingjunwu:
1. Fixed all kinds of bugs in concurrency-control layer;
2. Add GC support. By default, GC is disabled;
3. Refactor index structure. As the ItemPointer structure is 16 byte, CAS operation does not work. Spinlock is introduced to synchronize operations;
4.  YCSB benchmark (hard-code version) works now.


yingjunwu:
@jarulraj can you help fix this compilation error?

yingjunwu:
GC is working now. The current GC implementation can unlink garbage versions from its corresponding index bucket, further speeding up version-chain traversing. However, it does not reclaim memory from garbage versions. We are working on memory reclamation using Hekaton's epoch-based scheme.

jarulraj:
Is GC enabled by default ?

yingjunwu:
@jarulraj GC is disabled by default. We are tuning the performance.

It is ok to restrict the size of ItemPointer, but I find that `ItemPointerContainer` may still be useful for GC.

Let's consider the cooperative garbage collection. Now we assume that a version chain contains four versions: `v1` (oldest), `v2`, `v3`, and `v4` (newest), where `v1`, `v2`, and `v3` are garbage versions. Two transactions `T1` and `T2` may traverse the version chain at the same time. During the traversing, `T1` will find that `v1`, `v2`, and `v3` are garbage versions, and it will set the pointer of the corresponding index bucket to `v1`, `v2`, and `v3` accordingly.  Now the `T2` comes in, and it will perform the same action. That is, even if the pointer of the index bucket has already been set to `v3` by `T1`, `T2` will still set the pointer back to `v1` and `v2`. While eventually the pointer will be set to `v3` by `T2`, some other concurrent transactions may still see `v1` and `v2`, thus wasting time for traversing. Now if we maintain a timestamp in `ItemPointerContainer` recording the `begin_cid` of the header item pointer, such problem can be addressed.

Definitely the mechanism I described requires acquiring spinlock for pointer swap. I am not sure how the performance can be affected. 

Well, you can safely restrict the size of ItemPointer to 8, and I will figure out which kind of implementation yields the best performance.

jarulraj:
I have changed the type of `oid_t`. You should also not use `std::cout` in the code base.

yingjunwu:
@jarulraj `std::cout` removed. but I am using `printf` to print out throughput for convenience. will move it out later. 

Can you help fix compilation error?

jarulraj:
I have added support for `boost 1.55` on Travis CI. Can you do another merge and see if that works ?

yingjunwu:
@jarulraj I don't understand. We have already passed compilation as well as all the test cases. Why it still reports error?

jarulraj:
You can look at the Travis CI report for the errors : https://travis-ci.org/cmu-db/peloton/builds/125264504

yingjunwu:
@jarulraj nice, it works now :-)

jarulraj:
Is it possible to do GC without `ItemPointerContainer` ? That is a major change.

yingjunwu:
We must have `ItemPointerContainer` or similar structures to hold `ItemPointer`. The major reason is that we may need to maintain auxiliary variables, such as timestamp, in `ItemPointerContainer`, as was described in my previous post in this thread.

I think the key question one may want to ask is why we maintain `ItemPointerContainer*` instead of `ItemPointerContainer`. This is because the underlying tree structure (either B+tree or Bwtree) can copy or delete its leaf node due to structure change. Directly maintaining a stack-allocated `ItemPointerContainer` requires more complicated synchronization scheme. More specifically, when we change (or let's say swap) the value of `ItemPointerContainer` maintained in a certain index bucket `b`, some other transactions can insert or delete some tuples, causing structure change, which potentially copies or deletes `b`. To make the swap safe, synchronization primitives must be introduced to block index-level copies and deletes. However, if we maintain a heap-allocated structure that is pointed by `ItemPointerContainer`, then the updates of the value of `ItemPointerContainer` can be performed concurrently with the index structure change.

Another possible question is why we maintain `ItemPointerContainer*` instead of `std::shared_ptr<ItemPointerContainer>`. The answer is that the underlying tree structure is unaware of the semantics of the smart pointer. When copying or deleting a tree node, the content can be copied in binary instead of calling smart pointer's copy constructor.

To sum up,
1. we need `ItemPointerContainer` instead of `ItemPointer` for the extensibility of the system;
2. we need `ItemPointerContainer*` instead of `ItemPointerContainer` for less synchronization overhead;
3. we need `ItemPointerContainer*` instead of `std::smart_ptr<ItemPointerContainer>` as the underlying tree structure is unaware of the upper layer data structure.

jarulraj:
Let's discuss this in person sometime tomorrow or on Monday.

yingjunwu:
@jarulraj sure. drop by 9022 whenever it is convenient.

Issue #161, opened 2016-05-05
compile error with format string


larryxiao:
The CI doesn't seem to check debug info in build. When configured with

```
../configure CXXFLAGS="-O0 -g -ggdb" --enable-debug CPPFLAGS="-DLOG_LEVEL=LOG_LEVEL_TRACE"
```

Error:
```
In file included from ../../src/backend/storage/data_table.cpp:22:0:
../../src/backend/storage/data_table.cpp: In member function 'peloton::ItemPointer peloton::storage::DataTable::GetEmptyTupleSlot(const peloton::storage::Tuple*, bool)':
../../src/backend/common/logger.h:132:43: error: use of deleted function 'std::atomic<long unsigned int>::atomic(const std::atomic<long unsigned int>&)'
   ::fprintf(LOG_OUTPUT_STREAM, __VA_ARGS__);                          \
                                           ^
../../src/backend/storage/data_table.cpp:152:3: note: in expansion of macro 'LOG_TRACE'
   LOG_TRACE("tile group count: %lu, tile group id: %lu, address: %p",
   ^
In file included from ../../src/backend/concurrency/transaction.h:16:0,
                 from ../../src/backend/storage/data_table.h:22,
                 from ../../src/backend/storage/data_table.cpp:17:
/usr/include/c++/4.8/atomic:658:7: error: declared here
       atomic(const atomic&) = delete;
       ^
In file included from ../../src/backend/storage/data_table.cpp:22:0:
../../src/backend/common/logger.h:132:43: error: cannot pass objects of non-trivially-copyable type 'struct std::atomic<long unsigned int>' through '...'
   ::fprintf(LOG_OUTPUT_STREAM, __VA_ARGS__);                          \
                                           ^
../../src/backend/storage/data_table.cpp:152:3: note: in expansion of macro 'LOG_TRACE'
   LOG_TRACE("tile group count: %lu, tile group id: %lu, address: %p",
   ^
../../src/backend/common/logger.h:132:43: error: format '%lu' expects argument of type 'long unsigned int', but argument 3 has type 'std::atomic<long unsigned int>' [-Werror=format=]
   ::fprintf(LOG_OUTPUT_STREAM, __VA_ARGS__);                          \
                                           ^
../../src/backend/storage/data_table.cpp:152:3: note: in expansion of macro 'LOG_TRACE'
   LOG_TRACE("tile group count: %lu, tile group id: %lu, address: %p",
   ^
```

apavlo:
[...]

PR #162, opened 2016-05-05
update travis to compile all log message #161

larryxiao:
#161

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,6 +24,6 @@ script:
     # bootstrap
     - ./bootstrap
     - cd build
-    - ../configure CXXFLAGS="-O0" --enable-debug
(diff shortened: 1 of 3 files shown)


larryxiao:
Including the log messages at TRACE level creates too much log. Resulting in:

```
The log length has exceeded the limit of 4 MB (this usually means that the test suite is raising the same exception over and over).
The job has been terminated
```

Can we make the logging level a runtime option?

apavlo:
Let's disable trace for Travis and I will make a separate build task on
Jenkins.

It's better to have it as a preprocessor directive to reduce branches.

larryxiao:
Didn't think of that. Thank you!

PR #164, opened 2016-05-06
comment sending query

Michael-Tieying-Zhang:
Make the current networking model *un-affect* other models


normanyahq:
quick fix to [SeqScan crashed](https://github.com/cmu-db/peloton/issues/165)

jarulraj:
I removed all the networking stuff from peloton dml for now.

PR #167, opened 2016-05-07
Support for T_Param and T_ResultState type Subquery + assert fix on REAL type

normanyahq:
Subquery
-------------

DDL

> create table a (a int, b int);
> create table b (a int, b int);
> insert into a values(1, 2);
> insert into b values(1, 2);

Following are two different types of subqueries

> select a from a where a = (select min(a) from b) and b = (select max(a) from b);
> select a from a where a = (select min(a) from b) and a = (select max(a) from b);

The query plans are:

T_Param:

> postgres=# explain select a from a where a = (select min(a) from b) and b = (select max(a) from b);
>  Seq Scan on a  (cost=76.52..120.42 rows=1 width=4)
>    Filter: ((a = $0) AND (b = $1))
>    InitPlan 1 (returns $0)
>      ->  Aggregate  (cost=38.25..38.26 rows=1 width=4)
>            ->  Seq Scan on b  (cost=0.00..32.60 rows=2260 width=4)
>    InitPlan 2 (returns $1)
>      ->  Aggregate  (cost=38.25..38.26 rows=1 width=4)
>            ->  Seq Scan on b b_1  (cost=0.00..32.60 rows=2260 width=4)
> (8 rows)

For this type of subqueries, we need to evaluate them, get a value for each and push the values into param vector before the execution of main plan.

T_ResultState:
> postgres=# explain select a from a where a = (select min(a) from b) and a = (select max(a) from b);
>  Result  (cost=76.52..114.77 rows=11 width=4)
>    One-Time Filter: ($0 = $1)
>    InitPlan 1 (returns $0)
>      ->  Aggregate  (cost=38.25..38.26 rows=1 width=4)
>            ->  Seq Scan on b  (cost=0.00..32.60 rows=2260 width=4)
>    InitPlan 2 (returns $1)
>      ->  Aggregate  (cost=38.25..38.26 rows=1 width=4)
>            ->  Seq Scan on b b_1  (cost=0.00..32.60 rows=2260 width=4)
>    ->  Seq Scan on a  (cost=0.00..38.25 rows=11 width=4)
>          Filter: (a = $1)
> (10 rows)

For this type, we need to handle T_ResultState in mapper.cpp

@pmenon , as we discuss before, for these two types of queries, I think evaluate it and pass the result into param vector before the execution of main plan is correct, as you can see from the subplan names "InitPlan*", which is different from "SubPlan*"

And following is another type of subquery, which is expression type, and should be transformed into subquery expression and be evaluated at the last minute:

> SELECT * FROM STATION
> WHERE 50 < (SELECT AVG(TEMP_F) FROM STATS
> WHERE STATION.ID = STATS.ID);

> postgres=# explain SELECT * FROM STATION
> postgres-# WHERE 50 < (SELECT AVG(TEMP_F) FROM STATS
> postgres(# WHERE STATION.ID = STATS.ID);
>  Seq Scan on station  (cost=0.00..8897.43 rows=200 width=108)
>    Filter: ('50'::double precision < (SubPlan 1))
>    SubPlan 1
>      ->  Aggregate  (cost=14.79..14.80 rows=1 width=4)
>            ->  Bitmap Heap Scan on stats  (cost=4.22..14.76 rows=9 width=4)
>                  Recheck Cond: (station.id = id)
>                  ->  Bitmap Index Scan on stats_pkey  (cost=0.00..4.22 rows=9 width=0)
>                        Index Cond: (station.id = id)
> (8 rows)
> 

I will take at look into this case later.

REAL type
-------------
Currently we don't have a 4-byte float. Thus, all REAL type from postgres part are transformed into double   in peloton part, and will be transformed back to 4-byte float before sending back to postgres. Thus, some assertion checking whether the type is double should also tolerate real.


src/backend/bridge/dml/executor/plan_executor.cpp
```diff
@@ -35,15 +35,124 @@ executor::ExecutorContext *BuildExecutorContext(ParamListInfoData *param_list,
...
 void CleanExecutorTree(executor::AbstractExecutor *root);
 
+/*
+ * Execute the subplan and get a value from the result.
+ */
+Value PlanExecutor::ExecutePlanGetValue(const planner::AbstractPlan *plan,
+                                        const std::vector<Value> &params) {
+  peloton_status p_status;
+
+  //  if (plan == nullptr) return p_status;
```
jarulraj: Do we need this check  or not ? If not, please remove it and add a comment saying how we will handle empty plans.

src/backend/bridge/dml/executor/plan_executor.cpp
```diff
@@ -35,15 +35,124 @@ executor::ExecutorContext *BuildExecutorContext(ParamListInfoData *param_list,
...
+  }
+  assert(txn);
+
+  LOG_TRACE("Txn ID = %lu ", txn->GetTransactionId());
+  LOG_TRACE("Building the executor tree");
+
+  // Use const std::vector<Value> &params to make it more elegant for network
+  std::unique_ptr<executor::ExecutorContext> executor_context(
+      BuildExecutorContext(params, txn));
+  // auto executor_context = BuildExecutorContext(param_list, txn);
```
jarulraj: Do we need this ?

src/backend/bridge/dml/executor/plan_executor.cpp
```diff
@@ -35,15 +35,124 @@ executor::ExecutorContext *BuildExecutorContext(ParamListInfoData *param_list,
...
+      expression::ContainerTuple<executor::LogicalTile> cur_tuple(
+          logical_tile.get(), tuple_id);
+      ret_val = cur_tuple.GetValue(0);
+      break;
+    }
+  }
+
+  // Set the result
+  p_status.m_processed = executor_context->num_processed;
+//  p_status.m_result_slots = slots;
```
jarulraj: Why is this commented out ?

src/backend/bridge/dml/executor/plan_executor.cpp
```diff
@@ -79,8 +188,8 @@ peloton_status PlanExecutor::ExecutePlan(const planner::AbstractPlan *plan,
 
   // Use const std::vector<Value> &params to make it more elegant for network
   std::unique_ptr<executor::ExecutorContext> executor_context(
-  BuildExecutorContext(params, txn));
-  //auto executor_context = BuildExecutorContext(param_list, txn);
+      BuildExecutorContext(params, txn));
+  // auto executor_context = BuildExecutorContext(param_list, txn);
```
jarulraj: Remove this.

src/backend/bridge/dml/executor/plan_executor.h
```diff
@@ -38,15 +38,22 @@ class PlanExecutor {
   static void PrintPlan(const planner::AbstractPlan *plan,
                         std::string prefix = "");
 
-/* TODO: Delete this mothod
-  static peloton_status ExecutePlan(const planner::AbstractPlan *plan,
-                                    ParamListInfo m_param_list,
-                                    TupleDesc m_tuple_desc);
-*/
+  /* TODO: Delete this mothod
```
jarulraj: Please delete this method.

src/postgres/backend/postmaster/peloton.cpp
```diff
@@ -166,42 +161,59 @@ peloton_ddl(Node *parsetree) {
...
   }
 
   // Construct params
-  std::vector<peloton::Value> param_values = peloton::bridge::PlanTransformer::BuildParams(param_list);
-
-  //===----------------------------------------------------------------------===//
-  //   End for sending query
-  //===----------------------------------------------------------------------===//
+  std::vector<peloton::Value> param_values =
+      peloton::bridge::PlanTransformer::BuildParams(param_list);
```
jarulraj: Can we move the subsequent code to `peloton::bridge::PlanTransformer::BuildParams` ? Like pass in another argument to it, and do this processing within that function.

src/postgres/backend/postmaster/peloton.cpp
```diff
@@ -361,76 +358,74 @@ bool IsPelotonQuery(List *relationOids) {
...
 
-    return true;
+  return true;
 }
 
 /**
  * TODO: Deserialize
  */
 bool peloton_status::DeserializeFrom(peloton::SerializeInputBE &input) {
+  //    List *slots = NULL;
```
jarulraj: Please remove this code.

jarulraj:
Copying Heqing's mail :

I've handled some subquery types, and re-tested the NIST queries:
https://github.com/agershun/learnsql/tree/master/node_modules/alasql/test-sql/nist

Got following crashed:

14.sql
SELECT * FROM STATION
WHERE 50 < (SELECT AVG(TEMP_F) FROM STATS
WHERE STATION.ID = STATS.ID);

21.sql
COMMIT WORK;

24.sql
ROLLBACK WORK;

26.sql
UPDATE STATS SET RAIN_I = 4.50
WHERE ID = 44
AND MONTH = 7;
COMMIT WORK;


28.sql
DELETE FROM STATS
WHERE MONTH = 7
OR ID IN (SELECT ID FROM STATION
WHERE LONG_W < 90);
DELETE FROM STATION WHERE LONG_W < 90;
COMMIT WORK;
SELECT * FROM STATION;
SELECT * FROM STATS;
SELECT * FROM METRIC_STATS;

Except for 14.sql, all are related to COMMIT and ROLLBACK.

And for 14.sql, I have some discussion here:
https://github.com/cmu-db/peloton/pull/167
which should be transformed into subquery expression.

pmenon:
Thanks for submitting the PR, Heqing. Looking at it now, I'm still not sold on this approach. There isn't a clear separation of plan construction and execution. We've ended up copying a lot of code in order to *not* refactor existing code. This will also make code-gen more difficult because all the plan state isn't contained within the plan.

Now, since we'll be rewriting the front-end and this will probably have to be rewritten, I guess it's okay to merge it in to save the effort/time spent to refactor other parts.

src/backend/bridge/dml/executor/plan_executor.cpp
normanyahq: I shouldn't be empty. I change it into assert.

src/backend/bridge/dml/executor/plan_executor.cpp
normanyahq: It's executed as a new plan, I guess, yes?

src/backend/bridge/dml/executor/plan_executor.cpp
normanyahq: Because this method only returns the first value of of result tuple in execution, slots are for the main plan instead of initplan.

normanyahq:
There are two types of "subplan".
one is initplan, and another is subplan, as you can see the explain results I've shown in previous comments.

for initplan, we should get a value and put it into parameter -- this is what I'm handling.
for subplan, it's executed in runtime -- this should be transformed into subquery expression.

normanyahq:
@jarulraj I've modified according to your comments. You can take a look

normanyahq:
@pmenon I think it should be correct. Because they are executed only once and the result are used as params.

I think the case you are worrying about is not in this plan type, but the "SubPlanState"

pmenon:
What is the distinction between SubPlan and InitPlan, other than that's what Postgres does?

normanyahq:
@pmenon InitPlan generates only one value, and the execution is independent from main plan context.

I guess following two queries can show the differences.

Initplan

> select a from a where a = (select min(a) from b) and b = (select max(a) from b);

Subplan

> SELECT * FROM STATION
> WHERE 50 < (SELECT AVG(TEMP_F) FROM STATS
> WHERE STATION.ID = STATS.ID);

pmenon:
I still don't see why they both can't be represented as expressions in the main query plan, one evaluating the subplan and caching it (InitPlan), and one evaluating the subplan per-tuple (SubPlan).  Both return one value.

Now, the problem with correctness. Using your query as an example, it is possible for a transaction to insert a new minimum value of 'a' into b between the evaluation of the InitPlan, but before the execution of the main plan that should be visible to the main plan.

normanyahq:
That's true. They can be combined together, but that will involve a lot of work. The interfaces need to be changed to pass in the pointers of top-level plan state so that we can replace "T_Param" and "T_ResultState" with subquery expression, for now, this implementation is the only solution with least modification.

For the correctness, do you have any idea? I think this is the implementation problem. I don't know how I can put the subqueries into the same transection with main query, even if we transform it into subquery expression, the evaluation is still separate because it's another plan, right?

jarulraj:
@normanyahq nice work !
@pmenon Your thoughts on merging this PR ? Should we do more refactoring ?

pmenon:
@jarulraj I think we should refactor it. Let's do it right the first time.

@normanyahq Can you refactor it so that you pull out the Postgres subplans into some TransformationContext that is passed around during transformation? This way, when the ExpressionTransformer is called, it should have access to the context, and the subplans it references.

At a high-level, the goals should be:

1. Make PlanTransformer::TransformPlan() re-entrant.
2. Create a subclass of AbstractExpression for subqueries. Subquery expressions that only need to be evaluated once can cache their results.

We get a few benefits this way:

1. Both InitPlan and SubPlan can be expressed uniformly as expressions in the plan tree. This can be used even after the front-end rewrite.
2. The subplans are invoked/evaluated with the main plan within the same transaction, ensuring correctness.
3. We isolate ourselves from how Postgres does things.

I think this is the right approach, but please chime in if you want something else.

normanyahq:
OK, I will think about the details after my final.

jarulraj:
@normanyahq Can you refactor the code ?

normanyahq:
Yes, I'm working On it. Sorry, I've just come to Mount View, it takes sometime to setup everything. I should be able to finish soon.

PR #179, opened 2016-05-23
Index expression

amaliujia:
In this commit, complicated range search index expressions are supported. In order version of code, greater than, less than expression are implemented for scanning all index data.

Now, >=, >, <, <=, = can be combined on different columns and do efficient search. All index scan is not needed anymore. Expressions will create intervals and index only search intervals.

Index tests are also extended to test this new feature.


src/backend/common/value.cpp
```diff
@@ -85,6 +85,10 @@ Value &Value::operator=(const Value &other) {
           Varlen *src_sref = *reinterpret_cast<Varlen *const *>(other.m_data);
           Varlen *new_sref = Varlen::Clone(*src_sref, nullptr);
 
+          // TODO: Fix memory leak problem.
```
jarulraj: Is this leak still there ? Do we need this comment ?

src/backend/common/value.h
```diff
@@ -2512,6 +2522,12 @@ class Value {
     return GetAllocatedValue(VALUE_TYPE_VARCHAR, value, size, nullptr);
   }
 
+  static Value GetMaxTempStringValue(size_t size) {
```
jarulraj: We can fix INT_MAX as the max string size and remove the size argument.

amaliujia: Done

src/backend/common/value.cpp
amaliujia: If I released the string using the way in destructor, basically I could pass small unit test, but there were somewhere else broken. It may change a lot. So it is better to fix this one in another PR.

jarulraj: That makes sense

PR #187, opened 2016-06-08
Modified few planners and executers to cache some immutable variables

saifalharthi:
In this PR I started modifying couple of plans and executers to use immutable variables from the plans rather than giving a copy to the executers. The end goal is to cache the plans. This PR is for code review purposes.


.gitignore
```diff
@@ -307,3 +307,28 @@ docs/dev
...
+CTestTestfile.cmake
+DartConfiguration.tcl
+Doxyfile
+PelotonConfig.cmake
+PelotonTargets.cmake
+cmake/PelotonConfig.cmake
+cmake_install.cmake
+lib/
+peloton_config.h
+src/CMakeFiles/
```
jarulraj: Remove these files.

src/executor/hybrid_scan_executor.cpp
```diff
@@ -70,19 +70,19 @@ bool HybridScanExecutor::DInit() {
     index_done_ = false;
 
     column_ids_ = node.GetColumnIds();
-    key_column_ids_ = node.GetKeyColumnIds();
-    expr_types_ = node.GetExprTypes();
+    //key_column_ids_ = node.GetKeyColumnIds();
```
jarulraj: Use local variables.

src/include/executor/hybrid_scan_executor.h
```diff
@@ -82,13 +82,13 @@ class HybridScanExecutor : public AbstractScanExecutor {
   //===--------------------------------------------------------------------===//
   std::vector<oid_t> column_ids_;
```
jarulraj: You can remove the member variable from the base class.

src/include/executor/hybrid_scan_executor.h
```diff
@@ -82,13 +82,13 @@ class HybridScanExecutor : public AbstractScanExecutor {
...
 
-  std::vector<oid_t> key_column_ids_;
+ // std::vector<oid_t> key_column_ids_;
```
jarulraj: Don't comment out things.

PR #203, opened 2016-08-05
Frontend - Lin's work + index type specificaion

saifalharthi:
In this PR:
- Lin's work on TPCC
- Index type specification from the parser (Default is SKIPLIST)

From NewYork with love


wangziqi2013:
Commit c499042 changes the default logging level, which causes a failure in Travis CI because too many log messages are printed inside index_performance_test (inserting 1 million key). We could either consider disabling logging inside index performance test, or revert back to the normal default logging level

malin1993ml:
@wangziqi2013 That is fixed. Thanks!

PR #204, opened 2016-08-10
Adding a New Index Scan Optimizer & New Test Cases & Fix Bug in "class Value"

wangziqi2013:
Index scan with AND predicate now works correctly on all test cases, with the new index predicate scan optimizer that supports late binding, multiple binding

TODO: 

1. Further optimize bwtree scan which is not in its most efficient form with the new scan optimizer
2. Currently the predicate works together with IndexScanPlan and HybridScanPlan and their respective executors, in a single threaded environment (i.e. planning and binding must be done by only one thread at a time). I am not sure whether this is consistent with the design or not. If not maybe we could consider putting the index scan into the plan state, which is not shared between threads.


apavlo:
Even though we are single-threaded now, we are going to be multi-threaded in the future. So it would be good if we make the executors thread safe.

wangziqi2013:
The executor is indeed thread-safe right now. What was left undone is multi-threaded parameter binding, i.e. multiple threads trying to bind parameter values received from the wire protocol to an index scan plan for stored procedures. 

I followed Lin's way of doing the binding, and assumed that only one thread could bind parameter values to a plan at a time. If this is not the case then the index optimizer should be moved to the plan state rather than staying inside index scan plan.

src/catalog/schema.cpp
```diff
@@ -124,12 +125,74 @@ Schema *Schema::CopySchema(const Schema *schema) {
   return CopySchema(schema, set);
 }
 
-// Copy subset of columns in the given schema
+/*
+ * CopySchema() - Copies the schema into a new schema object with index_list
+ *                as indices to copy
+ *
+ * This function essentially does a "Gathering" operation, in a sense that it
+ * collects columns indexed by elemenys inside index_list in the given order
```
malin1993ml: elements

src/catalog/schema.cpp
```diff
@@ -124,12 +125,74 @@ Schema *Schema::CopySchema(const Schema *schema) {
...
+ *
+ * The returned schema is created by new operator, and the caller is responsible
+ * for destroying it.
+ */
 Schema *Schema::CopySchema(const Schema *schema,
-                           const std::vector<oid_t> &set) {
+                           const std::vector<oid_t> &index_list) {
+  std::vector<Column> column_list{};
+  
+  // Reserve some space to avoid multiple ma110��() calls
```
malin1993ml: Seems an encoding error here?

src/include/common/value.h
```diff
@@ -2966,19 +2984,39 @@ inline void Value::SerializeToTupleStorageAllocateForObjects(
...
-      if (GetObjectLengthWithoutNull() == INT_MAX) {
-        LOG_INFO("Variable length with INT_MAX");
+      
+      // Special treatment for VARCHAR: Always treat it as non-inlined value
+      // and allocate a special object only sufficient for holding
+      // the length of the object (should be 4 bytes)
+      if (GetObjectLengthWithoutNull() == VARCHAR_MAX_INDICATOR) {
+        LOG_INFO("SerializeToTupleStorageAllocateForObjects(): "
+                 "Variable length with VARCHAR_MAX_INDICATOR "
+                 "(VARCHAR Maximum Value)");
```
malin1993ml: Reduce the debug level.

src/include/common/value_factory.h
```diff
@@ -63,15 +63,20 @@ class ValueFactory {
...
+  /// be freed automatically by the destructor inside Value's destructor
+  /// otherwise, it will be left in the string pool and does not require
+  /// deallocation
+  ///
+  /// Both cases do not need the caller to perform any special action such
+  /// as freeing memory
   static inline Value GetStringValue(const char *value,
                                      VarlenPool *pool = NULL) {
     return Value::GetAllocatedValue(VALUE_TYPE_VARCHAR, value,
                                     (size_t)(value ? strlen(value) : 0), pool);
```
malin1993ml: Explicit nullptr check.

src/include/index/index_util.h
```diff
@@ -13,6 +13,9 @@
 
 #pragma once
 
+#include <map>
+#include "index/index.h"
```
malin1993ml: Has index.h to be included here? Can we use fast forward?

src/include/index/scan_optimizer.h
```diff
@@ -0,0 +1,604 @@
...
+ private:
+   
+  // This is the list holding indices of the upper bound and lower bound
+  // for the scan key. We only keep index in the corresponding Value[]
+  // since some values might not be able to bind when constructing the
+  // IndexScanPlan
+  //
+  // The length of this vector should equal the number of columns in
+  // index key
+  std::vector<std::pair<oid_t, oid_t>> value_index_list;
```
malin1993ml: Use underscore for Class Data Members. The same for all other members below. (https://google.github.io/styleguide/cppguide.html) Actually I know it was probably already there before you modified, but it would be nice if you can apply the same refactor to tuple schema and index metadata...

src/include/index/scan_optimizer.h
```diff
@@ -0,0 +1,604 @@
...
+
+#pragma once
+
+#include <vector>
+
+#include "index/index.h"
+#include "common/exception.h"
+#include "common/logger.h"
+#include "common/pool.h"
+#include "catalog/schema.h"
```
malin1993ml: I saw that you have some fast forwards below. You could remove some includes if they're not necessary. Or move the definition of complex functions to a .cpp file if they require many dependencies.

src/include/planner/hybrid_scan_plan.h
```diff
@@ -19,6 +19,8 @@
 #include "common/types.h"
 #include "expression/abstract_expression.h"
 
+#include "index/scan_optimizer.h"
```
malin1993ml: Remove the include and use fast forward.

src/include/planner/index_scan_plan.h
```diff
@@ -20,11 +20,13 @@
 #include "common/types.h"
 #include "expression/abstract_expression.h"
 #include "storage/tuple.h"
+#include "index/scan_optimizer.h"
```
malin1993ml: Remove if possible.

src/include/index/scan_optimizer.h
```diff
@@ -0,0 +1,604 @@
...
+  // IndexScanPlan
+  //
+  // The length of this vector should equal the number of columns in
+  // index key
+  std::vector<std::pair<oid_t, oid_t>> value_index_list;
+  
+  // This vector holds indices for those index key columns that have
+  // a free variable
+  // We use this to speed up value binding since we could just skip
+  // columns that do not have a free variabke
```
malin1993ml: variable

src/include/index/index.h
```diff
@@ -101,44 +149,73 @@ class IndexMetadata : public Printable {
...
+  // The mapping relation between key schema and tuple schema
   std::vector<oid_t> key_attrs;
-
-  // unique keys ?
+  
+  // The mapping relation between tuple schema and key schema
+  // i.e. if the column in tuple is not indexed, then it is set to INVALID_OID
+  //      if the column in tuple is indexed, then it is the index in index_key
+  //
+  // This vector has the same length as the tuple schema.columns
```
malin1993ml: These comments are very useful!

src/include/index/scan_optimizer.h
```diff
@@ -0,0 +1,604 @@
...
+/*
+ * class IndexScanPredicate - The predicate for index scan
+ *
+ * This class maintains planner information about index scan, especially
+ * columns involved in the index scan and their corresponding values.
+ *
+ * Please note that this class represents the most general form of scan
+ * predicate that does not have any other constraints
+ */
+class IndexScanPredicate {
```
malin1993ml: Actually, why are the two classes declared in the same header?

src/index/bwtree_index.cpp
```diff
@@ -15,6 +15,8 @@
 #include "index/index_key.h"
 #include "storage/tuple.h"
 
+#include "index/scan_optimizer.h"
+
```
malin1993ml: Fast forward.

src/planner/index_scan_plan.cpp
```diff
@@ -37,25 +40,43 @@ IndexScanPlan::IndexScanPlan(storage::DataTable *table,
...
   auto where = predicate_with_params_->Copy();
+  
   expression::ExpressionUtil::ConvertParameterExpressions(where, values, GetTable()->GetSchema());
   SetPredicate(where);
+  
   for (unsigned int i = 0; i < values_.size(); ++i) {
-  //for (auto &value : values_) {
-	auto &value = values_[i];
-	auto column_id = key_column_ids_[i];
+    //for (auto &value : values_) {
```
malin1993ml: Remove unused code.

test/index/index_util_test.cpp
```diff
@@ -0,0 +1,450 @@
...
+
+  // At this time low key and high key should be binded
+  LOG_INFO("Low key = %s", cl[0].GetLowKey()->GetInfo().c_str());
+  LOG_INFO("High key = %s", cl[0].GetHighKey()->GetInfo().c_str());
+
+
+  ///////////////////////////////////////////////////////////////////
+  // End of all tests
+  ///////////////////////////////////////////////////////////////////
+
```
malin1993ml: Tests look good.

src/include/planner/hybrid_scan_plan.h
```diff
@@ -79,6 +85,7 @@ class HybridScanPlan : public AbstractScan {
 
   std::shared_ptr<index::Index> index_;
 
+  index::IndexScanPredicate index_predicate;
```
malin1993ml: underscore

src/include/planner/index_scan_plan.h
```diff
@@ -35,37 +37,75 @@ namespace planner {
...
+
+    // The index object for scanning
+    //
+    // NOTE: For hybrid scan plans, even if there is no index required
+    // for a scan, an empty scan descriptor will still be passed in as
+    // argument. This is a bad design but currently we have to live with it
+    // In order to prevent the scan predicate optimizer from trying to
+    // optimizing the index scan while the index pointer is not valid
+    // this should be set to 0 for an empty initialization
+    std::shared_ptr<index::Index> index_obj;
```
malin1993ml: underscore for this and below

src/include/planner/index_scan_plan.h
```diff
@@ -118,18 +162,37 @@ class IndexScanPlan : public AbstractScan {
...
+  // also the lifetime of the IndexScanPlan object
   std::vector<Value> values_;
 
   const std::vector<expression::AbstractExpression *> runtime_keys_;
+  
+  // Currently we just support single conjunction predicate
+  //
+  // In the future this might be extended into an array of conjunctive
+  // predicates connected by disjunction
+  index::IndexScanPredicate index_predicate;
```
malin1993ml: underscore

src/catalog/schema.cpp
wangziqi2013: No it's not. Since the validator forbids any occurrence of malloc() so I changed to ma1(one)1(one)0(zero)c(cent)

src/catalog/schema.cpp
wangziqi2013: Done

src/include/common/value.h
wangziqi2013: Done

src/include/index/index_util.h
wangziqi2013: Removed it

src/include/common/value_factory.h
wangziqi2013: If you mean checking whether pool is nullptr -  it is done inside the Value object and is always checked

src/include/index/scan_optimizer.h
wangziqi2013: Done

src/include/index/scan_optimizer.h
wangziqi2013: Done

src/include/planner/hybrid_scan_plan.h
wangziqi2013: Done

src/include/planner/index_scan_plan.h
wangziqi2013: Since there is an in-lined function that uses the class defined inside scan_optimizer it is hard to remove

src/include/index/scan_optimizer.h
wangziqi2013: Done

src/include/index/scan_optimizer.h
wangziqi2013: Since we expect in the future that the predicate would consist of CNF formulas, class IndexScanPredicate represents the CNF formula itself, while class ConjunctionScanPredicate represents one term inside the CNF (i.e. conditions that are connected by implicit AND). Because now we only have the latter but not the full CNF representation, as you could see class IndexScanPredicate is primitive and is left deliberately for future extension

src/planner/index_scan_plan.cpp
wangziqi2013: Done (it was there before I changed it but I removed it)

PR #207, opened 2016-08-17
support threadpool

yingjunwu:


src/include/common/init.h
```diff
@@ -13,8 +13,12 @@
 
 #pragma once
 
+#include "common/thread_pool.h"
```
jarulraj: Can you use a forward declaration here ?

src/include/common/thread_pool.h
```diff
@@ -2,9 +2,9 @@
 //
 //                         Peloton
 //
-// thread_pool.h
+// worker_thread_pool.h
```
jarulraj: Name mismatch

test/common/thread_pool_test.cpp
```diff
@@ -25,11 +25,25 @@ class ThreadPoolTests : public PelotonTest {};
...
+  int var2 = 2;
+  int var3 = 3;
+  int var4 = 4;
+  thread_pool.SubmitTask([](int *var){*var = *var + *var;}, &var1);
+  thread_pool.SubmitTask([](int *var){*var = *var - *var;}, &var2);
+  thread_pool.SubmitTask([](int *var){*var = *var * *var;}, &var3);
+  thread_pool.SubmitTask([](int *var){*var = *var / *var;}, &var4);
+
+
+  std::this_thread::sleep_for(std::chrono::seconds(1));
```
jarulraj: Can you reduce it to a few ms ?

PR #213, opened 2016-08-22
reduce memory allocation

yingjunwu:
currently, Peloton suffers from high memory allocation overhead when performing update operations. The key reason is that `update_executor` will always create a temporal copy for holding results obtained from tuple projection. This PR avoids creating such copy by directly performing in-place update on tuple versions. Every time the DBMS updates a tuple, it will (1) acquire a version slot from data table; (2) perform projection; (3) install the version back to data table.

btw, our itempointer pointer design is not the key factor for high memory allocation overhead. In fact, such allocation is cheap.

### DIFF
--- a/src/executor/logical_tile.cpp
+++ b/src/executor/logical_tile.cpp
@@ -179,6 +179,13 @@ Value LogicalTile::GetValue(oid_t tuple_id, oid_t column_id) {
   }
 }
 
+// this function is designed for overriding pure virtual function.
+void LogicalTile::SetValue(Value &value UNUSED_ATTRIBUTE, 
+                           oid_t tuple_id UNUSED_ATTRIBUTE, 
+                           oid_t column_id UNUSED_ATTRIBUTE) {
+  PL_ASSERT(false);
+}
+
 /**
  * @brief Returns the number of visible tuples in this logical tile.
  *
--- a/src/executor/update_executor.cpp
+++ b/src/executor/update_executor.cpp
@@ -132,26 +132,35 @@ bool UpdateExecutor::DExecute() {
         }
         // if it is the latest version and not locked by other threads, then
         // insert a new version.
-        std::unique_ptr<storage::Tuple> new_tuple(
-            new storage::Tuple(target_table_->GetSchema(), true));
 
-        // Make a copy of the original tuple and allocate a new tuple
+        // acquire a version slot from the table.
+        ItemPointer new_location = target_table_->AcquireVersion();
+
+        auto &manager = catalog::Manager::GetInstance();
+        auto new_tile_group = manager.GetTileGroup(new_location.block);
+
+        expression::ContainerTuple<storage::TileGroup> new_tuple(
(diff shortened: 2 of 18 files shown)


src/executor/logical_tile.cpp
```diff
@@ -179,6 +179,13 @@ Value LogicalTile::GetValue(oid_t tuple_id, oid_t column_id) {
   }
 }
 
+// this function is designed for overriding pure virtual function.
+void LogicalTile::SetValue(Value &value UNUSED_ATTRIBUTE,
```
jarulraj: What's the purpose of this function ?

src/executor/update_executor.cpp
```diff
@@ -132,26 +125,35 @@ bool UpdateExecutor::DExecute() {
...
+        // this triggers in-place update, and we do not need to allocate another version.
+        project_info_->Evaluate(&new_tuple, &old_tuple, nullptr,
                                 executor_context_);
 
+        // get indirection.
         ItemPointer *indirection = tile_group_header->GetIndirection(old_location.offset);
-        // finally insert updated tuple into the table
-        ItemPointer new_location = target_table_->InsertVersion(new_tuple.get(), &(project_info_->GetTargetList()), indirection);
+        // finally install new version into the table
+        bool ret = target_table_->InstallVersion(&new_tuple, &(project_info_->GetTargetList()), indirection);
```
jarulraj: Can you reduce number of functions to install versions ?

src/include/expression/container_tuple.h
```diff
@@ -175,6 +183,12 @@ class ContainerTuple<std::vector<Value>> : public AbstractTuple {
 
     return (*container_)[column_id];
   }
+  
+  /** @brief Set the value at the given column id. */
+  // for override pure virtual function only.
+  void SetValue(oid_t column_id UNUSED_ATTRIBUTE, Value &value UNUSED_ATTRIBUTE) override {
```
jarulraj: Why is this function added ?

src/include/storage/data_table.h
```diff
@@ -260,7 +267,11 @@ class DataTable : public AbstractTable {
   // INDEX HELPERS
   //===--------------------------------------------------------------------===//
 
-  bool InsertInSecondaryIndexes(const storage::Tuple *tuple, 
+  // bool InsertInSecondaryIndexes(const storage::Tuple *tuple,
```
jarulraj: Why is this commented out ?

src/storage/data_table.cpp
```diff
@@ -362,19 +380,85 @@ bool DataTable::InsertInIndexes(const storage::Tuple *tuple,
   return true;
 }
 
-bool DataTable::InsertInSecondaryIndexes(const storage::Tuple *tuple,
+// bool DataTable::InsertInSecondaryIndexes(const storage::Tuple *tuple,
```
jarulraj: Why is this commented out ?

Issue #238, opened 2016-09-12
Master broken due to incorrect path of jdbc_validator.py


malin1993ml:
The current master brunch is failing because the directory was changed to "build" when running the tests, but the command that invokes jdbc_validator is:
"python ./script/validators/jdbc_validator.py"
which assumes root directory.

@sid1607 I believe that it's easy for you to fix this?

PR #240, opened 2016-09-13
Modified jdbc validator path in travis

sid1607:
Fix for #238


sid1607:
Latest patch will just run the basic jdbc tests. Once the existing issues with JDBC are resolved.

Issue #241, opened 2016-09-13
JDBC segfaults while doing an index_scan


sid1607:
Appears while using test_jdbc.sh

stacktrace 

`#0  0x00007ffff786912c in peloton::common::Type::GetTypeId (this=0x7fff00000005) at /home/siddharth/git/peloton/src/common/type.cpp:203
#1  0x00007ffff76bfd08 in peloton::common::Value::GetTypeId (this=0x7ffff1c1aaa0) at /home/siddharth/git/peloton/src/include/common/value.h:67
#2  0x00007ffff7771ba1 in peloton::expression::ConstantValueExpression::ConstantValueExpression (this=0x7ffff1c2fb00, value=...) at /home/siddharth/git/peloton/src/include/expression/constant_value_expression.h:32
#3  0x00007ffff7771cb2 in peloton::expression::ConstantValueExpression::Copy (this=0x7ffff1c2f800) at /home/siddharth/git/peloton/src/include/expression/constant_value_expression.h:45
#4  0x00007ffff7792b0f in peloton::expression::ComparisonExpression::Copy (this=0x7ffff1c2f980) at /home/siddharth/git/peloton/src/include/expression/comparison_expression.h:65
#5  0x00007ffff7783eed in peloton::planner::DeletePlan::DeletePlan (this=0x7ffff1cbe100, delete_statemenet=0x7ffff1c1aaa0) at /home/siddharth/git/peloton/src/planner/delete_plan.cpp:45
#6  0x00007ffff78cc762 in peloton::optimizer::SimpleOptimizer::BuildPelotonPlanTree (parse_tree=std::unique_ptr<peloton::parser::SQLStatementList> containing 0x7ffff1c1c180) at /home/siddharth/git/peloton/src/optimizer/simple_optimizer.cpp:385
#7  0x00007ffff792a4ea in peloton::tcop::TrafficCop::PrepareStatement (this=0x7ffff7dd9b28 <peloton::tcop::TrafficCop::GetInstance()::traffic_cop>, statement_name="", query_string="DELETE FROM A WHERE id = 1", error_message="") at /home/siddharth/git/peloton/src/tcop/tcop.cpp:108
#8  0x00007ffff78a04e7 in peloton::wire::PacketManager::ExecParseMessage (this=0x7ffff4448640, pkt=0x7ffff2bfb470, responses=std::vector of length 0, capacity 4) at /home/siddharth/git/peloton/src/networking/protocol.cpp:306
[...]

PR #245, opened 2016-09-15
[#241] Fix the copying of values when binding IndexScanPlan.

malin1993ml:
The values of a IndexScanPlan are copied when binding a new query. That was implicitly done when copying a vector with value objects.

Now since we're using value pointers, we need to do deep copy manually.


malin1993ml:
I can successfully run JDBC batch tests on my laptop, but they failed on Jenkins. So I commented them out for now. The same problem is reported by @sid1607 as well.

sid1607:
I'll try to look into why the batch tests fail on Jenkins

PR #253, opened 2016-09-16
revive logging benchmark

yingjunwu:
The logging benchmark was dead because of the refactor of tpcc and ycsb benchmarks. now it works again.

### DIFF
--- a/src/main/logger/logger_workload.cpp
+++ b/src/main/logger/logger_workload.cpp
@@ -25,6 +25,7 @@
 #include "common/logger.h"
 #include "common/timer.h"
 #include "logging/log_manager.h"
+#include "storage/storage_manager.h"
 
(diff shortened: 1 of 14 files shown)


src/main/logger/logger_workload.cpp
```diff
@@ -323,20 +324,15 @@ bool PrepareLogFile() {
 
   // Pick metrics based on benchmark type
   double throughput = 0;
-  double latency = 0;
```
MattPerron: Is there a good reason to remove latency experiments?

yingjunwu: no good reason. simply because I haven't started implementing latency measurement in tpcc and ycsb benchmark. will do it in the future.

Issue #254, opened 2016-09-16
Stat tables and per query stat


eric-haibin-lin:
- Currently statistics are stored in the aggregator object. Moving the stat to a database table makes it easier to manage these data, with the existing functionality peloton supports.
- Recording per query stat, including the number of tuples for read/update/delete/insert, and the latency of query.

Issue #246, opened 2016-09-15
Add support for database selection when selecting table


malin1993ml:
Currently Peloton supports for creating different databases, but doesn't support for accessing databases other than the DEFAULT_DB. All coming queries are assumed to access the DEFAULT_DB.

To handle this, we need to add database token to the parser when handling selection of tables. And then pass this token to wherever catalog is accessed.

MattPerron:
If this is complete, please close the issue.

PR #256, opened 2016-09-16
Stat PR #254

eric-haibin-lin:
Description: 
#254 

Todo:
* #246
* Add stat end to end test to JDBC validator script  

Commit messages in the past didn't refer to the issue number. I'll include the issue number for future commits.


src/statistics/backend_stats_context.cpp
```diff
@@ -24,13 +24,19 @@
 namespace peloton {
 namespace stats {
 
-BackendStatsContext& BackendStatsContext::GetInstance() {
+CuckooMap<std::thread::id, std::shared_ptr<BackendStatsContext>>
+    BackendStatsContext::stats_context_map_;
```
eric-haibin-lin: It was declared inside the class, here I am define it. For const static variable, you can both declare and define inside the class, but here it's not a constant.

src/include/statistics/backend_stats_context.h
```diff
@@ -109,15 +146,24 @@ class BackendStatsContext {
   std::unordered_map<TableMetric::TableKey, std::unique_ptr<TableMetric>>
       table_metrics_{};
 
-  // Index metrics
+  // TODO stl map is not thread safe. use cuckoomap instead
+  // CuckooMap<IndexMetric::IndexKey,std::shared_ptr<IndexMetric>>
+  //      index_metrics_{};
```
eric-haibin-lin: Not yet. The CuckooMap container doesn't have begin() and end() iterators, so we can't use it directly to replace stl map, since during aggregation we need to iterate over all elements in the map. We need to come up with some walk around (e.g. use a set to store all the keys, or use a RW lock, which slow down txn performance)

src/catalog/catalog.cpp
```diff
@@ -357,7 +389,8 @@ storage::Database *Catalog::GetDatabaseWithOid(const oid_t db_oid) const {
 storage::Database *Catalog::GetDatabaseWithName(const std::string database_name)
     const {
   for (auto database : databases_) {
-    if (database->GetDBName() == database_name) return database;
+    if (database != nullptr && database->GetDBName() == database_name)
```
eric-haibin-lin: This usually doesn't happen.. It's todo with catalog destruction. Previously I created catalog_db and default_db on catalog bootstrap, upon destruction if catalog_db is deleted first, then GetDatabaseWithName will segfault when looking up default_db.

Issue #236, opened 2016-09-10
LOG_LEVEL_TRACE broken due to new type/value system


malin1993ml:
Because the new type/value system has a slightly different interface than then old one, some LOG_TRACE statements that use the old system interface are broken.

This will not be checked through either Travis or Jenkins. I'm thinking whether it's a good idea to check whether the system can be built successfully with LOG_TRACE on Jenkins.

apavlo:
Maybe we should enable TRACE on Jenkins.

malin1993ml:
I think we could build the system twice on Jenkins. First try to build with TRACE to see whether it succeeds. The second time build and run "make check" in release mode to avoid giant log output.

ranxian:
@malin1993ml 

Is **broken** means compile error or something else? For me it is compile errors.

malin1993ml:
Yes. It is compile error.

apavlo:
I will update Jenkins tonight to do this.

apavlo:
Jenkins now compiles both `TRACE` and `RELEASE` targets. But it only builds the test cases with `RELEASE`.

@jarulraj Is there a way to build the test cases without having to actually run them?

sid1607:
@jessesleeping Could you look into this, so that the system can be prepared for the first public release.

jessesleeping:
Ok, I will look at this issue.

PR #257, opened 2016-09-16
Fix LOG_LEVEL_TRACE

jessesleeping:
#236

Fix the LOG_LEVEL_TRACE. Now we can compile with "-DLOG_LEVEL=LOG_LEVEL_TRACE".


sid1607:
How did Peloton compile before, despite having all these pointer dereferences wrong?

apavlo:
We weren't building it with LOG_LEVEL_TRACE on Jenkins. I thought that I
had enabled it this week but I guess not.

jessesleeping:
Because when you set a lower log level, the LOG_TRACE stuff won't be compiled.

PR #260, opened 2016-09-17
#246 Support database selection when selecting tables

eric-haibin-lin:
Description: #246 

Changes: 
Modified parser, planner and executor to get target table based on the optional database name in the query. Such as select * from catalog_db.table1;

It would be great if someone familiar with parser and planner can review this code before it's merged in.


src/include/parser/statement_drop.h
```diff
@@ -32,20 +33,29 @@ struct DropStatement : SQLStatement {
...
-        type(type),
-        name(NULL),
-        table_name(NULL),
-        missing(false) {}
+      : SQLStatement(STATEMENT_TYPE_DROP), type(type), missing(false) {}
 
   virtual ~DropStatement() {
-    free (name);
-    free (table_name);
+    free(database_name);
```
MattPerron: why use free and not delete here?

src/include/parser/statement_drop.h
```diff
@@ -32,20 +33,29 @@ struct DropStatement : SQLStatement {
...
+    free(index_name);
+    free(prep_stmt);
+    delete table_name;
+  }
+
+  inline std::string GetTableName() { return std::string(table_name->name); }
```
MattPerron: should we check for null here?

src/parser/table_ref.cpp
```diff
@@ -17,9 +17,9 @@ namespace peloton {
 namespace parser {
 
 TableRef::~TableRef() {
-  delete name;
-  free (alias);
-  free (schema);
+  delete table_name;
+  free(alias);
```
MattPerron: why free and not delete?

src/include/parser/statement_drop.h
eric-haibin-lin: The value of table_name is always initialized with parser, if we check null value, the more interesting question is what value do we return.. empty string / default table name?

src/include/expression/parser_expression.h
```diff
@@ -26,54 +26,58 @@ namespace expression {
...
-  : AbstractExpression(type) {
+  ParserExpression(ExpressionType type, char* name_, char* column_or_database_)
+      : AbstractExpression(type) {
     name = name_;
-    column = column_;
+    if (type == peloton::EXPRESSION_TYPE_COLUMN_REF) {
+      column = column_or_database_;
+    } else if (type == peloton::EXPRESSION_TYPE_TABLE_REF) {
+      database = column_or_database_;
+    }
```
malin1993ml: I can understand the logic here after reading other parts of the code, but is it possible to separate this to two different functions? This might lead to confusion.

eric-haibin-lin: Squashed recent changes into 2 commits.

Issue #265, opened 2016-09-20
Hanging when handling partial writes in socket_base.cpp


MattPerron:
The code to handle the case when `write` does not write a whole buffer causes a hang at https://github.com/cmu-db/peloton/blob/master/src/wire/socket_base.cpp#L185

MattPerron:
Also, even though the sockets are non-blocking, the read and write functions in this file block until a buffer is written.

sid1607:
Yes, it doesn't seem like this code uses all the non-blocking facilities that libevent has to offer. Any steps on how to reproduce this hang condition?

MattPerron:
When peloton returns large amounts of data (around 8MB) this happens almost every time.
Also, I already have a solution (to it hanging), I created this issue for bookkeeping purposes.

MattPerron:
Fixed with #266

PR #266, opened 2016-09-20
fix for hanging when a buffer only partially writes to the socket

MattPerron:
[#265]

### DIFF
--- a/src/wire/socket_base.cpp
+++ b/src/wire/socket_base.cpp
@@ -126,12 +126,12 @@ bool SocketManager<B>::RefillReadBuffer() {
 
 template <typename B>
 bool SocketManager<B>::FlushWriteBuffer() {
-  fd_set rset;
+  fd_set wset;
   ssize_t written_bytes = 0;
   wbuf.buf_ptr = 0;
   // still outstanding bytes
-  while ((int)wbuf.buf_size - written_bytes > 0) {
-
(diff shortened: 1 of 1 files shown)


src/wire/socket_base.cpp
```diff
@@ -126,12 +126,12 @@ bool SocketManager<B>::RefillReadBuffer() {
...
 bool SocketManager<B>::FlushWriteBuffer() {
-  fd_set rset;
+  fd_set wset;
   ssize_t written_bytes = 0;
   wbuf.buf_ptr = 0;
   // still outstanding bytes
-  while ((int)wbuf.buf_size - written_bytes > 0) {
-
+  while ((int)wbuf.buf_size > 0) {
+   written_bytes=0;
```
sid1607: How does this fix the problem during large data transfers?

MattPerron: Since we never reset written bytes to zero, if we write less than `wbuf.buf_size` bytes during the write, we will never enter the inner loop again (although we update `wbuf.buf_size` and `wbuf.buf_ptr` on 213) because written_bytes is never reset to zero. So we keep updating `wbuf.buf_size` and `wbuf.buf_ptr` even though no additional data has been written after the first write.

Resetting written_bytes to zero at the beginning of the outer loop ensures that we do not reuse the value from a previous iteration of the outer loop.

Issue #224, opened 2016-08-31
Add support for "SELECT...FOR UPDATE"


apavlo:
The `ReadModifyWrite` transaction in YCSB uses the `FOR UPDATE` feature in the first SELECT query.

https://github.com/oltpbenchmark/oltpbench/blob/master/src/com/oltpbenchmark/benchmarks/ycsb/procedures/ReadModifyWriteRecord.java#L30

We currently support this in the new parser and are able to set the flag properly in the query plan, but we need to add code to take the write lock in the `IndexScanExecutor` and `SeqScanExecutor` components.

@jarulraj - You may need to add utility code to make it easier to grab the write lock for a tuple.

apavlo:
The basic idea of how this works is that there will be a boolean flag in the `IndexScanPlan` and `SeqScanPlan` plan node objects. If a tuple satisfies the predicate during the scan (and will be added in executor's output result), then you will want to acquire a write-lock. 

https://github.com/cmu-db/peloton/blob/master/src/executor/seq_scan_executor.cpp#L148
https://github.com/cmu-db/peloton/blob/master/src/executor/seq_scan_executor.cpp#L161

@yingjunwu  Would it be easier if we modify `TransactionManager.PerformRead()` to take in a boolean flag that tells the manager to acquire the lock?

@saifalharthi Where is the "FOR UPDATE" flag in the plan nodes? I only see it in the parser code.

ranxian:
@apavlo 
There is actually an `AcquireOwnership()`([link](https://github.com/cmu-db/peloton/blob/master/src/concurrency/timestamp_ordering_transaction_manager.cpp#L292)) method that takes a write lock. We can implement as you said to maintain a boolean flag and call this method

yingjunwu:
@apavlo I think current API is sufficient enough to implement `FOR UPDATE`. As @ranxian said, there's several APIs for acquiring read or write locks. `IsOwnable()` (https://github.com/cmu-db/peloton/blob/master/src/executor/update_executor.cpp#L112) checks whether the current transaction owns the write lock, and `AcquireOwnership()` (https://github.com/cmu-db/peloton/blob/master/src/executor/update_executor.cpp#L118) attempts to acquire the write lock. `YieldOwnership()` (https://github.com/cmu-db/peloton/blob/master/src/executor/update_executor.cpp#L157) helps release the write lock if the version insertion to indexes fails (e.g., violation of foreign key or unique index constraints).

jarulraj:
[...]

PR #268, opened 2016-09-20
Support select for update #224

aquafemi:
* Added for update executor which acquires ownership of ownable tuples for the current transaction
* Returns true if success, false otherwise
* Added test case for a transaction trying to get ownership of tuple another transaction currently owns


test/executor/for_update_test.cpp
```diff
@@ -0,0 +1,123 @@
...
+
+//===--------------------------------------------------------------------===//
+// For Update Tests
+//===--------------------------------------------------------------------===//
+
+class ForUpdateTests : public PelotonTest {};
+
+TEST_F(ForUpdateTests, Updating) {
+
+//  LOG_INFO("Bootstrapping...");
```
yingjunwu: I think unused code should be removed.

yingjunwu:
I have doubts about the implementation of the `select_for_update` executor. I think creating a new executor is fine,  but the current implementation is, at least for me, questionable. We can consider the `select_for_update` executor as a special case of the `index_scan` (or `seq_scan`) executor. The difference is that, instead of acquiring a read lock on accessing a qualified tuple, we directly acquire a write lock. Hence, you may want to implement the `select_for_update` executor based on the logic of `index_scan` or `seq_scan` executor instead of the `update` executor. You could also check the microbenchmark (https://github.com/cmu-db/peloton/blob/master/src/main/ycsb/ycsb_mixed.cpp#L162) to see how an executor works in the system.

Issue #277, opened 2016-09-21
Revive the log manager


jessesleeping:
Currently the log manager is disabled in the transaction manager. Also, since we are using our own catalog table now, the recovery logic should also handle the catalog table.

jarulraj:
What's the current status of the log manager @jessesleeping  @yingjunwu @MattPerron ? Do we have test cases for the log manager ?

apavlo:
@jessesleeping is going to port their latest logging + epoch + GC code from their paper branch to the master next week. Then @YingjunWu will work on the next paper in January. 

We want their optimized logging code in order to do real TPC-C benchmarking.

Issue #222, opened 2016-08-31
Clean-up + Revive WAL Component


apavlo:
The WAL code was abused this summer. Nobody is sure whether it is still working (but probably not). Since @MattPerron wrote a lot of it last semester, he should go through and tidy things up to start the semester.

I also have a branch that I started with some fixes that we should eventually merge in. Not all of these items have been implemented:

* Add BEGIN/END markers to each log record
* Record a checksum for each block of log records
* Move all the serialization code out of the worker thread

We are also going to need to switch the serialization code to use the original H-Store SerializeIO class. We will need to coordinate on this.

yingjunwu:
Logging subsystem has already been revived. New logging subsystem is on the way.

PR #281, opened 2016-09-22
revive logging component

yingjunwu:
see #222 

This PR revives the logging component in Peloton. We temporarily do not support log recovery, as the current version of Peloton does not persist catalogs (see #277). This problem will be solved in the next PR.


src/concurrency/timestamp_ordering_transaction_manager.cpp
```diff
@@ -93,6 +93,10 @@ Transaction *TimestampOrderingTransactionManager::BeginTransaction() {
   auto eid = EpochManagerFactory::GetInstance().EnterEpoch(begin_cid);
   txn->SetEpochId(eid);
 
+  auto &log_manager = logging::LogManager::GetInstance();
+  log_manager.PrepareLogging();
```
MattPerron: @jessesleeping 
Don't you need to call this before generating a begin_commit id?

src/concurrency/timestamp_ordering_transaction_manager.cpp
```diff
@@ -93,6 +93,10 @@ Transaction *TimestampOrderingTransactionManager::BeginTransaction() {
...
+  log_manager.LogBeginTransaction(begin_cid);
```
MattPerron: If we write this record here, what do we do if we abort the transaction? Will we still have to clean up the log somehow?

src/concurrency/timestamp_ordering_transaction_manager.cpp
```diff
@@ -104,13 +108,18 @@ Transaction *TimestampOrderingTransactionManager::BeginTransaction() {
 
 void TimestampOrderingTransactionManager::EndTransaction(Transaction *current_txn) {
   EpochManagerFactory::GetInstance().ExitEpoch(current_txn->GetEpochId());
+  auto &log_manager = logging::LogManager::GetInstance();
```
MattPerron: Is EndTransaction only for read only transactions? and if so, do we need to log anything? 
I think we can just call `log_manager.DoneLogging()`

src/concurrency/timestamp_ordering_transaction_manager.cpp
```diff
@@ -777,6 +796,8 @@ Result TimestampOrderingTransactionManager::CommitTransaction(
 
         // add to gc set.
         gc_set->operator[](tile_group_id)[tuple_slot] = RW_TYPE_INS_DEL;
+
+        // no log is needed for this case
```
MattPerron: When do you write the commit record to the log?

jessesleeping: At EndTransaction().

src/concurrency/timestamp_ordering_transaction_manager.cpp
jessesleeping: No, it's for all non-readonly transactions.

src/concurrency/timestamp_ordering_transaction_manager.cpp
jessesleeping: If a transaction abort, the only log record it will generate it's a begin transaction record. It won't log down any tuple record because we only do that when committing a transaction.

src/concurrency/timestamp_ordering_transaction_manager.cpp
jessesleeping: I don't think there is any different though... we can talk about this offline...

yingjunwu:
@MattPerron Yes I know this problem. I did not coalesce these commits because our local commit history of this branch is in a mess due to the previous squash, and it's hard for us to rebase to the master branch. We will discard this branch and sync with the master branch after this PR. Sorry for the inconvenience incurred.

PR #286, opened 2016-09-25
Support index scan in Update and Delete statement

hzxa21:
Support index scan in update and delete. 

Pass all tests but may cause potential memory leak in
https://github.com/cmu-db/peloton/blob/master/src/optimizer/simple_optimizer.cpp#L664

I have traced this problem and found that it might caused by the previous codes in the simple optimizer. Could Ran help to check? @ranxian

### DIFF
--- a/src/planner/delete_plan.cpp
+++ b/src/planner/delete_plan.cpp
@@ -17,6 +17,11 @@
 #include "expression/expression_util.h"
 #include "catalog/catalog.h"
 
+#include "planner/abstract_plan.h"
+#include "planner/abstract_scan_plan.h"
+#include "planner/index_scan_plan.h"
+#include "planner/seq_scan_plan.h"
+
 namespace peloton {
 
 namespace expression {
@@ -29,8 +34,8 @@ DeletePlan::DeletePlan(storage::DataTable *table, bool truncate)
   LOG_TRACE("Creating a Delete Plan");
 }
 
-DeletePlan::DeletePlan(parser::DeleteStatement *delete_statemenet) {
-
+// Initializes the delete plan and retrieves coloum ids.
+void DeletePlan::Init(parser::DeleteStatement *delete_statemenet, std::vector<oid_t>& columns) {
   LOG_TRACE("Creating a Delete Plan");
   table_name_ = delete_statemenet->GetTableName();
   auto database_name = delete_statemenet->GetDatabaseName();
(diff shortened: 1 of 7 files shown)


src/planner/delete_plan.cpp
```diff
@@ -48,12 +53,43 @@ DeletePlan::DeletePlan(parser::DeleteStatement *delete_statemenet) {
     LOG_TRACE("Replacing COLUMN_REF with TupleValueExpressions");
     ReplaceColumnExpressions(target_table_->GetSchema(), expr_);
   }
-  std::vector<oid_t> column_ids = {};
+  columns = {};
```
ranxian: This argument is passed in and set to empty without using it, what's the point?

src/planner/update_plan.cpp
```diff
@@ -32,7 +39,7 @@ UpdatePlan::UpdatePlan(storage::DataTable *table,
   LOG_TRACE("Creating an Update Plan");
 }
 
-UpdatePlan::UpdatePlan(parser::UpdateStatement *parse_tree) {
+void UpdatePlan::Init(parser::UpdateStatement *parse_tree, std::vector<oid_t>& columns) {
```
ranxian: What is `columns`?

ranxian: The name `Init` is ambiguous. This code is actually building the update plan, and the constructor code is initializing the scan plan. Something like `BuildUpdatePlan` sounds better?

test/optimizer/rule_test.cpp
```diff
@@ -54,5 +69,176 @@ TEST_F(RuleTests, SimpleRuleApplyTest) {
   EXPECT_EQ(outputs.size(), 1);
 }
 
+// Test whether update stament will use index scan plan
+TEST_F(RuleTests, UpdateDelWithIndexScanTest) {
```
ranxian: Please separate it into two tests for update and delete respectively.

ranxian: This test suite is for rule.cpp|h, please add a test suite purely for simple_optimizer and move this test there.

ranxian: This test relies on a functional parser and planner, it is more like a integrate test instead of a unit test. When the test fails it would be hard to determine which part is wrong. To make it more straightforward and clearer, you can consider following this pattern:

1. Prepare the test table using code instead of SQL
2. Construct the test plan with code, see rule_test.cpp#SimpleRuleApplyTest
3. Feed the constructed plan directly to simple optimizer
4. Check the output plan of simple optimizer

For how to create and load a table using code, please have a look at `executor_tests_util.cpp` (and yes it would be great if we also add ab utility class for optimizer, given that we don't actually have a test for optimizer, it will be a good start)

test/optimizer/rule_test.cpp
```diff
@@ -54,5 +69,176 @@ TEST_F(RuleTests, SimpleRuleApplyTest) {
...
+  LOG_INFO("Updating a tuple...");
+  LOG_INFO(
+      "Query: UPDATE department_table SET dept_name = 'CS' WHERE student_id = 52");
+  statement.reset(new Statement(
+      "UPDATE",
+      "UPDATE department_table SET dept_name = 'CS' WHERE student_id = 52"));
+  LOG_INFO("Building parse tree...");
+  update_stmt = peloton_parser.BuildParseTree(
+      "UPDATE department_table SET dept_name = 'CS' WHERE student_id = 52");
+  LOG_INFO("Building parse tree completed!");
```
ranxian: Too many debug info, either use LOG_DEBUG or remove some of them

test/optimizer/rule_test.cpp
```diff
@@ -54,5 +69,176 @@ TEST_F(RuleTests, SimpleRuleApplyTest) {
...
+  // Check scan plan
+  ASSERT_FALSE(plan == nullptr);
+  if (plan->GetPlanNodeType() == PLAN_NODE_TYPE_UPDATE) {
+    auto& scanPlan = plan->GetChildren().front();
+    EXPECT_EQ(scanPlan->GetPlanNodeType(), PLAN_NODE_TYPE_INDEXSCAN);
+  }
+  LOG_INFO("Executing plan...");
+  result_format =
+      std::move(std::vector<int>(statement->GetTupleDescriptor().size(), 0));
+  status = bridge::PlanExecutor::ExecutePlan(statement->GetPlanTree().get(),
```
ranxian: Do not need to execute the plan if we are only testing the optimizer. Limit the scope of the test especially when it is unit test.

apavlo:
This is awesome. Nice work.

@ranxian Can you help him find the memory leak? Feel free to ask @malin1993ml  for help too.

ranxian:
Overall very nice work, and a good step towards a robust optimizer.

@apavlo Sure I will work with him to make it done
@hzxa21 Let's meet offline to chat these issues.

ranxian:
#289

PR #289, opened 2016-09-28
Support index scan in Update and Delete statement 

ranxian:
A replacement of https://github.com/cmu-db/peloton/pull/286 with memory leak fixed


ranxian:
Now all memory leaks are fixed, and the optimizer tests are passed. 

The memory leaks are because of the new type system, where we return a raw pointer to a newly allocated value when we call `Value::Copy()`. Thus, We need to be very careful to free the memory.

However the `jdbc_validate.py` fails. I have looked at the log message, it is because the `index_read` counter does not match. I guess this is because we previously not support index scan in delete and update, so that the counter is not updated. Here is the suspicious code in `PelotonTest.java`:

```java
    // Test Update
    UpdateByIndex(1);
    txn_committed++;
    table_update++;
    table_read++;
    // TODO this test doesn't examine index since update by index scan
    // is not supported yet
    //index_read++;

    stmt.execute(INDEXSCAN);
    table_read++;
    txn_committed++;
    index_read++;

    // TODO this test doesn't examine index since delete by index
    // scan is not supported
    stmt.execute(DELETE_A);
    table_read++;
    table_delete++;
    txn_committed++;
    // index_read++;
```

ranxian:
see 0272856, the JDBC test is passing now

apavlo:
This looks good. I think we're good to go, right?

ranxian:
@apavlo yes, this PR is ready to be merged (finally)

@malin1993ml @MattPerron @sid1607

PR #293, opened 2016-10-04
fix bug in tpcc benchmark.

yingjunwu:
this PR fixed a double-free bug in hard-coded tpcc benchmark.

### DIFF
--- a/src/main/tpcc/tpcc_workload.cpp
+++ b/src/main/tpcc/tpcc_workload.cpp
@@ -83,9 +83,9 @@ namespace tpcc {
 /////////////////////////////////////////////////////////
 
 #define STOCK_LEVEL_RATIO     0.04
-#define ORDER_STATUS_RATIO    0.04
-#define PAYMENT_RATIO         0.43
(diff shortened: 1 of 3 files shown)


src/main/tpcc/tpcc_workload.cpp
```diff
@@ -83,9 +83,9 @@ namespace tpcc {
 /////////////////////////////////////////////////////////
 
 #define STOCK_LEVEL_RATIO     0.04
-#define ORDER_STATUS_RATIO    0.04
-#define PAYMENT_RATIO         0.43
-#define NEW_ORDER_RATIO       0.45
+#define ORDER_STATUS_RATIO    0.0
+#define PAYMENT_RATIO         0.48
+#define NEW_ORDER_RATIO       0.48
```
MattPerron: Why change these ratios? Is order status not working?

yingjunwu: You are too smart... and yes there's a type conversion problem in the order status transaction, but since it's not that emergent at this time point, I did not fix that.

PR #294, opened 2016-10-04
#291 Fixed multi-tile group output collection

sid1607:
This diff pulls the output logical tiles even when Execute() status is true. This ensures that all the tiles will be returned for queries that generate output spanning over multiple logical tiles.

### DIFF
--- a/script/testing/jdbc/src/PelotonTest.java
+++ b/script/testing/jdbc/src/PelotonTest.java
@@ -14,7 +14,7 @@ import java.sql.*;
 import org.postgresql.util.*;
 
 public class PelotonTest {
-  private final String url = "jdbc:postgresql://localhost:5432/";
+  private final String url = "jdbc:postgresql://localhost:54321/";
   private final String username = "postgres";
   private final String pass = "postgres";
 
--- a/script/testing/jdbc/src/PelotonTest.java
+++ b/script/testing/jdbc/src/PelotonTest.java
@@ -14,7 +14,7 @@ import java.sql.*;
 import org.postgresql.util.*;
 
 public class PelotonTest {
-  private final String url = "jdbc:postgresql://localhost:54321/";
+  private final String url = "jdbc:postgresql://localhost:5432/";
   private final String username = "postgres";
   private final String pass = "postgres";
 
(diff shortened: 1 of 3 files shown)


src/executor/plan_executor.cpp
```diff
@@ -91,47 +91,41 @@ peloton_status PlanExecutor::ExecutePlan(
...
-            result.push_back(res);
+
+    std::unique_ptr<executor::LogicalTile> logical_tile(
+        executor_tree->GetOutput());
+    // Some executors don't return logical tiles (e.g., Update).
+    if (logical_tile.get() != nullptr) {
+      LOG_TRACE("Final Answer: %s",
+                logical_tile->GetInfo().c_str());  // Printing the answers
+      auto output_schema =
+          logical_tile->GetPhysicalSchema();  // Physical schema of the tile
```
MattPerron: can we make this a unique_ptr so we don't have to delete?

sid1607: Resolved

script/testing/jdbc/src/PelotonTest.java
```diff
@@ -14,7 +14,7 @@
 import org.postgresql.util.*;
 
 public class PelotonTest {
-  private final String url = "jdbc:postgresql://localhost:54321/";
+  private final String url = "jdbc:postgresql://localhost:5432/";
```
eric-haibin-lin: We should have a port number consistent with the shell script
https://github.com/cmu-db/peloton/blob/master/script/validators/jdbc_validator.py#L34

(Usually psql already occupies port 5432 on my local machine which causes the validator to fail so I changed it to 54321, but that's a minor point).

sid1607: I usually shutdown Postgres when I run Peloton. I guess we should create a new issue to fix a non-colliding port number for Peloton, once and for all.

apavlo: Maybe we should start using our own port number

sid1607: @eric-haibin-lin Also, could you add the ../script/testing/jdbc/stats_log file to gitignore?

Issue #205, opened 2016-08-16
Failing on TPC-C with OLTPBench


malin1993ml:
For now, Peloton can successfully run the NewOrder transaction in TPC-C.

For Payment, the transaction fails to execute until customerByNameSQL. The query can be successfully executed on the Peloton side. But the OLTPBench side cannot read the result back:
15:26:21,203 (Worker.java:403) DEBUG - Payment/02 Bad value for type timestamp/date/time: {1} 0 - 22007

For OrderStatus, it fails on the same customerByNameSQL:
16:17:03,244 (Worker.java:403) DEBUG - OrderStatus/03 Bad value for type timestamp/date/time: {1} 0 - 22007

For Delivery, we actually can already run it without any debugging! Surprise:)

For StockLevel, if fails on stockGetCountStockSQL. We need to be able to handle alias:
16:19:06,238 (Worker.java:403) DEBUG - StockLevel/05 The column name STOCK_COUNT was not found in this ResultSet. 0 - 42703

saifalharthi:
I pushed a commit to my fork for supporting aliases. Please pull the changes and test it on the last query.

saifalharthi:
This is done , I pushed the fix for that and OrderStatus runs now and Payment should run as well.

malin1993ml:
The two old issues on TPC-C was fixed by @saifalharthi, thanks!

Now we have some new issues since we can reach more queries after the above fixes.
StockLevel:
The query stockGetDistOrderIdSQL doesn't always retrieve correct the answer from Peloton. It will succeed on the first few queries and the return a random big number after that. Also, we seems to return NULL instead of 0 when there's no tuple satisfying the predicate of stockGetCountStockSQL. That should be fixed.

OrderStatus and Payment:
The transaction keeps retrying the execution for some reason.
19:41:10,583 (Worker.java:375) DEBUG - execute status: RETRY_DIFFERENT

saifalharthi:
I am working on fixing this and pushed a commit to my repo that runs StockLevel successfully ( according to OLTPBecnhmark). However , there are still some problems.

In stockGetDistOrderId the query is correctly executed from Peloton side , yet OLTPBench seems to get and assign a different result. I made sure that the result left peloton (at least reached the function where it sends the results to he client) is correct. So I will try to use wireshark and see to determine if its still a problem from our wire protocol or from OLTPBench.

[...]

PR #297, opened 2016-10-05
Order by fix

malin1993ml:
This PR fixed the wrong answer problem with OrderByPlan mentioned in [#205].

Now we're able to run NewOrder, Payment, OrderStatus, and StockLevel. However, we're now failing on Delivery on:
UPDATE ORDER_LINE SET OL_DELIVERY_D = $1 WHERE OL_O_ID = $2 AND OL_D_ID = $3 AND OL_W_ID = $4
That issue was not revealed because we had a wrong OrderByPlan before, so we selected nothing to deliver. I'll work on fixing that (hopefully) last issue. I checked other transactions in TPC-C. They at least all run to the end and generate results.

### DIFF
--- a/src/include/planner/abstract_plan.h
+++ b/src/include/planner/abstract_plan.h
@@ -12,19 +12,19 @@
 
 #pragma once
 
-#include <iostream>
-#include <memory>
 #include <cstdint>
-#include <vector>
+#include <iostream>
 #include <map>
+#include <memory>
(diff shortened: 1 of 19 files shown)


malin1993ml:
@sid1607 The problem was at binding stage. PSQL sends all queries as raw queries, so the binding logic  would not be invoked.

PR #301, opened 2016-10-06
psql & jdbc error message 

eric-haibin-lin:
In this PR, error messages are propagated to the front end when 
* A query accesses a table/database which doesn't exist 
* It fails to parse the query 
* Query planner is unimplemented (e.g. join queries, select for update)

Now when accessing a table which doesn't exist, an exception will be thrown and possibly propagate up to the front end. Throwing an exception is slow, but this should not happen very often. 

End-to-end test cases are added to `PelotonTest.java`. 

However, it doesn't cover all the unimplemented features yet. It would take a while for me to figure out all features that are not implemented. **If you know some feature not supported off-hand and not covered by my code, please kindly leave a comment.** Thanks! 

@sid1607  Please review what I changed in the wire protocol code. 

TODO:
* also throw exception on columns that don't exist.

### DIFF
--- a/src/catalog/catalog.cpp
+++ b/src/catalog/catalog.cpp
@@ -15,6 +15,8 @@
 #include "catalog/catalog.h"
 #include "catalog/manager.h"
 #include "index/index_factory.h"
+#include "common/exception.h"
+#include "common/macros.h"
 
 namespace peloton {
 namespace catalog {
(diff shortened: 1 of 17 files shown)


src/catalog/catalog.cpp
```diff
@@ -119,37 +121,45 @@ void Catalog::InsertDatabaseIntoCatalogDatabase(oid_t database_id,
...
-    catalog::InsertTuple(
-        databases_[START_OID]->GetTableWithName(TABLE_CATALOG_NAME),
-        std::move(tuple), txn);
-    return Result::RESULT_SUCCESS;
-  } else {
-    LOG_TRACE("Could not find a database with name %s", database_name.c_str());
+    catch (CatalogException &e) {
+      // Table doesn't exist, now create it
+      oid_t database_id = database->GetOid();
+      storage::DataTable *table = storage::TableFactory::GetDataTable(
```
sid1607: Use smart pointer here?

src/catalog/catalog.cpp
```diff
@@ -291,8 +301,9 @@ index::Index *Catalog::GetIndexWithOid(const oid_t database_oid,
 Result Catalog::DropDatabaseWithName(std::string database_name,
                                      concurrency::Transaction *txn) {
   LOG_TRACE("Dropping database %s", database_name.c_str());
-  storage::Database *database = GetDatabaseWithName(database_name);
-  if (database != nullptr) {
+  try {
+    storage::Database *database = GetDatabaseWithName(database_name);
```
sid1607: smart pointer here as well?

src/networking/protocol.cpp
```diff
@@ -304,8 +304,9 @@ void PacketManager::ExecParseMessage(Packet *pkt, ResponseBuffer &responses) {
   statement =
       tcop.PrepareStatement(statement_name, query_string, error_message);
   if (statement.get() == nullptr) {
+    skipped_stmt_ = true;
     SendErrorResponse({{'M', error_message}}, responses);
-    SendReadyForQuery(txn_state_, responses);
```
sid1607: Doesn't a ReadyForQuery packet need to be sent even when a query fails?

PR #303, opened 2016-10-07
Disable stat cuckoomap initialization

eric-haibin-lin:
Callgrind reports that stat cuckoomap constructor & destructor have high cost (even just called once). Removed it temporarily so that we can benchmark tpcc first.


eric-haibin-lin:
@yingjunwu please review this pr so that you can benchmark tpcc without stat. By default stat is completely turned off now.

eric-haibin-lin:
@MattPerron @malin1993ml @sid1607 can you help merge it? Looks like @yingjunwu cannot authorize to merge this PR.

Issue #302, opened 2016-10-07
TPCC fails to insert more than 1 tuples for history table 


eric-haibin-lin:
In tpcc, history table doesn't have any primary keys. After loading all the tables, the history table only contains 1 tuple. All insertion fails after the first one. 

We can reproduce it problem with the following 3 queries:

```
postgres=# create table foo(id integer, year double);
postgres=# insert into foo values(72, 42);
insert into foo values(72, 42) 1
postgres=# insert into foo values(7, 4);
insert into foo values(7, 4) 0
```

The second insertion failed. 

The problem is caused by the default primary index of the table. Currently when a table is created, a primary index is **always created**, even though no primary key is specified. This leads to a primary index with empty primary key schema. 

https://github.com/cmu-db/peloton/blob/master/src/catalog/catalog.cpp#L138

When inserting a tuple to table, updating this primary key index fails after the first tuple is inserted:

https://github.com/cmu-db/peloton/blob/master/src/storage/data_table.cpp#L346

We could disable primary index creation for a table with no primary key specified, so all queries on the table results in seq scans. Not sure if other components are built based the assumption that a table always has a primary key index. 

If tables must have primary index, then for the ones with no primary keys specified, we can create an implicit column of row IDs, which is hidden from the user's perspective and used to create the primary index. Definitely more memory is required for each table in this case.

apavlo:
TPC-C never reads from the HISTORY table. It is append only. I say disable the pkey for now when one is not defined and then see what breaks.

malin1993ml:
Fixed in #307

PR #307, opened 2016-10-12
[#302] Remove primary key index creation when there's no primary key.

malin1993ml:
Fix for #302.


malin1993ml:
@eric-haibin-lin Can you review this fix?

malin1993ml:
@eric-haibin-lin It's done.

PR #308, opened 2016-10-12
Add support for "SELECT ... FOR UPDATE" #224

jessesleeping:
Modify the logic in transaction manager layer to support "select for update". Also add a new test case.

### DIFF
--- a/src/concurrency/timestamp_ordering_transaction_manager.cpp
+++ b/src/concurrency/timestamp_ordering_transaction_manager.cpp
@@ -165,12 +165,15 @@ bool TimestampOrderingTransactionManager::IsOccupied(
   bool invalidated = (current_txn->GetBeginCommitId() >= tuple_end_cid);
 
   // there are exactly two versions that can be owned by a transaction.
-  // unless it is an insertion.
+  // unless it is an insertion/select for update.
   if (own == true) {
     if (tuple_begin_cid == MAX_CID && tuple_end_cid != INVALID_CID) {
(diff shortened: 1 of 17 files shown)


src/concurrency/timestamp_ordering_transaction_manager.cpp
```diff
@@ -357,6 +374,25 @@ bool TimestampOrderingTransactionManager::PerformRead(
...
   auto tile_group_header = tile_group->GetHeader();
 
+  // Check if it's select for update before we check the ownership and modify the
+  // last reader tid
+  if (acquire_ownership == true
+      && IsOwner(current_txn, tile_group_header, tuple_id) == false) {
+    // Acquire ownership if we haven't
+    if (IsOwnable(current_txn, tile_group_header, tuple_id) == false) {
+      // Can not own
+      printf("[JX] Read Own Txn IsOwnerable failed\n");
```
jessesleeping: Sorry I forgot to remove it .....

malin1993ml:
There're two PRs that merged into master today. Can you rebase your change on top of the current master?

Also, can you run TPC-C through OLTPBench to make sure this doesn't break TPC-C before we merge it?

jessesleeping:
Sure, I can do the rebase.

For the TPC-C, currently I can not run it on our master branch. It seems like there is a SQL parse error when creating the table. Also, I think it would be better if @ranxian or @yingjunwu can review my code before we merge it...

malin1993ml:
Okay. Sounds reasonable.

PR #309, opened 2016-10-14
support scalable tile group allocation

yingjunwu:
for now, the number of preallocated group is set to the number of threads / 2.


MattPerron:
Why divide by 2?

yingjunwu:
Preallocating one tile group for each thread is too memory consuming. Our current implementation is to allow two threads to share a same tile group. We wanna use memory efficiently and avoid scalability bottleneck. We may automatically tune the number of preallocated tile groups in the future.

Issue #312, opened 2016-10-18
Add executor thread pool


sid1607:
Currently the client's thread pool is responsible for query execution as well. This limits the number of concurrent clients that we can support without over-comitting to the number of cores and is (potentially?) responsible for some of the performance problems we face with TPCC. Decouple the client thread pool from the execution thread pool by handing over the query plan from the client thread to an executor thread and synchronizing using future-promise constructs.

sid1607:
Performance tests showed that this approach made NOP/empty query execution 1.5x slower. This is because boost threadpool uses mutexes and condition variables to assign tasks from the queue to threads in the pool. Conclusion: We need a custom-made thread pool for query execution.

PR #314, opened 2016-10-18
Decouple Client thread pool from Executor thread pool

sid1607:
#312


sid1607:
Check #312 for why this PR was cancelled.

Issue #311, opened 2016-10-18
Excessive buffer flush in wire protocol


eric-haibin-lin:
For each message peloton receives, peloton processes it and flush the response buffer. To execute a query, JDBC sends over parse, bind, describe, execute and sync (5 messages) all together in one packet, and waits for response. Currently peloton flushes the buffer after processing each of these messages. Instead flushing 5 times, we should batch the 5 responses and just flush once. 

@sid1607 @malin1993ml

malin1993ml:
That's a cool finding. Does that improve our networking performance?

eric-haibin-lin:
Used no-op benchmark script to stress the networking in peloton, and got the following result (unit: queries / second). Batch flush is the improve version of peloton. 

We have significant improvement on local tests, but for remote(cross-machine) test the improvement is not as significant. 

| No-op query | Timesten | Peloton | Postgres | Peloton (Batch Flush) 
| --- | --- | --- | --- | --- | 
| Remote | 8554 | 8024 | 8906 | 10183 | 
| Local | 26455 | 26533 | 33858 | 51578 |

| Prep Stmt | Timesten | Peloton | Postgres | Peloton (Batch Flush) 
| --- | --- | --- | --- | --- | 
| Remote | 4827 | 4476 | 4245 | 4632 | 
| Local | 18158 | 12073 | 13896 | 17774 |

** No op query = “;”
** Prep stmt = “select * from A where id = ?”. Table A is empty. 
** Remote = JDBC on dev2, database on dev1
** Local = Both JDBC & database on dev1

ranxian:
This is cool, I can run timesten ycsb+tpcc right after this is merged

PR #315, opened 2016-10-19
Reduce the number of write buffer flush in networking code

eric-haibin-lin:
I and @sid1607's improvement for Issue #311


src/networking/protocol.cpp
```diff
@@ -721,7 +734,7 @@ bool PacketManager::ManageStartupPacket() {
     return false;
   }
   status = ProcessStartupPacket(&pkt, responses);
-  if (!WritePackets(responses, &client_) || !status) {
+  if (!WritePackets(responses, &client_, true) || status == false) {
```
sid1607: Pass the third boolean argument of WirePackets as default argument for better readability?

PR #317, opened 2016-10-20
Move Heirarchy from Value to Type and return Value rather than Value * 

MattPerron:
We no longer have to "new" each Value.

This is a significant PR but the most crucial changes are to the values and types. 
All other changes handle the change of the interface.


MattPerron:
@pmenon @wangziqi2013  can you do a code review on this?

pmenon:
As a first step, this looks good.  We've managed to get rid of the new/delete's that was the primary reason for undergoing this change.

There are a few problems, but that probably don't need to be fixed now. I see a lot of switch statements in each of the type classes, which means that adding a new type will be a giant headache.  

I think the way to get around this to have some form of type promotion table to quickly promote coercible types. This sort of centralizes the switch statements into a table indexed by type id (or something to that effect). Methods in `Type` use this promotion table to perform, for example, arithmetic operations.

This is off the top of my head, we should discuss in detail.

Issue #316, opened 2016-10-20
Handle empty queries in wire protocol


eric-haibin-lin:
When peloton receives empty queries, it should send back `parse complete` for the parse msg, `no data` for the describe msg. This fix is missing in the PR merged today morning.

eric-haibin-lin:
Resolved by PR #319

PR #319, opened 2016-10-20
Wire protocol bug fix for empty queries (parse & bind) #316

eric-haibin-lin:
@sid1607


src/networking/protocol.cpp
```diff
@@ -298,16 +298,22 @@ void PacketManager::ExecParseMessage(Packet *pkt, ResponseBuffer &responses) {
     responses.push_back(std::move(response));
     return;
   }
+
+  // For empty queries, just send parse complete response
+  if (query_string == "") {
+    skipped_stmt_ = true;
+    std::unique_ptr<Packet> response(new Packet());
+    response->msg_type = '1';
```
sid1607: Instead of duplicating code to send complete response, why not branch the rest of the code so that this case directly moves on to line 347?

apavlo: We should also make the message types be constants in types.h instead of just putting '1'

Issue #321, opened 2016-10-25
column vector is copied many times for each bind parameter call


eric-haibin-lin:
e.g. here: https://github.com/cmu-db/peloton/blob/master/src/planner/update_plan.cpp#L146

PR #322, opened 2016-10-25
Fix column vector copying during ExecBind

eric-haibin-lin:
#321 
We should also run full TPCC & YCSB again after this PR.

### DIFF
--- a/script/testing/jdbc/src/PelotonTest.java
+++ b/script/testing/jdbc/src/PelotonTest.java
@@ -585,7 +585,7 @@ public class PelotonTest {
   }
 
 
-  public void Batch_Insert() throws SQLException{
+  public void Batch_Insert() throws Exception{
     PreparedStatement stmt = conn.prepareStatement(TEMPLATE_FOR_BATCH_INSERT);
 
     conn.setAutoCommit(false);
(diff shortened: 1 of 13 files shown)


jessesleeping:
Good news on TPC-C!

We got (another) 15% better in the remote case and about 25% better running peloton and oltpbench on the same machine. I have updated the TPC-C graphs in [the google doc](https://docs.google.com/spreadsheets/d/1Xu3K6q80yMgRfGWgU27egsyTG7Yo789gK5E0SbrrWgs/edit#gid=0).

YCSB is still running and we will get the result soon.

jessesleeping:
No difference in YCSB...

malin1993ml:
That's pretty good for TPCC. I think for YCSB it's kind of expected.
According to the instrumentation, TPC-C has 50% of the time spent on
networking and 50% of the time spent on processing. So if you can speed up
the processing part by 30%, then you can get 15% improvement in total.

However for YCSB, over 85% of the time was spent on reading the socket. If
we can also speed up the processing by 30%, we can only get 4.5%
improvement in total.

On Tue, Oct 25, 2016 at 23:56 Jiexi Lin <notifications@github.com> wrote:

No difference in YCSB...

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/cmu-db/peloton/pull/322#issuecomment-256243414>, or mute
the thread

Issue #332, opened 2016-11-03
GetVarcharValue.ToString lost the last character


eric-haibin-lin:
This test will fail: 

>   common::Value v9 = common::ValueFactory::GetVarcharValue("hello");
>   EXPECT_EQ(v9.ToString(), "hello");

v9.ToString() returns "hell". 
Because `GetVarcharValue` didn't count the \0 at the end, while ToString considers the last character to be \0. 

https://github.com/cmu-db/peloton/blob/master/src/include/common/value_factory.h#L61
https://github.com/cmu-db/peloton/blob/master/src/common/varlen_type.cpp#L139

PR #333, opened 2016-11-03
Fix getVarchar value length  #332

eric-haibin-lin:
#332


malin1993ml:
I just merged Yingjun's fix in another PR. Can you rebase on top of that? Thanks!

PR #336, opened 2016-11-07
[#328] Multi insert

MattPerron:
Supports inserting multiple tuples in the same query. See [#328]

### DIFF
--- a/src/planner/insert_plan.cpp
+++ b/src/planner/insert_plan.cpp
@@ -112,17 +112,8 @@ InsertPlan::InsertPlan(parser::InsertStatement *parse_tree)
 
           // If the column does not exist, insert a null value
           if (pos >= query_columns->size()) {
-            //switch (elem.GetType()) {
-              //case common::Type::VARCHAR:
-              //case common::Type::VARBINARY: {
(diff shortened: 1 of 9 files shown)


MattPerron:
@malin1993ml Can you take a look at this?

malin1993ml:
@MattPerron Sure.

src/planner/insert_plan.cpp
```diff
@@ -61,95 +60,101 @@ InsertPlan::InsertPlan(parser::InsertStatement *parse_tree,
...
+          // If the column does not exist, insert a null value
+          if (pos >= query_columns->size()) {
+            //switch (elem.GetType()) {
+              //case common::Type::VARCHAR:
+              //case common::Type::VARBINARY: {
+              //  tuple->SetValue(col_cntr, *common::ValueFactory::GetNullStringValue(),
+              //                  GetPlanPool());
+              //  break;
+              //}
+              //default:
```
malin1993ml: If these code are no longer useful, we can just delete them.

MattPerron:
@malin1993ml It should already be rebased on top of master

src/planner/insert_plan.cpp
MattPerron: Sure, This is not code I added but it can be removed.

PR #337, opened 2016-11-11
Refactor the binding mechanism

malin1993ml:
The old binding mechanism was copying parameter values into predicates and changing plans in place. That was unclean software engineering and could introduce extra copies of parameters.

After this refactoring, Peloton is going to save parameters values in the per-connection cache, and pass that to the executor context for the evaluation in the execution.

**Note:** There're still two places where we're modifying the plan in-place right now: inserted tuples in the insert plan and index scan keys in in the index scan plan. We may also want to refactor those two places in the future.

### DIFF
--- a/src/common/portal.cpp
+++ b/src/common/portal.cpp
@@ -10,33 +10,25 @@
 //
 //===----------------------------------------------------------------------===//
 
-
 #include "common/portal.h"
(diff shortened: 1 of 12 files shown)


malin1993ml:
@MattPerron Can you review this?

Issue #330, opened 2016-11-02
Implement COPY command 


eric-haibin-lin:
Implement the copy query command so that we can dump the query_metric table into a csv file with deserialized values.  

https://www.postgresql.org/docs/9.2/static/sql-copy.html

Issue #325, opened 2016-10-30
Collect parameters' value & type in query metrics table for prep stmt


eric-haibin-lin:
(add description later)

PR #338, opened 2016-11-11
[stat] implement COPY 

eric-haibin-lin:
#330  #325 
* store prepared stmt parameter values and types in query_metric table 
* support COPY stmt which export query metric table to csv. 

The copy stmt is implemented to perform table export on server side. This doesn't involve the [copy networking protocol](https://www.postgresql.org/docs/9.1/static/protocol-flow.html#PROTOCOL-COPY) specified in postgres. 

@malin1993ml  I'll add some sample csv parsing code in peloton-tf soon.


eric-haibin-lin:
Ah the validation failed multiple times because I was using memcpy..

malin1993ml:
@eric-haibin-lin Can you try to reword something and force-push again? The build failed due to some weird reason I think we can try it again.

eric-haibin-lin:
@malin1993ml  I've updated the csv parser script

src/include/planner/copy_plan.h
```diff
@@ -0,0 +1,59 @@
...
+  CopyPlan(const CopyPlan &) = delete;
+  CopyPlan &operator=(const CopyPlan &) = delete;
+  CopyPlan(CopyPlan &&) = delete;
+  CopyPlan &operator=(CopyPlan &&) = delete;
+
+  explicit CopyPlan(parser::CopyStatement *parse_tree);
+
+  inline PlanNodeType GetPlanNodeType() const { return PLAN_NODE_TYPE_COPY; }
+
+  void SetParameterValues(UNUSED_ATTRIBUTE std::vector<common::Value> *values) {
```
malin1993ml: I think we still want to call SetParameterValues on child plans. If fact you can remove this override and use the function in the base class:
https://github.com/cmu-db/peloton/blob/master/src/planner/abstract_plan.cpp#L65

src/include/statistics/backend_stats_context.h
```diff
@@ -17,6 +17,7 @@
 #include <sstream>
 #include <unordered_map>
 
+#include "common/statement.h"
```
malin1993ml: If you just use Statement as a pointer, I think you can fast forward the declaration and include the header in the cpp file. Or it doesn't work with shared_ptr?

src/planner/copy_plan.cpp
```diff
@@ -0,0 +1,73 @@
...
+
+  // Construct select stmt
+  std::unique_ptr<parser::SelectStatement> select_stmt(
+      new parser::SelectStatement());
+  select_stmt->from_table = table_ref;
+  select_stmt->select_list = select_list;
+
+  // Add the child seq scan plan
+  std::unique_ptr<planner::SeqScanPlan> child_SelectPlan(
+      new planner::SeqScanPlan(select_stmt.get()));
```
malin1993ml: We're not going to have this constructor method for SeqScanPlan after Hao's refactor. If fact, I think a better way is to construct a SQLStatementList and the directly call SimpleOptimizer::BuildPelotonPlanTree. (A even better approach is to refactor the SimpleOptimier a little bit and abstract out a method which can directly take a SQLStatement* as the argument...) What do you think?

test/statistics/stats_test.cpp
```diff
@@ -365,11 +369,17 @@ TEST_F(StatsTest, PerQueryStatsTest) {
   LOG_INFO("Table created!");
 
   auto backend_context = stats::BackendStatsContext::GetInstance();
+  // Get a query param object
+  std::shared_ptr<uchar> type_buf;
+  std::shared_ptr<uchar> format_buf;
+  std::shared_ptr<uchar> val_buf;
+  auto query_params =
+      StatsTestsUtil::GetQueryParams(type_buf, format_buf, val_buf);
```
malin1993ml: Is it possible to test the serialization and deserialization of query_params? It seems to be hard to do it with ctest. But is it feasible to implement a simple python validator to do that?

eric-haibin-lin: check stats_validator.py in peloton/scripts/validators

malin1993ml: I see. For some reason I thought it's for another purpose. That looks good!

src/include/statistics/backend_stats_context.h
eric-haibin-lin: Good point. Works with shared_ptr.

src/planner/copy_plan.cpp
```diff
@@ -0,0 +1,73 @@
...
+#include "parser/statement_copy.h"
+
+namespace peloton {
+namespace planner {
+
+/**
+ * XXX Copy plan by default assumes a seq scan plan node as it's child
+ * This can be extended when we support nested queries in COPY query
+ */
+CopyPlan::CopyPlan(parser::CopyStatement* copy_parse_tree)
```
eric-haibin-lin: Yes will do

src/include/catalog/column.h
```diff
@@ -29,14 +29,20 @@ class Column : public Printable {
...
+         oid_t column_offset = INVALID_OID)
       : column_type(value_type),
         column_name(column_name),
         is_inlined(is_inlined),
         column_offset(column_offset) {
     SetInlined();
 
+    // We should not have an inline value of length 0
+    if (is_inlined && column_length == 0) {
+      assert(false);
```
eric-haibin-lin: done

PR #342, opened 2016-11-15
Peloton Networking with Libevent Callbacks

sid1607:
Presently, Peloton uses Libevent for accepting new connections and relies on select() to implement non-blocking network I/O. Further, a client thread can only support a single connection.

This PR gets rid of select() to have a complete Libevent callback-based network state machine. Additional features include -

- Every client thread can support multiple active sockets and the libevent callback mechanism implicitly schedules sockets to be processed. **The number of clients that Peloton can support is no longer constrained by the number of cores in the system.**

- Previously, data from the socket's read buffer were copied into a packet buffer, before applying the protocol's processing logic. Now, references are directly maintained to the read buffer when the packet fits in the read buffer. Data is only copied when the packet exceeds read buffer size (ex: blob transfers)

TPCC and YCSB numbers can be found here (look for the **Peloton-libevent-bind*** lines) - https://docs.google.com/spreadsheets/d/1Xu3K6q80yMgRfGWgU27egsyTG7Yo789gK5E0SbrrWgs/edit#gid=1173468315


eric-haibin-lin:
Also the number for networking microbenchmark is at https://docs.google.com/spreadsheets/d/10U5EQBF0BzlmzwincFxZyMbJlFvbXPbELuQ3z0QI7Ig/edit#gid=1306029145

eric-haibin-lin:
I am still modifying some code re: Lin's comments. Please don't merge it in.

eric-haibin-lin:
@malin1993ml done editing.

Issue #347, opened 2016-11-17
configurable thread count


yingjunwu:
The current implementation set the number of threads for transaction processing to `std::thread::hardware_concurrency()` by default. This may not be correct. We should preserve some threads for garbage collection and logging. Also, the system must know the number of threads assigned for each component (e.g., transaction processing, garbage collection, and logging) prior to execution. To solve this problem, I introduced some variables in `types.h` and users should set these variables when booting Peloton.

apavlo:
We should talk about this with @malin1993ml. It's good that this parameter is no longer hardcoded, but we want to be able to make it dynamic so that we can tweak it at runtime. This is part of a larger theme of having all of the configuration knobs dynamic so that Lin's work can change it.

This is a good start though.

malin1993ml:
Sounds good. We can talk about this at some point.

PR #348, opened 2016-11-17
configurable thread count

yingjunwu:
#347 

I will start rewriting epoch manager (which supports read-only transaction) once this PR is merged.


malin1993ml:
@yingjunwu Can you rebase this PR on top of the current master and fix the failing tests? Thanks!

yingjunwu:
@malin1993ml it has passed all the tests.

malin1993ml:
There're duplicated commit messages in you local branch:
https://github.com/yingjunwu/peloton/commits/new-epoch

Several commit messages starting from 3a2ec0d4561260067b1ccd062894cc6103487e47 are the same as those starting from 5d010bdb13265a23f7b69fa5b14d9fe6112e7151. Can you clean them up? 

There're 30 commits in total in this PR, which is not correct. I'm afraid something will be messed up if I directly merge them.

PR #359, opened 2016-11-26
Add pre-declared read only support

jessesleeping:
Use the BeginReadOnlyTransaction() method to create a pre-declared read-only transaction.

Signed-off-by: Jiexi Lin <jxlin39@gmail.com>

### DIFF
--- a/test/include/concurrency/transaction_tests_util.h
+++ b/test/include/concurrency/transaction_tests_util.h
@@ -173,10 +173,12 @@ struct TransactionSchedule {
   std::vector<int> results;
   int stored_value;
   int schedule_id;
-  TransactionSchedule(int schedule_id_)
+  bool declared_ro;
+  TransactionSchedule(int schedule_id_, bool ro = false)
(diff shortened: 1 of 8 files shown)


jessesleeping:
@sid1607 This is the read-only feature you need.

test/include/concurrency/transaction_tests_util.h
```diff
@@ -332,13 +341,18 @@ class TransactionThread {
...
-                       concurrency::TransactionManager *txn_manager_)
+                       concurrency::TransactionManager *txn_manager_,
+                       bool first_as_ro = false)
       : txn_manager(txn_manager_),
         table(datatable_),
         time(0),
         concurrent(false) {
     for (size_t i = 0; i < num_txn; i++) {
-      schedules.emplace_back(i);
+      if (first_as_ro && i == 0) {
```
malin1993ml: Why only the first transaction can be read_only? Is that a requirement for batched transactions in our system?

PR #360, opened 2016-11-28
Support simple equal-join

hzxa21:
- Support simple equal-join with single join condition
   Example: SELECT * FROM A join B ON A.a = B.b;  SELECT A.c, B.c FROM A join B ON A.a = B.b;
- Modify tcop.cpp to support SELECT * in join query
- Not yet supported features:
   - Apply predicates in hash_join_executor (Fail to get right_tuple in the logical tile)
   - Aggregation in join query
   - Extract subexpression that can be pushed down predicates into the scan plan

@malin1993ml


malin1993ml:
@hzxa21 Some tests have failed. Can you try to fix them?

hzxa21:
@malin1993ml I notice that. Will look into it.

src/optimizer/simple_optimizer.cpp
```diff
@@ -1023,5 +1025,153 @@ std::shared_ptr<const peloton::catalog::Schema> CreateHackJoinSchema() {
...
+		  else {
+			  tl.push_back(Target(i, expr->Copy()));
+			  output_table_columns.push_back(
+				  catalog::Column(expr->GetValueType(),
+					  common::Type::GetTypeSize(expr->GetValueType()),
+					  "expr" + std::to_string(i)));
+		  }
+		  i++;
+	  }
+  }
```
malin1993ml: Instead of another set of separate logic determining output columns and mapping list, we should reuse the logic here:
https://github.com/cmu-db/peloton/blob/master/src/include/expression/expression_util.h#L299

I know that utility function can only handle the single table case, so you probably have to refactor that logic a little bit. I think replacing the single schema reference in the parameter to a reference of a vector of schemas would do. You also need to supplement the logic in that function (e.g., get output columns for star expression).

I think if we bite the bullet to refactor that utility function right now, it will also save us lots of time later. You can talk to me if you're not sure what you should to do.

test/optimizer/hash_join_test.cpp
```diff
@@ -0,0 +1,260 @@
...
+  // txn_manager.CommitTransaction(txn);
+
+  // LOG_INFO("Before Join\n\n\n\n\n\n\n\n\n\n\n");
+
+  txn = txn_manager.BeginTransaction();
+  LOG_INFO("Join ...");
+  LOG_INFO(
+      "Query: SELECT * FROM table_a INNER JOIN table_b ON aid = bid;");
+  statement.reset(new Statement("SELECT",
+                                "SELECT * FROM table_a INNER JOIN table_b ON aid = bid;"));
```
hzxa21: Got it. Thank you for pointing out. Will fix them.

malin1993ml:
The refactoring looks good! Is there any other remaining changes that you want to push to this PR?

hzxa21:
@malin1993ml No. Done already with the simple equi-join.

malin1993ml:
Cool! Can you rebase again real quick and then I'll merge the PR?

PR #364, opened 2016-11-28
Fixes for Boolean Value Type

apavlo:
This commit fixes a lot of the problems with the Boolean type:

1. We were not able to convert VARCHARs to BOOLEANs.
2. And then even if we could, we had mispelled 'true' and 'false' so it wouldn't have worked anyway.
3. `BooleanType::CompareEquals` would always return true if you compared a NULL value...
4. This was because it would call `ValueFactory::GetBooleanValue` which only supported bool inputs. That means `PELOTON_BOOLEAN_NULL` was always converted to true
5. And then because I was disappointed that our code coverage has not moved up in a few months, I busted out new boolean test cases to make sure that all of this is working properly.

Note that **clang-format** reformatted some of the the source code files when I edited them so that's why there are a bunch of changes...

### DIFF
--- a/src/common/array_type.cpp
+++ b/src/common/array_type.cpp
@@ -16,14 +16,14 @@
 #include "common/decimal_type.h"
 #include "common/numeric_type.h"
 #include "common/timestamp_type.h"
-#include "common/varlen_type.h"
 #include "common/type.h"
+#include "common/varlen_type.h"
 
 namespace peloton {
 namespace common {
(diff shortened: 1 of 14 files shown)


apavlo:
Actually hold on merging this. Apparently we don't support "TRUE >= TRUE". I will fix after class

apavlo:
This is ready to merge

Issue #343, opened 2016-11-15
NestedLoopIndexJoin Executor


apavlo:
We need the `NestedLoopIndexJoin` executor for TPC-C StockLevel. The current implementation is a regular dirty nested-loop without the join probe. You are also going to need to fix up `SimpleOptimizer::CreateHackingJoinPlan` to use the new executor instead of the hash join:

https://github.com/cmu-db/peloton/blob/master/src/optimizer/simple_optimizer.cpp#L764

domino-succ:
Yes. We need to replace HashJoin in current simple_optimizer.cpp with the fixed nested-loop join.

PR #366, opened 2016-11-30
NestedLoopJoin for TPC-C

Michael-Tieying-Zhang:
#343 Now we can run TPC-C without Aggregation issue by using NestedLoopJoin. The throughput is improved around 30% for full TPC-C. We need to add IndexPredicate update in the near future.


src/include/executor/index_scan_executor.h
```diff
@@ -88,6 +105,8 @@ class IndexScanExecutor : public AbstractScanExecutor {
   std::vector<expression::AbstractExpression *> runtime_keys_;
 
   bool key_ready_ = false;
+
+  planner::IndexScanPlan *plan_node_;
```
malin1993ml: I don't think we should use another plan_node_ here for IndexScanPlan, since we already have a node_ in the base class. If you already have the UpdatePredicate() method, can you call that method using node_ instead of have another non-const copy of the plan?

src/optimizer/simple_optimizer.cpp
```diff
@@ -103,7 +104,8 @@ std::shared_ptr<planner::AbstractPlan> SimpleOptimizer::BuildPelotonPlanTree(
         try {
           catalog::Catalog::GetInstance()->GetTableWithName(DEFAULT_DB_NAME,
                                                             "order_line");
-          auto child_SelectPlan = CreateHackingJoinPlan();
+          // auto child_SelectPlan = CreateHackingJoinPlan();
```
malin1993ml: If CreateHackingJoinPlan() is no longer used anymore, we should just delete it. You don't have to do this now, but make sure you clean it up after we finish all the stuff for NestedLoopJoin.

src/executor/nested_loop_join_executor.cpp
```diff
@@ -74,26 +76,23 @@ bool NestedLoopJoinExecutor::DInit() {
  * to join the left and right result.
  *
  */
+
+// For this version, the work flow is that we first lookup the left table, and
+// use the result to lookup right table. If left table is done that means right
+// table is also done. So we only keep the left_child_done_ as the sign.
 bool NestedLoopJoinExecutor::DExecute() {
```
malin1993ml: Does this method still work with non-index nested loop join (sequential scan nested loop join) after your modification? Did you test that?

src/include/executor/index_scan_executor.h
Michael-Tieying-Zhang: We have discussed this problem. Since node_ is const, we cannot update plan node using node_. That's why I add plan_node_.

src/executor/nested_loop_join_executor.cpp
Michael-Tieying-Zhang: Yes, since UpdatePredicate can be overloaded by each Executor including SeqScan. But I plan to implement index predicate with seq predicate together, so for now it doesn't work. I will implement it in the next version (with index predicate).

Issue #375, opened 2016-12-04
Min aggregation on integer is not working


malin1993ml:
Max on integer is working. Min and max on double are also working.

```sql
malin199@dev1:~/oltpbench$ psql -U postgres -h localhost -p 15721
psql (9.5devel)
Type "help" for help.

postgres=# create table test(id int, value double);
create table test(id int, value double) 0
postgres=# insert into test values(1, 2);
insert into test values(1, 2) 1
postgres=# insert into test values(3, 4);
insert into test values(3, 4) 1
postgres=# select * from test;                                                                                                                 
 id |  value   
----+----------
  1 | 2.000000
  3 | 4.000000
(2 rows)

postgres=# select min(id) from test;                                                                                                           
 min 
-----
   3
(1 row)

postgres=# select max(id) from test;
 max 
-----
   3
(1 row)

postgres=# select min(value) from test;
   min    
----------
 2.000000
(1 row)

postgres=# select max(value) from test;
   max    
----------
 4.000000
(1 row)
```

PR #378, opened 2016-12-05
Min integer fix

malin1993ml:
Fix for #375.


apavlo:
How hard would it be to make this test more generic and test the different types?

malin1993ml:
@apavlo It's not hard. It's just right now I'm reusing the mock execution system where there're only 4 columns in the table. I don't think I should change that since that affects lots of testing of other executors. I'll just write a separate test class for that.

malin1993ml:
@apavlo The test case is done. Though I commented out the test for varchar and timestamp for now because we can't do min/max aggregate on them right now. I'll create a separate issue for that.

Issue #384, opened 2016-12-07
UPDATE allows you to violate primary key constraint


apavlo:
We don't let you INSERT a duplicate key:
```
postgres=# create table foo(id integer PRIMARY KEY, year integer);
create table foo(id integer PRIMARY KEY, year integer) 0
postgres=# insert into foo values(1, 100);
insert into foo values(1, 100) 1
postgres=# insert into foo values(1, 100);
insert into foo values(1, 100) 0
postgres=# SELECT * FROM foo;
 id | year 
----+------
  1 |  100
(1 row)
```

Well now that's good. But you better believe that we let you update an existing tuple with a duplicate primary key:
```
postgres=# insert into foo values(2, 200);
insert into foo values(2, 200) 1
postgres=# SELECT * FROM foo;
 id | year 
----+------
  1 |  100
  2 |  200
(2 rows)

postgres=# UPDATE foo SET id = 1 WHERE year = 200;
UPDATE foo SET id = 1 WHERE year = 200 1
postgres=# SELECT * FROM foo;
 id | year 
----+------
  1 |  100
  1 |  200
(2 rows)
```
In the above example, we should not allow you to have two of the same ids.

Michael-Tieying-Zhang:
Do we support UNIQUE INDEX now?

apavlo:
Yes. See the first example above where I try to INSERT the same key twice and it fails on the second one (as expected).

Michael-Tieying-Zhang:
Got you. In that case, it's easy to fix.

Michael-Tieying-Zhang:
The problem also exists for UNIQUE SECONDARY INDEX: 

`michael=# UPDATE foo SET age = 1000 WHERE year = 100;
UPDATE foo SET age = 1000 WHERE year = 100 1
michael=# select * from foo;
 id | age  | year 
----+------+------
  2 |   20 |  200
  1 | 1000 |  100
(2 rows)
`
The code should examine UNIQUE INDEX, but it did nothing: 
[https://github.com/cmu-db/peloton/blob/master/src/storage/data_table.cpp#L439](https://github.com/cmu-db/peloton/blob/master/src/storage/data_table.cpp#L439)

PR #389, opened 2016-12-09
Fix secondary index issue: cannot find tuple after updating

Michael-Tieying-Zhang:
This PR fixed the secondary index issue: cannot find tuple after updating, that is described in the comment of #384 (but #384 is still there).

### DIFF
--- a/src/executor/plan_executor.cpp
+++ b/src/executor/plan_executor.cpp
@@ -44,9 +44,8 @@ void CleanExecutorTree(executor::AbstractExecutor *root);
  * @return status of execution.
  */
 peloton_status PlanExecutor::ExecutePlan(
-    const planner::AbstractPlan *plan,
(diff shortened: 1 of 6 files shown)


apavlo:
I don't understand what problem this is solving? Can you make a separate issue with a more complete example?

Michael-Tieying-Zhang:
Pls see #390

malin1993ml:
@Michael-Tieying-Zhang The fix is good. Can you write a test case for that? You can use [SQLTestsUtil](https://github.com/cmu-db/peloton/blob/master/test/sql/sql_tests_util.cpp) to write a sql test. See how I wrote the sql test for min/max aggregation here:
https://github.com/cmu-db/peloton/tree/master/test/sql

Michael-Tieying-Zhang:
@malin1993ml Done

PR #391, opened 2016-12-13
Refactor insert and delete plan

Michael-Tieying-Zhang:
This PR rewrote insert and delete plan constructor. It uses member variable as the input of constructor rather than insert statement or delete statement. This PR also fixed some bugs of insert test and delete test.


malin1993ml:
Could you delete the old constructors for inset and delete plan which use statements, if not used anymore?

Michael-Tieying-Zhang:
Don't you think we can keep the old constructor to make some tests convenient, like this: [https://github.com/cmu-db/peloton/blob/master/test/planner/planner_test.cpp#L71](https://github.com/cmu-db/peloton/blob/master/test/planner/planner_test.cpp#L71)

malin1993ml:
I don't think we should keep deprecated APIs for the convenience of tests. That can cause trouble later on. Could you help replace the test with the new API?

Michael-Tieying-Zhang:
Agree!

apavlo:
@malin1993ml is correct. Delete old API calls if the code doesn't use them anymore.

Michael-Tieying-Zhang:
Done.

Issue #394, opened 2016-12-14
BeginTransaction is called for every query that is executed


sid1607:
To execute a query, protocol.cpp invokes tcop.ExecuteStatement(). 

TrafficCop::ExecuteStatement in-turn invokes PlanExecutor::ExecutePlan

PlanExecutor::ExecutePlan calls txn_manager.BeginTransaction().

The current implementation is not aware of when a BEGIN/COMMIT query is received and creates transactions for individual queries.

MattPerron:
closed with #400

PR #400, opened 2016-12-14
#394: Added frontend txn handling bug fix

sid1607:
Features introduced by this PR include:

- Instantiation of tcop per client
- Adding state in tcop to track active tansactions for a given connection
- Support nested transactions by maintaining tcop state as stack
- Reuse same transaction object for all queries within the same BEGIN-COMMIT block

### DIFF
--- a/src/tcop/tcop.cpp
+++ b/src/tcop/tcop.cpp
@@ -49,6 +49,15 @@ TrafficCop::~TrafficCop() {
   }
 }
 
+/* Singleton accessor
+ * NOTE: Used by in unit tests ONLY
+ */
+TrafficCop& TrafficCop::GetInstance() {
+  static TrafficCop tcop;
+  tcop.Reset();
+  return tcop;
+}
+
 TrafficCop::TcopTxnState& TrafficCop::GetDefaultTxnState() {
   static TcopTxnState default_state;
   default_state = std::make_pair(nullptr, Result::RESULT_INVALID);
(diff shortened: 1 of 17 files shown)


malin1993ml:
Hi Siddharth @sid1607 , what is the status for this PR? Can you fix the failing test? Also, can you rebase since there're some conflicts with the latest merged PRs?

sid1607:
This PR is on hold because I'm facing an issue when I run TPCC with multiple terminals - After a few seconds the transactions begin to fail because index scan executor is not able to acquire ownership of the tuples it is trying to read. This is because the tuple end commit ID is -1 and current transaction's begin commit ID is a non-zero valid timestamp. I will discuss this further with @jessesleeping  today to find out why this is happening.

src/tcop/tcop.cpp
```diff
@@ -81,28 +163,99 @@ Result TrafficCop::ExecuteStatement(
...
   try {
-    bridge::PlanExecutor::PrintPlan(statement->GetPlanTree().get(), "Plan");
-    bridge::peloton_status status = bridge::PlanExecutor::ExecutePlan(
-        statement->GetPlanTree().get(), params, result, result_format);
-    LOG_TRACE("Statement executed. Result: %d", status.m_result);
-    rows_changed = status.m_processed;
-    return status.m_result;
-  } catch (Exception &e) {
+    auto query_str = boost::to_upper_copy<std::string>(statement->GetQueryString());
+    if (query_str == "BEGIN")
```
MattPerron: I think we should avoid string matching here. There should be a way to get the state of the transaction from the statement object. Also I think "ABORT" will never be sent. The actual call is "ROLLBACK" according to sql_parser.y

Issue #396, opened 2016-12-14
Support reference count for varlen columns


jessesleeping:
When creating a new version of a tuple in the update executor, if a column:
- is of VARCHAR or VARBINARY type
- and is not inlined
- and is not updated
then, instead of allocating a copy of the varlen value, we can add to its reference count in the varlen pool.

To support this, we need to
- support reference count in the varlen pool
- support copy by reference in update executor/project info
- (maybe) make garbage collection manager aware of the reference count

Related issues: 
- #349. After we have inlined varchar, extra check needs to be done in the executor to determine whether we should copy by reference or not.
- #395. When primary key is of varlen type, we need to modify its update logic too.

jessesleeping:
While implementing this feature, I had a feeling that we need to rethink our design of the type system.

In order to copy

@MattPerron @apavlo

apavlo:
Why do you think we need to re-implement our type system?

jessesleeping:
I think we may need to rewrite the code for varlen type. 

Currently, we memcpy the varlen value whenever we call GetValue. I think what we can do is to copy varlen value by reference and make it as our default setting. Whenever we do need a physical copy of it, we can use something like a deep copy method.

In my local branch, I have partially supported this feature. Now the update executor will not memcpy a varlen value when projecting it directly. I will send a PR once I have fully tested it.

jessesleeping:
As stated in the PR, now we can fully support reference count in varlen type. However, it's kind of hacking the type system to support the shallow copy feature we want.

I have profiled the performance gain from this feature with the YCSB micro-benchmark. Here is the [result](https://docs.google.com/spreadsheets/d/1nL0-rUCTE_niGRLvMt3paW1RZruc_Cavblno98sG99o/edit#gid=0).

Basically, we get 10x throughput at 10 thread compared to the original varlen pool. The performance is closed to the YCSB using integer instead of string. Compared to our old varlen pool, the new one is way too slow and not scalable.

PR #405, opened 2016-12-17
Support reference count in varlen type

jessesleeping:
In this PR, we implemented the following features:
- Support reference count in varlen pool. 
- Make the update executor aware of the reference count and perform a shallow copy on varlen column
- Make the GC manager aware of the reference count and reclaim memory in the varlen pool.

#396

### DIFF
--- a/src/common/value.cpp
+++ b/src/common/value.cpp
@@ -410,6 +410,19 @@ Value Value::CastAs(const Type::TypeId type_id) const {
 // Access the raw variable length data
 const char *Value::GetData() const { return type_->GetData(*this); }
 
+// Access the raw variable length data from a pointer pointed to a tuple storage
+char *Value::GetDataFromStorage(Type::TypeId type_id, char *storage) {
+  switch (type_id) {
+    case Type::VARCHAR:
+    case Type::VARBINARY: {
+      return Type::GetInstance(type_id)->GetData(storage);
+    }
+    default:
+      throw Exception(EXCEPTION_TYPE_INCOMPATIBLE_TYPE,
+                      "Invalid Type for getting raw data pointer");
+  }
+}
+
 // Get the length of the variable length data
 uint32_t Value::GetLength() const { return type_->GetLength(*this); }
 
--- a/src/common/value.cpp
+++ b/src/common/value.cpp
@@ -394,6 +394,12 @@ void Value::SerializeTo(SerializeOutput &out) const {
   type_->SerializeTo(*this, out);
 }
 
+// Perform a shallow copy from a serialized varlen value to another serialized varlen value
+// Only support VARCHAR/VARBINARY
+void Value::ShallowCopyTo(char *dest, char *src, const Type::TypeId type_id, bool inlined, VarlenPool *src_pool) {
+  Type::GetInstance(type_id)->DoShallowCopy(dest, src, inlined, src_pool);
+}
+
 // Create a copy of this value
 Value Value::Copy() const { return type_->Copy(*this); }
 
--- a/src/planner/project_info.cpp
+++ b/src/planner/project_info.cpp
@@ -95,7 +95,8 @@ bool ProjectInfo::Evaluate(expression::ContainerTuple<storage::TileGroup> *dest,
 
   // (B) Execute direct map
   if (inplace == false) {
-    // Only copy un-modified value if not in-place update
+    // For update that creates a new version, we copy all unmodified columns
+    // to the new version. Note that for varlen column, we perform shallow copy.
     for (auto dm : direct_map_list_) {
       // whether left tuple or right tuple ?
       auto tuple_index = dm.second.first;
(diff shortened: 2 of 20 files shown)


jessesleeping:
Now we can fully support reference count in varlen type. However, it's kind of hacking the type system to support the shallow copy feature we want.

I have profiled the performance gain from this feature with the YCSB micro-benchmark. Here is the [result](https://docs.google.com/spreadsheets/d/1nL0-rUCTE_niGRLvMt3paW1RZruc_Cavblno98sG99o/edit#gid=0).

Basically, we get 10x throughput at 10 thread compared to the original varlen pool. The performance is closed to the YCSB using integer instead. Compared to our old varlen pool, the new one is way too slow and not scalable.

jessesleeping:
Now we can fully support reference count in varlen type. However, it's kind of hacking the type system to support the shallow copy feature we want.

I have profiled the performance gain from this feature with the YCSB micro-benchmark. Here is the [result](https://docs.google.com/spreadsheets/d/1nL0-rUCTE_niGRLvMt3paW1RZruc_Cavblno98sG99o/edit#gid=0).

Basically, we get 10x throughput at 10 thread compared to the original varlen pool. The performance is closed to the YCSB using integer instead. Compared to our old varlen pool, the new one is way too slow and not scalable.

jessesleeping:
Also, I think we need to make shallow copy as the default copy method for varlen type. Currently a simple `GetValue` call to a `ConatinerTuple` will incur 3~4 memory copy of a varlen value.

malin1993ml:
@jessesleeping What exactly are the "original varlen pool", "old varlen pool", and "new one" you're referring to in [this comment](https://github.com/cmu-db/peloton/pull/405#issuecomment-267752155)? Are the "original varlen pool" and the "old varlen pool" the same thing? Also, are your results gotten with OLTPBench or micro-benchmark?

jessesleeping:
Sorry for the confusing wording. "The original" one means the one in our current

jessesleeping:
Sorry for the confusing wordings... I have corrected them in the commend. I tested it with the micro benchmark.

malin1993ml:
Got it. Thanks!

src/common/value.cpp
```diff
@@ -394,6 +394,12 @@ void Value::SerializeTo(SerializeOutput &out) const {
   type_->SerializeTo(*this, out);
 }
 
+// Perform a shallow copy from a serialized varlen value to another serialized varlen value
+// Only support VARCHAR/VARBINARY
+void Value::ShallowCopyTo(char *dest, char *src, const Type::TypeId type_id, bool inlined, VarlenPool *src_pool) {
+  Type::GetInstance(type_id)->DoShallowCopy(dest, src, inlined, src_pool);
+}
```
malin1993ml: Is there a deep copy version for this? Or the Copy() function below will just do the deep copy?

src/planner/project_info.cpp
```diff
@@ -80,30 +82,30 @@ bool ProjectInfo::Evaluate(storage::Tuple *dest, const AbstractTuple *tuple1,
...
+    for (auto dm : direct_map_list_) {
+      // whether left tuple or right tuple ?
+      auto tuple_index = dm.second.first;
+      auto src_col_id = dm.second.second;
+
+      PL_ASSERT(dm.first == dm.second.second);
+      PL_ASSERT(tuple_index == 0);
+      if (tuple_index == 0) {
+        src->CopyColumnTo(dest, src_col_id);
+      }
```
malin1993ml: I don't quite understand the logic here. The comment said "Only copy un-modified value if not in-place update", but it seems that the code suggested if "inplace == false", then we call CopyColumnTo(), which calls the shallow copy function. So should we do or not do a deep copy when it's in-place update? Also, shouldn't there be another branch of logic to do the deep copy on the other condition?

src/common/value.cpp
jessesleeping: The Copy() function below is the deep copy one.

src/common/value.cpp
```diff
@@ -404,6 +410,19 @@ Value Value::CastAs(const Type::TypeId type_id) const {
...
+// Access the raw variable length data from a pointer pointed to a tuple storage
+char *Value::GetDataFromStorage(Type::TypeId type_id, char *storage) {
+  switch (type_id) {
+    case Type::VARCHAR:
+    case Type::VARBINARY: {
+      return Type::GetInstance(type_id)->GetData(storage);
+    }
+    default:
+      throw Exception(EXCEPTION_TYPE_INCOMPATIBLE_TYPE,
+                      "Invalid Type for constructor");
```
jessesleeping: Fixed. Thank you for pointing it out.

src/planner/project_info.cpp
malin1993ml: Got it. That makes sense.

PR #406, opened 2016-12-17
#392: Skip parser rebuilding, use configuration file, and fix benchmarks

jarulraj:
This PR:

- Skips rebuilding parser every single time
- Moves the configuration folder to the root directory
- Expands the Google flags configuration file
- Fixes compilation of benchmarks

Let's start using the configuration file for setting runtime parameters.


configuration/peloton.conf
```diff
@@ -0,0 +1,119 @@
+# -----------------------------
```
malin1993ml: Where did you load this file for the server? I just didn't find the code... Did you specify it somewhere in the CMake file?

configuration/peloton.conf
```diff
@@ -0,0 +1,119 @@
...
+#                GB = gigabytes                     s   = seconds
+#                                                   min = minutes
+#                                                   h   = hours
+
+#------------------------------------------------------------------------------
+# FILE LOCATIONS
+#------------------------------------------------------------------------------
+
+# Location of data directory
+--data_directory='data'
```
malin1993ml: For parameter like this, how the system can accept it as a command-line argument? I was using GFLAG to parser and accept arguments like "stats_mode" or "port" from command line. But I didn't see the GFLAG declaration of "data_directory".

configuration/peloton.conf
```diff
@@ -0,0 +1,119 @@
...
+--socket_family=AF_INET
+
+#------------------------------------------------------------------------------
+# RESOURCE USAGE
+#------------------------------------------------------------------------------
+
+# - Memory -
+
+# Shared memory pool
+--shared_pool=128MB
```
malin1993ml: Same question here.

configuration/peloton.conf
jarulraj: It is loaded at runtime like this:

`./bin/peloton -flagfile ../configuration/peloton.conf`

configuration/peloton.conf
jarulraj: I have not yet added some of those parameters or used them inside peloton. 

But, they should eventually be declared in `src/include/configuration/configuration.h` and defined in `src/configuration/configuration.cpp`. They can then be used in any module within peloton like a regular GFLAGs parameter.

malin1993ml: Okay.

malin1993ml:
You may have to rebase now since there're some conflicts with the latest merged PR.

jarulraj:
Sure, I have rebased it.

jarulraj:
I am not sure why the jdbc validator failed.

Issue #183, opened 2016-05-27
temp table


yingjunwu:
A transaction needs to create a temp table if an aggregation operator is performed. Our current implementation treats such a temp table equally as a regular database table, which means that every temp table has to be registered to the catalog, causing unnecessary synchronization overhead. Meanwhile, some aggregation operations, such as sum, count, max, min, generates only one output tuple. using a regular database table to maintain the query results wastes memory spaces, as every data table in our implementation has to have at least one tile group, which can maintain 1,000 tuples.

We should develop a new `temp_table` class to maintain results for aggregation queries.

poojanilangekar:
Fixed in #409

PR #409, opened 2016-12-20
TempTable + AggregationExecutor Fix 💩

apavlo:
This has a working version of the new `storage::TempTable` class for #183. I have also updated `executor::AggregateExecutor` to use it.

I have **not** tested TPC-C with this yet. It was a pain to rebase because of #406 so I just want to get it in now. 

I'm dedicating this PR to @eric-haibin-lin. Stay strong, son.


src/storage/temp_table.cpp
```diff
@@ -0,0 +1,133 @@
...
+
+  // Set tuple location and increase our counter
+  ItemPointer location(tile_group_id, tuple_slot);
+  IncreaseTupleCount(1);
+
+  // Make sure that we mark the tuple as active in the TileGroupHeader too
+  // If you don't do this, then anybody that tries to use a LogicalTile wrapper
+  // on this TempTable will not have any active tuples.
+  auto tile_group_header = tile_group->GetHeader();
+  tile_group_header->SetTransactionId(location.offset, INITIAL_TXN_ID);
```
jessesleeping: I think it's not enough to just set the transaction id field in the header. For insert executor, it will later call the `TransactionManager.PerformInsert()` to set up the header of a newly inserted tuple. Here you may need to set the begin timestamp of this tuple as MIN_CID and the end timestamp as MAX_CID to ensure it pass the visibility test in the upper layer.

src/executor/aggregate_executor.cpp
```diff
@@ -67,11 +67,15 @@ bool AggregateExecutor::DInit() {
...
-  output_table = storage::TableFactory::GetDataTable(
-      INVALID_OID, INVALID_OID, output_table_schema, "aggregate_temp_table",
-      DEFAULT_TUPLES_PER_TILEGROUP, own_schema, adapt_table);
+  //  bool own_schema = false;
+  //  bool adapt_table = false;
+
+  //  output_table = storage::TableFactory::GetDataTable(
+  //      INVALID_OID, INVALID_OID, output_table_schema, "aggregate_temp_table",
+  //      DEFAULT_TUPLES_PER_TILEGROUP, own_schema, adapt_table);
+
```
apavlo: Fixed.

src/include/storage/temp_table.h
```diff
@@ -0,0 +1,137 @@
...
+
+class Tuple;
+class TileGroup;
+class TileGroupHeader;
+
+// FIXME
+const size_t TEMPTABLE_DEFAULT_SIZE = 100;  // # of tuples
+
+// FIXME
+const oid_t TEMPTABLE_TILEGROUP_ID = 99999;
```
jessesleeping: OK, I see.

src/executor/aggregate_executor.cpp
```diff
@@ -169,37 +173,31 @@ bool AggregateExecutor::DExecute() {
       } else {
         tuple->SetAllNulls();
       }
-      auto location = output_table->InsertTuple(tuple.get());
-      PL_ASSERT(location.block != INVALID_OID);
-
-      auto &manager = catalog::Manager::GetInstance();
```
jessesleeping: The `tuple` object here is a local variable and gets destructed right after we leave the if clause. It seems that we just allocate it and set it as zero without inserting it into any tile or table. Is there something wrong? Or we just don't need the tuple for empty group by result?

src/storage/temp_table.cpp
jessesleeping: My bad... The scan executor does not check visibility when reading a logical tile. Your code is correct.

src/executor/aggregate_executor.cpp
apavlo: Hmmmm you might be right. Let me check on this.

src/storage/temp_table.cpp
```diff
@@ -0,0 +1,133 @@
...
+  // We only want to instantiate a single TileGroup
+  AddDefaultTileGroup();
+}
+
+TempTable::~TempTable() {
+  // Nothing to see, nothing to do
+}
+
+ItemPointer TempTable::InsertTuple(const Tuple *tuple,
+                                   concurrency::Transaction *transaction,
```
jessesleeping: The `transaction` and the `index_entry_ptr` should be tagged as `UNUSED_ATTRIBUTE`, otherwise it will failed the release build.

apavlo:
Merging now because @jarulraj is ruining my marriage....

PR #412, opened 2016-12-21
Fix layout tuner test

jarulraj:
This PR reenables the layout tuner test.


jarulraj:
@malin1993ml can you check this fix?

PR #413, opened 2016-12-21
Remove PlanExecutor::Print() and replace it with PlanUtil::GetInfo()

apavlo:
This PR makes it so that we don't call the pretty-print function on the query plan *every* time we execute a query.

We never saw this because `PlanExecutor::Print()` was using `LOG_TRACE`. This is not a huge win but it does remove wasted cycles.


apavlo:
I'm merging this because everyone is so busy doing other things. It's a minor change.

PR #418, opened 2016-12-22
Redundant predicate fix and some sql_test refactor

malin1993ml:
This PR fix the redundant predicate terms in index_scan_plan when some of the scanning condition can be satisfied by index search.

E.g.:
Let's say we have one table: "create table foo(id int primary key, value, int);".
Then "select * from foo where id = ?" doesn't need to check any predicate after primary key index search.

This PR also refactored the SQLTestsUtil a little bit to make it easier to check the execution results.


test/include/sql/sql_tests_util.h
```diff
@@ -66,6 +66,19 @@ class SQLTestsUtil {
   // A another simpler wrapper around ExecuteSQLQuery
   static Result ExecuteSQLQuery(const std::string query);
 
+  // Get the return value of one column as string at a given position
+  // NOTE: Result columns across different rows are unfolded into a single
+  // vector (vector<ResultType>).
+  static std::string GetResultValueAsString(
+      const std::vector<ResultType> &result, size_t idx) {
+    std::string value;
+    for (auto character : result[idx].second) {
```
apavlo: You can do the same thing with just this:

```
 std::string value(result[idx].second.begin(), result[idx].second.end());
```

src/include/expression/expression_util.h
```diff
@@ -101,6 +101,87 @@ class ExpressionUtil {
...
+   * searchable. That means there are only "and" conjunctions.
+   */
+  static AbstractExpression *RemoveTermsWithIndexedColumns(
+      AbstractExpression *expression, std::shared_ptr<index::Index> index) {
+    LOG_TRACE("Expression Type --> %s",
+              ExpressionTypeToString(expression->GetExpressionType()).c_str());
+
+    size_t children_size = expression->GetChildrenSize();
+    // We shouldn't have a expression that has more than two children
+    PL_ASSERT(children_size <= 2);
```
apavlo: Ask @MattPerron about this. I think that you can have more than two children for some functions.

src/include/expression/expression_util.h
```diff
@@ -101,6 +101,87 @@ class ExpressionUtil {
...
   }
 
+  /*
+   * This function removes all the terms related to indexed columns within an
+   * expression.
+   *
+   * NOTE: This should only be called after we check that the predicate is index
+   * searchable. That means there are only "and" conjunctions.
+   */
+  static AbstractExpression *RemoveTermsWithIndexedColumns(
```
apavlo: Thank you for putting this in `ExpressionUtil` and not in `SimpleOptimizer`

test/include/sql/sql_tests_util.h
malin1993ml: Thanks for pointing out. That's an interesting way.

src/include/expression/expression_util.h
malin1993ml: I've checked with Matt. You're right and I've fixed that.

Issue #419, opened 2016-12-22
Fix performance issue in value/type system


jessesleeping:
Several issues in the value/type system need to fix:
- `IsNull()` check is expensive and we can get rid of it by adding a null flag for every value type.
- Simple value/type functions should be inlined.
- `CheckComparable()` and `CheckInteger()` should be replaced by some assert.
- Varlen value type should allow "shallow copy" to save unnecessary memcpy overhead.

jessesleeping:
Fixed

PR #420, opened 2016-12-22
Fix value/type system issues

jessesleeping:
In this PR, we:
- Replace all `CheckComparable()` and `CheckInteger()` with asserts
- Add a null flag to value type and simplify the `IsNull()`'s check
- Inline simple functions in type/value
- Support shallow copy in varlen type

#419

### DIFF
--- a/src/include/type/value_factory.h
+++ b/src/include/type/value_factory.h
@@ -85,29 +85,40 @@ class ValueFactory {
   }
 
   static inline Value GetNullValueByType(Type::TypeId type_id) {
+    Value ret_value;
     switch (type_id) {
       case Type::BOOLEAN:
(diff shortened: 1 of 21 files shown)


src/include/type/value_factory.h
```diff
@@ -63,9 +63,9 @@ class ValueFactory {
   }
 
   static inline Value GetVarcharValue(
-      const char *value, UNUSED_ATTRIBUTE VarlenPool *pool = nullptr) {
+      const char *value, bool manage_data, UNUSED_ATTRIBUTE VarlenPool *pool = nullptr) {
```
apavlo: You should add a function `@param` comment that explains what `manage_data` means in case anybody calls it. Do this for all of the Get*Value methods.

PR #423, opened 2016-12-23
Value comparisons return enum (TRUE, FALSE, NULL)

MattPerron:


apavlo:
When the build passes, I will merge.

Issue #426, opened 2016-12-23
TPC-C gets super slow when customer index is enabled


jessesleeping:
When we enable the customer index, TPC-C throughput drops dramatically and become not scalable at all.

malin1993ml:
@jessesleeping In the email you said you tried to allocate the key tuple’s varlen in the executor context varlen pool, instead of the index’s one. But the number is still really bad. I don't quite understand why the number is still really bad in this case. There's no contention here, since one executor context can only be used by a single execution thread at one time, right?

jessesleeping:
When using executor context's varlen pool, the number is bad but scalable. The problem is varlen pool itself is much slower than the plain malloc.

jarulraj:
Jiexi -- If we use a latch free per-connection pool, would the performance
improve?

Or do you think that a new implementation is needed? If so, what are the
key differences in its design with respect to the current one?

On Dec 27, 2016 5:57 AM, "Jiexi Lin" <notifications@github.com> wrote:

When using executor context's varlen pool, the number is bad but scalable.
The problem is varlen pool itself is much slower than the plain malloc.

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/cmu-db/peloton/issues/426#issuecomment-269251486>, or mute
the thread

jessesleeping:
I think per connection varlen pool or per thread varlen pool can help improve our performance in other cases (e.g. allocating varlen for inserted titles), but not the problem stated here. 

A lock free implementation may help when background gc is freeing memory in varlen pool while the worker thread is allocating memory in its local pool.

Small things we can do is to optimize the usage of std::list and bitmap in current implementation. I think @MattPerron and I have the feeling that they can be more efficient. But I think what we really need to do is to re-consider what we want from the varlen pool and how can we distinguish it from general purpose malloc implementation?

MattPerron:
"When using executor context's varlen pool, the number is bad but scalable. The problem is varlen pool itself is much slower than the plain malloc."

Then we should just use pure malloc in these cases.

"But I think what we really need to do is to re-consider what we want from the varlen pool and how can we distinguish it from general purpose malloc implementation?"

If we are going to re-implement Varlen Pool we should do it right from the ground up and fix all the problems we have found so far. 

[...]

PR #427, opened 2016-12-23
Fix varlen pool usage in secondary index scan

jessesleeping:
From callgrind result on cmu-db/peloton/master (9f1c88f55e979a124a05ea44fc0420d7dd839cea), the problem is the varlen pool allocate. By tracing the code I find that [this line](https://github.com/cmu-db/peloton/blob/master/src/executor/index_scan_executor.cpp#L403) is allocating a temp varlen value for a tuple key from a per index varlen pool. It causes unnecessary contention over the index's varlen pool.

After switch to the executor context varlen pool, we get the performance back to normal level.

#426


jessesleeping:
.. It seems that I ran the wrong benchmark... with that fix tpc-c is still super slow...

jessesleeping:
Better than before... after the fix it can scale well...

Issue #425, opened 2016-12-23
Send error code for Aborted Txns


sid1607:
Rejected txns are happening because the server is aborting them when there is a conflict. The number is likely high for us because we are reading more tuples than we need to without that CUSTOMER index.

Send back a SQL State code when this occurs so that we know it's because of a conflict. Postgres sends status code "40001" which is defined as "serialization failure".

sid1607:
@apavlo - With this commit Oltpbench for tpcc now catches the conflict aborts as an **expected** error -

````
22:08:47,606 (NewOrder.java:397) DEBUG - Caught an expected error in New Order
org.postgresql.util.PSQLException: 
	at org.postgresql.core.v3.QueryExecutorImpl.receiveErrorResponse(QueryExecutorImpl.java:2310)
	at org.postgresql.core.v3.QueryExecutorImpl.processResults(QueryExecutorImpl.java:2023)
	at org.postgresql.core.v3.QueryExecutorImpl.execute(QueryExecutorImpl.java:217)
	at org.postgresql.jdbc.PgConnection.executeTransactionCommand(PgConnection.java:778)
	at org.postgresql.jdbc.PgConnection.rollback(PgConnection.java:811)
	at com.oltpbenchmark.benchmarks.tpcc.TPCCWorker.executeWork(TPCCWorker.java:81)
	at com.oltpbenchmark.api.Worker.doWork(Worker.java:381)
	at com.oltpbenchmark.api.Worker.run(Worker.java:291)
	at java.lang.Thread.run(Thread.java:745)
22:08:47,607 (Worker.java:412) DEBUG - NewOrder/02  0 - 40001
22:08:47,607 (Worker.java:438) DEBUG - Postgres serialization
````

However, YCSB with Read-Update workload fails because the benchmark simply throws every Exception, unlike the TPCC implementation - https://github.com/oltpbenchmark/oltpbench/blob/master/src/com/oltpbenchmark/benchmarks/ycsb/procedures/ReadRecord.java#L34

Should I modify YCSB to handle aborts?

apavlo:
What kind of abort do you mean?

I would probably focus on the things that we need to fix in the network layer.

sid1607:
I mean conflicting aborts - if read and update occur simultaneously over common records then the transaction manager might abort one of them. While TPCC-oltpbench can handle such serialization failures (error code: '40001'), the YCSB-oltpbench code has no mechanism of catching and handling such failures.
[...]

Issue #411, opened 2016-12-21
Support 'C' packet in the wire protocol


sid1607:
This packet is used to inform the server to close a statement or portal. While this doesn't affect TPCC's correctness, it should be implemented for completeness.

sid1607:
Fixed by #430

PR #430, opened 2016-12-23
[wire] Close + Serialization error reporting

sid1607:
Details in #411 and #425

### DIFF
--- a/src/type/types.cpp
+++ b/src/type/types.cpp
@@ -1275,4 +1275,9 @@ QuantifierType StringToQuantifierType(std::string str) {
   return QUANTIFIER_TYPE_NONE;
 }
 
+//===--------------------------------------------------------------------===//
+// Network Message types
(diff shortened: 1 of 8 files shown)


src/type/types.cpp
```diff
@@ -1275,4 +1275,9 @@ QuantifierType StringToQuantifierType(std::string str) {
   return QUANTIFIER_TYPE_NONE;
 }
 
+//===--------------------------------------------------------------------===//
+// Network Message types
+//===--------------------------------------------------------------------===//
+const std::string SqlStateErrorCodes::serialization_error =  "40001";
```
apavlo: Maybe we should make this an enum? There are going to be a lot more that we need to support:

https://www.postgresql.org/docs/9.6/static/errcodes-appendix.html

sid1607: Alright, I'll fix this.

Issue #431, opened 2016-12-24
IndexFactory only uses GenericKeys


apavlo:
The IntsKey stuff is commented out. I will fix.

apavlo:
All of the PRIMARY KEY indexes are getting stored with the same name:

```
[DEBUG] [src/index/index_factory.cpp:277:GetInstance()] 2016-12-24 08:42:58 - Index 'customer_pkey' => BWTREE::IntsKey<2>(s_w_id, s_i_id)
[DEBUG] [src/index/index_factory.cpp:277:GetInstance()] 2016-12-24 08:42:58 - Index 'customer_pkey' => BWTREE::IntsKey<2>(d_w_id, d_id)
[DEBUG] [src/index/index_factory.cpp:277:GetInstance()] 2016-12-24 08:42:58 - Index 'customer_pkey' => BWTREE::IntsKey<3>(no_w_id, no_d_id, no_o_id)
[DEBUG] [src/index/index_factory.cpp:277:GetInstance()] 2016-12-24 08:42:58 - Index 'customer_pkey' => BWTREE::IntsKey<2>(s_w_id, s_i_id)
```

wangziqi2013:
Is this fixed?

apavlo:
Yes, this is fixed in #432

PR #432, opened 2016-12-24
Index IntsKey Fixes

apavlo:
This PR re-enables the specialized IntsKey comparators that were commented out for the indexes. I have not tested whether this is fully working but I added a basic test case.

I also fixed it so that we don't hardcode the name of **every** to be `CUSTOMER_PKEY` ☹

This is for #431

### DIFF
--- a/script/oltpbenchmark/testbed/.gitignore
+++ b/script/oltpbenchmark/testbed/.gitignore
@@ -1,0 +1,3 @@
+outputfile*
+oltpbench
+*_config.xml
--- a/script/oltpbenchmark/testbed/measure-performance-tpcc.py
+++ b/script/oltpbenchmark/testbed/measure-performance-tpcc.py
@@ -5,14 +5,15 @@ import sys
 import time
 import subprocess
 import re
+import argparse
 from subprocess import call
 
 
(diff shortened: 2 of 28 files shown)


wangziqi2013:
We called GetTupleForComparison() only in Scan(). This function has not yet been implemented yet. However it should just return the pointer passed to it in SetFromKey()

apavlo:
@wangziqi2013 Not sure what you mean that it this function return the pointer from `SetFromKey()`? That would mean we are storing the pointer to the tuple and not the compact form of the key.

wangziqi2013:
@apavlo I mean if we re-construct IntsKey then it should return exactly the same tuple as what is used as the argument of SetFromKey(). Right now we call SetFromKey() in InsertEntry() and DeleteEntry().

apavlo:
@wangziqi2013 Yes, I am working on this. IntsKey's memory boundaries don't match with GenericKey. I am fixing IndexFactory.

apavlo:
I am here with @wangziqi2013 + @Michael-Tieying-Zhang + @malin1993ml. They all agree this should be merged. So in it goes!

PR #433, opened 2016-12-25
Adding Interface to Support Efficient Equi-Join with Index Scan

wangziqi2013:
This change adds a new function SetTupleColumnValue() which allows for the index scan executor to change a field inside the scan key, which is synthesized by index scan optimizer

Prior to this it is impossible to change a field that is not marked as "needs binding" during the initialization of IndexScanPlan, because we do not save information about columns not requiring a binding. After this interface is added the scan key is exposed to the programmer, and any value could be plugged into the scan key with an appropriate index column ID and value type. 

This change is added to support efficient equi-join on tables, where the scan key of the inner table has to be changed every time a new tuple is fetched from the outer table. Such dynamic alternation of scan key is not realized before.


wangziqi2013:
Please do not merge right now - let me add another two interfaces for updating low key and high key respectively. These two are needed to update the key for nonequi-join

apavlo:
Cool stuff. Let me know when you want me to review this.

wangziqi2013:
@apavlo @malin1993ml 

If this get passed then I'm done (Michael should call these three functions on his part to complete nested loop join on stock level table)

I am now working on IntsKey. I will add IntsKey to the BwTree repo first and then port it to Peloton

wangziqi2013:
@malin1993ml Good point.... Will add it tonight

wangziqi2013:
@malin1993ml I passed all test cases but failed the validator for using memset(). I will fix this and run another CI instance. After that we could merge this.

This PR contains CompactIntegerKey as well as what was stated above

wangziqi2013:
@malin1993ml @apavlo Done!

Michael-Tieying-Zhang:
@malin1993ml @apavlo Please take a look at this RP. Since Ziqi has changed a lot of code, I can't merge his brunch yesterday.

src/include/index/ints_key.h
```diff
@@ -20,213 +20,417 @@ namespace index {
...
+  }
+  
+ public:
+  
+  /*
+   * PrintRawData() - Prints the content of this key
+   *
+   * All contents are printed to stderr
+   */
+  void PrintRawData() {
```
apavlo: Please change this to `GetInfo()`

src/include/index/ints_key.h
```diff
@@ -236,281 +440,230 @@ class IntsKey {
...
- private:
 };
 
-/** comparator for Int specialized indexes. */
-template <std::size_t KeySize>
+/*
+ * class IntsComparator - Compares two compact integer key
+ */
+template <size_t KeySize>
 class IntsComparator {
```
apavlo: Rename to `CompactIntsComparator`

src/include/index/ints_key.h
```diff
@@ -236,281 +440,230 @@ class IntsKey {
...
 
-/**
- *
+/*
+ * class IntsEqualityChecker - Compares whether two integer keys are 
+ *                             equivalent
  */
-template <std::size_t KeySize>
+template <size_t KeySize>
 class IntsEqualityChecker {
```
apavlo: Rename to `CompactIntsEqualityChecker`

src/include/index/ints_key.h
```diff
@@ -236,281 +440,230 @@ class IntsKey {
...
-struct IntsHasher : std::unary_function<IntsKey<KeySize>, std::size_t> {
-  IntsHasher(index::IndexMetadata *metadata)
-      : schema(metadata->GetKeySchema()) {}
-
-  inline size_t operator()(IntsKey<KeySize> const &p) const {
-    size_t seed = 0;
-    for (size_t ii = 0; ii < KeySize; ii++) {
-      boost::hash_combine(seed, p.data[ii]);
+template <size_t KeySize>
+class IntsHasher {
```
apavlo: Rename to `CompactIntsHasher`

Issue #435, opened 2016-12-27
Transaction ROLLBACK doesn't work


apavlo:
If you rollback a txn from the terminal, it does not remove any of the changes that the txn made to the database:

```
postgres=# CREATE TABLE foo(id integer PRIMARY KEY, year integer);
CREATE TABLE foo(id integer PRIMARY KEY, year integer) 0
postgres=# INSERT INTO foo VALUES(1, 100);
INSERT INTO foo VALUES(1, 100) 1
postgres=# SELECT * FROM foo;
 id | year 
----+------
  1 |  100
(1 row)

postgres=# BEGIN;
BEGIN
postgres=# INSERT INTO foo VALUES(2, 200);
INSERT INTO foo VALUES(2, 200) 1
postgres=# SELECT * FROM foo;
 id | year 
----+------
  1 |  100
  2 |  200
(2 rows)

postgres=# ROLLBACK;
ROLLBACK
postgres=# SELECT * FROM foo;
 id | year 
----+------
  1 |  100
  2 |  200
(2 rows)
```

As you can see from this, the second tuple that we inserted during of the txn is still there even though we call `ROLLBACK`. If you log out and then log back in, it's still there:

```
$ psql "sslmode=disable" -U postgres -h dev4.db.pdl.cmu.local -p 15721
psql (9.5.5, server 9.5devel)
Type "help" for help.

postgres=# SELECT * FROM foo;
 id | year 
----+------
  1 |  100
  2 |  200
(2 rows)
```

apavlo:
Fixed.

PR #437, opened 2016-12-27
Transaction Rollback Fix in psql

sid1607:
Tcop uses the query type (i.e. the first word in the query string) to identify if the query is BEGIN, COMMIT or ROLLBACK to handle the transaction in the backend appropriately. Until now this was done using statement->SetQueryType(). However, the query type is not set while executing the simple query protocol (i.e. while interacting with psql). For this reason, these Transaction switches went unnoticed by tcop. This PR fixes this issue by setting query type on statement construction for the simple query protocol. #435

### DIFF
--- a/src/common/statement.cpp
+++ b/src/common/statement.cpp
@@ -10,6 +10,7 @@
 //
 //===----------------------------------------------------------------------===//
 
+#include <cstdio>
 #include "common/statement.h"
 #include "common/logger.h"
 #include "planner/abstract_plan.h"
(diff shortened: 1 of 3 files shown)


src/common/statement.cpp
```diff
@@ -18,49 +19,53 @@ namespace peloton {
...
+    : statement_name_(statement_name), query_string_(query_string) {
+  ParseQueryType(query_string_, query_type_);
+}
 
 Statement::~Statement() {}
 
-std::vector<FieldInfoType> Statement::GetTupleDescriptor() const {
-  return tuple_descriptor;
+void Statement::ParseQueryType(const std::string &query_string,
+                               std::string &query_type) {
```
apavlo: Are you passing query_type by reference here to avoid having to allocate a new variable?

src/include/common/statement.h
```diff
@@ -68,22 +69,22 @@ class Statement {
...
-  std::string statement_name;
+  std::string statement_name_;
 
   // query string
-  std::string query_string;
+  std::string query_string_;
 
   // first token in query
-  std::string query_type;
+  std::string query_type_;
```
apavlo: Should we use `type::StatementType` instead of a string?

apavlo:

src/common/statement.cpp
sid1607: Yes

src/include/common/statement.h
sid1607: Yes, this is the long-term goal. As I had mentioned to you that day, using type::StatementType for situations like these requires us to support multi-query statements first. Once I get that patch in, all these things can be made cleaner.

sid1607: Since it will take me a little longer (perhaps after the new year) to get that patch in, we will have to live with this hack till then?

apavlo: Yes this is fine.

PR #439, opened 2016-12-28
Special case comparison on first/last tuple in index range scan

malin1993ml:
This PR adds the logic to handle left/right open range index scan in the IndexScanExecutor. Special flags will be set when constructing the IndexScanPlan in the optimizer, and then IndexScanExecutor will take the flags in the plan and perform pruning on the head/tail of the return tuple list if the flags indicate so.

We have to do that because the raw index can only return tuples for a close-boundary range. The logic was handled by index::Compare(), but that function will be removed soon by @wangziqi2013 .

@wangziqi2013 After you remove the index::Compare(), it seems to me that we no longer need IndexMetadata::GetTupleToIndexMapping and some related interfaces. Can you double check that and remove unnecessary code? I cannot remove them now since the compilation would fail when we still have index::Compare().


apavlo:
I will review and merge in the morning.

PR #441, opened 2016-12-29
Refactoring memory pool module

jarulraj:
This PR:

- Adds a new abstract pool class
- Adds an ephemeral pool class that is a wrapper around jemalloc and cleans up after itself

### DIFF
--- a/src/gc/gc_manager.cpp
+++ b/src/gc/gc_manager.cpp
@@ -11,51 +11,10 @@
 //===----------------------------------------------------------------------===//
 #include "gc/gc_manager.h"
 
-#include "catalog/schema.h"
-#include "type/types.h"
-#include "type/value.h"
-#include "type/varlen_pool.h"
-#include "storage/tile.h"
-#include "storage/tile_group.h"
-
 namespace peloton {
 namespace gc {
 
-// Check a tuple and reclaim all varlen field
-void GCManager::CheckAndReclaimVarlenColumns(storage::TileGroup *tg, oid_t tuple_id) {
-  oid_t tile_count = tg->tile_count;
-  oid_t tile_col_count;
-  type::Type::TypeId type_id;
-  char *tuple_location;
-  char *field_location;
-  char *varlen_ptr;
 
-  for (oid_t tile_itr = 0; tile_itr < tile_count; tile_itr++) {
-    const catalog::Schema &schema = tg->tile_schemas[tile_itr];
-
-    tile_col_count = schema.GetColumnCount();
-
-    storage::Tile *tile = tg->GetTile(tile_itr);
-    PL_ASSERT(tile);
-    for (oid_t tile_col_itr = 0; tile_col_itr < tile_col_count; ++tile_col_itr) {
-      type_id = schema.GetType(tile_col_itr);
-
-      if ((type_id != type::Type::TypeId::VARCHAR && type_id != type::Type::TypeId::VARBINARY)
-          || (schema.IsInlined(tile_col_itr) == true)) {
-        // Not of varlen type, or is inlined, skip
-        continue;
-      }
-      // Get the raw varlen pointer
-      tuple_location = tile->GetTupleLocation(tuple_id);
-      field_location = tuple_location + schema.GetOffset(tile_col_itr);
-      varlen_ptr = type::Value::GetDataFromStorage(type_id, field_location);
-      // Call the corresponding varlen pool free
-      if (varlen_ptr != nullptr) {
-        tile->pool->Free(varlen_ptr);
-      }
-    }
-  }
-}
 
 }
 }
--- a/src/include/type/ephemeral_pool.h
+++ b/src/include/type/ephemeral_pool.h
@@ -1,0 +1,64 @@
+//===----------------------------------------------------------------------===//
+//
+//                         Peloton
+//
+// ephemeral_pool.h
+//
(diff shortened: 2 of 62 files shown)


src/include/type/varlen_pool.h
```diff
@@ -96,31 +84,11 @@ class VarlenPool {
...
-private:
+public:
   // All these fields are implementation specific.
   // This class must be thread-safe, very very fast and provide some form of
   // compaction or garbage-collection.
 
-  // Internal memory allocation
-  void *AllocateHelper(size_t size);
-
-  // Internal memory deallocation
```
jessesleeping: I don't think we need to remove all the old varlen code including the ref count optimization. A better way would be to abstract an interface, keep the old code and implement a new "ephemeral" pool.

src/gc/gc_manager.cpp
```diff
@@ -11,51 +11,10 @@
...
-      type_id = schema.GetType(tile_col_itr);
-
-      if ((type_id != type::Type::TypeId::VARCHAR && type_id != type::Type::TypeId::VARBINARY)
-          || (schema.IsInlined(tile_col_itr) == true)) {
-        // Not of varlen type, or is inlined, skip
-        continue;
-      }
-      // Get the raw varlen pointer
-      tuple_location = tile->GetTupleLocation(tuple_id);
-      field_location = tuple_location + schema.GetOffset(tile_col_itr);
```
jessesleeping: We definitely need this piece of code after we stop hacking the varlen pool because it help the gc manager reclaim memory from the varlen pool. It seems that the "ephemeral" pool does not free memory at all.

src/include/type/varlen_pool.h
jarulraj: I agree that ref count optimization can improve performance, but I think that we need to think about the design more carefully. For instance, when I switched over from `varlen pool` to global memory allocator (`nullptr`) in executor context, things broke. 
Furthermore, as you had noted in another issue, this is not a trivial change and can introduce memory leaks if not done carefully. In fact,

1) I think that none of the abstract pool implementations must clean up after themselves -- otherwise folks will start leaking memory all over the place as is already the case.

2) I am trying to fix such leaks right now in the `TileGroup`. In `SetValue`, when we overwrite an `varlen` field, we should `automatically` release the memory associated with the old value.

src/gc/gc_manager.cpp
jarulraj: Right now, it actually does clean up after itself. I am not fixing leaks, so that it doesn't have to.

jessesleeping: My bad, I missed the location vector inside the new pool. 

I think that before we have tile group compaction, gc may still need to care about such low level things.

src/include/type/ephemeral_pool.h
```diff
@@ -0,0 +1,64 @@
...
+
+  // Returns the provided chunk of memory back into the pool
+  void Free(UNUSED_ATTRIBUTE void *ptr){
+
+  }
+
+public:
+
+  // Buffer lists
+  std::vector<char*> locations_;
```
jessesleeping: So this pool is not concurrent?

src/include/type/ephemeral_pool.h
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+  }
+
+  // Allocate a contiguous block of memory of the given size. If the allocation
+  // is successful a non-null pointer is returned. If the allocation fails, a
+  // null pointer will be returned.
+  void *Allocate(size_t size){
+    auto location = new char[size];
+    locations_.push_back(location);
```
jessesleeping: Do we ensure that this pool is not used concurrently?

apavlo:
@jessesleeping Is the right person to review this since he was the last one in this part of the code.

jarulraj:
@jessesleeping A problem that I encountered is that neither a tuple nor a varlen field has any reference to the pool from which memory was allocated. So, reclaiming memory incrementally is hard unless we add some tracking. 

How is the GC reclaiming memory w.r.t.:
(1) tile pool
(2) index pool

Unless the tile or index is deleted, varlen fields of unreferenced versions are reclaimed by the GC?
I suspect that this is the reason behind issue #442. 

@apavlo Sure, I am discussing this with him. Wanted to also keep you all in the loop.

src/include/type/ephemeral_pool.h
jarulraj: Yes, we should.

src/include/type/ephemeral_pool.h
jarulraj: Yes. This is designed for executor context.

src/include/type/varlen_pool.h
jarulraj: I agree that this is an unnecessary complication and may be a reason for memory leaks. What do you think, @yingjunwu ?

jessesleeping:
Under our current implementation, the GC manager can get the per tile varlen pool from a tuple's item pointer and the column id (item pointer -> tile group -> tile -> varlen pool). In fact, that's what we did in the GCManager::CheckAndReclaimVarlenColumns(). 

For the pool in index, I think in current master branch no one is using that pool. May be we can remove that pool from index. Currently the GC manager deletes entries of recycled tuple in all indexes via the index's Delete API.

jarulraj:
This is good to go if it passes all the checks.

Issue #442, opened 2016-12-29
DBMS Memory Usage Always Increases


apavlo:
The DBMS process keeps growing in memory even if the logical database size does not increase.

For this test, I first loaded a 1GB YCSB database (scalefactor=1000) with OLTP-Bench. This was the size of the peloton process in htop before I started executing txns (yes I know there are better ways to get this measurement):
```
VIRT: 3708M
RES:  2227M
```

I then had a single worker thread execute only the `UpdateRecord` transaction for 20 minutes. The throughput was 11440 txn/sec. Afterwards, the DBMS size was this:
```
VIRT: 22.9G
RES:  21.4G
```

This should not happen since `UpdateRecord` never inserts or deletes a tuple. To make sure that the database was the same size, I ran a simple count query and it matched what should be there:

```
postgres=# SELECT COUNT(*) FROM USERTABLE;
 count(*) 
----------
  1000000
(1 row)
```

Something is obviously not right.

apavlo:
@yingjunwu  - Latest version is still not reclaiming memory. CPU is not spiked at 100% anymore. Tested with 1fc29ad10ac86b7abaf3493a9ed24122e596189b 

**BEFORE**
```
VIRT: 3628M
RES:  3140M
```

**AFTER:***
```
VIRT: 20.6G
RES:  20.2G
```

AFAIK, the GC is supposed to be turned on:

https://github.com/cmu-db/peloton/blob/master/src/gc/gc_manager_factory.cpp#L19

yingjunwu:
I'll take a look at it.

PR #443, opened 2016-12-30
turn on garbage collection

yingjunwu:
#442

### DIFF
--- a/script/testing/jdbc/src/PelotonTest.java
+++ b/script/testing/jdbc/src/PelotonTest.java
@@ -688,15 +688,15 @@ public class PelotonTest {
       stmt.setString(1,"Cool");
       stmt.setInt(2,i);
       stmt.addBatch();
+    }
+    System.out.println("All Good Here");
+    int[] res = stmt.executeBatch();
+    for(int i=0; i < res.length; i++) {
+      if (res[i] < 0) {
+        throw new SQLException("Query "+ (i+1) +" returned " + res[i]);
       }
-      System.out.println("All Good Here");
-      int[] res = stmt.executeBatch();
-      for(int i=0; i < res.length; i++) {
-        if (res[i] < 0) {
-          throw new SQLException("Query "+ (i+1) +" returned " + res[i]);
-        }
-        System.out.println(res[i]);
-      }
+      System.out.println(res[i]);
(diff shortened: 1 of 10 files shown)


apavlo:
New YCSB Update Numbers:

**BEFORE**:
```
VIRT: 3736M
RES:  2275M
```

**AFTER**:
```
VIRT: 3772M
RES:  2412M
```

But if you just let the system sit idle after running the benchmark, there is one CPU spiked at 100% and the memory seems to be increasing. After letting it run for 10 minutes, the memory usage increased to 2667M.

So I think there is a problem with the GC thread.

yingjunwu:
You mean the memory keeps increasing even if the system is idle??? It's
reasonable to observe one thread is always at 100% because it's the GC
thread and it keeps checking whether there's expired versions.

apavlo:
Yes, the memory keeps increasing when idle. It doesn't happen when you just turn the system on without executing anything.

I don't think it's a good idea for the GC to spike the CPU at 100% if there are no txns. That looks bad. You may want to consider adding a simple counter that keeps track of the # of times that it ran but no txns were executed and then sleep for a few milliseconds.

yingjunwu:
How do you let the system sit idle after running the benchmark? I think in
OLTP bench, the system will be closed once the benchmark finishes running.

Originally we let the gc thread to sleep periodically, but I find that in
this case the gc thread may not catch up with the execution thread under
write-intensive workload. I will tweak the gc code and let it sleep.

apavlo:
We're not building a system just to run benchmarks. Think about if somebody was really using this. They turn on our system and it immediately spikes to 100% before they do anything. That looks bad.

Only sleep for a little bit if the last 10 times it ran it did not find anything. Then when txns start executing, it will not sleep and should be able to keep up.

Long term we are going to be able to dynamically adjust the # of threads that we use for GC.

apavlo:
@yingjunwu As part of this, can you write some test cases for `TransactionLevelGCManager`. This code is completely untested right now.

yingjunwu:
#447

yingjunwu:
@apavlo you can merge this PR now.

apavlo:
We forgot to rebase :-(

PR #445, opened 2016-12-30
StockLevel fix with NestedLoopJoin and IndexPredicate

Michael-Tieying-Zhang:
This PR guarantees the correctness of TPC-C StockLevel txns and is tested using JDBC clients. Currently, stocklevel is transformed as two indexscan executors. I will add seqscan support for NestedLoopJoin later.


apavlo:
Thanks! You say "tested with JDBC clients". Where is that test case? I only see C++ files.

Michael-Tieying-Zhang:
That is not a unit test. It is a JDBC client and I examine the correctness using both JDBC and PSQL. Do you wanna me to commit the JDBC client code? and to where?

apavlo:
How hard would it be to rewrite your tests to use the `SQLTestsUtil::ExecuteSQLQuery()` in C++? That would help improve our coverage.

Right now we really don't have a full testing suite for JDBC that is integrated. There is only one file that we execute.

Michael-Tieying-Zhang:
I have seen there is `SQLTestsUtil::ExecuteSQLQuery()`, like this 

>   SQLTestsUtil::ExecuteSQLQuery("SELECT min(b) from test", result,
>                                 tuple_descriptor, rows_affected, error_message);
>   // Check the return value
>   EXPECT_EQ(result[0].second[0], '1');

It seems we directly get the result from `SQLTestsUtil::ExecuteSQLQuery`. But we need to execute it like the JDBC code: 

> pst = con.prepareStatement(query3);
> 
> pst.setInt(1, 1);   // OL_W_ID
> pst.setInt(2, 1);   // OL_D_ID
> pst.setInt(3, 2);   // OL_O_ID
> pst.setInt(4, 1);   // OL_O_ID
> pst.setInt(5, 1);   // S_W_ID
> pst.setInt(6, 100);
> 
> boolean isResult = pst.execute();
> 
> rs = pst.getResultSet();
>

apavlo:
Does it not work if you use constants?

On Friday, December 30, 2016 12:27:57 PM EST Michael wrote:
> I have seen there is `SQLTestsUtil::ExecuteSQLQuery()`, like this 
> 
> >   SQLTestsUtil::ExecuteSQLQuery("SELECT min(b) from test", result,
> >                                 tuple_descriptor, rows_affected, error_message);
> >   // Check the return value
> >   EXPECT_EQ(result[0].second[0], '1');
> 
> It seems we directly get the result from `SQLTestsUtil::ExecuteSQLQuery`. But we need to execute it like the JDBC code: 
> 
> > pst = con.prepareStatement(query3);
> > 
> > pst.setInt(1, 1);   // OL_W_ID
> > pst.setInt(2, 1);   // OL_D_ID
> > pst.setInt(3, 2);   // OL_O_ID
> > pst.setInt(4, 1);   // OL_O_ID
> > pst.setInt(5, 1);   // S_W_ID
> > pst.setInt(6, 100);
> > 
> > boolean isResult = pst.execute();
> > 
> > rs = pst.getResultSet();
> > 
> 
> 


-- 
Andy Pavlo
pavlo@cs.cmu.edu

Michael-Tieying-Zhang:
You mean directly use sql request rather than prepare statement? I suppose it doesn't work. Because simple optimizer use the parameters in prepare statement: [https://github.com/cmu-db/peloton/blob/master/src/optimizer/simple_optimizer.cpp#L940](https://github.com/cmu-db/peloton/blob/master/src/optimizer/simple_optimizer.cpp#L940). But I'll try.

Michael-Tieying-Zhang:
I have tried that. It doesn't work. 

> SELECT * FROM ORDER_LINE, STOCK WHERE OL_W_ID = 1 AND OL_D_ID = 1 AND OL_O_ID < 2 AND OL_O_ID >= 1 AND S_W_ID = 1 AND S_I_ID=OL_I_ID AND AND S_QUANTITY < 200;
> Error parsing SQL statement

apavlo:
Check it into `script/testing/jdbc/src/`

Michael-Tieying-Zhang:
System doesn't crash, but the result is  not right.

malin1993ml:
I see. Just remember to fix it. And I've thought about the join_column_ids_right_ field in the nested_loop_join plan. I understand why you store the physical ID there, but after some consideration I think I have another comment on that. Please wait a bit.

Michael-Tieying-Zhang:
Looling forward that!

src/optimizer/simple_optimizer.cpp
```diff
@@ -1014,26 +1006,14 @@ SimpleOptimizer::CreateHackingNestedLoopJoinPlan(
...
-  expression::TupleValueExpression* right_table_attr_1 =
-      new expression::TupleValueExpression(type::Type::INTEGER, 1, 0);
-
-  std::unique_ptr<const expression::AbstractExpression> join_predicate(
-      new expression::ComparisonExpression(EXPRESSION_TYPE_COMPARE_EQUAL,
-                                           left_table_attr_4,
-                                           right_table_attr_1));
-
   std::vector<oid_t> join_column_ids_left = {0};   // I_I_ID in the result
   std::vector<oid_t> join_column_ids_right = {1};  // S_I_ID in the table
```
malin1993ml: So instead of storing the physical offset in the table of the right child operator, I think it's better to store the logical offset of the output schema here as well, just as with the left child operator. In that case the two variables are consistent, and I think it's better to unify the logic for both index NestedLoopJoin and seq NestedLoopJoin. Otherwise in the seqscan version of NestedLoopJoin, you have to know the schema of the base table of the right child operator to perform the join here correctly, which is not desirable.

Well if you store the logical offset, you can still get the physical offset in the index_scan_plan with IndexScanPlan::column_ids_. The only disadvantage here is that we have to select the join column of the right child to be one of its output schema columns, even though sometimes we don't need to output the join column. But I think returning one extra column here is acceptable (we have to scan it anyway).

Michael-Tieying-Zhang: The goal of storing  the physical column id is to pass them to UpdatePredicate: [https://github.com/cmu-db/peloton/pull/445/files#diff-847f567350162b7e387ad2c64d313b7eR130](https://github.com/cmu-db/peloton/pull/445/files#diff-847f567350162b7e387ad2c64d313b7eR130). 

Even for seqscan, we should use the same API (UpdatePredicate) to update seqscan's predicate which still requires physical column id.

malin1993ml: Yes I understand that. I'm saying even though we store the logical id here, we can still get the physical column id inside [this function](https://github.com/cmu-db/peloton/pull/445/files#diff-eefe560c257a7f54a22b8f0de670bd01R703) with [this field](https://github.com/cmu-db/peloton/blob/master/src/include/planner/index_scan_plan.h#L172). Basically the logical id in NestedLoopPlan stores the offset in IndexScanPlan::column_ids_, and the IndexScanPlan::column_ids_ stores the physical id. That shouldn't require lots of changes.

I think in general it's not good to store an offset with respect to the input of the child executor. That requires extra exposure of information unless we hard-code it. All other executors only refer the logical offset with respect to the output schema of their children, and don't need to know how the input of their children look like. I basically don't want to add that exposure just for this executor.

malin1993ml: Also, consider the the senario when there's multi-way join or nested query. In that case the right child itself might be a join or some other executors. So there might not be "physical ids" in the first place.

Michael-Tieying-Zhang: It's reasonable.

malin1993ml: Cool:)

malin1993ml:
@Michael-Tieying-Zhang Looks good. Can you rebase and then I'll merge?

PR #446, opened 2016-12-30
Remove Index::Compare Calls + MaskedTuple Enhancement

apavlo:
Ok here's what we got in this one:

1. Remove the call to `Index::Compare` inside of `BwTreeIndex::Scan`. This works because of @malin1993ml's fixes in #439 

2. New `MaskedTuple` wrapper that takes in an `AbstractTuple` and a mask of column offsets and then converts any calls to `AbstractTuple::GetValue` to use the proper offset. This is a way to "fake" a tuple that has exactly the schema that is needed for a key look-up in an index. Using `MaskedTuple` completely removes the creation of `Value` objects inside of `IndexScanExecutor::ExecSecondaryIndexLookup`.

I am going to check whether we can use `MaskedTuple` in the `GenericComparator` next.

As an aside, I am not sure whether `MaskedTuple` is the right name to call this. I was thinking maybe `MappedTuple` but that doesn't sound right either...


wangziqi2013:
How about calling it "VirtualTuple"?

apavlo:
When this build passes, I am going to merge this PR into the master. This has a ton of fixes and code clean up that I want to merge now. I will then create a separate PR for the new `FastGenericComparator`.

apavlo:
ok I am merging this mofo now...

Issue #453, opened 2017-01-03
Fix B+Tree Index or Remove it


apavlo:
I think that we should keep the `BTreeIndex` around in case we need something more simple than the `BwTreeIndex`. But it currently does not work with any of our test cases. I think that @jarulraj also has a `SkipListIndex` somewhere too.

We should either fix these indexes and re-enable them in `IndexPerformanceTests` or remove them.

apavlo:
Here are the errors if you re-enable `INDEX_TYPE_BTREE` in the `IndexPerformanceTests`:

```
/home/pavlo/Documents/Peloton/Github/peloton/test/performance/index_performance_test.cpp:288: Failure
Value of: location_ptrs.size()
  Actual: 0
Expected: num_thread * num_key
Which is: 1048576
[INFO ] [test/performance/index_performance_test.cpp:293:TestIndexPerformance()] 2017-01-03 09:46:47 - InsertTest1 :: Type=BTREE; Duration=0.21
[INFO ] [test/performance/index_performance_test.cpp:314:TestIndexPerformance()] 2017-01-03 09:46:47 - DeleteTest1 :: Type=BTREE; Duration=0.79
/home/pavlo/Documents/Peloton/Github/peloton/test/performance/index_performance_test.cpp:330: Failure
Value of: location_ptrs.size()
  Actual: 0
Expected: num_thread * num_key
Which is: 1048576
[INFO ] [test/performance/index_performance_test.cpp:335:TestIndexPerformance()] 2017-01-03 09:46:47 - InsertTest2 :: Type=BTREE; Duration=1.02
[INFO ] [test/performance/index_performance_test.cpp:356:TestIndexPerformance()] 2017-01-03 09:46:48 - DeleteTest :: Type=BTREE; Duration=1.74
[  FAILED  ] IndexPerformanceTests.BTreeMultiThreadedTest (1746 ms)
```

wangziqi2013:
The B+Tree index could be used as a backup plan, in a sense that if the fanout for a key in a table is too large (e.g. > 100) then BwTree has miserable performance on Insert() and Delete() because it has to keep operations atomic on a large set of records. In this case maybe using a B+Tree might be beneficial. This could be made as a knob for tuning and let the user make decision

apavlo:
Right, but we would have to modify it to make it concurrent.

tomasic:
Large fan out is important in disk based b tree indexing. That's because
processing the index page is cheap compared to a disk read. (100th the
[...]

Issue #429, opened 2016-12-23
Potential problem of varlen pool with reference count


jessesleeping:
Currently every tile has its own varlen pool. With reference count, memory addresses in a tile's varlen pool can be referenced by tuples in a different tile group. However, the tile/varlen pool has no knowledge about these external references. It's possible that a tile group destructs (due to GC or compaction) and destroys its varlen pools, while tuples in other tile groups are still referencing  certain addresses in those pools.

apavlo:
We need to have a long discussion about how we want to maintain VarlenPools in the future. I feel like we keep throwing band-aids on it without ever really thinking it through carefully.

jessesleeping:
I think we are not the first one to find this problem. Other in-memory databases must have the same issue. I feel like we are reimplementing malloc in our varlen pool...

jarulraj:
@jessesleeping: when are you performing a shallow copy of a varlen?

jessesleeping:
At ProjectInfo's Evaluate method. I overloaded the Evaluate method to specialized it for update executor and did shallow copy there. To support shallow copy, I added API all the way down from ContainerTuple to Tile. At Tile I directly performed copy without creating a Value object. You can check out the diff of the PR to see the detailed implementation.

PR #454, opened 2017-01-03
Removed varlen pool, btree, and libcds

jarulraj:
This PR:
- Resolves issue #453 by removing unused data structures.
- Resolves issue #429 by getting rid of varlen pool and associated shallow copying


jarulraj:
@jessesleeping Before we merge this, can you check `oltpbench`?

jarulraj:
I think that we should run oltpbench with multiple terminals as part of our travis-based CI.

test/type/varlen_pool_test.cpp
```diff
@@ -1,391 +0,0 @@
...
-#include <limits.h>
-#include <pthread.h>
-#include "type/varlen_pool.h"
-#include "gtest/gtest.h"
-#include "common/harness.h"
-
-namespace peloton {
-namespace test {
-
-class VarlenPoolTests : public PelotonTest {};
```
apavlo: Why not keep this for the new `EphemeralPool`? The API should be the same, right?

CMakeLists.txt
```diff
@@ -13,8 +13,8 @@ project(Peloton CXX C)
 include(CTest)
 
 # ---[ Peloton version
-set(PELOTON_TARGET_VERSION "0.0.3" CACHE STRING "Peloton logical version")
```
apavlo: Why do we want to bump the version number? This seem premature

test/type/varlen_pool_test.cpp
jarulraj: Sure, I adapted that test case for `EphemeralPool`.

CMakeLists.txt
jarulraj: We need to add a release system. There have been 1000+ commits since `0.0.3`. We should either stop versioning, or bump it periodically.

Issue #457, opened 2017-01-04
Adding Scan Limit Support


wangziqi2013:
This proposal adds an extra interface to the index class called ScanLimit()

ScanLimit() takes the same set of argument as Scan(), with the addition of another argument called limit. The index is responsible for identifying whether ScanLimit() has the semantics of max() or min().

apavlo:
@wangziqi2013 I think this is complete, right?

wangziqi2013:
@apavlo Yes. I will close this.

PR #461, opened 2017-01-05
Fix Limit issue and merge with new BwTree

Michael-Tieying-Zhang:
This PR optimizes the sql statement with order by limit 1 and also merged ziqi's new BwTree. A simple performance test is done and the result shows that the performance is improved a lot! Let's test the true performance after merging this PR with master.


wangziqi2013:
#457

PR #462, opened 2017-01-05
drop sql test

AngLi-Leon:


apavlo:
Why did you close this?

AngLi-Leon:
I find a mistake in my test case. Will change it and PR again.

apavlo:
No, just keep the same PR and push a new commit. No need to close/open a new one.

test/sql/drop_sql_test.cpp
```diff
@@ -0,0 +1,76 @@
...
+#include "planner/create_plan.h"
+
+#include "sql/sql_tests_util.h"
+
+namespace peloton {
+namespace test {
+
+class DropSQLTests : public PelotonTest {};
+
+TEST_F(DropSQLTests, DropSQLTest) {
```
apavlo: Rename to DropTableTest

test/sql/drop_sql_test.cpp
```diff
@@ -0,0 +1,76 @@
...
+
+  // Insert and query from that table
+  SQLTestsUtil::ExecuteSQLQuery("INSERT INTO test VALUES (1, 10);", result,
+                                tuple_descriptor, rows_affected, error_message);
+  SQLTestsUtil::ExecuteSQLQuery("SELECT * FROM test;", result, tuple_descriptor,
+                                rows_affected, error_message);
+  EXPECT_EQ(result[0].second[0], '1');
+
+  // Drop the table
+  SQLTestsUtil::ExecuteSQLQuery("DROP TABLE test;");
```
apavlo: Does this query return a result? If so, then we should check it.

test/sql/drop_sql_test.cpp
```diff
@@ -0,0 +1,76 @@
...
+TEST_F(DropSQLTests, DropSQLTest) {
+  catalog::Catalog::GetInstance()->CreateDatabase(DEFAULT_DB_NAME, nullptr);
+
+  // Create a table first
+  SQLTestsUtil::ExecuteSQLQuery("CREATE TABLE test(a INT PRIMARY KEY, b INT);");
+
+  // Check the table in catalog
+  EXPECT_EQ(catalog::Catalog::GetInstance()
+                ->GetDatabaseWithName(DEFAULT_DB_NAME)
+                ->GetTableCount(),
```
apavlo: You should look for the actual Table instead of just getting the count in case we start adding internal tables.

Issue #358, opened 2016-11-26
Pre-declared read only transaction support


jessesleeping:
We need the following functionality to support pre-declared read only transaction

- The **epoch manager** should assign a proper timestamp to the read-only transaction to ensure that this transaction reads a consistent read-only snapshot of the database.
- The **transaction manager** should ignore all the read performed by a read-only transaction. Since the transaction is read-only and it's running on a consistent read-only snapshot, there is no rw conflicts or ww conflicts.
- The **garbage collection manager** should coordinate with the epoch manager to ensure that there is always an available read only snapshot for incoming read only transactions.

apavlo:
Merged with #359

PR #463, opened 2017-01-05
Partial Bugfix #358

PauloAmora:
In `SimpleOptimizer`: Adding the order by column explicitely into the column map, to allow materialization and sorting.

In `OrderByExecutor`: Use only the columns in select for output_schema.


apavlo:
Thanks for sending this! It looks like you have a lot of redundant stuff from our latest commits. We also don't want to pollute our history with your qmake files.

Can you checkout a clean copy of the project and then just add the patches for your fix without the other extra stuff?

PauloAmora:
Sure, I'll look into that.

apavlo:
I'm not sure why this is marked as merged?

Issue #385, opened 2016-12-07
ORDER BY is Broken


apavlo:
We don't support `ORDER BY` with a `SELECT *` query:
```
postgres=# create table foo(id integer PRIMARY KEY, year integer);
create table foo(id integer PRIMARY KEY, year integer) 0
postgres=# insert into foo values(1, 100);
insert into foo values(1, 100) 1
postgres=# insert into foo values(2, 200);
insert into foo values(2, 200) 1
postgres=# insert into foo values(3, 150);
insert into foo values(3, 150) 1
postgres=# SELECT * FROM foo ORDER BY year;
server closed the connection unexpectedly
```
The server gives the following error:
```
[ERROR] [src/optimizer/simple_optimizer.cpp:285:BuildPelotonPlanTree()] 2016-12-06 22:13:27 - Not supported: Ordering column is not an element of select list!
```
But if you make the `SELECT` include the column you want to sort on, then it works:
```
postgres=# SELECT id, year FROM foo ORDER BY year;
 id | year 
----+------
  1 |  100
  3 |  150
  2 |  200
(3 rows)
```
When you remove `year` from the output list, we crash again:
```
postgres=# SELECT id FROM foo ORDER BY year;
server closed the connection unexpectedly
```

sid1607:
I will look into this

PauloAmora:
I've been working on this for a while, and I think I may have a partial fix, for the SELECT id FROM foo ORDER BY year; case. I managed to get order by working and am currently doing code checks for committing soon.
Screenshot is attached.
![screenshot peloton](https://cloud.githubusercontent.com/assets/2302077/21663959/33fe1932-d2c2-11e6-9bca-a11a2e76bb8a.PNG)

apavlo:
@PauloAmora Great! I think it should be something wrong in the `SimpleOptimizer`, right?

PauloAmora:
Not quite wrong, I changed the order by logic to look to all columns in the schema instead of only the select-projected ones and in the OrderBy executor changed the input_schema_ to be composed of only the columns contained in the output_columns contained in the order_by plan.

PauloAmora:
@apavlo How can I run more comprehensive tests so I can make sure I won't break anything on your end?

apavlo:
(1) Run ``make check`` locally.

(2) Submit a pull request. That will fire off a complete test/build on Travis.

Michael-Tieying-Zhang:
[...]

PR #465, opened 2017-01-05
#385 Partial Bugfix

PauloAmora:
This PR resolves issue #385.

In` SimpleOptimizer`: Adding the order by column explicitely into the column map, to allow materialization and sorting.

In `OrderByExecutor`: Use only the columns in select for output_schema.

As requested, only the necessary commit.


PauloAmora:
I just figured that the modification done on `SimpleOptimizer` must be applied in 2 places. There is an `order and limit` condition besides the order condition. I tested order and limit with the modification and it is working.

src/executor/order_by_executor.cpp
```diff
@@ -57,7 +57,8 @@ bool OrderByExecutor::DExecute() {
   PL_ASSERT(input_tiles_.size() > 0);
 
   // Returned tiles must be newly created physical tiles,
-  // which have the same physical schema as input tiles.
+  // which not always have the same physical schema as input tiles.
```
jarulraj: We can rephrase this as: "NOTE: the schema of these tiles might not match the input tiles when some of the order by columns are not be part of the output schema"

src/optimizer/simple_optimizer.cpp
```diff
@@ -288,11 +297,13 @@ std::shared_ptr<planner::AbstractPlan> SimpleOptimizer::BuildPelotonPlanTree(
               ((expression::TupleValueExpression*)select_stmt->order->expr)
                   ->GetColumnName());
           for (size_t column_ctr = 0;
-               column_ctr < select_stmt->select_list->size(); column_ctr++) {
-            std::string col_name((
-                (expression::TupleValueExpression*)select_stmt->select_list->at(
-                    column_ctr))->GetColumnName());
-            if (col_name == sort_col_name) key.push_back(column_ctr);
+               column_ctr < target_table->GetSchema()->GetColumns().size(); column_ctr++) {
```
jarulraj: The original code looks ugly with long lines. Can you break it up?
Like `auto schema_size = target_table->GetSchema()->GetColumns().size()` and `column_ctr < schema_size` etc.
`target_table->GetSchema()->GetColumns().at(column_ctr)`.

src/optimizer/simple_optimizer.cpp
```diff
@@ -288,11 +297,13 @@ std::shared_ptr<planner::AbstractPlan> SimpleOptimizer::BuildPelotonPlanTree(
...
+            std::string col_name(target_table->GetSchema()->GetColumns().at(column_ctr).GetName());
+            if (col_name == sort_col_name) {
+                //The column_ctr is not reliable anymore given that we are looking to the whole schema.
+                //Since the columns were added in the column_ids, it is safe to retrieve only those indexes.
+                key.push_back(find(column_ids.begin(), column_ids.end(), column_ctr) - column_ids.begin());
```
jarulraj: Store the iterator `find(column_ids.begin(), column_ids.end(), column_ctr)` in a variable `auto iter`.
Then use `std::distance(column_ids.begin(), iter)` instead of `iter - column_ids.begin()` to find the `offset`
Then push the `offset` into the `vector`.

jarulraj: This PR would be perfect if you can also add a test case using `SQLTestsUtil` -- a query where a couple of `ORDER BY` columns are not part of the `SELECT` list.

PauloAmora:
Ok, I'll work on that.

src/optimizer/simple_optimizer.cpp
PauloAmora: Where can I write those tests? I saw a `OptmizerSqlTest`, but I'm not sure if that is the correct class. I tried to find documentation on how to add test cases but couldn't find anything.

jarulraj: Hey @PauloAmora -- here's an example: https://github.com/AngLi-Leon/peloton/blob/6c11ff7489d088d31b47cf99a12a6ce0ebb52626/test/sql/drop_sql_test.cpp. We can modify it based on our requirements.

PauloAmora:
Something went wrong in my end, I'll add a new pull request

PR #466, opened 2017-01-05
Enabled IndexTuner + CIDR Demo Script

apavlo:
I will merge this after it passes.


apavlo:
I forgot that I was waiting for this build. I will stop pushing and wait for it to pass so we can merge.

PR #467, opened 2017-01-06
Layout Tuner

jarulraj:
This PR: 

- Launches layout tuner based on a flag.


apavlo:
You are pushing duplicate commits again. Fetch and rebase from cmu-db:master.

jarulraj:
I already did that, this is effect of recursive merge.

PR #473, opened 2017-01-06
Added sql interface to load stats for the index tuner

MattPerron:
THIS IS A HACK. After the demo we should change this
to use the ALTER keyword instead.

The following now works from psql, I am about to test with JDBC
`load stats from 'test';`


MattPerron:
I'm not sure why coveralls decreased so much, I haven't changed much

apavlo:
Travis is being weird today because we have so much going on. I will just restart the build

apavlo:
I think the jdbc validator is inconsistent

MattPerron:
Did it fail again?

apavlo:
@MattPerron Yes

PR #468, opened 2017-01-06
Layout Tuner Integration

jarulraj:
This PR launches layout tuner when enabled.

PR #475, opened 2017-01-06
Server-side PreparedStatement Invalidation

apavlo:
You can now tell a `PacketManager` that it needs to invalidate the cached query plan for all of its PreparedStatements and then replan them the next time that they are invoked. I am assuming that there is only one `PacketManager` per thread.

This code is nasty. I'm not pleaed with myself. But the `PacketManager` is a mess and there is no easy way to break out the PreparedStatement cache from it for now.

There is no easy way to write a test for this either because the `PacketManager` requires you to start up the libevent code. So I tested it manually (for now) but inserting this chunk of code into `IndexScanExecutor::DExecute()`

```
  // HACK
  LOG_DEBUG(
      "HACK: Telling PacketManagers to invalidate the PreparedStatements for "
      "'%s'",
      table_->GetName().c_str());
  for (auto pm : wire::PacketManager::GetPacketManagers()) {
    pm->InvalidatePreparedStatements(table_->GetOid());
  } 
```

Again, I told you that I am not boasting about this. I just had to get it done.

### DIFF
--- a/src/brain/index_tuner.cpp
+++ b/src/brain/index_tuner.cpp
@@ -207,25 +207,25 @@ bool SampleFrequencyMapEntryComparator(sample_frequency_map_entry a,
 std::vector<sample_frequency_map_entry> GetFrequentSamples(
     const std::vector<brain::Sample>& samples) {
   std::unordered_map<brain::Sample, double> sample_frequency_map;
-  double total_metric = 0;
+  double total_weight = 0;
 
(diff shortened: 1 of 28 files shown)


apavlo:
This now includes of @jarulraj's commits from the busted #468 PR

MattPerron:
https://www.youtube.com/watch?v=S7FsAgPuVwU

PR #485, opened 2017-01-08
set reserved field size to 0

yingjunwu:
simple yet effective method for reducing memory usage.


apavlo:
Why did you close this?

PR #490, opened 2017-01-09
OrderBy Partial Fix Bug #385 and unit tests

PauloAmora:
This PR resolves issue #385.

InSimpleOptimizer: Adding the order by column explicitely into the column map, to allow materialization and sorting.

In OrderByExecutor: Use only the columns in select for output_schema.

As requested, with unit tests.


jarulraj:
This is perfect. Nice fix, @PauloAmora!

PR #493, opened 2017-01-10
Adding namespace and README

wangziqi2013:


wangziqi2013:
This does not require a code review because it changes no functional part

Issue #498, opened 2017-01-15
Make source_validator.py a real pre-commit hook


apavlo:
This one should be easy to do:

http://codeinthehole.com/writing/tips-for-using-a-git-pre-commit-hook/

AngLi-Leon:
I think the pre-commit hook have to be installed locally. Do I need to add an installation instruction under wiki?

apavlo:
Hmmm... I guess we can't just include the symlink in the repo since git doesn't do that.

Yes, I think that we should add it to the repo instructions.

apavlo:
Fixed in #499

PR #499, opened 2017-01-15
add instructions for installing pre-commit hook

AngLi-Leon:
#498

### DIFF
--- a/script/pre_commit/pre_commit.md
+++ b/script/pre_commit/pre_commit.md
@@ -1,0 +1,6 @@
+# Git pre-commit
+
+### Installation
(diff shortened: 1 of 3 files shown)


script/pre_commit/pre_commit.md
```diff
@@ -0,0 +1,6 @@
+# Git pre-commit
+
+### Installation
+Under peloton root directory, run `ln -s ../../script/pre_commit/pre-commit.sh .git/hooks/pre-commit` for once to install locally.
```
apavlo: I think `../../` path is wrong if you are saying that they should be in the root directory.

AngLi-Leon: Though it is weird, actually I tested that the first path is relative to `.git/hooks/` (where symlink resides) and second path is relative to the current working path (the root).

PR #502, opened 2017-01-17
Fixed 'select *' case, added tests in 'order_by_sql'

PauloAmora:
From the previous PR

Fixes bug #385 completely, added tests for limit in `order_by_sql_tests` and new tests for `ORDER BY DESC`


Michael-Tieying-Zhang:
@apavlo @jarulraj The code sounds good to me.

PR #509, opened 2017-01-19
Switch to Enums Class (1)

AllisonWang:
This PR switched the following enum to enum class:
- PayloadType
- AggregateType
- QuantifierType
- TableReferenceType
- InsertType
- CopyType
- TaskPriorityType

The rest will be done in separate PRs to make each PR smaller.


apavlo:
@AllisonWang We run with `-Wall` and `-Werror`. That's why your build is failing.

AllisonWang:
The build issue is related to the string utility function of `CopyType`, which doesn't really have a default or invalid type, and is not assigned to integers. So I removed the CopyType's string utility functions.

Issue #382, opened 2016-12-07
Min/Max aggregation doesn't work with varchar/timestamp


malin1993ml:
We get a core dump with this:
```sql
postgres=# create table test(a int, b varchar, c timestamp);
create table test(a int, b varchar, c timestamp) 0
postgres=# insert into test values(0, 'Hi', '2016-12-06 00:00:01-04');
insert into test values(0, 'Hi', '2016-12-06 00:00:01-04') 1
postgres=# select * from test;
 a | b  |               c               
---+----+-------------------------------
 0 | Hi | 2016-12-06 00:00:01.000000-04
(1 row)

postgres=# select min(a) from test;
 min 
-----
   0
(1 row)

postgres=# select min(b) from test;
server closed the connection unexpectedly
	This probably means the server terminated abnormally
	before or while processing the request.
The connection to the server was lost. Attempting reset: Failed.
```

With timestamp, the core dump will happen when there're more than one values in the table:
```sql
postgres=# create table test(a int, b varchar, c timestamp);
create table test(a int, b varchar, c timestamp) 0
postgres=# insert into test values(0, 'Hi', '2016-12-06 00:00:01-04');
insert into test values(0, 'Hi', '2016-12-06 00:00:01-04') 1
postgres=# select min(c) from test;
              min              
-------------------------------
 2016-12-06 00:00:01.000000-04
(1 row)

postgres=# insert into test values(1, 'Hi', '2016-12-06 00:00:02-04');
insert into test values(1, 'Hi', '2016-12-06 00:00:02-04') 1
postgres=# select * from test;
 a | b  |               c               
---+----+-------------------------------
 0 | Hi | 2016-12-06 00:00:01.000000-04
 1 | Hi | 2016-12-06 00:00:02.000000-04
(2 rows)

postgres=# select min(c) from test;
server closed the connection unexpectedly
	This probably means the server terminated abnormally
	before or while processing the request.
The connection to the server was lost. Attempting reset: Failed.
```

elvismt:
Classes `type::TimestampType` and `type::VarlenType` do not reimplement `Min()`, `Max()` and a few other methods from the `type::Type` interface. So when `executor::MinAggregator` and `executor::MaxAggregator` call these methods they reach the empty implementation from the base class.

I will propose an implementation by issuing a pull request in the next few days, this will probably fix this issue.

elvismt:
Check pull request #511

poojanilangekar:
[...]

PR #511, opened 2017-01-20
Fix Min/Max aggregations for Timestamps. Issue #382

elvismt:
This change Fixes Min/Max aggregations for Timestamps by adding the missing methods.

It is not a total fix for issue #382 because the system still crashes for VARCHAR Min/Max aggregations.

In addition I uncommented the SQL tests for Min/Max aggregations on Timestamps.

Now focusing on varchar.


elvismt:
Coverage decreased overall because the tests for Varchar are still commented out since it doesn't work yet. On the other hand the coverage for timestamp_type.cpp increased. I will now focus on varchar.

elvismt:
Issue #382 fully fixed. Changing the default value of `is_inlined` for aggregations in `optimizer::SimpleOptimizer` was really all we needed.

Added SQL tests for Min/Max aggregations over VARCHARs.

jarulraj:
Looks good to me.

Issue #517, opened 2017-01-24
ORDER BY is incorrect for derived values


apavlo:
ORDER BY is still broken. It does not work correctly if we have values that are computed from expressions:

Let's first create a table with a single column:
```
postgres=# CREATE TABLE foo (val INTEGER PRIMARY KEY);
CREATE TABLE foo (val INTEGER PRIMARY KEY) 0
postgres=# INSERT INTO foo VALUES (10);
INSERT INTO foo VALUES (10) 1
postgres=# INSERT INTO foo VALUES (20);
INSERT INTO foo VALUES (20) 1
```

We then need to set one of the values to be negative. We have to do it this way because we don't support inserting negative values directly:
```
postgres=# UPDATE foo SET val = val - 20 WHERE val = 10;
UPDATE foo SET val = val - 20 WHERE val = 10 1

postgres=# SELECT * FROM foo ORDER BY val;
 val 
-----
 -10
  20
(2 rows)
```

Now we will have a query that computes an expression on the `val` column but also has an `ORDER BY` clause on the `val` column:
```
postgres=# SELECT (val * -1) AS newVal FROM foo ORDER BY val ASC;
 newval 
--------
    -20
     10
(2 rows)
``` 

The above result is incorrect. It is sorting on `newVal` instead of `val`. If you add back `val` in the SELECT output, you get a correct sort order:
```
postgres=# SELECT val, (val * -1) AS newVal FROM foo ORDER BY val ASC;
 val | newval 
-----+--------
 -10 |     10
  20 |    -20
(2 rows)
```

**[SQL script is available here](https://github.com/cmu-db/peloton/files/728065/orderby-bug.txt)**

PauloAmora:
I'll take a look into it.

PauloAmora:
@apavlo  -- Is this correct?
```sql
postgres=# UPDATE foo SET val = val - 20 WHERE val = 10;
UPDATE foo SET val = val - 20 WHERE val = 10 1

postgres=# SELECT * FROM foo ORDER BY val;
 val 
-----
 -10
  20
(2 rows)
```

apavlo:
@PauloAmora Yes, that's correct. It turns out you can actually just insert negatives if you put them in quotes:

```
INSERT INTO foo VALUES ('-10');
```

PauloAmora:
I misunderstood, sorry about that.

[...]

PR #520, opened 2017-01-25
Bugfix #517

PauloAmora:
This PR solves Bug #517

The bug was caused by the order of operations in `SimpleOptimizer`, since `ProjectPlan` was always placed below `OrderByPlan` in the plan tree, the `OrderByPlan` would order only the output.

The solution was to check if there was an `order_by` before creating the `ProjectPlan`, and take care to not recreate it in a later point. This required moving the logic of `order_by` and `order_by limit` to separate functions.


Michael-Tieying-Zhang:
@malin1993ml Lin, the code (`ProjectPlan`) @PauloAmora modified is the stuff in the original simple_optimizer. Could you review the code?

src/optimizer/simple_optimizer.cpp
```diff
@@ -1508,5 +1363,164 @@ bool SimpleOptimizer::UnderlyingSameOrder(planner::AbstractPlan* select_plan,
...
+        "Underlying plan has the same ordering output with"
+        "order_by plan without limit");
+    order_by_plan->SetUnderlyingOrder(true);
+  }
+  return std::move(order_by_plan);
+}
+
+std::unique_ptr<planner::AbstractPlan> SimpleOptimizer::CreateOrderByLimitPlan(
+    parser::SelectStatement* select_stmt, planner::AbstractPlan* child_plan,
+    catalog::Schema* schema, std::vector<oid_t> column_ids, bool is_star) {
```
malin1993ml: This function has lots of code in common with CreateOrderByPlan(). Can you remove the duplicated code, reuse CreateOrderByPlan() here to generate the order_by_plan, and then set the limit flags and add the limit plan?

PauloAmora: Got it, I tried not to mess around with the already present code, but I'll do that.

PauloAmora:
Sorry about that, I messed up the rebase (forgot to extract my commits as patches). I'll submit another PR with the revision fixes.

Issue #380, opened 2016-12-07
Update cannot convert integer to double automically


malin1993ml:
In the Update query, if we pass a integer value to a double field in the table, that will always be converted to 0.

```sql
postgres=# create table test(a int primary key, b double);
create table test(a int primary key, b double) 0
postgres=# insert into test values(0, 1);
insert into test values(0, 1) 1
postgres=# select * from test;
 a |    b     
---+----------
 0 | 1.000000
(1 row)

postgres=# update test set b = 2.0 where a = 0;
update test set b = 2.0 where a = 0 1
postgres=# select * from test;
 a |    b     
---+----------
 0 | 2.000000
(1 row)

postgres=# update test set b = 2 where a = 0;
update test set b = 2 where a = 0 1
postgres=# select * from test;
 a |    b     
---+----------
 0 | 0.000000
(1 row)
```

PR #524, opened 2017-01-30
Interpret correctly update values

danielpraciano:
This PR solves bug #380 

The bug was caused by an incompatibility between the table attribute type and the update value type. The update value must be reinterpreted as the table attribute to allow correct input.

### DIFF
--- a/src/optimizer/simple_optimizer.cpp
+++ b/src/optimizer/simple_optimizer.cpp
@@ -577,6 +577,44 @@ std::shared_ptr<planner::AbstractPlan> SimpleOptimizer::BuildPelotonPlanTree(
           updateStmt->table->GetDatabaseName(),
           updateStmt->table->GetTableName());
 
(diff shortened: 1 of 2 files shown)


apavlo:
@danielpraciano Can you rebase? I will then merge:

```bash
git remote add upstream git://github.com/cmu-db/peloton.git
git fetch upstream
git rebase -i upstream/master
git push --force
```

PauloAmora:
@apavlo Sorry to intrude on another PR, but I did the rebase command in my PR as you say and it didn't go well, probably because the commits were already remotely pushed.

The reason I'm asking this is because @danielpraciano is our undergrad research student, and the same issues may occur.

apavlo:
AFAIK, you have to do the rebase and not a merge. If you merge then that messes everything up.

danielpraciano:
@apavlo I did the rebase. It seems successful.

PR #527, opened 2017-01-31
Bugfix #517

PauloAmora:
This PR solves Bug #517

The bug was caused by the order of operations in SimpleOptimizer, since ProjectPlan was always placed below OrderByPlan in the plan tree, the OrderByPlan would order only the output.

The solution was to check if there was an order_by before creating the ProjectPlan, and take care to not recreate it in a later point. This required moving the logic of order_by and order_by limit to separate functions.

The previous PR review is also addressed in this.


apavlo:
@PauloAmora I will merge when it passes. Nice work.

PauloAmora:
All good now.

Issue #362, opened 2016-11-28
Internal Catalog Overhaul


apavlo:
This is a more long-term plan, so this is not going to happen this semester. But this is where I think that we need to go early next year to clean up our internal architecture. Like the networking overhaul, I think that we are going to have to bite the bullet and spend some time re-factoring our internal catalog management sub-system. This is necessary in the long run to make the code easier to understand, more understandable, recoverable, and to support a more sophisticated query optimizer.

1. There should be an abstract `CatalogObject` class that all catalog objects have to inherit. Each instance of this class will be backed by a `storage::DataTable`. This will ensure that they are logged and durable like regular tables. Since the `CatalogObject`'s are backed by regular data tables, we then be able to expose them to the user through the `INFORMATION_SCHEMA` views.

2. The tricky part in all of this is that if we keep the object interface to the catalog, then we need a way to make sure that if some part of the system modifies the catalog object, then those updates need to get propagated to their `storage::DataTable` as part of a transaction. I think that this is why Postgres has their catalog code tightly coupled with their storage manager. There is no way to avoid this. 

3. All catalog objects should have a globally unique 32-bit `oid`. We can come up with a scheme where we pack in the hostname, catalog type, and local counter (per type) into the bits. All other parts of the system should reference an object by its `oid` and not its name. For example, `TupleValueExpression` has both table/column name fields and what I think it supposed to be the `oid` (`value_idx_`?). This will make it easier to support `ALTER TABLE` without having to refresh **every** cached object in the system.

4. We need to make sure that we always reference catalog objects using their database prefix. This will ensure that we can support multi-tenant databases without having to do major refactoring later. Right now I think that a lot of the code is written assuming that there is a single database. We may want to add a simple ACL shim to provide basic security protection.

We should discuss this over the next month and come up with a plan for January.

PR #532, opened 2017-02-02
index catalog refactor

AngLi-Leon:
#362 first step


apavlo:
@AngLi-Leon can you rebase and resubmit?

apavlo:
@AngLi-Leon We should also remove the raw pointers in `IndexCatalogObject`.

PR #534, opened 2017-02-04
refactor microbenchmark

yingjunwu:
Currently, our microbenchmarks are in a bad shape -- nobody uses it to test the system. However, I think it is still a very important tool for evaluating backend performance. In this PR, I refactored the microbenchmark code, and preliminary results indicate that parallel database loading is unscalable.


apavlo:
@yingjunwu Why did the overall coverage go down by almost 1%? Our coverage for `src/gc/transaction_level_gc_manager.cpp` is still really bad too (0.75% of lines).

yingjunwu:
@apavlo the coverage is recovered now. Previously the GC threads were not created from the thread pool. I fixed that part but forgot to change the test cases. Now all the test cases have passed and hence the coverage is recovered.

PR #536, opened 2017-02-05
transaction pool

yingjunwu:
In this PR, I introduce transaction pool to Peloton.

A transaction pool is a concurrent data structure that contains a fixed number of transaction objects. The transation manager of the DBMS can concurrently acquire/release transaction objects from/to the transaction pool. More specifically, a transaction manager acquires a new transaction object from the transaction pool when invoking `BeginTransaction` function, and returns the transaction object back to the transaction pool when invoking `EndTransaction` function. 

Transaction pool brings several advantages to the system:

1. reduce memory allocation by recycling transaction objects;
2. eliminate the necessity of creating `txn_id` (which is a centralized contention bottleneck) for every newly incoming transaction;
3. lay the foundation for decentralized epoch manager that will be ported into the system in the future. (as we can fix the max # of concurrent transactions now.)
4. simplify the decoupling of connection threads and execution threads.


MattPerron:
I will review when Travis passes.

yingjunwu:
@MattPerron Sure. It will pass soon :-) Thanks!

src/wire/libevent_server.cpp
```diff
@@ -28,7 +28,7 @@ std::vector<std::unique_ptr<LibeventSocket>>
     &LibeventServer::GetGlobalSocketList() {
   static std::vector<std::unique_ptr<LibeventSocket>>
       // 2 fd's per thread for pipe and 1 listening socket
-      global_socket_list(FLAGS_max_connections + QUERY_THREAD_COUNT * 2 + 1);
+      global_socket_list(FLAGS_max_connections + CONNECTION_THREAD_COUNT * 2 + 1);
```
MattPerron: magic numbers 
(I know you didn't add these but if you are already here, then this is a good time to fix them)

src/include/concurrency/transaction_manager.h
```diff
@@ -172,11 +183,14 @@ class TransactionManager {
   // first value is exclusive, last value is inclusive
   std::pair<cid_t, cid_t> dirty_range_ =
       std::make_pair(INVALID_CID, INVALID_CID);
+  
+  bool is_from_transaction_pool_;
```
MattPerron: Is this just to make sure asserts fail in the correct cases? If so, it should be clear that it is slated for removal.

src/common/init.cpp
```diff
@@ -31,15 +31,16 @@ namespace peloton {
 ThreadPool thread_pool;
 
 void PelotonInit::Initialize() {
-  QUERY_THREAD_COUNT = std::thread::hardware_concurrency();
+  CONNECTION_THREAD_COUNT = std::thread::hardware_concurrency();
   LOGGING_THREAD_COUNT = 1;
   GC_THREAD_COUNT = 1;
   EPOCH_THREAD_COUNT = 1;
+  MAX_CONCURRENCY = 10;
```
MattPerron: magic number,

yingjunwu: yes it is. the users should be responsible for setting the maximum number of connections (threads) in the system.

src/common/init.cpp
```diff
@@ -31,15 +31,16 @@ namespace peloton {
...
 
   // set max thread number.
   thread_pool.Initialize(0, std::thread::hardware_concurrency() + 3);
 
-  int parallelism = (std::thread::hardware_concurrency() + 1) / 2;
+  int parallelism = (std::thread::hardware_concurrency() + 3) / 4;
```
yingjunwu: yes. this means that every four threads races for a single tile group.

src/concurrency/timestamp_ordering_transaction_manager.cpp
```diff
@@ -105,11 +110,69 @@ Transaction *TimestampOrderingTransactionManager::BeginTransaction() {
...
 
 Transaction *TimestampOrderingTransactionManager::BeginReadonlyTransaction() {
+  PL_ASSERT(is_from_transaction_pool_ == false);
+  
+  auto &epoch_manager = EpochManagerFactory::GetInstance();
+
   txn_id_t txn_id = READONLY_TXN_ID;
+  cid_t begin_cid = epoch_manager.GetReadOnlyTxnCid();
+  
+  Transaction *txn = new Transaction(txn_id);
```
yingjunwu: yes. I am waiting for the thread decoupling patch. Once it is done, we can remove all the deprecated code.

src/concurrency/transaction_pool.cpp
```diff
@@ -0,0 +1,23 @@
...
+// Copyright (c) 2015-16, Carnegie Mellon University Database Group
+//
+//===----------------------------------------------------------------------===//
+
+#include "concurrency/transaction_pool.h"
+
+namespace peloton {
+namespace concurrency {
+
+size_t TransactionPool::max_concurrency_ = 10;
```
yingjunwu: yes.

src/type/types.cpp
```diff
@@ -33,10 +33,11 @@ int DEFAULT_TUPLES_PER_TILEGROUP = 1000;
 int TEST_TUPLES_PER_TILEGROUP = 5;
 
 // For threads
-size_t QUERY_THREAD_COUNT = 1;
+size_t CONNECTION_THREAD_COUNT = 1;
```
yingjunwu: for now, we don't have any file for keeping these contants. we should create one.

src/wire/libevent_server.cpp
yingjunwu: we should have a discussion on this.

src/include/concurrency/transaction_manager.h
yingjunwu: it's temporarily added. I maintain this so that we can still use the older code for running JDBC transactions and use the new code for running microbenchmarks.

yingjunwu: this variable will be removed once the thread decoupling is done.

src/common/init.cpp
MattPerron: Then it should be in a macro or configurable

src/common/init.cpp
MattPerron: set a macro or make it configurable.

src/concurrency/timestamp_ordering_transaction_manager.cpp
MattPerron: Is there a dependency on the networking threads being separated from worker threads?

yingjunwu:
After talking to @MattPerron and @sid1607, I will refactor the entire epoch manager based on current implementation.

PR #547, opened 2017-02-18
Binder

hzxa21:
This PR adds a Binder to extend the parser tree for name binding. It adds members in the parser tree node and can support:

- Bind column name to the corresponding table name/alias in the query;
- Bind column name to (DB_ID, TABLE_ID, COL_POSITION) tuple;
- Syntax checking for invalid names and ambiguous names.

The Binder now can support binding in nested query, which means the inner query can reference alias in the outer query if it is not overridden. But the following case is not yet supported (the outer query references a column alias in the inner query):
`select * from (select a as c from A) as C where C.c = 1;` 

Lin, Could you help to review this PR? @malin1993ml


src/binder/bind_node_visitor.cpp
```diff
@@ -0,0 +1,149 @@
...
+void BindNodeVisitor::Visit(const parser::LimitDescription *) {}
+void BindNodeVisitor::Visit(const parser::CopyStatement *) {}
+void BindNodeVisitor::Visit(const parser::CreateStatement *) {}
+void BindNodeVisitor::Visit(const parser::InsertStatement *) {}
+void BindNodeVisitor::Visit(const parser::DropStatement *) {}
+void BindNodeVisitor::Visit(const parser::PrepareStatement *) {}
+void BindNodeVisitor::Visit(const parser::ExecuteStatement *) {}
+void BindNodeVisitor::Visit(const parser::TransactionStatement *) {}
+
+void BindNodeVisitor::Visit(expression::TupleValueExpression *expr) {
```
malin1993ml: Do we handle the aliases for column names? I didn't seem to see the logic. Does the parser handle that in the first place?

src/include/expression/tuple_value_expression.h
```diff
@@ -78,6 +79,18 @@ class TupleValueExpression : public AbstractExpression {
 
   std::string GetColumnName() const { return col_name_; }
 
+  void SetTableName(std::string table_alias) { table_name_ = table_alias; }
+
+  // Binder stuff
+  bool isObjectBound = false;
+  std::tuple<oid_t, oid_t, oid_t> BoundObjectId;
```
malin1993ml: Please follow the naming convention for member variables here.

test/binder/binder_test.cpp
```diff
@@ -0,0 +1,200 @@
...
+  parse_tree = parser.BuildParseTree(selectSQL);
+  selectStmt = (parser::SelectStatement*)(parse_tree->GetStatements().at(0));
+  binder->BindNameToNode(selectStmt);
+  tupleExpr =
+      (expression::TupleValueExpression*)(selectStmt->select_list->at(0));
+  EXPECT_EQ(tupleExpr->BoundObjectId, make_tuple(db_oid, tableA_oid, 0));
+  tupleExpr =
+      (expression::TupleValueExpression*)(selectStmt->select_list->at(1));
+  EXPECT_EQ(tupleExpr->BoundObjectId, make_tuple(db_oid, tableB_oid, 1));
+
```
malin1993ml: Can you add a test case for something like "select * from A, A as AA where A.a1 = AA.a2"?

src/binder/bind_node_visitor.cpp
hzxa21: No. Because the alias of a column is defined in the select clause, it should be invisible in other clauses of the same level. However, if the inner query defines a column alias, the outer query can see it but it is not supported now.

src/include/binder/binder_context.h
```diff
@@ -0,0 +1,133 @@
...
+
+#include <unordered_map>
+#include <string>
+#include "catalog/catalog.h"
+#include "type/types.h"
+
+namespace peloton {
+namespace binder {
+
+class BinderContext;
```
hzxa21: Done.

src/include/binder/binder_context.h
```diff
@@ -0,0 +1,133 @@
...
+
+class BinderContext {
```
hzxa21: Done.

src/include/binder/binder_context.h
```diff
@@ -0,0 +1,133 @@
...
+            table_ref->GetDatabaseName(), table_ref->GetTableName());
+
+    auto id_tuple = std::make_tuple(table->GetDatabaseOid(), table->GetOid());
+
+    std::string alias = table_ref->GetTableAlias();
+    if (table_ref->alias != nullptr) {
+      alias = table_ref->alias;
+    } else {
+      alias = table_ref->GetTableName();
+    }
```
hzxa21: Done.

src/include/binder/binder_context.h
```diff
@@ -0,0 +1,133 @@
...
+    return false;
+  }
+
+ private:
+  //  struct comp {
+  //    bool operator() (const std::string& lhs, const std::string& rhs) const {
+  //      return strcasecmp(lhs.c_str(), rhs.c_str()) < 0;
+  //    }
+  //  };
```
hzxa21: Done.

src/include/expression/tuple_value_expression.h
hzxa21: Done.

test/binder/binder_test.cpp
hzxa21: Done.

src/binder/bind_node_visitor.cpp
malin1993ml: Sure. I understand.

PR #549, opened 2017-02-24
port decentralized epoch manager

yingjunwu:
This is a major rewrite of the epoch manager. Previously, Peloton implements a centralized epoch manager, which is unscalable, especially for logging. This PR introduced a decentralized epoch manager we designed to resolve potential centralized contention points. The next step is to rewrite the logging component, which is a missing piece in Peloton.

I also fixed some random bugs in the networking layer, but please note that the jdbc bug still exists.

@MattPerron @jessesleeping as discussed before, could you please help review this PR?


src/include/concurrency/decentralized_epoch_manager.h
```diff
@@ -0,0 +1,303 @@
...
+    read_write_count_(epoch.read_write_count_.load()) {}
+
+  void Init() {
+    read_only_count_ = 0;
+    read_write_count_ = 0;
+  }
+};
+
+struct LocalEpochContext {
+
```
jessesleeping: For better cache performance, we need to force the LocalEpochContext to take up a cache line. I think for now sizeof(LocalEpochContext) is greater than 64 bytes. But I suggest that either we enforce it in the code or leave a comment about it.

src/include/concurrency/decentralized_epoch_manager.h
```diff
@@ -0,0 +1,303 @@
...
+    // very long transactions must be timeout
+    PL_ASSERT(epoch_id - tail_epoch_id_ <= epoch_buffer_size_);
+
+    size_t epoch_idx = (size_t) epoch_id % epoch_buffer_size_;
+    epoch_buffer_[epoch_idx].read_write_count_++;
+
+    return true;
+  }
+
+  // for now, we do not support read-only transactions
```
jessesleeping: Make it an inlined TODO.

src/concurrency/decentralized_epoch_manager.cpp
```diff
@@ -0,0 +1,69 @@
...
+namespace peloton {
+namespace concurrency {
+
+
+  // enter epoch with thread id
+  cid_t DecentralizedEpochManager::EnterEpochD(const size_t thread_id) {
+
+    PL_ASSERT(local_epoch_contexts_.find(thread_id) != local_epoch_contexts_.end());
+
+    while (true) {
```
jessesleeping: Please add comment explaining the retry logic in EnterLocalEpoch.

src/include/concurrency/decentralized_epoch_manager.h
```diff
@@ -0,0 +1,303 @@
...
+    // in this case, we should reject entering local epoch.
+    // this is essentially a validation scheme.
+    if (epoch_id < head_epoch_id_) {
+      return false;
+    }
+    
+    PL_ASSERT(epoch_id >= head_epoch_id_);
+
+    // update head epoch id.
+    head_epoch_id_ = epoch_id;
```
jessesleeping: Line 71 and line 78 should be done in an atomic way. Otherwise it will cause a race condition between GC thread and transaction worker.

src/include/concurrency/decentralized_epoch_manager.h
```diff
@@ -0,0 +1,303 @@
...
+
+    // when exiting a local epoch, we must check whether it can be reclaimed.
+    IncreaseTailEpochId();
+  }
+
+  // this function can be invoked by both the epoch thread and the local worker thread.
+  void IncreaseTailEpochId() {
+    lock_.Lock();
+    // in the best case, tail_epoch_id can be increased to (head_epoch_id - 1)
+    while (tail_epoch_id_ < head_epoch_id_ - 1) {
```
jessesleeping: I think we can use a priority heap to replace this while loop.

src/common/init.cpp
```diff
@@ -31,15 +31,16 @@ namespace peloton {
...
   LOGGING_THREAD_COUNT = 1;
   GC_THREAD_COUNT = 1;
   EPOCH_THREAD_COUNT = 1;
+  MAX_CONCURRENCY = 10;
 
   // set max thread number.
   thread_pool.Initialize(0, std::thread::hardware_concurrency() + 3);
 
-  int parallelism = (std::thread::hardware_concurrency() + 1) / 2;
+  int parallelism = (std::thread::hardware_concurrency() + 3) / 4;
```
yingjunwu: we should have a discussion about this in the group meeting.

src/concurrency/timestamp_ordering_transaction_manager.cpp
```diff
@@ -104,15 +120,32 @@ Transaction *TimestampOrderingTransactionManager::BeginTransaction() {
...
+    cid_t begin_cid = epoch_manager.GetReadOnlyTxnCid();
+    
+    txn = new Transaction(begin_cid, thread_id, true);
+
+    auto eid = epoch_manager.EnterReadOnlyEpoch(begin_cid);
+    txn->SetEpochId(eid);
+
+
+  } else {
+
```
yingjunwu: we support read-only txn now.

src/concurrency/timestamp_ordering_transaction_manager.cpp
```diff
@@ -104,15 +120,32 @@ Transaction *TimestampOrderingTransactionManager::BeginTransaction() {
...
+    cid_t begin_cid = EpochManagerFactory::GetInstance().EnterReadOnlyEpochD(thread_id);
 
-  cid_t begin_cid = epoch_manager.GetReadOnlyTxnCid();
-  Transaction *txn = new Transaction(txn_id, begin_cid, true);
+    txn = new Transaction(begin_cid, thread_id);
```
yingjunwu: I will remove centralized epoch manager in the future.

src/include/concurrency/centralized_epoch_manager.h
```diff
@@ -0,0 +1,214 @@
...
+      is_running_(false) {}
+
+
+  static CentralizedEpochManager &GetInstance() {
+    static CentralizedEpochManager epoch_manager;
+    return epoch_manager;
+  }
+
+  virtual void Reset(const size_t &current_epoch) override {
+    current_epoch_ = current_epoch;
```
yingjunwu: TODO added. Will fix this problem when porting logging stuffs.

src/include/concurrency/centralized_epoch_manager.h
```diff
@@ -0,0 +1,214 @@
...
+  }
+
+  virtual void StartEpoch(std::unique_ptr<std::thread> &epoch_thread) override {
+    LOG_TRACE("Starting epoch");
```
yingjunwu: No. The current implemtation is that each component must acquire (dedicated) thread from thread pool.

src/include/concurrency/epoch_manager.h
```diff
@@ -16,289 +16,61 @@
...
-
-/*
-Epoch queue layout:
- current epoch               queue tail                reclaim tail
-/                           /                          /
-+--------+--------+--------+--------+--------+--------+--------+-------
-| head   | safety |  ....  |readonly| safety |  ....  |gc usage|  ....
-+--------+--------+--------+--------+--------+--------+--------+-------
-New                                                   Old
-
```
yingjunwu: no. It's already moved to centralized_epoch_manager.h

src/include/concurrency/transaction.h
```diff
@@ -140,6 +153,8 @@ class Transaction : public Printable {
   size_t insert_count_;
 
   bool declared_readonly_;
+
+  size_t pool_hint_;
```
yingjunwu: removed.

src/type/types.cpp
```diff
@@ -33,10 +33,11 @@ int DEFAULT_TUPLES_PER_TILEGROUP = 1000;
 int TEST_TUPLES_PER_TILEGROUP = 5;
 
 // For threads
-size_t QUERY_THREAD_COUNT = 1;
+size_t CONNECTION_THREAD_COUNT = 1;
```
yingjunwu: it's set to a large number at init.cpp

src/include/concurrency/decentralized_epoch_manager.h
yingjunwu: it's much larger than 64 bytes..

src/include/concurrency/decentralized_epoch_manager.h
yingjunwu: just rewrote the code.

src/include/concurrency/decentralized_epoch_manager.h
yingjunwu: just rewrote the code.

src/include/concurrency/decentralized_epoch_manager.h
yingjunwu: done.

src/concurrency/decentralized_epoch_manager.cpp
yingjunwu: the logic is much clearer now.

yingjunwu:
@jessesleeping please review for the second time.

yingjunwu:
Please note that I have fully removed the older version of the epoch manager.

src/gc/transaction_level_gc_manager.cpp
```diff
@@ -53,7 +53,10 @@ void TransactionLevelGCManager::Running(const int &thread_id) {
     auto &txn_manager = concurrency::TransactionManagerFactory::GetInstance();
     auto max_cid = txn_manager.GetMaxCommittedCid();
 
-    PL_ASSERT(max_cid != MAX_CID);
+    if (max_cid == MAX_CID) {
```
jessesleeping: I assume that the previous assert prevent some bad things from happening. In what case will max_cid == MAX_CID?

src/concurrency/decentralized_epoch_manager.cpp
```diff
@@ -0,0 +1,84 @@
...
+    local_epochs_.at(thread_id)->ExitEpoch(epoch_id);
+ 
+  }
+
+
+  uint64_t DecentralizedEpochManager::GetMaxCommittedEpochId() {
+    uint64_t global_max_committed_eid = UINT64_MAX;
+    
+    // for all the local epoch contexts, obtain the minimum max committed epoch id.
+    for (auto &local_epoch : local_epochs_) {
```
jessesleeping: I think it may be better to rename "local_epoch" as "local_epoch_itr"

src/concurrency/decentralized_epoch_manager.cpp
```diff
@@ -0,0 +1,84 @@
...
+      uint64_t local_max_committed_eid = local_epoch.second->GetMaxCommittedEpochId(current_global_epoch_);
+      
+      if (local_max_committed_eid < global_max_committed_eid) {
+        global_max_committed_eid = local_max_committed_eid;
+      }
+    }
+
+    if (global_max_committed_eid != UINT64_MAX && 
+        global_max_committed_eid >= current_global_epoch_ro_) {
+      current_global_epoch_ro_ = global_max_committed_eid + 1;
```
jessesleeping: The "global_max_committed_eid + 1" is related to how we generate txn_id for  read only txn. Please leave a comment to explain it.

src/concurrency/local_epoch.cpp
```diff
@@ -0,0 +1,118 @@
...
+    if (epoch_lower_bound_ == UINT64_MAX) {
+      epoch_lower_bound_ = epoch_id - 1;
+    } else if (epoch_lower_bound_ >= epoch_id) {
+    // epoch_lower_bound_ has already been updated by the GC.
+    // have to grab a newer epoch_id.
+      epoch_lock_.Unlock();
+      return false;
+    } 
+    // check whether the corresponding epoch exists.
+    if (epoch_map_.find(epoch_id) == epoch_map_.end()) {
```
jessesleeping: We can refactor the code to save a map lookup.

src/concurrency/local_epoch.cpp
```diff
@@ -0,0 +1,118 @@
...
+
+    epoch_lock_.Unlock();
+    return true;
+  }
+  
+  // the input parameter is the global read-only epoch id.
+  void LocalEpoch::EnterEpochRO(const uint64_t epoch_id) {
+    epoch_lock_.Lock();
+    // if this thread is never used or GC'd
+    if (epoch_lower_bound_ == UINT64_MAX) {
```
jessesleeping: Why do we also need to update epoch_lower_bound_ here when creating RO txn? I assume that the epoch_id passed to this function is of a "stale" epoch.

src/gc/transaction_level_gc_manager.cpp
yingjunwu: please note that the assertion has been commented out. When the system is started but it never processes any transaction, then we may see max_cid == MAX_CID.

Comments added.

src/concurrency/decentralized_epoch_manager.cpp
yingjunwu: updated.

src/concurrency/decentralized_epoch_manager.cpp
yingjunwu: done.

src/concurrency/local_epoch.cpp
yingjunwu: how?

src/concurrency/local_epoch.cpp
yingjunwu: epoch_lower_bound_ is for guarantee the epoch's logic is still correct even if some threads are starving.

src/concurrency/local_epoch.cpp
jessesleeping: Store the result of find(epoch_id) in a local variable and use it if it's valid.

yingjunwu:
@apavlo @jessesleeping @MattPerron This PR can be merged now. I will port the logging component in the next PR.

apavlo:
65% coverage :-1:

MattPerron:
I know this is already merged, but does this actually work when the # connections > # networking threads?
Do you have a test for this?

src/common/init.cpp
MattPerron: Did this discussion ever happen? and if so, why are there still magic numbers here?

src/wire/libevent_callbacks.cpp
```diff
@@ -152,7 +152,7 @@ void StateMachine(LibeventSocket *conn) {
           conn->pkt_manager.is_started = true;
         } else {
           // Process all other packets
-          status = conn->pkt_manager.ProcessPacket(&conn->rpkt);
+          status = conn->pkt_manager.ProcessPacket(&conn->rpkt, (size_t)conn->thread_id);
```
MattPerron: The thread_id you are getting here can be used on multiple txns, are you sure this is what you want?

yingjunwu: @MattPerron see the private msg I sent to you in slack.

Issue #545, opened 2017-02-17
UPDATE crashes on non-indexed predicates


PauloAmora:
The UPDATE command crashes if used with a non-index predicate or without one.

Start by creating a table and adding some data:
```SQL
CREATE TABLE foo (a INTEGER, b INTEGER);
INSERT INTO foo VALUES (1, 200);
INSERT INTO foo VALUES (2, 150);
INSERT INTO foo VALUES (3, 400);
```

If you execute the following command:
```SQL
UPDATE foo SET b = 100 WHERE a = 2;
```

The system crashes. It also happens with queries like this:
```SQL
UPDATE foo SET b = 100;
```

Test script is here: http://pastebin.com/zq7dtPac

seojungmin:
Attached is a quick patch to the issue.  The crash comes from asserting (index_entry_ptr != nullptr).  index_entry_ptr may be equal to nullptr if there is no index.  Since the previous codefor performUpdate() does not align with handling index_entry_ptr in perfromDelete(), I assume that this is a bug. 

[peloton-545.diff.txt](https://github.com/cmu-db/peloton/files/802186/peloton-545.diff.txt)

apavlo:
@seojungmin Can you send a PR? We don't do direct patches.

poojanilangekar:
Fixed in #558

PR #558, opened 2017-02-27
BugFix #545 - Crash with Update

seojungmin:
This is a patch on BugFix #545. The code is aligned with handling index_entry_ptr in PerformDelete().


apavlo:
@seojungmin Does this pass @PauloAmora's test script in #545? We should add it to `script/testing/dml/`

Issue #553, opened 2017-02-27
Support Conditional Expressions


malin1993ml:
Here's what we want to support:
https://www.postgresql.org/docs/9.1/static/functions-conditional.html

We have parsing error right now, but I don't think our executor supports that, either.
```sql
postgres=# create table a(id int, value varchar);
create table a(id int, value varchar) 0
postgres=# insert into a values(1, 'hi');
insert into a values(1, 'hi') 1
postgres=# insert into a values(2, 'ha');
insert into a values(2, 'ha') 1
postgres=# select * from a;
 id | value 
----+-------
  1 | hi
  2 | ha
(2 rows)

postgres=# select id, case when id=1 then 'one' when id=2 then 'two' else 'other' end from a;
Error parsing SQL statement
postgres=# 
```

malin1993ml:
We also want to support NULLIF there.

Issue #548, opened 2017-02-20
Revive CASE code + tests


apavlo:
We use to support `CASE` but then it got removed by accident over the summer. The original commit is here:

https://github.com/cmu-db/peloton/commit/c537df4a016920fbd4b2fda3b12da4c82e5f34c0

There are additional commits made afterwards for fixing the test cases:

https://github.com/cmu-db/peloton/commit/a2307f7cc5df3043acebed0e2c1625b61f96989f

Somebody needs to dig through the history and find the latest version of `CaseExpression` and restore it. You're not going to be able to just copy the old file into the codebase because we rewrote the type system and expression interface, but it should be a relatively easy task.

There is going to be a bunch of bridge code that we used to have for converting the Postgres plan to a Peloton plan. We may also want to bring this back later (for #521) but for now we don't need it. Note also that the parser does not support `CASE` so all the testing will have to be done procedurally for now.

apavlo:
FYI, @thejdeep is working on this. I can't assign him officially without adding him to the team.

IanQS:
Not sure if this helps but the last place that case_expression.h existed is

https://github.com/cmu-db/peloton/commit/d0c1e97530ec49cd251f1d3a1ff3c7e37f5b886c

seojungmin:
Fixed in #559

PR #559, opened 2017-03-02
Issue #548, #553 - CASE works with SELECT

seojungmin:
This patch adds a fix on the current parser.  The previous test scenario in #548 is tested and I added some scenarios including those mentioned in #553.

### DIFF
--- a/src/include/expression/case_expression.h
+++ b/src/include/expression/case_expression.h
@@ -1,0 +1,107 @@
+//===----------------------------------------------------------------------===//
+//
+//                         PelotonDB
+//
+// case_expression.h
+//
+// Identification: src/include/expression/case_expression.h
+//
+// Copyright (c) 2015-17, Carnegie Mellon University Database Group
+//
+//===----------------------------------------------------------------------===//
+
+#pragma once
+
+#include "expression/abstract_expression.h"
+#include "expression/comparison_expression.h"
+#include "common/sql_node_visitor.h"
+
+namespace peloton {
+namespace expression {
+
+//===----------------------------------------------------------------------===//
+// CaseExpression
+//===----------------------------------------------------------------------===//
+
+class CaseExpression : public AbstractExpression {
+ public:
+  using AbsExprPtr = std::unique_ptr<AbstractExpression>;
+  using WhenClause = std::pair<AbsExprPtr,AbsExprPtr>;
+
+  CaseExpression(type::Type::TypeId type_id, AbsExprPtr argument,
+                 std::vector<WhenClause> &when_clauses,
+                 AbsExprPtr dflt_expression)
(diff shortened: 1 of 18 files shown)


pmenon:
This doesn't look right. 

1. It's a bad idea to use/throw exceptions to indicate conditions in expression/operator logic.
2. Why are two separate classes needed?
3. Why aren't smart pointers being used?

The design of CASE from [here](https://github.com/cmu-db/peloton/pull/119) is the correct approach. Use it as a starting point.

seojungmin:
I was unaware that this code part that I ported has been refactored once.  The refactored design you mentioned could be applied. Thanks.

apavlo:
What is the status of this PR?

apavlo:
@seojungmin Why did the coverage go down so much?

seojungmin:
@apavlo I am trying to see whether the coverage may differ if I rebase.

src/include/expression/case_expression.h
```diff
@@ -0,0 +1,114 @@
...
+    // Otherwise, we assume that the When Clauses provided are sufficient
+    if (argument_ != nullptr) {
+      for (uint32_t i = 0; i < clauses_.size(); i++)
+		clauses_[i].first.reset(
+           new peloton::expression::ComparisonExpression(
+              peloton::ExpressionType::COMPARE_EQUAL, argument_->Copy(),
+              clauses_[i].first->Copy()));
+    }
+  }
+  type::Value Evaluate(const AbstractTuple *tuple1, const AbstractTuple *tuple2,
```
pmenon: I think we should have two constructors: one with an argument and one without an argument. This makes it's use a little clearer. Thoughts?

test/expression/expression_test.cpp
```diff
@@ -245,5 +246,188 @@ TEST_F(ExpressionTests, ExtractDateTests) {
...
+  tuple->SetValue(0, type::ValueFactory::GetIntegerValue(2), nullptr);
+  tuple->SetValue(1, type::ValueFactory::GetIntegerValue(1), nullptr);
+  result = case_expression->Evaluate(tuple, nullptr, nullptr);
+  expected = type::ValueFactory::GetIntegerValue(3);
+  EXPECT_EQ(type::CmpBool::CMP_TRUE, expected.CompareEquals(result));
+
+  delete case_expression;
+  delete schema;
+  delete tuple;
+}
```
pmenon: Can you modify these tests to use smart pointers?

pmenon:
Overall, it looks good 👍 Just address the reviews, but otherwise LGTM.

src/include/expression/case_expression.h
seojungmin: Agreed.  I am a natural born C programmer, and I hate introducing functions. BUT, I will take your suggestion.

seojungmin:
@apavlo Can you merge this? I think it is ready.

PR #564, opened 2017-03-05
Link repo work dir on host into VM instead of cloning

summivox:
This allows working on the host without having to sync every change to VM.

### DIFF
--- a/script/installation/Vagrantfile
+++ b/script/installation/Vagrantfile
@@ -1,5 +1,5 @@
 # -*- mode: ruby -*-
-# vi: set ft=ruby :
+# vi: set ft=ruby et:
 
 # Customize VM resources
 VM_CPUS = "4"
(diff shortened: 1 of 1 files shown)


summivox:
Wait. Seems that `make` is not happy with building in the shared folder (at least when host is Windows).

summivox:
It works if I use a VM-local build dir (instead of one inside the shared repo work dir):

```bash
mkdir -p ~/build
cd ~/build
cmake -DCMAKE_BUILD_TYPE=Release /peloton
make
make check
```

So this is still good. Of course needs some extra documentation, but I feel the benefit of not having to `git pull` all the time far outweighs the one-time up-front cost.

summivox:
@jarulraj : I think it needs some extra documentation (otherwise people would be very confused when the build instruction doesn't work). When I have time I'll try update wiki.

PR #565, opened 2017-03-05
add editorconfig for consistent indent/trailing space

summivox:


jarulraj:
I am not sure if we want to check in editor configurations, since people use different editors with different configurations.

summivox:
This is an editor-agnostic file designed to provide basic style guidance.

See http://editorconfig.org/

jarulraj:
Sounds good, that should work.

PR #572, opened 2017-03-17
Query compilation

pmenon:
This PR adds code generation and compilation using LLVM. Right now, plan execution first checks if a given query can be compiled. If so, it goes through the compilation and execution process. If the plan can't be compiled, it falls back to interpreted execution.

Some high-level points:

- Table scans, order-by's, hash joins, and aggregations are supported.
- Most of the current expression subsystem is supported. Some exist in the new engine, but not in the interpreted engine (e.g., case statements).
- The general ROF framework is enabled. Prefetching is working, but disabled by default. SIMD vectorization is still being ported over.
- Single-threaded execution.
- String functionality is a little shotty, as is NULL handling.
- Plenty of tests, but not nearly sufficient.
- The beginnings of an embedded TPCH benchmark.


pmenon:
Ugh, I'm closing this PR. I'll create a new one when I figure out what's up with Travis.

PR #573, opened 2017-03-17
PostgresParser

haojin2:
This PR is for the new Pg_Parser that use libpg_query for parsing and transform to Peloton parse tree. Important changes includes:
Modified original libpg_query library for direct access to Postgres parse tree
Support for several select queries
Several miscellaneous changes for some bugs in old code

### DIFF
--- a/src/include/expression/star_expression.h
+++ b/src/include/expression/star_expression.h
@@ -38,7 +38,7 @@ class StarExpression : public AbstractExpression {
   virtual void Accept(SqlNodeVisitor *v) { v->Visit(this); }
 
  protected:
-  StarExpression(const FunctionExpression &other) : AbstractExpression(other) {}
+  StarExpression(const AbstractExpression &other) : AbstractExpression(other) {}
 
  private:
 };
--- a/src/include/parser/nodes.h
+++ b/src/include/parser/nodes.h
@@ -1,0 +1,668 @@
+/*-------------------------------------------------------------------------
+ *
+ * nodes.h
+ *	  Definitions for tagged nodes.
+ *
+ *
+ * Portions Copyright (c) 1996-2015, PostgreSQL Global Development Group
+ * Portions Copyright (c) 1994, Regents of the University of California
+ *
+ * src/include/nodes/nodes.h
+ *
+ *-------------------------------------------------------------------------
+ */
+#ifndef NODES_H
+#define NODES_H
+
+#ifdef __cplusplus
+;
+#else
+typedef char bool;
+#endif
+
+/*
+ * The first field of every node is NodeTag. Each node created (with makeNode)
+ * will have one of the following tags as the value of its first field.
+ *
+ * Note that the numbers of the node tags are not contiguous. We left holes
+ * here so that we can add more tags without changing the existing enum's.
+ * (Since node tag numbers never exist outside backend memory, there's no
+ * real harm in renumbering, it just costs a full rebuild ...)
+ */
+typedef enum NodeTag
+{
+	T_Invalid = 0,
+
+	/*
+	 * TAGS FOR EXECUTOR NODES (execnodes.h)
+	 */
+	T_IndexInfo = 10,
+	T_ExprContext,
+	T_ProjectionInfo,
+	T_JunkFilter,
+	T_ResultRelInfo,
+	T_EState,
+	T_TupleTableSlot,
+
+	/*
+	 * TAGS FOR PLAN NODES (plannodes.h)
+	 */
+	T_Plan = 100,
+	T_Result,
+	T_ModifyTable,
+	T_Append,
+	T_MergeAppend,
+	T_RecursiveUnion,
+	T_BitmapAnd,
+	T_BitmapOr,
+	T_Scan,
+	T_SeqScan,
+	T_SampleScan,
+	T_IndexScan,
+	T_IndexOnlyScan,
+	T_BitmapIndexScan,
+	T_BitmapHeapScan,
+	T_TidScan,
+	T_SubqueryScan,
+	T_FunctionScan,
+	T_ValuesScan,
+	T_CteScan,
+	T_WorkTableScan,
+	T_ForeignScan,
+	T_CustomScan,
+	T_Join,
+	T_NestLoop,
+	T_MergeJoin,
+	T_HashJoin,
+	T_Material,
+	T_Sort,
+	T_Group,
+	T_Agg,
+	T_WindowAgg,
+	T_Unique,
+	T_Hash,
+	T_SetOp,
+	T_LockRows,
+	T_Limit,
+	/* these aren't subclasses of Plan: */
+	T_NestLoopParam,
+	T_PlanRowMark,
+	T_PlanInvalItem,
+
+	/*
+	 * TAGS FOR PLAN STATE NODES (execnodes.h)
+	 *
+	 * These should correspond one-to-one with Plan node types.
+	 */
+	T_PlanState = 200,
+	T_ResultState,
+	T_ModifyTableState,
+	T_AppendState,
+	T_MergeAppendState,
+	T_RecursiveUnionState,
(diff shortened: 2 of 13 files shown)


src/include/parser/parsenodes.h
```diff
@@ -0,0 +1,249 @@
+#include "nodes.h"
```
apavlo: You are missing the standard file header

haojin2: Fixed

src/include/parser/parsenodes.h
```diff
@@ -0,0 +1,249 @@
...
+#include "pg_list.h"
+
+typedef enum SetOperation
+{
```
apavlo: This doesn't follow the Peloton style guidelines

haojin2: Where can I find that guideline?

src/include/parser/pg_parser.h
```diff
@@ -0,0 +1,104 @@
...
+
+#pragma once
+
+#include "parser/statements.h"
+#include "parser/pg_query.h"
+#include "parser/parser.h"
+#include "parser/parsenodes.h"
+
+namespace peloton {
+namespace pgparser {
```
haojin2: changed to parser

src/include/parser/nodes.h
```diff
@@ -0,0 +1,668 @@
...
+ * Portions Copyright (c) 1994, Regents of the University of California
+ *
+ * src/include/parser/nodes.h
+ *
+ *-------------------------------------------------------------------------
+ */
+#ifndef NODES_H
+#define NODES_H
+
+#ifdef __cplusplus
```
haojin2: This was the original code from libpg_query, I can try changing it and see if that causes any problem.

test/parser/pg_parser_test.cpp
```diff
@@ -0,0 +1,77 @@
...
+#include "parser/parser.h"
+
+namespace peloton {
+namespace test {
+
+//===--------------------------------------------------------------------===//
+// Parser Tests
+//===--------------------------------------------------------------------===//
+
+class PgParserTests : public PelotonTest {};
```
haojin2: Changed this to Pg_ParserTests.

apavlo: Actually, what if you just renamed it to `PostgresParserTests`? Use `Postgres` instead of `Pg` so that it is more clear.

src/include/parser/parsenodes.h
haojin2: I could not find any format.py under peloton folder, are there any written guides?

haojin2: I guess the format.py has been renamed to formatter.py

PauloAmora: There is a formatter at script/formatting/formatter.py

third_party/libpg_query/src/pg_query_parse.c
```diff
@@ -9,6 +9,8 @@
 #include <unistd.h>
 #include <fcntl.h>
 
+#define PARSE_STRING 1
```
haojin2: This is the control flag for enabling debug output for postgres parsetree. The control is now added in the same file on line 78

src/include/expression/star_expression.h
```diff
@@ -38,7 +38,7 @@ class StarExpression : public AbstractExpression {
   virtual void Accept(SqlNodeVisitor *v) { v->Visit(this); }
 
  protected:
-  StarExpression(const FunctionExpression &other) : AbstractExpression(other) {}
+  StarExpression(const AbstractExpression &other) : AbstractExpression(other) {}
```
jarulraj: Why this change?

src/include/parser/postgresparser.h
```diff
@@ -0,0 +1,107 @@
...
+  // transform helper for JoinExpr parsenodes
+  parser::JoinDefinition* JoinTransform(JoinExpr* root);
+
+  // transform helper for from clauses
+  parser::TableRef* FromTransform(List* root);
+
+  // transform helper for select targets
+  std::vector<expression::AbstractExpression*>* TargetTransform(List* root);
+
+  // transform helper for A_Expr nodes
```
jarulraj: Good documentation. Might want to expand what A_Expr means etc.

src/parser/parser_utils.cpp
```diff
@@ -110,21 +110,22 @@ void GetExpressionInfo(const expression::AbstractExpression* expr,
     case ExpressionType::STAR:
       inprint("*", num_indent);
       break;
-    case ExpressionType::COLUMN_REF:
-      // TODO: Fix this
-      inprint((expr)->GetExpressionName(), num_indent);
-      //if (expr->GetColumn() != NULL) inprint((expr)->GetColumn(), num_indent);
+    case ExpressionType::VALUE_TUPLE:
+      inprint((expr)->GetInfo().data(), num_indent);
```
jarulraj: Can we skip using `inprint` and directly use `LOG_TRACE`? Delete `inprint`.

test/parser/postgresparser_test.cpp
```diff
@@ -0,0 +1,273 @@
...
+TEST_F(PostgresParserTests, BasicTest) {
+  std::vector<std::string> queries;
+
+  // Simple select
+  queries.push_back("SELECT * FROM foo;");
+
+  auto parser = parser::PostgresParser::GetInstance();
+  // auto ref_parser = parser::Parser::GetInstance();
+  // Parsing
+  UNUSED_ATTRIBUTE int ii = 0;
```
jarulraj: Nice job with testing stuff!

test/parser/postgresparser_test.cpp
```diff
@@ -0,0 +1,273 @@
...
+  std::vector<std::string> queries;
+
+  // Select with functional call
+  queries.push_back("SELECT COUNT(*) FROM foo;");
+  queries.push_back("SELECT COUNT(DISTINCT id) FROM foo;");
+  queries.push_back("SELECT MAX(*) FROM foo;");
+  queries.push_back("SELECT MIN(*) FROM foo;");
+
+  auto parser = parser::PostgresParser::GetInstance();
+  // auto ref_parser = parser::Parser::GetInstance();
```
jarulraj: Don't forget to delete these references after removing the old parser.

third_party/libpg_query/src/pg_query_parse.c
```diff
@@ -92,7 +94,11 @@ PgQueryInternalParsetreeAndError pg_query_parse(const char* input)
 
 	parsetree_and_error = pg_query_raw_parse(input);
 
-	pg_query_exit_memory_context(ctx);
+#if PARSE_STRING == 1
```
jarulraj: Adding some comments in this function would help.

src/include/expression/star_expression.h
haojin2: If I remember correctly, this old version reported the following error when I was adding support for star expressions: "error: no matching function for call to ‘peloton::expression::AbstractExpression::AbstractExpression(const peloton::expression::FunctionExpression&)’
"

src/include/parser/postgresparser.h
haojin2: There're some documentations in the original libpg_query code already. If I add more it's gonna be a bit redundant.

test/parser/postgresparser_test.cpp
haojin2: I'm keeping those for the time being since sometimes I still want to compare with the Peloton parser to check my transformation. I can submit a version without the references but keep those locally, do you prefer that?

src/parser/parser_utils.cpp
haojin2: I think inprint is a nice helper function since it also handles the number of indents to give a nicer output? We may have to handle the indents manually if we use LOG_TRACE.

Issue #11, opened 2015-07-01
Remove table/column names from TupleValueExpression


apavlo:
I can't think of any reason that we need to store the table and column names in `TupleValueExpression` (other than for debugging). We should remove these attributes.

jarulraj:
Removed it.

Issue #10, opened 2015-06-30
Not picking up libmm14 on Ubuntu 14.04


apavlo:
> $ ./configure
> checking for mm_version in -lmm... no
> configure: error: Please install libmm library : libmm14

This is after installing the `libmm14` package.

mindbergh:
I passed the configure after installing `libmm14 libmm-dev libmm-dbg` .

PR #577, opened 2017-03-23
Add Worker Pool/ Networking Thread, worker thread separation

MattPerron:
This PR is not ready to merge, but I'd like to get feedback from @yingjunwu and other interested parties.

The current implementation has the following known issues:
1. A Client disconnection while worker threads are running queries result in a segfault.
2. The Worker Thread Pool has a single task queue, this may have a negative effect on scalability.
3. The transactional consistency of the separation of networking and worker threads is not well tested.
4. More testing is needed, particularly for the Worker Thread Pool. (Only simple functionality has been tested manually)
5. The number of worker threads/network threads needs tuning.

### DIFF
--- a/src/include/task/worker_pool.h
+++ b/src/include/task/worker_pool.h
@@ -25,6 +25,8 @@
 namespace peloton{
 namespace task{
 
+class WorkerPool;
+
 //task to submit to the queue
 class Task {
   friend class WorkerThread;
@@ -43,10 +45,10 @@ private:
 class WorkerThread{
 public:
 
-  void StartThread();
+  void StartThread(WorkerPool* worker_pool);
 
   // poll work queue, until exiting
-  static void PollForWork(WorkerThread* current_thread);
+  static void PollForWork(WorkerThread* current_thread, WorkerPool* current_pool);
 
   // wait for the current task to complete and shutdown the thread;
   void Shutdown();
@@ -70,13 +72,13 @@ public:
 
   void Shutdown();
 
-private:
-  WorkerPool(size_t num_threads, size_t task_queue_size);
 
+  WorkerPool(size_t num_threads, size_t task_queue_size);
+private:
   //TODO: use Bili's work stealing queues?
   peloton::LockFreeQueue<Task> task_queue_;
 
-  std::vector<WorkerThread> worker_threads_;
+  std::vector<std::unique_ptr<WorkerThread>> worker_threads_;
 
 };
 
--- a/src/include/task/worker_pool.h
+++ b/src/include/task/worker_pool.h
@@ -1,0 +1,85 @@
+//===----------------------------------------------------------------------===//
+//
+//                         Peloton
+//
+// worker_pool.h
+//
+// Identification: src/include/task/worker_pool.h
+//
+// Copyright (c) 2015-16, Carnegie Mellon University Database Group
+//
+//===----------------------------------------------------------------------===//
+
+#pragma once
+
+#include <vector>
+#include <thread>
+#include <memory>
+#include "container/lock_free_queue.h"
+
+//TODO: should be configurable
+#define DEFAULT_NUM_WORKER_THREADS (std::thread::hardware_concurrency())
+//TODO: make configurable, choose some more reasonable default value
+#define DEFAULT_TASK_QUEUE_LENGTH (std::thread::hardware_concurrency()*4)
+
+namespace peloton{
+namespace task{
+
+//task to submit to the queue
+class Task {
+  friend class WorkerThread;
+public:
+  inline Task() {};
+  //
+  inline Task(void(*func_ptr)(void *), void *func_args) :
+          func_ptr_(func_ptr), func_args_(func_args) {};
+
+private:
+  void(*func_ptr_)(void *);
+  void *func_args_;
+};
+
+// Data structure wraping worker thread
+class WorkerThread{
+public:
+
+  void StartThread();
+
+  // poll work queue, until exiting
+  static void PollForWork(WorkerThread* current_thread);
+
+  // wait for the current task to complete and shutdown the thread;
+  void Shutdown();
+
+private:
+  volatile bool shutdown_thread_;
+
+  std::thread worker_thread_;
+
+};
+
+// generic pool for executing arbitrary tasks asynchronously
+class WorkerPool {
+
+  friend class WorkerThread;
+public:
+  static WorkerPool& GetInstance();
+
+  // submit a task for asynchronous execution.
+  void SubmitTask(void(*func_ptr)(void *), void *func_args);
+
+  void Shutdown();
+
+private:
+  WorkerPool(size_t num_threads, size_t task_queue_size);
+
+  //TODO: use Bili's work stealing queues?
+  peloton::LockFreeQueue<Task> task_queue_;
+
+  std::vector<WorkerThread> worker_threads_;
+
+};
+
+
+}//task
+}//peloton
--- a/src/task/worker_pool.cpp
+++ b/src/task/worker_pool.cpp
@@ -16,17 +16,17 @@
 namespace peloton{
 namespace task{
 
-void WorkerThread::StartThread(){
+void WorkerThread::StartThread(WorkerPool* current_pool){
   shutdown_thread_ = false;
-  worker_thread_ = std::thread(WorkerThread::PollForWork, this);
+  worker_thread_ = std::thread(WorkerThread::PollForWork, this, current_pool);
 }
 
-void WorkerThread::PollForWork(WorkerThread* current_thread){
+void WorkerThread::PollForWork(WorkerThread* current_thread, WorkerPool* current_pool){
   size_t empty_count = 0;
   Task t;
-  while(!current_thread->shutdown_thread_){
+  while(!current_thread->shutdown_thread_ || !current_pool->task_queue_.IsEmpty()){
     // poll the queue
-    if(!WorkerPool::GetInstance().task_queue_.Dequeue(t)){
+    if(!current_pool->task_queue_.Dequeue(t)){
       empty_count++;
       if (empty_count == 10){
         empty_count = 0;
@@ -49,8 +49,8 @@ WorkerPool::WorkerPool(size_t num_threads, size_t task_queue_size) :
         task_queue_(task_queue_size){
   for (size_t thread_id = 0; thread_id < num_threads; thread_id++){
     // start thread on construction
-    WorkerThread worker = WorkerThread();
-    worker.StartThread();
+    std::unique_ptr<WorkerThread> worker(new WorkerThread());
+    worker->StartThread(this);
     worker_threads_.push_back(std::move(worker));
   }
 }
@@ -66,8 +66,9 @@ void WorkerPool::SubmitTask(void(*func_ptr)(void *), void *func_args) {
 
 void WorkerPool::Shutdown() {
   for (auto& thread : worker_threads_){
-    thread.Shutdown();
+    thread->Shutdown();
   }
+  worker_threads_.clear();
   return;
 }
 
--- a/src/task/worker_pool.cpp
+++ b/src/task/worker_pool.cpp
@@ -1,0 +1,76 @@
+//===----------------------------------------------------------------------===//
+//
+//                         Peloton
+//
+// worker_pool.cpp
+//
+// Identification: src/task/worker_pool.cpp
+//
+// Copyright (c) 2015-16, Carnegie Mellon University Database Group
+//
+//===----------------------------------------------------------------------===//
+
+#include "include/task/worker_pool.h"
+#include <unistd.h>
+
+namespace peloton{
+namespace task{
+
+void WorkerThread::StartThread(){
+  shutdown_thread_ = false;
+  worker_thread_ = std::thread(WorkerThread::PollForWork, this);
+}
+
+void WorkerThread::PollForWork(WorkerThread* current_thread){
+  size_t empty_count = 0;
+  Task t;
+  while(!current_thread->shutdown_thread_){
+    // poll the queue
+    if(!WorkerPool::GetInstance().task_queue_.Dequeue(t)){
+      empty_count++;
+      if (empty_count == 10){
+        empty_count = 0;
+        usleep(1);
+      }
+      continue;
+    }
+    // call the task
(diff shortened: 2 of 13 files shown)


test/task/worker_pool_test.cpp
```diff
@@ -0,0 +1,40 @@
...
+#include "task/worker_pool.h"
+
+namespace peloton{
+namespace test{
+
+class WorkerPoolTest : public PelotonTest {};
+
+static void blah(UNUSED_ATTRIBUTE void* args){
+  sleep(2);
+  printf("whatup\n");
```
yingjunwu: remove this line and you will pass the validator test.

src/include/task/worker_pool.h
```diff
@@ -0,0 +1,87 @@
...
+//===----------------------------------------------------------------------===//
+
+#pragma once
+
+#include <vector>
+#include <thread>
+#include <memory>
+#include "container/lock_free_queue.h"
+
+//TODO: should be configurable
```
yingjunwu: we should discuss where to put these tunable variables.

MattPerron:
@yingjunwu See my comment above.
I am aware of a number of issues, and I agree that it is not ready to merge. 
I'm looking for feedback about the possibility of bugs with concurrency control. 
Can you comment on that?

yingjunwu:
I read your code again, and I don't think this PR can affect the correctness of the concurrency control protocol. 

The purpose of this PR is to seprate the worker thread from the networking thread. Originally, a networking thread should be responsible for both message handling and query processing. On receiving a message from the client side, a networking thread had to perform query parsing, query planning, and query execution sequentially, prior to handling the next message. By separating the worker thread from the networking thread, the DBMS can simply leverage networking thread for handling message, and all the query-related tasks can be assigned to a background worker thread.

I think the thread separation discussed above is easy to understand and will not cause any concurrent bugs. But I would like to ask some other (irrelavant) questions:

1. Are we trying to achieve intra-_transaction_ parallelism? That is, if there are two operations in a transaction, e.g., `READ(A), READ(B)`, are we going to parallelize the execution of these two operations? Originally we should execute them one by one.

2. We definitely want to achieve intra-_operation_ parallelism. For example, we would like to parallelize the execution of full table scan and aggregations. Then how many threads should we use to parallelize a single analytical operation? And how to coordinate these threads? (e.g., which part of table should a certain thread scan?) I think @pmenon's code should be responsible for that.

apavlo:
@yingjunwu Do we have any tests that check for concurrency control correctness from JDBC/SQL? That way we can know whether @MattPerron's changes break anything.

yingjunwu:
@apavlo No we don't have that. We even don't have any test cases for JDBC.

MattPerron:
@yingjunwu I don't see the distinction between intra-transaction and intra-operation parallelism from the perspective of concurrency control. The effect is the same right? if the seq scan operation on a single table is done in parallel, then you will necessarily have READ(A) and READ(B) happening in parallel.  

The questions of "how many threads should we use to parallelize a single analytical operation? And how to coordinate these threads?" I think are best left to another discussion, unless there is a compelling reason they should coordinate with the concurrency control mechanism.

apavlo:
Did anyone test this with OLTP-Bench or JDBC? It's crashing when you try to load in the TPC-C database:

```
2017-03-29 20:48:12 [src/wire/packet_manager.cpp:405:ExecParseMessage] DEBUG - PrepareStatement[S_1] => BEGIN
2017-03-29 20:48:12 [src/wire/packet_manager.cpp:405:ExecParseMessage] DEBUG - PrepareStatement[S_2] => INSERT INTO ITEM VALUES ($1, $2, $3, $4, $5)
2017-03-29 20:48:12 [src/tcop/tcop.cpp:302:PrepareStatement] DEBUG - Statement Prepared: Statement[S_2] -> INSERT INTO ITEM VALUES ($1, $2, $3, $4, $5) (TablesRef={21}, ReplanNeeded=0, QueryType=INSERT)
2017-03-29 20:48:12 [src/wire/packet_manager.cpp:405:ExecParseMessage] DEBUG - PrepareStatement[S_3] => COMMIT
[err] event_queue_remove: 0x7fffc602a020(fd 57) not on queue 8
2017-03-29 20:48:24 [src/wire/packet_manager.cpp:79:~PacketManager] DEBUG - Removed PacketManager [count=4]
2017-03-29 20:48:24 [src/wire/packet_manager.cpp:79:~PacketManager] DEBUG - Removed PacketManager [count=3]
2017-03-29 20:48:24 [src/wire/packet_manager.cpp:79:~PacketManager] DEBUG - Removed PacketManager [count=2]
2017-03-29 20:48:24 [src/wire/packet_manager.cpp:79:~PacketManager] DEBUG - Removed PacketManager [count=1]
2017-03-29 20:48:24 [src/wire/packet_manager.cpp:79:~PacketManager] DEBUG - Removed PacketManager [count=0]
terminate called without an active exception

Thread 14 "peloton" received signal SIGABRT, Aborted.
[Switching to Thread 0x7fffe37fd700 (LWP 26997)]
0x00007ffff45f2428 in __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:54
54      ../sysdeps/unix/sysv/linux/raise.c: No such file or directory.
```

MattPerron:
I used the JDBC test, but not OLTP-Bench
This complaint from libevent seems to be a problem.
`[err] event_queue_remove: 0x7fffc602a020(fd 57) not on queue 8`

Let me take a look.

apavlo:
Yes, I just tested with `testing/jdbc/test_jdbc.sh` and it seems to work.

src/task/worker_pool.cpp
```diff
@@ -0,0 +1,77 @@
...
+}
+
+void WorkerThread::PollForWork(WorkerThread* current_thread, WorkerPool* current_pool){
+  size_t empty_count = 0;
+  Task t;
+  while(!current_thread->shutdown_thread_ || !current_pool->task_queue_.IsEmpty()){
+    // poll the queue
+    if(!current_pool->task_queue_.Dequeue(t)){
+      empty_count++;
+      if (empty_count == 10){
```
Michael-Tieying-Zhang: @MattPerron Why there is a magic number?

src/task/worker_pool.cpp
```diff
@@ -0,0 +1,77 @@
...
+        empty_count = 0;
+        usleep(1);
+      }
+      continue;
+    }
+    // call the task
+    t.func_ptr_(t.func_args_);
+    empty_count = 0;
```
Michael-Tieying-Zhang: I guess you wanna break dead lock or similar stuff. If so, probably using instruction is more efficient, say `mm_pause()`. We can make it more robust like this: 
```
#if defined(USE_mm_pause)
#define PAUSE _mm_pause
#else
#define PAUSE usleep
#endif
```

src/wire/packet_manager.cpp
```diff
@@ -293,6 +295,19 @@ bool PacketManager::HardcodedExecuteFilter(std::string query_type) {
   return true;
 }
 
+// Execute the Simple Query Protocol with a worker thread
+void PacketManager::ExecQueryMessageAsync(void* conn_ptr){
+  LOG_DEBUG("Exec Execute Message Async starting");
```
Michael-Tieying-Zhang: `LOG_DEBUG("Exec Query Message Async starting");`

src/wire/packet_manager.cpp
```diff
@@ -769,6 +784,19 @@ bool PacketManager::ExecDescribeMessage(InputPacket *pkt) {
...
+// Execute the exec message in a worker thread
+void PacketManager::ExecExecuteMessageAsync(void* conn_ptr){
+  LOG_DEBUG("Exec Execute Message Async starting");
+  wire::LibeventSocket * conn = (wire::LibeventSocket *)conn_ptr;
+  InputPacket *pkt = &conn->rpkt;
+  PacketManager *pkt_mgr = &conn->pkt_manager;
+  pkt_mgr->ExecExecuteMessage(pkt, (size_t)conn->thread_id);
+  conn->worker_executing = false;
+  // tell the networking thread it can write
+  event_active(conn->event, EV_WRITE, 0);
```
Michael-Tieying-Zhang: Why here is 0, but above is 1 (in ExecQueryMessageAsync)

src/wire/packet_manager.cpp
```diff
@@ -769,6 +784,19 @@ bool PacketManager::ExecDescribeMessage(InputPacket *pkt) {
...
+  LOG_DEBUG("Exec Execute Message Async done, activating the event");
+}
+
 void PacketManager::ExecExecuteMessage(InputPacket *pkt, const size_t thread_id) {
```
seojungmin: I think we could remove ExecExecuteMessage() and ExecQueryMessage(), since these are no longer called.

haojin2:
Seem like the error may be caused by a segfault, when I used gdb to run peloton and tpcc benchmark, I got the following seg fault in peloton::wire::LibeventSocket::BufferWriteBytesHeader:
Thread 8 "peloton" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7fffe17ff700 (LWP 6679)]
0x00007ffff780bb68 in peloton::wire::LibeventSocket::BufferWriteBytesHeader (
    this=0x7fffdf00f980, pkt=0x0)
    at /home/michael/Documents/peloton/src/wire/libevent_socket.cpp:391
391	  if (pkt->skip_header_write) {
(gdb) bt
#0  0x00007ffff780bb68 in peloton::wire::LibeventSocket::BufferWriteBytesHeader
    (this=0x7fffdf00f980, pkt=0x0)
    at /home/michael/Documents/peloton/src/wire/libevent_socket.cpp:391
#1  0x00007ffff780b4f1 in peloton::wire::LibeventSocket::WritePackets (
    this=0x7fffdf00f980)
    at /home/michael/Documents/peloton/src/wire/libevent_socket.cpp:182
#2  0x00007ffff7816054 in peloton::wire::StateMachine (conn=0x7fffdf00f980)
    at /home/michael/Documents/peloton/src/wire/libevent_callbacks.cpp:182
#3  0x00007ffff7815bd1 in peloton::wire::EventHandler (connfd=36, ev_flags=2, 
    arg=0x7fffdf00f980)
    at /home/michael/Documents/peloton/src/wire/libevent_callbacks.cpp:72
#4  0x00007ffff53cd4c9 in event_base_loop ()
   from /usr/lib/x86_64-linux-gnu/libevent-2.0.so.5
#5  0x00007ffff780d89b in peloton::wire::LibeventMasterThread::StartWorker (
    worker_thread=0x7fffee457900)
    at /home/michael/Documents/peloton/src/wire/libevent_thread.cpp:72
#6  0x00007ffff7815758 in std::_Bind_simple<void (*(peloton::wire::LibeventWorkerThread*))(peloton::wire::LibeventWorkerThread*)>::_M_invoke<0ul>(std::_Index_tuple<0ul>) (this=0x7fffee6c8ee8) at /usr/include/c++/5/functional:1531
#7  0x00007ffff7815344 in std::_Bind_simple<void (*(peloton::wire::LibeventWorkerThread*))(peloton::wire::LibeventWorkerThread*)>::operator()() (
    this=0x7fffee6c8ee8) at /usr/include/c++/5/functional:1520
#8  0x00007ffff7814fea in std::thread::_Impl<std::_Bind_simple<void (*(peloton::---Type <return> to continue, or q <return> to quit---
wire::LibeventWorkerThread*))(peloton::wire::LibeventWorkerThread*)> >::_M_run() (this=0x7fffee6c8ed0) at /usr/include/c++/5/thread:115
#9  0x00007ffff4a6dc80 in ?? () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#10 0x00007ffff5b4c6ba in start_thread (arg=0x7fffe17ff700)
    at pthread_create.c:333
#11 0x00007ffff41d382d in clone ()
    at ../sysdeps/unix/sysv/linux/x86_64/clone.S:109

haojin2:
By comparing the outputs of two runs of oltpbench with this PR merged, I found out that the system fails randomly during the loading of tables since the traces of the two runs have different lengths. This suggests that there may be a race condition.

haojin2:
The SIGABRT is due to that when the "not on queue 8" error occurs, libevent calls exit so the thread is terminated without being joined first, and thus we get SIGABRT when the thread is terminated.
To handle this, we need to add a destuctor for WorkerThread to first handle the SIGABRT:
https://github.com/haojin2/peloton/blob/pool/src/task/worker_pool.cpp#L20
On the other hand, it may be useful to reinstall libevent library from original source code to get debug symbols for gdb: git@github.com:libevent/libevent.git.

seojungmin:
@haojin2 Have you looked at this issue after compiling with -fsanitize=thread set? This will pick up some data races.

haojin2:
@seojungmin  How do I set this flag? Is that specified in some CMake files?

seojungmin:
@haojin2 One possible way to add this option is to set `CMAKE_CXX_FLAGS:STRING=-fsanitize=thread -fuse-ld=gold` in build/CMakeCache.txt once you cmake in the build directory. I tried this just now, and it detected some data races over the packet buffers, and over some others, with oltpbench.  Also be aware that the sanitizer may also produce false positives like all other detectors.
Hope this helps.

src/wire/packet_manager.cpp
seojungmin: AFAIK, this option is deprecated, at least in libevent 2.0 or above.

src/wire/libevent_callbacks.cpp
```diff
@@ -152,10 +152,14 @@ void StateMachine(LibeventSocket *conn) {
           conn->pkt_manager.is_started = true;
         } else {
           // Process all other packets
-          status = conn->pkt_manager.ProcessPacket(&conn->rpkt, (size_t)conn->thread_id);
+          status = conn->pkt_manager.ProcessPacket(conn);
         }
-
-        if (status == false) {
+        if (conn->worker_executing){
```
ChTimTsubasa: Why would it fall into this state while the 'worker_executing' thread is still false? Shouldn't it be set to true in "void PacketManager::ExecExecuteMessageAsync(void* conn_ptr)" already?

apavlo:
We decided to abandon this PR and start over. Thanks everyone for helping out with it. We'll hopefully get multi-threading working soon!

PR #583, opened 2017-03-24
Support Insert statement for the new parser

malin1993ml:
Three different types of INSERT statements are all supported:
(1) INSERT INTO foo VALUES (1, 2, 3);
(2) INSERT INTO foo VALUES (1, 2, 3), (4, 5, 6);
(3) INSERT INTO foo select * from bar where id = 5;

table_info_ is added in InsertStatement temporarily to make the old parser be compiled.

### DIFF
--- a/src/include/parser/insert_statement.h
+++ b/src/include/parser/insert_statement.h
@@ -26,11 +26,12 @@ namespace parser {
  */
 struct InsertStatement : SQLStatement {
   InsertStatement(InsertType type)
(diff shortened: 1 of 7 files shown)


malin1993ml:
Please wait from merging this PR. There's another bug that I need to fix before merging.

apavlo:
OK. Let me know.

malin1993ml:
I already fixed the bug and rebased. Should be good after the build passes.

PR #585, opened 2017-03-25
Support different isolation levels

yingjunwu:
In this PR, I attempt to introduce different isolation levels, i.e., `READ_ONLY`, `SERIALIZABLE`, `REPEATABLE_READS`, `READ_COMMITTED`, into Peloton.

Please note that, for now, our `SERIALIZABLE` is equivalent to `REPEATABLE_READS`, as we do not solve phantom problem. We do not support `READ_UNCOMMITTED` as it is too trival. Following Oracle's [article](http://www.oracle.com/technetwork/testcontent/o65asktom-082389.html), I treat `READ_ONLY` as a special isolation level.

I think @jessesleeping is the best person for reviewing this PR, but as he's enjoying life at Facebook, we need to find another person to review it.


src/concurrency/local_epoch.cpp
```diff
@@ -16,17 +16,29 @@
...
     } else if (epoch_lower_bound_ >= epoch_id) {
-    // epoch_lower_bound_ has already been updated by the GC.
-    // have to grab a newer epoch_id.
-      epoch_lock_.Unlock();
-      return false;
-    } 
+    
+      if (is_snapshot_read == true) {
+    
+        epoch_lower_bound_ = epoch_id - 1;
```
jessesleeping: Why do we set the epoch_lower_bound_ to a smaller epoch? I think it's supposed to be a non-decreasing number, otherwise the GC process will be confused.

src/concurrency/timestamp_ordering_transaction_manager.cpp
```diff
@@ -397,83 +163,188 @@ bool TimestampOrderingTransactionManager::AcquireOwnership(
 // It should not be called if the tuple is in the write set as commit and abort
 // will release the write lock anyway.
 void TimestampOrderingTransactionManager::YieldOwnership(
-    UNUSED_ATTRIBUTE Transaction *const current_txn, const oid_t &tile_group_id,
+    UNUSED_ATTRIBUTE Transaction *const current_txn,
```
jessesleeping: Remove the UNUSED_ATTRIBUTE tag. The argument is actually used.

src/concurrency/decentralized_epoch_manager.cpp
```diff
@@ -18,33 +18,34 @@ namespace concurrency {
 
 
   // enter epoch with thread id
-  cid_t DecentralizedEpochManager::EnterEpoch(const size_t thread_id) {
+  cid_t DecentralizedEpochManager::EnterEpoch(const size_t thread_id, const bool is_snapshot_read) {
```
jessesleeping: Does is_snapshot_read == true mean ``SNAPSHOT`` or ``READONLY``?

src/concurrency/timestamp_ordering_transaction_manager.cpp
```diff
@@ -397,83 +163,188 @@ bool TimestampOrderingTransactionManager::AcquireOwnership(
...
+        current_txn->RecordReadOwn(location);
+      }
+      // if the ownership has already been acquired, 
+      // then no need to update read/write set.
+      return true;
+    
+    } else {
+      // if it's not select for update, then directly return true.
+      // no need to update read/write set.
+      return true;
```
jessesleeping: Why don't we update the last read timestamp here? What if the read version is write-locked by a transaction that has a timestamp earlier than the reader?

src/concurrency/decentralized_epoch_manager.cpp
yingjunwu: yes.

src/concurrency/timestamp_ordering_transaction_manager.cpp
yingjunwu: it's used in assertion. the compiler will complain when compiling under release mode.

src/concurrency/timestamp_ordering_transaction_manager.cpp
yingjunwu: because it's for snapshot isolation. Under this isolation level, a transaction can read a version without setting the last read timestamp.

src/concurrency/decentralized_epoch_manager.cpp
jessesleeping: Which one? ``SNAPSHOT ISOLATION`` or ``READONLY``?

yingjunwu:
This PR requires careful code review. In addition to implementing all isolation levels, I refactored the epoch manager code to make it easier to understand. I removed all the old logging code and prepared APIs for the new logging component. I commented out the storage manager code because it is tightly coupled with the old logging code. According to our plan, we will refactor the storage management component in May. As discussed before, I also temporarily commented out some logging-related test cases.

src/concurrency/decentralized_epoch_manager.cpp
yingjunwu: both. Whenever a transaction is executed under snapshot isolation or declared as read-only, it will directly access a snapshot version.

src/concurrency/timestamp_ordering_transaction_manager.cpp
jessesleeping: Let's say we have the following txn:
- txnA (ts = 6)
- txnB (ts = 5)

And we have the following tuples/versions:
- tuple0_v0 (begin_ts = 3, end_ts = inf)
- tuple1_v0 (begin_ts = 4, end ts = inf)

Now we have the following operations:
1. txnA reads tuple0 -> txnA read tuple0_v0
2. txnB write locks tuple1
3. txnA reads tuple1 -> txnA read tuple1_v0
4. txnB updates tuple1 -> tuple1_v0 (begin_ts = 4, end_ts = 5), tuple1_v1 (begin_ts = 5, end_ts = inf)
5. txnB commits
6. txnA commits

In this case, A will end up reading tuple0_v0 and tuple1_v0. But for a consistent snapshot, it should read tuple0_v0 and tuple1_v1.

jessesleeping:
My dear Yingjun, how can I review a PR of +3,912/−9,540.....

jessesleeping:
Please add more test cases for different isolation levels.

yingjunwu:
@jessesleeping I think you could skip the logging part, because it's just some function declarations. You may start from reviewing epoch manager as well as the GC manager. I refactored them and unified API. I think we need to rewrite all the related test cases as the current test cases do not really test different isolation levels...

src/concurrency/timestamp_ordering_transaction_manager.cpp
jessesleeping: @yingjunwu ?

jessesleeping: Never mind.. I think that's an anomaly.

src/gc/transaction_level_gc_manager.cpp
```diff
@@ -50,18 +52,19 @@ void TransactionLevelGCManager::Running(const int &thread_id) {
...
-    auto &txn_manager = concurrency::TransactionManagerFactory::GetInstance();
-    auto max_cid = txn_manager.GetMaxCommittedCid();
+    auto &epoch_manager = concurrency::EpochManagerFactory::GetInstance();
+    
+    auto expired_eid = epoch_manager.GetExpiredEpochId();
     
     // When the DBMS has started working but it never processes any transaction,
-    // we may see max_cid == MAX_CID.
-    if (max_cid == MAX_CID) {
+    // we may see expired_eid == MAX_EID.
```
jessesleeping: Please explain why we will observe that equation.

src/gc/transaction_level_gc_manager.cpp
```diff
@@ -81,13 +84,13 @@ void TransactionLevelGCManager::Running(const int &thread_id) {
 }
 
 
-void TransactionLevelGCManager::RecycleTransaction(std::shared_ptr<GCSet> gc_set, const cid_t &timestamp) {
```
jessesleeping: Why did we use timestamp to hash the context? I think I may be the one who wrote this.. but I cannot remember why I did that..

src/gc/transaction_level_gc_manager.cpp
```diff
@@ -135,27 +138,28 @@ int TransactionLevelGCManager::Unlink(const int &thread_id, const cid_t &max_cid
     }
   }  // end for
 
-  auto safe_max_cid = concurrency::TransactionManagerFactory::GetInstance().GetNextCommitId();
+  eid_t safe_expired_eid = concurrency::EpochManagerFactory::GetInstance().GetNextEpochId();
```
jessesleeping: May be we should add a COMPILER FENCE before acquiring this epoch id? We need to reach a barrier here.

src/concurrency/local_epoch.cpp
jessesleeping: @yingjunwu ?

src/concurrency/timestamp_ordering_transaction_manager.cpp
yingjunwu: Let me think about this.

yingjunwu: Please read the original snapshot isolation paper https://www.microsoft.com/en-us/research/wp-content/uploads/2016/02/tr-95-51.pdf first, and then we discuss offline.

src/concurrency/local_epoch.cpp
yingjunwu: @jessesleeping The local thread is unaware of the progress of any other threads. It is possible that thread `TH1` is already at epoch 100, but another thread `TH2` is still at epoch 10. In this case, the centralized epoch manager should tell `TH1` that the progress from a global view is still at epoch 10.

yingjunwu:
@jessesleeping I agree with you that we need to spend more time testing the system backend, including transaction manager, epoch manager, GC, and logging. It's not that easy and we should review the code carefully.

src/gc/transaction_level_gc_manager.cpp
yingjunwu: I think I've already explained it. Initially, `expired_eid` is set to `MAX_EID`. So it is possible for the GC to observe the equation.

src/gc/transaction_level_gc_manager.cpp
yingjunwu: You were reviewing the removed code. Originally we use timestamp to hash the context. But now I have changed it to `thread_id`. We no longer pass timestamp to the GC, because the GC only needs to know the `epoch_id` of a GC context.

src/gc/transaction_level_gc_manager.cpp
yingjunwu: why? The only requirement is that `safe_expired_eid` is larger than the current `epoch_id`. Even if the compiler/hardware reorders the instruction, it does not affect the correctness of the GC.

apavlo:
Disable logger :-1:

yingjunwu:
@jessesleeping I have finished implementing `REPEATABLE_READS` and `SNAPSHOT_ISOLATION`. Please perform another round of code review. Thanks.

src/concurrency/timestamp_ordering_transaction_manager.cpp
```diff
@@ -397,83 +184,286 @@ bool TimestampOrderingTransactionManager::AcquireOwnership(
...
+    LOG_TRACE("PerformRead (%u, %u)\n", location.block, location.offset);
+    auto &manager = catalog::Manager::GetInstance();
+    auto tile_group_header = manager.GetTileGroup(tile_group_id)->GetHeader();
+
+    // Check if it's select for update before we check the ownership 
+    // and modify the last reader cid
+    if (acquire_ownership == true) {
+
+      // get the latest version of this tuple.
+      location = *(tile_group_header->GetIndirection(location.offset));
```
chenboy: I don't quite understand here. why can we acquire ownership of a version that is newer than the snapshot version?

src/executor/update_executor.cpp
```diff
@@ -139,44 +139,67 @@ bool UpdateExecutor::DExecute() {
...
 
     LOG_TRACE("Visible Tuple id : %u, Physical Tuple id : %u ",
               visible_tuple_id, physical_tuple_id);
 
-    bool is_owner = transaction_manager.IsOwner(current_txn, tile_group_header,
-                                                physical_tuple_id);
+    ///////////////////////////////////////////////////////////
+    // if running at snapshot isolation, 
+    // then we need to retrieve the latest version of this tuple.
+    if (current_txn->GetIsolationLevel() == IsolationLevelType::SNAPSHOT) {
```
chenboy: Same here. I don't understand why can we update a version that is newer than the snapshot version.

yingjunwu:
@apavlo I've talked to @pmenon before making this change. The reason is that the original design of storage manager is heavily dependent on the logging component, which is unreasonable. I think that the storage manager requires a complete redesign. As `sorter` only needs memory, we can safely replace it with malloc and free.

apavlo:
Only the core parts of the storage system should be calling malloc/free directly. I don't see why this change is necessary.

yingjunwu:
@apavlo I understand. But anyway we have to completely refactor the storage manager code. It depends on logging stuff, which is wired. Meanwhile, it also assumes the underlying NVRAM is an emulated one, which is unreasonable for a real-world DBMS. I think we can have a discussion on the storage manager part after SIGMOD.

src/concurrency/timestamp_ordering_transaction_manager.cpp
```diff
@@ -84,280 +83,64 @@ void TimestampOrderingTransactionManager::InitTupleReserved(
...
 bool TimestampOrderingTransactionManager::IsWritten(
-    Transaction *const current_txn,
+    UNUSED_ATTRIBUTE Transaction *const current_txn,
     const storage::TileGroupHeader *const tile_group_header,
     const oid_t &tuple_id) {
-  return IsOwner(current_txn, tile_group_header, tuple_id) &&
-         tile_group_header->GetBeginCommitId(tuple_id) == MAX_CID;
+  auto tuple_begin_cid = tile_group_header->GetBeginCommitId(tuple_id);
+  
+  return tuple_begin_cid == MAX_CID;
```
yingjunwu: done.

src/concurrency/timestamp_ordering_transaction_manager.cpp
yingjunwu: it's `select_for_update`. We read the snapshot version and update the latest version. We need to acquire write locks to prevent write-write conflicts. See Section VIII.B in the paper: Multi-version Concurrency via Timestamp Range Conflict Management.

src/concurrency/timestamp_ordering_transaction_manager.cpp
```diff
@@ -397,83 +184,286 @@ bool TimestampOrderingTransactionManager::AcquireOwnership(
...
-    PL_ASSERT(GetLastReaderCommitId(tile_group_header, tuple_id) <=
-              current_txn->GetBeginCommitId());
-    // Increment table read op stats
-    if (FLAGS_stats_mode != STATS_TYPE_INVALID) {
-      stats::BackendStatsContext::GetInstance()->IncrementTableReads(
-          location.block);
+  } // end SNAPSHOT
+
+  //////////////////////////////////////////////////////////
+  //// handle READ_COMMITTED
```
yingjunwu: I think we'd better keep them in separate so as to make the logic clean enough for review.

src/concurrency/transaction.cpp
```diff
@@ -122,16 +116,17 @@ void Transaction::RecordUpdate(const ItemPointer &location) {
       return;
     }
     PL_ASSERT(false);
+  } else {
```
yingjunwu: It's possible. Just consider the `select_for_update` case. The DBMS locks the version without recording it into read/write set.

Comments added.

src/concurrency/transaction.cpp
```diff
@@ -172,7 +165,8 @@ bool Transaction::RecordDelete(const ItemPointer &location) {
     }
     PL_ASSERT(false);
   } else {
-    PL_ASSERT(false);
+    rw_set_[tile_group_id][tuple_id] = RWType::UPDATE;
```
yingjunwu: good catch here. should change `RWType::UPDATE` to `RWType::DELETE`. Similar as the previous one, a DBMS can delete a tuple after it `select_for_update` this tuple.

src/executor/update_executor.cpp
yingjunwu: Similarly, because we always need to update the latest version of a tuple, we can hold a lock on a version that is newer than the snapshot one.

Another paper for your reference: a critique of ansi sql isolation levels.

yingjunwu:
@apavlo So I have revised the code based on the review comments. The snapshot isolation part is done. However, the logging component is still unfinished, as it depends on the PR #577. Also, the storage management component should be redesigned.

apavlo:
@yingjunwu Let's discuss in person. I need to figure out what PRs we are going to need to merge.

yingjunwu:
I think once @seojungmin fixed the `optimizer_sql_test` bug, the PR can be merged.

src/gc/transaction_level_gc_manager.cpp
```diff
@@ -96,18 +100,19 @@ int TransactionLevelGCManager::Unlink(const int &thread_id,
...
-        if (res == true) {
-          DeleteFromIndexes(garbage_ctx);
-          // Add to the garbage map
-
-          garbages.push_back(garbage_ctx);
-          tuple_counter++;
-        }
-        return res;
-      });
+    [this, &garbages, &tuple_counter, expired_eid]
```
haoxianghua: 'this' value is never used in the lambda function

src/gc/transaction_level_gc_manager.cpp
```diff
@@ -96,18 +100,19 @@ int TransactionLevelGCManager::Unlink(const int &thread_id,
 
   // First iterate the local unlink queue
   local_unlink_queues_[thread_id].remove_if(
```
Michael-Tieying-Zhang: The function for `local_unlink_queues_` is that if an item in `unlink_queue_` is not garbage for now, that item should be put back again for the future gc. The reason you separate `local_unlink_queues_` from `unlink_queue_` is that `unlink_queue_` uses `LockFreeQueue`, but `local_unlink_queues_` uses `std::List`. So why `unlink_queue_` can not use  `std::List` as well, then we don't need `local_unlink_queues_`?

src/gc/transaction_level_gc_manager.cpp
```diff
@@ -116,14 +121,14 @@ int TransactionLevelGCManager::Unlink(const int &thread_id,
       break;
```
haoxianghua: In this part, we have unlink_queues_ and local_unlink_queues_. Is that right because unlink_queues_ uses peloton::LockFreeQueue and this queue doesn't support push the unrequired garbage_ctx back, so you maintain local_unlink_queues using std::list to save unrequired garbage_ctx. If so, can I use std::list instead? Because every thread has its own queue so we don't need to consider concurrency issue.

src/gc/transaction_level_gc_manager.cpp
yingjunwu: done.

src/include/concurrency/decentralized_epoch_manager.h
```diff
@@ -86,30 +100,29 @@ class DecentralizedEpochManager : public EpochManager {
...
-  }
+  virtual eid_t GetCurrentEpochId() override {
+    return current_global_epoch_id_.load();
+  }  
+
+private:
+
 
   inline uint32_t GetNextTransactionId() {
     return next_txn_id_.fetch_add(1, std::memory_order_relaxed);
```
yingjunwu: I need Matt's PR #577 to fully decentralize the transaction id generation.

src/include/gc/transaction_level_gc_manager.h
```diff
@@ -34,20 +34,20 @@ namespace gc {
 
 
 struct GarbageContext {
```
yingjunwu: There's no functon in this struct.

src/include/gc/transaction_level_gc_manager.h
```diff
@@ -91,7 +91,7 @@ class TransactionLevelGCManager : public GCManager {
     this->is_running_ = false;
   }
```
yingjunwu: done.

src/concurrency/decentralized_epoch_manager.cpp
```diff
@@ -18,69 +18,70 @@ namespace concurrency {
...
-      if (rt == true) {
-    
-        uint32_t next_txn_id = GetNextTransactionId();
+      local_epochs_.at(thread_id)->EnterEpoch(snapshot_global_epoch_id_, ts_type);
 
-        return (epoch_id << 32) | next_txn_id;
-      }
-    }
-  }
+      return (snapshot_global_epoch_id_ << 32) | 0x0;
```
yingjunwu: Yes. They are snapshot transactions. So we can assign them with same commit id.

src/concurrency/local_epoch.cpp
```diff
@@ -16,61 +16,53 @@
...
-    // have to grab a newer epoch_id.
-      epoch_lock_.Unlock();
-      return false;
-    } 
+  bool LocalEpoch::EnterEpoch(const eid_t epoch_id, const TimestampType ts_type) {
 
-    auto epoch_map_itr = epoch_map_.find(epoch_id);
-    // check whether the corresponding epoch exists.
-    if (epoch_map_itr == epoch_map_.end()) {
+    epoch_lock_.Lock();
```
yingjunwu: The global epoch manager thread needs to read data from LocalEpoch. So we must add lock to guarantee atomicity.

src/concurrency/local_epoch.cpp
```diff
@@ -16,61 +16,53 @@
...
 
-      std::shared_ptr<Epoch> epoch_ptr(new Epoch(epoch_id, 1));
+    // if this thread is never used or has been GC'd
+    if (epoch_id_lower_bound_ == UINT64_MAX) {
 
-      epoch_queue_.push(epoch_ptr);
-      epoch_map_[epoch_id] = epoch_ptr;
+      epoch_id_lower_bound_ = epoch_id - 1;
+    
+    } else if (epoch_id_lower_bound_ >= epoch_id) {
```
yingjunwu: The local epoch progress can be faster than the global one. In this case, `epoch_id_lower_bound_` can be larger than `epoch_id`.

src/concurrency/local_epoch.cpp
```diff
@@ -82,35 +74,37 @@ namespace concurrency {
         epoch_map_.erase(epoch_ptr->epoch_id_);
         epoch_queue_.pop();
       } else {
-        PL_ASSERT(epoch_ptr->epoch_id_ > epoch_lower_bound_);
-        epoch_lower_bound_ = epoch_ptr->epoch_id_ - 1;
+        PL_ASSERT(epoch_ptr->epoch_id_ > epoch_id_lower_bound_);
+        epoch_id_lower_bound_ = epoch_ptr->epoch_id_ - 1;
```
yingjunwu: The global epoch manger should also be responsible for updating local epoch. Otherwise, if there's no transaction assigned to a certain thread, its corresponding epoch manager will never be updated.

src/concurrency/local_epoch.cpp
```diff
@@ -82,35 +74,37 @@ namespace concurrency {
...
+    // which indicates that this thread is never used or has been GC'd for some time.
     if (epoch_queue_.size() == 0) {
-      epoch_lower_bound_ = epoch_id - 1;
+      epoch_id_lower_bound_ = epoch_id - 1;
     } else {
       while (epoch_queue_.size() != 0) {
+        // get the smallest epoch.
         auto &epoch_ptr = epoch_queue_.top();
+        // if there's no transaction in the epoch
         if (epoch_ptr->txn_count_ == 0) {
```
yingjunwu: The global epoch manger should also be responsible for updating local epoch. Otherwise, if there's no transaction assigned to a certain thread, its corresponding epoch manager will never be updated.

src/concurrency/local_epoch.cpp
yingjunwu: The global epoch manager can update the lower bound.

src/gc/transaction_level_gc_manager.cpp
yingjunwu: `LockFreeQueue` is thread-safe, while `List` is not.

src/gc/transaction_level_gc_manager.cpp
```diff
@@ -134,29 +139,31 @@ int TransactionLevelGCManager::Unlink(const int &thread_id,
...
-      concurrency::TransactionManagerFactory::GetInstance().GetNextCommitId();
-  for (auto &item : garbages) {
-    reclaim_maps_[thread_id].insert(std::make_pair(safe_max_cid, item));
+  // once the current epoch id is expired, then we know all the transactions
+  // that are active at this time point will be committed/aborted.
+  // at that time point, it is safe to recycle the version.
+  eid_t safe_expired_eid = concurrency::EpochManagerFactory::GetInstance().GetCurrentEpochId();
+
+  for(auto& item : garbages){
+      reclaim_maps_[thread_id].insert(std::make_pair(safe_expired_eid, item));
```
yingjunwu: It's a multi-map.

src/gc/transaction_level_gc_manager.cpp
yingjunwu: `local_unlink_queue` is thread-local and will be updated by a single GC thread, while `unlink_queue` is not because worker thread can push to-be-recycled tuple versions into it.

src/concurrency/local_epoch.cpp
Michael-Tieying-Zhang: > The global epoch manger should also be responsible for updating local epoch.


Updating `txn_count_`? Where is that?

src/gc/transaction_level_gc_manager.cpp
Michael-Tieying-Zhang: `std::multimap` is ordered. I guess you mean `std::unordered_multimap`.

apavlo:
Other than @Michael-Tieying-Zhang's [comment](https://github.com/cmu-db/peloton/pull/585#discussion_r120016263), is everything else ready to go in this PR?

If so, then @yingjunwu can rebase and I will merge.

src/concurrency/local_epoch.cpp
yingjunwu: While it is true that we do not update  `txn_count_` in `GetExpiredEpochId()`, I suggest leaving the code segment there as we may further optimize the code to remove the `epoch_lock_`. In that case data race can occur and we need to check `txn_count_` in `GetExpiredEpochId()`.

src/gc/transaction_level_gc_manager.cpp
yingjunwu: We should not use `std::unordered_multimap`. Our design is to sort the garbage versions according to epoch id. Versions with the oldest epoch id can be reclaimed first. Please check the code in `Reclaim()` (https://github.com/yingjunwu/peloton/blob/f5ec56f1c7f6e95f6e19a2409857f05e7ed6bb61/src/gc/transaction_level_gc_manager.cpp#L155). I have comments there.

yingjunwu:
@apavlo I suggest we should perform a performance test before or after merging the PR. Now we can support different isolation levels. We can test whether varying the isolation levels help improve the performance.

src/concurrency/local_epoch.cpp
domino-succ: Ok. Please add TODO tag in the code.

ChTimTsubasa:
I would do a test no later than tomorrow as soon as I got the account.

apavlo:
@yingjunwu I don't think we should worry about performance at this point. We should worry about correctness and stability. Can you rebase?

@ChTimTsubasa What test are you referring to? If you mean OLTP-Bench, then it will crash because of the bug that's fixed in #672. We can't run OLTP-Bench without that.

yingjunwu:
Please refer to PR #691.

PR #586, opened 2017-03-28
[FIX] proper parameter indexing in insert plans

akamaus:
Looks like there is a mistake in building parameter_vector_ for insert plan in case insert statement contains column list and value list with constants and parameters. This case wasn't covered by test so I added one.


apavlo:
@hzxa21 Can you check whether this works with the new parser?

hzxa21:
@apavlo This should work with the new parser.

apavlo:
@hzxa21 Let's merge this PR after we merge #587 

@akamaus This is awesome. Thanks!

apavlo:
@akamaus This breaks the build now. I thought I could just rebase from github but it merged it too, so I had to roll this back. Can you submit a new PR?

Issue #551, opened 2017-02-27
Support NULL value insertion


malin1993ml:
Right now we cannot insert a tuple with NULL values into table:
```sql
postgres=# create table a(id int, value varchar);
create table a(id int, value varchar) 0
postgres=# insert into a values(1, 'hi');
insert into a values(1, 'hi') 1
postgres=# insert into a values(1, NULL);
Error parsing SQL statement
```

Issue #373, opened 2016-12-02
Invalid PRIMARY KEY DDL Syntax


apavlo:
We can only declare a multi-attribute primary key by add it to each attribute definition:

```sql
CREATE TABLE new_order (
  no_w_id int NOT NULL PRIMARY KEY,
  no_d_id int NOT NULL PRIMARY KEY,
  no_o_id int NOT NULL PRIMARY KEY
);
```

This is invalid SQL. It should be:

```sql
CREATE TABLE new_order (
  no_w_id int NOT NULL,
  no_d_id int NOT NULL,
  no_o_id int NOT NULL,
  PRIMARY KEY (no_w_id,no_d_id,no_o_id)
);
```

Michael-Tieying-Zhang:
@apavlo What's the status for this issue? We cannot run the latest OLTP-Bench to test Peloton due to this issue.

apavlo:
What DBTYPE are you using? Don't use 'postgres'  use 'peloton'

Michael-Tieying-Zhang:
Got you

apavlo:
@haojin2 Let's discuss this tomorrow. This might be because of the problems with our parser.

apavlo:
We support both the incorrect and correct syntax for this:

**CORRECT SYNTAX**
```
postgres=# create table foo(a int, b varchar(32), primary key (a, b));
create table foo(a int, b varchar(32), primary key (a, b)) 0
postgres=# SELECT * FROM foo;
 a | b 
---+---
(0 rows)
```

**INCORRECT SYNTAX**
```
postgres=# DROP TABLE foo;
DROP TABLE foo 0
postgres=# create table foo(a int primary key, b varchar(32) primary key);
create table foo(a int primary key, b varchar(32) primary key) 
postgres=# SELECT * FROM foo;
 a | b 
---+---
(0 rows)
```

PR #587, opened 2017-03-28
New Parser

hzxa21:
This PR tears down the old parser and replaces it with the new parser based on libpg_query (resolve issue #521 ). It supports all functionalities the old parser supports and the TPC-C benchmark can successfully run on top of it. In addition, the new parser supports:

- DROP TABLE IF EXIST ...
- ORDER BY multiple expressions: `SELECT * FROM foo ORDER BY id, a+b, name DESC;`
- NULL value insertion (resolve issue #551): `insert into a values(1, NULL);` 
- Valid PRIMARY KEY DDL Syntax (resolve issue #373)

All test cases are refactored and passed except that there are memory leaks in pg_query_parse() in stats_test. We have examined this memory leaks for a while and it may because libpg is not thread_safe in our compiled version. @haojin2 @jarulraj Can you help to look into this and review this PR?


hzxa21:
@haojin2 https://github.com/cmu-db/peloton/blob/master/third_party/libpg_query/src/pg_query_parse.c#L103
Why comment out this line? I guess it will cause some memory leaks without this call. Is there any particular concern?

haojin2:
If I remember correctly, freeing the memory context will also cause the Postgres parse tree to be freed, and valgrind will then report problems on our reads from the parse tree since it is marked as freed memory if you do make the call to free memory context. I think the current postgresparsertest is not reporting any "definitely lost" kind of memory leak upon exit, but you may still see "still reachable" blocks if you run "valgrind ./test/postgresparsertest" from the build folder.

hzxa21:
Got you. But I think we should free all the malloc memory explicitly in our codes. Even though I have done that, when we create worker thread to call the parser, the memory allocated in libpg MemoryContextCreate still leaks. That is what stats_test did. You can try and run the stats_test to see what is going on.

haojin2:
I guess on solution to this is we can free a parsenode whenever we finish the transform of it. We may be able to alter the original version of memory context exit to let it only free the memory other than the internal parse tree.

apavlo:
The last travis failure might be because of me. I restarted the build.

Issue #326, opened 2016-11-01
Possible to create tables with columns of the same name


MattPerron:
```
matt=# create table foo (blah INTEGER, blah INTEGER);
create table foo (blah INTEGER, blah INTEGER) 0
matt=# select * from foo;
 blah | blah 
------+------
(0 rows)

matt=# 
```

We currently do not error when creating a table which already exists.

MattPerron:
```
matt=# insert into foo values (0, 1);
insert into foo values (0, 1) 1
matt=# insert into foo values (1, 2);
insert into foo values (1, 2) 1
matt=# select * from foo;
 blah | blah 
------+------
    0 |    1
    1 |    2
(2 rows)

matt=# select * from foo where blah = 1;
 blah | blah 
------+------
    1 |    2
(1 row)

matt=# select * from foo where blah = 2;
 blah | blah 
------+------
(0 rows)

matt=# select * from foo where blah = 0;
 blah | blah 
------+------
    0 |    1
(1 row)

matt=# 

```

This is also troubling.

camellyx:
Thanks! We have fixed it in the new catalog.
We will also add tests that check this issue.

poojanilangekar:
@dengyuchenkit and @ChTimTsubasa this issue is similar to #903. The catalog throws the exception but it isn't handled appropriately.

PR #592, opened 2017-03-30
Catalog Checkpoint

AngLi-Leon:
This PR contains a preliminary implementation of the new catalog system (#362). We support 4 catalog tables, namely` pg_database`, `pg_table`, ` pg_index` and `pg_attribute` and a `catalog manager` with same API as the old catalog system. Besides, we have fixed following issue related to the new catalog:
- #326: Prevent to create table with same column names

Note that some of the APIs in the current implementation should be deprecated and we are going to resolve dependency of these APIs gradually.

We have added QueryMetricCatalog as an example new catalog object. It is in src/include/catalog/query_metrics_catalog.h, src/catalog/query_metrics_catalog.cpp.


src/catalog/abstract_catalog.cpp
```diff
@@ -0,0 +1,198 @@
...
+  executor.Init();
+  bool status = executor.Execute();
+
+  if (single_statement_txn) {
+    txn_manager.CommitTransaction(txn);
+  }
+  return status;
+}
+
+// @brief   Delete a tuple using index scan
```
apavlo: You want to use `/* ... */` for the function comments

src/catalog/catalog.cpp
```diff
@@ -30,413 +30,764 @@ Catalog *Catalog::GetInstance(void) {
...
+  // Create indexes on catalog tables, insert them into pg_index
+  // TODO: This should be hash index rather than tree index?? (but postgres use
+  // btree!!)
+
+  CreatePrimaryIndex(CATALOG_DATABASE_OID, DATABASE_CATALOG_OID);
+  CreatePrimaryIndex(CATALOG_DATABASE_OID, TABLE_CATALOG_OID);
+
+  CreateIndex(CATALOG_DATABASE_OID, DATABASE_CATALOG_OID,
+              std::vector<std::string>({"database_name"}),
+              std::string(DATABASE_CATALOG_NAME) + "_skey0", true,
```
apavlo: We should make a helper function for generating index names.

src/catalog/catalog.cpp
```diff
@@ -30,413 +30,764 @@ Catalog *Catalog::GetInstance(void) {
...
-  // Update catalog_db with this database info
-  auto tuple =
-      GetDatabaseCatalogTuple(
-          databases_[START_OID]->GetTableWithName(DATABASE_CATALOG_NAME)->GetSchema(),
-          database_id, database_name, pool_);
-  catalog::InsertTuple(
-      databases_[START_OID]->GetTableWithName(DATABASE_CATALOG_NAME),
-      std::move(tuple), txn);
-}
+  if (txn == nullptr) {
```
apavlo: This seems weird to start the txn for a calling method if they don't provide us one. Is this just for convenience? Why not force anybody that calls you to start a txn on their own?

src/catalog/catalog.cpp
```diff
@@ -30,413 +30,764 @@ Catalog *Catalog::GetInstance(void) {
...
+    LOG_TRACE(
+        "Cannot find the database. Return "
+        "RESULT_FAILURE.");
+    return ResultType::FAILURE;
+  }
+
+  oid_t table_oid =
+      TableCatalog::GetInstance()->GetTableOid(table_name, database_oid, txn);
+  if (table_oid != INVALID_OID) {
+    LOG_TRACE(
```
apavlo: @jarulraj Should we be throwing exceptions here instead?

src/catalog/catalog.cpp
```diff
@@ -30,413 +30,764 @@ Catalog *Catalog::GetInstance(void) {
...
+    CreatePrimaryIndex(database_name, table_name, txn);
+
+    if (single_statement_txn) {
+      txn_manager.CommitTransaction(txn);
+    }
+
+    return ResultType::SUCCESS;
   } catch (CatalogException &e) {
-    LOG_ERROR("Could not find a database with name %s", database_name.c_str());
+    LOG_TRACE("Can't found database. Return RESULT_FAILURE");
```
apavlo: Why not include the database name like in the previous error message? It would make it easier for debugging.

src/catalog/catalog.cpp
```diff
@@ -30,413 +30,764 @@ Catalog *Catalog::GetInstance(void) {
...
 }
 
 ResultType Catalog::CreatePrimaryIndex(const std::string &database_name,
-    const std::string &table_name) {
+                                       const std::string &table_name,
+                                       concurrency::Transaction *txn) {
+  auto &txn_manager = concurrency::TransactionManagerFactory::GetInstance();
+  bool single_statement_txn = false;
+
+  if (txn == nullptr) {
```
apavlo: Same comment as above. Should we just force everyone to invoke these methods in a txn?

src/catalog/catalog.cpp
```diff
@@ -30,413 +30,764 @@ Catalog *Catalog::GetInstance(void) {
...
-    key_schema = catalog::Schema::CopySchema(schema, key_attrs);
-    key_schema->SetIndexedColumns(key_attrs);
+      if (key_attrs.empty()) return ResultType::FAILURE;
 
-    std::string index_name = table->GetName() + "_PKEY";
+      key_schema = catalog::Schema::CopySchema(schema, key_attrs);
+      key_schema->SetIndexedColumns(key_attrs);
 
-    bool unique_keys = true;
+      std::string index_name = table->GetName() + "_pkey";
```
apavlo: Use an index name helper. Maybe make an IndexUtil method?

src/catalog/catalog.cpp
```diff
@@ -30,413 +30,764 @@ Catalog *Catalog::GetInstance(void) {
...
+  }
+
+  try {
+    auto database = GetDatabaseWithOid(database_oid);
+    try {
+      auto table = database->GetTableWithOid(table_oid);
+
+      std::vector<oid_t> key_attrs;
+      catalog::Schema *key_schema = nullptr;
+      index::IndexMetadata *index_metadata = nullptr;
```
apavlo: Stupid question: Are we eventually going to remove `IndexMetadata`?

src/catalog/catalog.cpp
```diff
@@ -521,254 +904,188 @@ std::unique_ptr<storage::DataTable> Catalog::CreateMetricsCatalog(
...
-  name_column.AddConstraint(
-      catalog::Constraint(ConstraintType::NOTNULL, not_null_constraint_name));
-
-  std::unique_ptr<catalog::Schema> database_schema(new catalog::Schema( {
-      id_column, name_column }));
-  return database_schema;
-}
-
-// Initialize database catalog schema
 std::unique_ptr<catalog::Schema> Catalog::InitializeDatabaseMetricsSchema() {
```
apavlo: We may want to put this code in a separate file so that it is easier to find.

src/include/catalog/catalog.h
```diff
@@ -70,85 +71,101 @@ class Catalog {
...
-                     const std::string &table_name,
-                     std::vector<std::string> index_attr,
-                     std::string index_name, bool unique, IndexType index_type);
+                         const std::string &table_name,
+                         const std::vector<std::string> &index_attr,
+                         const std::string &index_name, bool unique,
+                         IndexType index_type,
+                         concurrency::Transaction *txn = nullptr,
+                         bool is_catalog = false);
+  ResultType CreateIndex(oid_t database_oid, oid_t table_oid,
```
apavlo: Since this has a lot of input parameters, you should include comments that say what they all mean.

src/include/catalog/column_catalog.h
```diff
@@ -0,0 +1,70 @@
...
+  // Global Singleton
+  static ColumnCatalog *GetInstance(storage::Database *pg_catalog = nullptr,
+                                    type::AbstractPool *pool = nullptr);
+
+  ~ColumnCatalog();
+
+  inline oid_t GetNextOid() { return oid_++ | COLUMN_OID_MASK; }
+
+  // Write related API
+  bool InsertColumn(oid_t table_oid, const std::string &column_name,
```
apavlo: Again, we should include a description of what the parameters mean in the function header.

src/include/type/catalog_type.h
```diff
@@ -0,0 +1,77 @@
+//===----------------------------------------------------------------------===//
+//
+//                         Peloton
+//
+// catalog_type.h
+//
+// Identification: src/include/type/catalog_type.h
```
apavlo: @jarulraj We should discuss where to put this file. I think that it should probably be in 'src/include/catalog' not 'src/include/type'. Maybe we should rename it to `catalog_defaults.h` or something else. These are not really "types".

src/storage/data_table.cpp
```diff
@@ -800,7 +800,7 @@ void DataTable::DropIndexWithOid(const oid_t &index_oid) {
 
   for (std::size_t index_itr = 0; index_itr < index_count; index_itr++) {
     index = indexes_.Find(index_itr);
-    if (index->GetOid() == index_oid) {
+    if (index && index->GetOid() == index_oid) {
```
apavlo: Should this be `index != nullptr` instead of just `index`?

test/concurrency/testing_transaction_util.cpp
```diff
@@ -56,8 +56,7 @@ storage::DataTable *TestingTransactionUtil::CreateCombinedPrimaryKeyTable() {
   auto table_name = "TEST_TABLE";
   size_t tuples_per_tilegroup = 100;
   auto table = storage::TableFactory::GetDataTable(
-      INVALID_OID, INVALID_OID, table_schema, table_name, tuples_per_tilegroup,
-      true, false);
+     CATALOG_DATABASE_OID, 2, table_schema, table_name, tuples_per_tilegroup, true, false);
```
apavlo: What is '2'?

src/catalog/catalog.cpp
mengranwo: The reasons we start the txn for a calling method are:
(1) some of the test methods just use nullptr for convenience, but we got that fixed.
(2) in initialization part(https://github.com/cmu-db/peloton/blob/master/src/common/init.cpp#L69), it will pass nullptr to create database(since there is no transaction going right now). But I think you're right that we should force anyone to start txn before calling functions in catalog.

src/catalog/catalog.cpp
mengranwo: I'm not sure about the differences between throw catalog exceptions and set transaction's result type as failure. 
e.g. when user try to create a table with duplicate column names, catalog will throw exception or just set transaction result as failure?

src/catalog/catalog.cpp
mengranwo: [Those functions](https://github.com/AngLi-Leon/peloton/blob/master/src/catalog/catalog.cpp#L862) in catalog.cpp are for metrics, we didn't change anything in those functions and nobody seems to use them right now.

test/concurrency/select_for_update_txn_test.cpp
```diff
@@ -14,21 +14,21 @@
 //#include "common/harness.h"
 //#include "concurrency/transaction_tests_util.h"
 //
-//namespace peloton {
+// namespace peloton {
```
jarulraj: @yingjunwu Is this test still needed?

test/logging/recovery_test.cpp
```diff
@@ -273,7 +276,7 @@ TEST_F(RecoveryTests, RestartTest) {
   status = logging::LoggingUtil::RemoveDirectory(dir_name.c_str(), false);
   EXPECT_EQ(status, true);
 
-  catalog->DropDatabaseWithOid(DEFAULT_DB_ID);
+  catalog->DropDatabaseWithOid(DEFAULT_DB_ID, nullptr);
```
jarulraj: Can you explain how this is working with a `nullptr` transaction?

test/parser/parser_test.cpp
```diff
@@ -109,7 +109,7 @@ TEST_F(ParserTests, BasicTest) {
   queries.push_back("EXECUTE prep;");
 
   queries.push_back(
-      "COPY catalog_db.query_metric TO '/home/user/output.csv' DELIMITER ',';");
+      "COPY pg_catalog.query_metric TO '/home/user/output.csv' DELIMITER ',';");
```
jarulraj: Is this valid PostgreSQL syntax?

src/include/catalog/abstract_catalog.h
```diff
@@ -0,0 +1,87 @@
...
+//
+// Identification: src/include/catalog/abstract_catalog.h
+//
+// Copyright (c) 2015-17, Carnegie Mellon University Database Group
+//
+//===----------------------------------------------------------------------===//
+
+#pragma once
+
+#include "catalog/column.h"
```
jarulraj: Can you move some of these headers to the relevant implementation files? Try to avoid putting several headers in header files since that increases coupling and compilation time.

src/catalog/catalog.cpp
```diff
@@ -30,421 +30,714 @@ Catalog *Catalog::GetInstance(void) {
...
+      IndexType::BWTREE, IndexConstraintType::PRIMARY_KEY, true,
+      std::vector<oid_t>({0, 1}), pool_.get(), nullptr);
+  IndexCatalog::GetInstance()->InsertIndex(
+      COLUMN_CATALOG_SKEY0_OID, COLUMN_CATALOG_NAME "_skey0",
+      COLUMN_CATALOG_OID, IndexType::BWTREE, IndexConstraintType::UNIQUE, true,
+      std::vector<oid_t>({0, 2}), pool_.get(), nullptr);
+  IndexCatalog::GetInstance()->InsertIndex(
+      COLUMN_CATALOG_SKEY1_OID, COLUMN_CATALOG_NAME "_skey1",
+      COLUMN_CATALOG_OID, IndexType::BWTREE, IndexConstraintType::DEFAULT,
+      false, std::vector<oid_t>({0}), pool_.get(), nullptr);
```
hzxa21: Should the updates in IndexCatalog and the CreateIndex be placed in the same transaction? Would it be a problem if CreateIndex succeeds but InsertIndex fails because of transaction aborts?

src/catalog/catalog.cpp
```diff
@@ -30,421 +30,714 @@ Catalog *Catalog::GetInstance(void) {
...
+//===----------------------------------------------------------------------===//
+// CREATE FUNCTIONS
+//===----------------------------------------------------------------------===//
 
-  // Create table for index metrics
-  auto index_metrics_catalog = CreateMetricsCatalog(default_db_oid,
-      INDEX_METRIC_NAME);
-  default_db->AddTable(index_metrics_catalog.release(), true);
+ResultType Catalog::CreateDatabase(const std::string &database_name,
+                                   concurrency::Transaction *txn) {
```
hzxa21: Check txn != nullptr and enforce a transaction?

src/catalog/catalog.cpp
```diff
@@ -30,421 +30,714 @@ Catalog *Catalog::GetInstance(void) {
...
-    } else {
-      index_metadata = new index::IndexMetadata(index_name.c_str(),
-          GetNextOid(), table->GetOid(), database->GetOid(), index_type,
-          IndexConstraintType::UNIQUE, schema, key_schema, key_attrs, true);
-    }
+/*
+* only for test purposes
+*/
+ResultType Catalog::DropDatabaseWithName(const std::string &database_name,
+                                         concurrency::Transaction *txn) {
```
hzxa21: Same comments. Check txn != nullptr and enforce a transaction?

src/catalog/catalog.cpp
```diff
@@ -30,421 +30,714 @@ Catalog *Catalog::GetInstance(void) {
...
-    std::shared_ptr<index::Index> key_index(
-        index::IndexFactory::GetIndex(index_metadata));
-    table->AddIndex(key_index);
+  return DropDatabaseWithOid(database_oid, txn);
+}
 
-    LOG_TRACE("Successfully add index for table %s", table->GetName().c_str());
-    return ResultType::SUCCESS;
+ResultType Catalog::DropDatabaseWithOid(oid_t database_oid,
+                                        concurrency::Transaction *txn) {
```
hzxa21: Same. Check txn!=nullptr.

src/catalog/catalog.cpp
mengranwo: Yeach, we are currently working on that. Since binder needs to read record from catalog tables, dose that mean transaction should begin in binder instead of executor??

apavlo:
Are we close to being able to merge this?

src/catalog/abstract_catalog.cpp
```diff
@@ -44,6 +48,31 @@ AbstractCatalog::AbstractCatalog(std::string catalog_table_name,
...
 
+AbstractCatalog::AbstractCatalog(const std::string &catalog_table_name,
+                                 const std::string &catalog_table_ddl,
+                                 concurrency::Transaction *txn) {
+  // get catalog table schema
+  auto &peloton_parser = parser::PostgresParser::GetInstance();
+  // optimizer::SimpleOptimizer optimizer;
+  auto create_stmt = peloton_parser.BuildParseTree(catalog_table_ddl);
+  auto create_plan = std::dynamic_pointer_cast<planner::CreatePlan>(
+      optimizer::SimpleOptimizer().BuildPelotonPlanTree(create_stmt));
```
AngLi-Leon: Please don't hardcode SimpleOptimizer...

src/catalog/query_metrics_catalog.cpp
```diff
@@ -26,7 +26,22 @@ QueryMetricsCatalog *QueryMetricsCatalog::GetInstance(
 }
 
 QueryMetricsCatalog::QueryMetricsCatalog(concurrency::Transaction *txn)
-    : AbstractCatalog(QUERY_METRICS_CATALOG_NAME, InitializeSchema().release(),
+    : AbstractCatalog(QUERY_METRICS_CATALOG_NAME,
```
AngLi-Leon: Actually, parameter QUERY_METRICS_CATALOG_NAME is passed twice and can be eliminated...

PR #594, opened 2017-03-31
Add libpqxx-dev package

hanli32:
Add libpqxx-dev package, which is the official C++ client API for ​PostgreSQL.
Add simple test that uses libpqxx to connect to Peloton and execute simple queries.


apavlo:
Your new test is leaking memory. It might be the full system (which we haven't tested). I am going to disable the test for now to get it to pass. We can then look into it further.

PR #609, opened 2017-04-07
New Optimizer Checkpoint

luochenUmich:
In this PR, the new optimizer supports most queries on single table and we refactor the **Property** in the framework to be more general. In details, we implemented the following functionalities

- Group By
- Aggregation with/without Group By
- Distinct
- Completely eliminate the use of expression::ExpressionUtil::TransformExpression in new optimizer. Instead, we implement an expression map at each operator level to do column mapping. After having this mechanism, we now support arbitrary expression at any part of the query
- Set the correct expression name in optimizer. i.e. expr_name of "a + b" will be "a + b" not "expr1"

Here is a sample query we support in this PR
`SELECT b + c, SUM(a * c + 10) FROM test GROUP BY b,c ORDER BY b + c`

@malin1993ml Can you help review this?


malin1993ml:
Sounds good. I'll do it tomorrow after it passes tests.

src/optimizer/operator_to_plan_transformer.cpp
```diff
@@ -22,203 +22,266 @@
...
+  auto col_prop = requirements_->GetPropertyOfType(PropertyType::COLUMNS)
+      ->As<PropertyColumns>();
+
+  PL_ASSERT(col_prop != nullptr);
+
+  output_plan_ = move(GenerateAggregatePlan(col_prop, AggregateType::SORTED,
+                                            &op->columns, op->having));
+}
+
+// TODO: Remove duplicate codes
```
malin1993ml: Which codes are duplicate? Should we remove them now?

src/include/optimizer/operator_to_plan_transformer.h
```diff
@@ -13,6 +13,8 @@
 #pragma once
 
 #include "optimizer/operator_visitor.h"
+#include "properties.h"
+#include "planner/aggregate_plan.h"
```
hzxa21: The last one is unnecessary. Will remove all the unnecessary includes in our codes.

src/include/optimizer/operators.h
```diff
@@ -141,39 +151,38 @@ class LogicalLimit : public OperatorNode<LogicalLimit> {
...
 // Update
 //===--------------------------------------------------------------------===//
-class LogicalUpdate: public OperatorNode<LogicalUpdate> {
+class LogicalUpdate : public OperatorNode<LogicalUpdate> {
  public:
-  static Operator make(const parser::UpdateStatement* update_stmt);
+  static Operator make(const parser::UpdateStatement *update_stmt);
 
-  const parser::UpdateStatement* update_stmt;
+  const parser::UpdateStatement *update_stmt;
```
hzxa21: Because the planner accepts a pointer to the parse tree as an argument for the update_plan constructor. I will refactor the update_plan and the optimizer.

src/optimizer/operator_to_plan_transformer.cpp
hzxa21: Already remove the duplicate codes. Will delete this comment.

malin1993ml:
How's the status on this PR? There's still two review comments that have not been fixed (abstract_expression.h and types.h).

luochenUmich:
@malin1993ml I am changing it right now. Will fix this today

seojungmin:
A compilation error occurs at line 86 in optimizer/properties.cpp  when the TRACE option is set. The variable named _columns__ no longer exists.

luochenUmich:
@seojungmin You can just comment it out for now. We will fix it

PR #613, opened 2017-04-11
Basic infrastructures to support triggers

nblintao:
## Major Updates
- Update data structure for `CreateStatement` to support trigger
- Convert Postgres parsing tree for Peloton statement for trigger
- Update data structure for `CreatePlan` to support trigger
- Convert `CreateStatement`  to `CreatePlan`  for trigger
- Data structures for `Trigger` and `TriggerList` and related operations
- Update `DataTable` with `TriggerList` information
- Trigger operations for one kind of trigger (other kinds are going to be added in the future)


nblintao:
I have just created a new branch [trigger-dev](https://github.com/nblintao/peloton/tree/trigger-dev) for team members' development. So the further commits from now will not be synchronized to this PR automatically. I hope this could avoid populating the timeline >_<

src/commands/trigger.cpp
```diff
@@ -0,0 +1,104 @@
...
+#include "parser/pg_trigger.h"
+
+namespace peloton {
+namespace commands {
+Trigger::Trigger(const peloton::planner::CreatePlan& plan) {
+  trigger_name = plan.GetTriggerName();
+  trigger_funcname = plan.GetTriggerFuncName();
+  trigger_args = plan.GetTriggerArgs();
+  trigger_columns = plan.GetTriggerColumns();
+  trigger_when = plan.GetTriggerWhen();
```
CHUYilei: I am confused about what "trigger_when" is for: is it the trigger time? Can more comment be added for such naming?

src/commands/trigger.cpp
```diff
@@ -0,0 +1,104 @@
...
+ */
+void TriggerList::AddTrigger(Trigger trigger) {
+  triggers.push_back(trigger);
+  UpdateTypeSummary(trigger.GetTriggerType());
+}
+
+/*
+ * Update a type summary when a new type of trigger is added
+ */
+void TriggerList::UpdateTypeSummary(int16_t type) {
```
CHUYilei: Each line is in the same format like

`types_summary[XX] |= TRIGGER_TYPE_MATCHES(type, TRIGGER_TYPE_XX, TRIGGER_TYPE_AFTER, TRIGGER_TYPE_XX);`

Since it contains much repeated work, can it be implemented using map? Let Map store parameters in .h file. (just personal suggestion, maybe my way makes code more confusing)

src/commands/trigger.cpp
```diff
@@ -0,0 +1,104 @@
...
+/**
+ * execute trigger on each row before inserting tuple.
+ */
+storage::Tuple* TriggerList::ExecBRInsertTriggers(storage::Tuple *tuple) {
+  unsigned i;
+  LOG_INFO("enter into ExecBRInsertTriggers");
+
+  //check valid type
+  if (!types_summary[EnumTriggerType::BEFORE_INSERT_ROW]) {
+    return nullptr;
```
CHUYilei: I observer that when handling invalid type, nullptr is returned. I suggest printing some log info here. Also, is throwing an exception and letting it bubble up a better idea or not? Still it is a discussion not a 'suggestion' ^^

src/storage/data_table.cpp
```diff
@@ -1088,6 +1088,26 @@ column_map_type DataTable::GetDefaultLayout() const{
...
+void DataTable::AddTrigger(commands::Trigger new_trigger) {
+  trigger_list.AddTrigger(new_trigger);
+}
+
+int DataTable::GetTriggerNumber() {
+  return trigger_list.GetTriggerListSize();
+}
+
+commands::Trigger* DataTable::GetTriggerByIndex(int n) {
+  if (trigger_list.GetTriggerListSize() <= n)
```
CHUYilei: The corner case handling again needs to consider n<0, and maybe some LOG print or exception throwing would be helpful for debugging.

src/storage/data_table.cpp
```diff
@@ -1088,6 +1088,26 @@ column_map_type DataTable::GetDefaultLayout() const{
...
+    return nullptr;
+  return trigger_list.Get(n);
+}
+
+commands::TriggerList* DataTable::GetTriggerList() {
```
CHUYilei: Do you think singleton pattern for establishing one is better or not?

src/include/commands/trigger.h
```diff
@@ -0,0 +1,88 @@
...
+#include "common/logger.h"
+#include "storage/tuple.h"
+
+namespace peloton {
+namespace commands {
+
+class Trigger {
+ public:
+  Trigger(const planner::CreatePlan& plan);
+  Trigger(const Trigger& that) {
```
CHUYilei: Em... the naming of "that"

src/include/commands/trigger.h
```diff
@@ -0,0 +1,88 @@
...
+
+class TriggerList {
+ public:
+  inline bool HasTriggerType(EnumTriggerType type) const {
+    return types_summary[type];
+  }
+  inline int GetTriggerListSize() { return static_cast<int>(triggers.size()); }
+  void AddTrigger(Trigger trigger);
+  void UpdateTypeSummary(int16_t type);
+  Trigger* Get(int n) { return &triggers[n]; }  // get trigger by index
```
CHUYilei: It should handle corner cases when n<0 or n>=size like you did above.

src/include/parser/pg_trigger.h
```diff
@@ -0,0 +1,81 @@
+#ifndef PG_TRIGGER_H
+#define PG_TRIGGER_H
+
+/* ----------------
+ *		compiler constants for pg_trigger
+ * ----------------
+ */
+#define Natts_pg_trigger 15
```
CHUYilei:
```
#define Natts_pg_trigger 15
#define Anum_pg_trigger_tgqual 15
```
Are these two related? Otherwise why do you choose 15 for Natts_pg_trigger? More documentation needed.

src/include/planner/create_plan.h
```diff
@@ -33,6 +34,11 @@ class CreatePlan : public AbstractPlan {
   CreatePlan &operator=(const CreatePlan &) = delete;
   CreatePlan(CreatePlan &&) = delete;
   CreatePlan &operator=(CreatePlan &&) = delete;
+  ~CreatePlan() {
+    if (trigger_when) {
+      delete trigger_when;
```
CHUYilei: It is observed that in `src/include/parser/create_statement.h` you also have 
```
if (trigger_when) {
      delete trigger_when;
}
```
Then would it be possible that `~CreatePlan` has double free problem?

test/commands/trigger_test.cpp
```diff
@@ -0,0 +1,87 @@
...
+
+  // Create plans
+  const planner::CreatePlan plan1(create_trigger_stmt1);
+
+  commands::Trigger trigger1(plan1);
+  EXPECT_EQ("check_update", trigger1.GetTriggerName());
+  int16_t trigger_type1 = trigger1.GetTriggerType();
+  EXPECT_TRUE(TRIGGER_FOR_ROW(trigger_type1));
+  EXPECT_TRUE(TRIGGER_FOR_BEFORE(trigger_type1));
+  EXPECT_TRUE(TRIGGER_FOR_UPDATE(trigger_type1));
```
CHUYilei: Why the first three expectations are all true? Thanks!

CHUYilei:
Hi, I write the code review on behalf of my whole group. Here is my general comment:
You did an awesome job for trigger! The code is clean and easy to understand. You have careful design of structure and tests. The documentations are helpful and TODOs show your goals clearly. 
My comment on code is mainly on error handling, debug info print, confusion on certain test cases and some personal point of views on certain design that I would like to discuss with you. ^^

src/commands/trigger.cpp
```diff
@@ -0,0 +1,104 @@
...
+      type, TRIGGER_TYPE_ROW, TRIGGER_TYPE_AFTER, TRIGGER_TYPE_DELETE);
+  types_summary[AFTER_DELETE_STATEMENT] |= TRIGGER_TYPE_MATCHES(
+      type, TRIGGER_TYPE_STATEMENT, TRIGGER_TYPE_AFTER, TRIGGER_TYPE_DELETE);
+}
+
+/**
+ * execute trigger on each row before inserting tuple.
+ */
+storage::Tuple* TriggerList::ExecBRInsertTriggers(storage::Tuple *tuple) {
+  unsigned i;
```
hanli32: It is better to use a descriptive name here.

src/executor/create_executor.cpp
```diff
@@ -79,10 +81,40 @@ bool CreateExecutor::DExecute() {
...
+    target_table->AddTrigger(newTrigger);
+
+    // if (current_txn->GetResult() == ResultType::SUCCESS) {
+    //   LOG_TRACE("Creating trigger succeeded!");
+    // } else if (current_txn->GetResult() == ResultType::FAILURE) {
+    //   LOG_TRACE("Creating trigger failed!");
+    // } else {
+    //   LOG_TRACE("Result is: %s", ResultTypeToString(
+    //     current_txn->GetResult()).c_str());
+    // }
```
hanli32: commented out code

src/include/parser/pg_trigger.h
```diff
@@ -0,0 +1,81 @@
...
+#define Anum_pg_trigger_tgisinternal 6
+#define Anum_pg_trigger_tgconstrrelid 7
+#define Anum_pg_trigger_tgconstrindid 8
+#define Anum_pg_trigger_tgconstraint 9
+#define Anum_pg_trigger_tgdeferrable 10
+#define Anum_pg_trigger_tginitdeferred 11
+#define Anum_pg_trigger_tgnargs 12
+#define Anum_pg_trigger_tgattr 13
+#define Anum_pg_trigger_tgargs 14
+#define Anum_pg_trigger_tgqual 15
```
hanli32: Is it better to use enum here to make code more clean?

src/include/parser/parsenodes.h
```diff
@@ -228,172 +222,182 @@ typedef struct SelectStmt {
...
+  RangeVar *relation; /* relation trigger is on */
+  List *funcname;     /* qual. name of function to call */
+  List *args;         /* list of (T_String) Values or NIL */
+  bool row;           /* ROW/STATEMENT */
+  /* timing uses the TRIGGER_TYPE bits defined in catalog/pg_trigger.h */
+  int16_t timing; /* BEFORE, AFTER, or INSTEAD */
+  /* events uses the TRIGGER_TYPE bits defined in catalog/pg_trigger.h */
+  int16_t events;    /* "OR" of INSERT/UPDATE/DELETE/TRUNCATE */
+  List *columns;     /* column names, or NIL for all columns */
+  Node *whenClause;  /* qual expression, or NULL if none */
```
hanli32: It is better to follow Peloton's lower_case_with_underscores variable names.

src/commands/trigger.cpp
huangminchn: Thanks for pointing it out. We will add more comments to explain the trigger data structures. "trigger_when" specifies when the trigger is fired, before/after insert/update/delete for each row/statement.

src/commands/trigger.cpp
huangminchn: Thanks for suggestion. We want to keep the trigger as light weight as possible. We will try to find a way to simplify the codes.

src/commands/trigger.cpp
huangminchn: Thanks. Sure we will prints some logs here. It's acceptable that some tables don't have a specific type of triggers or don't have any triggers, so we won't consider throwing exceptions for now.

src/storage/data_table.cpp
huangminchn: Thanks. Indeed, this function is not defensive enougth. We will fix it.

test/commands/trigger_test.cpp
huangminchn: Because the trigger we created in the test would fire "before any updates for each row in the target table"

src/include/parser/parsenodes.h
huangminchn: Thanks for suggestion. We will check our code style.

src/storage/data_table.cpp
nblintao: Each `DateTable` will have its own `TriggerList`. I can't see how does singleton pattern relate to the problem here. Could you explain more? Thanks~

src/include/commands/trigger.h
nblintao: Thanks. Yes, this should be more defensive. Actually, I don't think it reasonable to expose a `Get` interface here. This function might be created by a team member for debugging. I'll try to remove this in the future.

src/include/parser/pg_trigger.h
nblintao: This file is copied from Postgres parser. We need to keep the macros the same to interpret the parsing tree generated by the Postgres parser.

src/include/planner/create_plan.h
nblintao: When CreateStatement is converted to CreatePlan, this field is copy constructed. So the two `trigger_when` actually point to different positions. But thanks for your sharp mind. Memory is always tricky:)

src/include/parser/pg_trigger.h
nblintao: This file is copied from Postgres parser. We need to keep the macros the same to interpret the parsing tree generated by the Postgres parser. Actually, you don't have to review the newly imported head files under parser (like this one and `parsenodes.h`).

src/include/parser/parsenodes.h
nblintao: Yeah, keeping the naming style consistent is really important and thanks for your suggestions. But I believe we will not change this file because it is copied from Postgres' parser without any change. Updating this part may confuse other developers in the future.

apavlo:
Development moved: https://github.com/cmu-db/peloton/pull/613#issuecomment-293469473

PR #614, opened 2017-04-11
p3 - Extending Query Compilation

phisiart:
Our code is based on @pmenon's codegen branch. @pmenon's codegen has one test case failure, and so does ours.

You might want to look at the comparison between @pmenon's codebase and our code:
https://github.com/pmenon/peloton-1/compare/codegen...phisiart:p3

Our group's tasks are:
- Parameterization @avahuang1 
  See this for high-level explanation:
  https://github.com/phisiart/peloton-p3/wiki/LLVM-Query-Compilation-Development
  Parameterization should be working already.

- Caching @bechob 
  See this for high-level explanation:
  https://github.com/phisiart/peloton-p3/wiki/LLVM-Query-Compilation-Development
  Plan comparison has been implemented.

- DELETE @MatrixCW 
  A `DELETE` query performs a table scan, and then calls some functions of `DataTable` and `TransactionManager` to perform the deletion.
  We just leverage the current `TableScanTranslator` to perform the scan, and mimic the `DeleteExecutor` to perform deletion.

- INSERT @phisiart 
  There are 3 flavors of INSERT. Currently the first has been implemented.
  - Insert an array of "raw" tuples (`storage::Tuple`). This is the foundation of all flavors. We have implemented it and turned it on. Now INSERT of this flavor will no longer use the interpreted executor. It seems that tests are passing. There's a memory leak in the "insert_translator_test", which is not related to the implementation of INSERT. It's just a leak of the test case - we will look at it and figure out the reason.
  - Insert an array of `ProjectInfo`. We need to turn `ProjectInfo`s into "raw" tuples.
  - Insert tuples from a scan of another table. We need to "materialize" tuples.

- UPDATE
  Not Implemented.


src/include/codegen/query_cache.h
```diff
@@ -1,16 +1,46 @@
...
+#pragma once
+namespace peloton {
+namespace codegen {
+class QueryCache {
+public:
+  static QueryCache& Instance() {
+    static  QueryCache instance;
+    return instance;
+  }
+  bool FindPlan(const planner::AbstractPlan & key) {
```
seojungmin: Can we combine FindPlan() and GetQuery() instead of having them separately, unless there is any special reason?

src/include/codegen/query_cache.h
```diff
@@ -1,16 +1,46 @@
...
+    cache.insert(key, val);
+  }
+
+private:
+  struct ComparePlan {
+    bool operator()(const  planner::AbstractPlan & a, const planner::AbstractPlan & b) const {
+      return (codegen::PlanComparator::Compare(A, B) < 0);
+    }
+  };
+  std::map<planner::AbstractPlan, Query, ComparePlan> cache;
```
seojungmin: A limit on the cache size would be nice to have since we would not have an infinite size of cache.  Having said this, I wonder what would be the best or a reasonable cache policy.

phisiart: Makes sense! We should add a cache size and eviction policy later.

src/codegen/delete/delete_translator.cpp
```diff
@@ -0,0 +1,191 @@
...
+    CodeGen &codegen) {
+  const std::string &fn_name = GetFunctionName();
+
+  // Has the function already been registered?
+  llvm::Function *llvm_fn = codegen.LookupFunction(fn_name);
+  if (llvm_fn != nullptr) {
+    return llvm_fn;
+  }
+  llvm::FunctionType *int32_type =
+      llvm::FunctionType::get(codegen.Int32Type(), false);
```
tq5124: `llvm::FunctionType` is the llvm::Type that describe a function. I think `codegen.Int32Type()` is enough to describe a llvm::Type.

src/codegen/delete/delete_translator.cpp
```diff
@@ -0,0 +1,191 @@
...
+  llvm::FunctionType *pointer_type =
```
tq5124: This pointer type is a char pointer. Because you need a `concurrency::Transaction *txn` and `storage::DataTable *table`, you probably need `DataTableProxy ::GetType(codegen)->GetPointerTo()`.

src/codegen/delete/delete_translator.cpp
```diff
@@ -0,0 +1,191 @@
+//===----------------------------------------------------------------------===//
+//
+//                         Peloton
+//
+// transaction_manager.h
```
tq5124: Please fix headers here and all other new files.

src/codegen/arithmetic_translator.cpp
```diff
@@ -0,0 +1,88 @@
...
+#include "expression/operator_expression.h"
+
+namespace peloton {
+namespace codegen {
+
+// Constructor
+ArithmeticTranslator::ArithmeticTranslator(
+    const expression::OperatorExpression &arithmetic,
+    CompilationContext &context)
+    : ExpressionTranslator(arithmetic, context), ctx_(context) {
```
homelesscmu: You do not need to store the compilation context because you can use GetCompilationContext() to retrieve it later

src/codegen/constant_translator.cpp
```diff
@@ -0,0 +1,108 @@
...
+#include "codegen/value_proxy.h"
+#include "codegen/parameter.h"
+
+namespace peloton {
+namespace codegen {
+
+// Constructor
+ConstantTranslator::ConstantTranslator(
+    const expression::ConstantValueExpression &exp, CompilationContext &ctx)
+    : ExpressionTranslator(exp, ctx), ctx_(ctx) {
```
homelesscmu: No need to store ctx

src/codegen/negation_translator.cpp
```diff
@@ -0,0 +1,70 @@
...
+#include "codegen/parameter.h"
+
+namespace peloton {
+namespace codegen {
+
+// Constructor
+NegationTranslator::NegationTranslator(
+    const expression::OperatorUnaryMinusExpression &unary_minus_expression,
+    CompilationContext &ctx)
+    : ExpressionTranslator(unary_minus_expression, ctx), ctx_(ctx) {
```
homelesscmu: No need to store ctx

src/codegen/parameter_translator.cpp
```diff
@@ -0,0 +1,75 @@
...
+namespace expression {
+  class ParameterValueExpression;
+}  // namespace planner
+
+namespace codegen {
+
+// Constructor
+ParameterTranslator::ParameterTranslator(
+    const expression::AbstractExpression &exp, CompilationContext &ctx)
+    : ExpressionTranslator(exp, ctx), ctx_(ctx) {
```
homelesscmu: No need to store ctx

src/include/codegen/transaction_manager/transaction_manager.h
```diff
@@ -0,0 +1,80 @@
...
+// transaction_manager.h
+//
+// Identification: src/include/codegen/transaction_manager/transaction_manager.h
+//
+// Copyright (c) 2015-16, Carnegie Mellon University Database Group
+//
+//===----------------------------------------------------------------------===//
+
+/**
+ * This file contains the C API of the transaction manager.
```
homelesscmu: What is the purpose of this class ..? More comments will be nice ...

src/codegen/arithmetic_translator.cpp
avahuang1: I'm sorry but that API is for operator translator; there is no such API for expression translator; better check before writing reviews.

src/codegen/constant_translator.cpp
avahuang1: You shouldn't be writing same comment as well.

src/codegen/negation_translator.cpp
avahuang1: See above.

src/codegen/parameter_translator.cpp
avahuang1: See above.

src/codegen/arithmetic_translator.cpp
homelesscmu: Oh ok :) You know your code better than I do, and I was trying to read, understand and help with your problem.

src/codegen/constant_translator.cpp
homelesscmu: Again, I was trying to help. Meticulously writing these same comments may prevent you from missing a case. (though it turns out to be perfectly fine now)

src/include/codegen/plan_comparator.h
```diff
@@ -0,0 +1,73 @@
...
+namespace codegen {
+
+//===----------------------------------------------------------------------===//
+// ExpressionComparator provides compare function for all expressions that are
+// currently supported by codegen.
+//===----------------------------------------------------------------------===//
+class ExpressionComparator {
+
+public:
+  ExpressionComparator() {};
```
lihao98722: All functions are **static**, which means no instance is needed. So I wonder if the constructor needs to be disabled (ExpressionComparator() = delete;) in case someone uses it unintentionally.

src/include/codegen/transaction_manager/transaction_manager.h
phisiart: Oh, these are created because the C++ calling convention is more complicated. We will try to call into C++ though, so these might well get deleted later.

src/codegen/arithmetic_translator.cpp
seojungmin: @avahuang1 Can we modify ExpressionTranslator to store the compilation context so that we do not have to add this to all the children of ExpressionTranslator?

avahuang1: Sure. I left it there coz I'm going to abandon all expression translators, coz current schema cannot support parameterization. All these expr translators will be abandon and redesigned soon.

PR #618, opened 2017-04-12
Database Compression

rohit-cmu:
## Files Added

### Other Existing Files Modified

## Strategy for Compressing

* Once a tile gets full (all slots occupied by tuples), the CompressTile is created.
* Currently we can compress SMALLINT and INTEGER and BIGINT. TINYINT is already 1 byte so we do not compress it.
* We are yet to add support for decimal values.
* This median is also stored as the metadata to later retrieve the original value

## Testing

* This test inserts 25 tuples
* Each tuple is of the form (i, i*100) where i belongs to (0,25)
* Since each tile group contains 1 tile and 10 tuples per tile, there are 3 tile groups formed.
* The first 2 tile groups are full and get compressed.
* The third tile group has 5 slots vacant and is still not full and is uncompressed.

### Compression Size Test:
* The integers get compressed to TINYINT making the tuple sizes now (1+ 1+ 8+ 8) = 18 bytes


Vagrantfile
```diff
@@ -0,0 +1,71 @@
+# -*- mode: ruby -*-
```
wangxiaoyou11: It seems Vagrantfile shouldn't be pushed to the repo?

wangxiaoyou11:
We recommend not formatting other files that you didn't modify.

Vagrantfile
apavlo: Correct. Be sure that it is excised from the commit history instead of just deleting the file.

https://help.github.com/articles/removing-sensitive-data-from-a-repository/

src/include/storage/compressed_tile.h
```diff
@@ -0,0 +1,156 @@
...
+class TupleIterator;
+class Tile;
+
+/**
+ * Represents a Tile.
+ *
+ * Tiles are only instantiated via TileFactory.
+ *
+ * NOTE: MVCC is implemented on the shared TileGroupHeader.
+ */
```
wangxiaoyou11: The comments are not correct.  Are they directly copied from `storage/tile.h` ?

src/include/storage/compressed_tile.h
```diff
@@ -0,0 +1,156 @@
...
+
+  void SetValueFast(const type::Value &value, const oid_t tuple_offset,
+                    const size_t column_offset, const bool is_inlined,
+                    const size_t column_length);
+
+  //===--------------------------------------------------------------------===//
+  // Utility Functions
+  //===--------------------------------------------------------------------===//
+
+  bool IsCompressed() { return is_compressed; }
```
wangxiaoyou11: It can be declared as an inline function.

src/storage/compressed_tile.cpp
```diff
@@ -0,0 +1,301 @@
...
+                               const catalog::Schema &tuple_schema,
+                               TileGroup *tile_group, int tuple_count)
+    : Tile(backend_type, tile_header, tuple_schema, tile_group, tuple_count) {
+  is_compressed = false;  // tile is currently uncompressed.
+  compressed_columns_count = 0;
+}
+
+CompressedTile::~CompressedTile() {}
+
+void CompressedTile::CompressTile(Tile *tile) {
```
wangxiaoyou11: There are no comments in this file. We recommend adding comments for each function and those long code blocks.

src/include/storage/tile_group.h
```diff
@@ -52,6 +52,7 @@ class TileGroupHeader;
 class AbstractTable;
 class TileGroupIterator;
 class RollbackSegment;
+// class CompressedTile;
```
vanryan: It should be deleted.

src/include/storage/compressed_tile.h
```diff
@@ -0,0 +1,156 @@
...
+  //===--------------------------------------------------------------------===//
+  // Data members
+  //===--------------------------------------------------------------------===//
+
+  bool is_compressed;
+  oid_t compressed_columns_count;
+  int tuple_count;
+
+  std::map<oid_t, std::pair<type::Type::TypeId, type::Value>>
+      compressed_column_map;
```
wangxiaoyou11: Could you add comments describing what the key and value are in the `compressed_column_map` ?

src/storage/compressed_tile.cpp
```diff
@@ -0,0 +1,301 @@
...
+  auto allocated_tuple_count = tile->GetAllocatedTupleCount();
+  std::vector<type::Value> new_column_values(allocated_tuple_count);
+  std::vector<std::vector<type::Value>> new_columns(column_count);
+  std::vector<catalog::Column> columns;
+  auto tile_schema = tile->GetSchema();
+
+  for (oid_t i = 0; i < column_count; i++) {
+    LOG_TRACE("Compressing Column %s",
+              tile_schema->GetColumn(i).GetName().c_str());
+    switch (tile_schema->GetType(i)) {
```
vanryan: There are other integer types in peloton, such as `TINYINT` or `TIMESTAMP`, which can be found in `src/include/type/types.h` or `value_factory.h`. It might be better to include all the cases.

src/storage/compressed_tile.cpp
```diff
@@ -0,0 +1,301 @@
...
+    LOG_TRACE("Peloton does not support insert into compressed tiles.");
+    PL_ASSERT(false);
+  } else {
+    Tile::InsertTuple(tuple_offset, tuple);
+  }
+}
+
+type::Value CompressedTile::GetValue(const oid_t tuple_offset,
+                                     const oid_t column_id) {
+  type::Value deserizedValue = Tile::GetValue(tuple_offset, column_id);
```
vanryan: Possible typo. `deserized` -> `deserialized`.

src/storage/tile_group.cpp
```diff
@@ -314,6 +317,27 @@ oid_t TileGroup::InsertTupleFromCheckpoint(oid_t tuple_slot_id,
   return tuple_slot_id;
 }
 
+void TileGroup::CompressTiles() {
+  oid_t num_tiles = tiles.size();
+  for (oid_t i = 0; i < num_tiles; i++) {
+    LOG_TRACE("Compressing Tile %d in Tile Group", i);
+    Tile *old_tile = &*tiles[i];
```
wangxiaoyou11: The correct way to get the pointer from `shared_ptr` is to use `shared_ptr::get` function.
This can be rewritten as `Tile *old_tile = tiles[i].get()`.

src/include/storage/compressed_tile.h
rohit-cmu: Yes, Will fix the comments of the files and possibly add more comments detailing function use, parameters and uses.

src/storage/compressed_tile.cpp
rohit-cmu: Yes, its a typo

src/storage/compressed_tile.cpp
rohit-cmu: Yes, We hop to cover the entire spectrum of value types by the end of the semester.

test.txt
```diff
@@ -0,0 +1 @@
+testing for slack integration of github on #code
```
wangxiaoyou11: No use?

src/storage/compressed_tile.cpp
```diff
@@ -0,0 +1,301 @@
...
+    return deserizedValue;
+  }
+
+  type::Type::TypeId compressed_type = GetCompressedType(column_id);
+
+  PL_ASSERT(compressed_type != type::Type::INVALID);
+
+  type::Value original = GetUncompressedValue(column_id, deserizedValue);
+
+  return original;
```
vanryan: Usually compilers do RVO but it might be a more universal practice to use `std::move` in return or just return `GetUncompressedValue(column_id, deserizedValue);` to avoid copy.

test/storage/compression_test.cpp
```diff
@@ -0,0 +1,74 @@
...
+#include "concurrency/transaction_manager_factory.h"
+
+namespace peloton {
+namespace test {
+
+//===--------------------------------------------------------------------===//
+// Compression Tests
+//===--------------------------------------------------------------------===//
+
+std::unique_ptr<storage::DataTable> data_table_test_table;
```
wangxiaoyou11: This `data_table_test_table` doesn't need to be declared as a global variable here. And the variable name can be simplified as `data_table`.

src/storage/compressed_tile.cpp
```diff
@@ -0,0 +1,301 @@
...
+    return original;
+  }
+
+  return deserizedValue;
+}
+
+void CompressedTile::SetValue(const type::Value &value,
+                              const oid_t tuple_offset, const oid_t column_id) {
+  if (IsCompressed()) {
+    if (GetCompressedType(column_id) != type::Type::INVALID) {
```
vanryan: The two `if` should be combined together.

src/storage/data_table.cpp
```diff
@@ -172,14 +171,24 @@ ItemPointer DataTable::GetEmptyTupleSlot(const storage::Tuple *tuple) {
   // check if there are recycled tuple slots
   auto &gc_manager = gc::GCManagerFactory::GetInstance();
   auto free_item_pointer = gc_manager.ReturnFreeSlot(this->table_oid);
+  bool valid_slot = true;
+  ;
```
vanryan: This semicolon must feel lonely.

test/storage/compression_test.cpp
```diff
@@ -0,0 +1,74 @@
...
+
+TEST_F(CompressionTests, SizeTest) {
+  const int tuple_count = 100;
+  UNUSED_ATTRIBUTE oid_t tuple_size = 24;
+  UNUSED_ATTRIBUTE oid_t tuples_per_tile_group = 10;
+  auto &txn_manager = concurrency::TransactionManagerFactory::GetInstance();
+  auto txn = txn_manager.BeginTransaction();
+
+  data_table_test_table.reset(
+      TestingExecutorUtil::CreateTable(tuple_count, false));
```
wangxiaoyou11: Check the definition of `TestingExecutorUtil::CreateTable` here:
https://github.com/cmu-db/peloton/blob/master/test/include/executor/testing_executor_util.h#L70
The first argument is actually `tuples_per_tile_group`.

test/storage/compression_test.cpp
```diff
@@ -0,0 +1,74 @@
...
+    }
+  }
+  std::cout << "Tuples Per Tile Group: " << tuples_per_tile_group << "\n";
+  std::cout << "Number of Tile Groups: " << num_tile_groups << "\n";
+  std::cout << "Tuples size in bytes: " << tuple_size << "\n";
+  std::cout << "Uncompressed_Size in bytes: " << uncompressed_size << "\n";
+  std::cout << "Compressed_Size in bytes: " << compressed_size << "\n";
+
+  auto data_table_pointer = data_table_test_table.release();
+  delete data_table_pointer;
```
wangxiaoyou11: You don't need to explicitly delete the object. Just put that `unique_ptr` inside the test function. The object's memory will be automatically released after this function is finished.
Check the usage of `unique_ptr`: http://en.cppreference.com/w/cpp/memory/unique_ptr

test/storage/compression_test.cpp
```diff
@@ -0,0 +1,74 @@
...
+  TestingExecutorUtil::PopulateTable(data_table_test_table.get(), tuple_count,
+                                     false, false, true, txn);
+
+  txn_manager.CommitTransaction(txn);
+  auto num_tile_groups = data_table_test_table->GetTileGroupCount();
+  auto uncompressed_size = tuple_count * tuple_size;
+  auto compressed_size = 0;
+  for (oid_t i = 1; i < num_tile_groups; i++) {
+    storage::TileGroup *tile_group =
+        &*(data_table_test_table->GetTileGroupById(i));
```
wangxiaoyou11: Same. Use `shared_ptr::get`.

src/storage/compressed_tile.cpp
```diff
@@ -0,0 +1,301 @@
...
+
+  if (!IsCompressed()) {
+    return deserizedValue;
+  }
+
+  if (GetCompressedType(column_id) == type::Type::INVALID) {
+    return deserizedValue;
+  }
+
+  type::Type::TypeId compressed_type = GetCompressedType(column_id);
```
vanryan: It seems that a use in `PL_ASSERT` is not considered by the compiler in the given vagrant box, because the compilation using just `make` (rather than the tests) cannot run through.

> error: unused variable ‘compressed_type’

src/storage/data_table.cpp
```diff
@@ -172,14 +171,24 @@ ItemPointer DataTable::GetEmptyTupleSlot(const storage::Tuple *tuple) {
...
   if (free_item_pointer.IsNull() == false) {
     // when inserting a tuple
-    if (tuple != nullptr) {
+    if ((tuple != nullptr)) {
       auto tile_group =
           catalog::Manager::GetInstance().GetTileGroup(free_item_pointer.block);
-      tile_group->CopyTuple(tuple, free_item_pointer.offset);
+
+      if (tile_group->GetCompressionStatus() == true) {
+        valid_slot = false;
```
vanryan: You can just return here and avoid the use of `valid_slot`.

src/storage/compressed_tile.cpp
```diff
@@ -0,0 +1,301 @@
...
+
+std::vector<type::Value> CompressedTile::CompressColumn(
+    Tile *tile, oid_t column_id,
+    type::Type::TypeId compression_type = type::Type::TINYINT) {
+  oid_t num_tuples = tile->GetAllocatedTupleCount();
+  auto tile_schema = tile->GetSchema();
+  bool is_inlined = tile_schema->IsInlined(column_id);
+  size_t column_offset = tile_schema->GetOffset(column_id);
+  auto column_info = tile_schema->GetColumn(column_id);
+  auto column_type = column_info.GetType();
```
wangxiaoyou11: You can directly call `tile_schema->GetType(column_id)`.

src/storage/compressed_tile.cpp
```diff
@@ -0,0 +1,301 @@
...
+              compression_type);
+      for (oid_t k = 0; k < num_tuples; k++) {
+        modified_values[k] =
+            actual_values[k].Subtract(median).CastAs(compression_type);
+      }
+
+      LOG_TRACE("Base value selected = %s", median.ToString().c_str());
+      break;
+    } catch (Exception &e) {
+      if (column_type == (compression_type + 1)) {
```
wangxiaoyou11: What does `compression_type + 1` mean?

src/storage/compressed_tile.cpp
```diff
@@ -0,0 +1,301 @@
...
+  std::sort(column_values.begin(), column_values.end(), CompareLessThanBool);
+
+  type::Value median = column_values[num_tuples / 2];
+  std::vector<type::Value> modified_values(num_tuples);
+
+  while (true) {
+    try {
+      type::Value min_diff =
+          column_values[0].Subtract(median).CastAs(compression_type);
+      type::Value max_diff =
```
wangxiaoyou11: How are `min_diff` and `max_diff` are used?

vanryan: In addition, why is this code block in a try/catch clause pair?

src/storage/compressed_tile.cpp
```diff
@@ -0,0 +1,301 @@
...
+
+  return deserizedValue;
+}
+
+void CompressedTile::SetValue(const type::Value &value,
+                              const oid_t tuple_offset, const oid_t column_id) {
+  if (IsCompressed()) {
+    if (GetCompressedType(column_id) != type::Type::INVALID) {
+      LOG_TRACE("Peloton does not support SetValue on a Compressed Class");
+      PL_ASSERT(false);
```
vanryan: What does `PL_ASSERT(false)` mean here? Doesn't this incur an exit? (As in the former travis CI build log for job 2231.1)

vanryan: Also running `join_test` locally I saw the `JoinTests.ComplicatedTest` failing.

src/storage/compressed_tile.cpp
poojanilangekar: The code seems to be compiling on my local vagrant box.

src/storage/compressed_tile.cpp
poojanilangekar: @wangxiaoyou11 Kindly refer to the strategy for compression in the comments of the PR. 
@vanryan The code is present in a try/catch pair because value_factory.cpp would throw an exception if the CastAs()  function fails, if a particular type fails, we re-try compression using a bigger data type.

src/storage/compressed_tile.cpp
poojanilangekar: compression_type + 1 is the next bigger type. We only re-try compression when the next bigger type will not be equal to the current type. (This is because we compress only if the compression will lead to the use of a smaller data type.)

src/storage/compressed_tile.cpp
poojanilangekar: Once a tile has been compressed, its values can't be modified since the new value might be out of the compression range. In this case the entire tile will have to be uncompressed at runtime (leading to poor performance). The PL_ASSERT(false) will have to stay. We will be changing the call to the compressTiles() function in the tile_group.cpp file to handle this. The current code was only used to check the correctness of the compression. (Inserts and selects)

src/storage/data_table.cpp
poojanilangekar: No, returning at this point would return an invalid pointer to the calling executor function. Once valid_slot is set to false, the data_table would create a new empty slot (by allocating a new tile_group)  and return it to the executor.

src/storage/compressed_tile.cpp
poojanilangekar: Will fix it.

test/storage/compression_test.cpp
poojanilangekar: Will change that.

src/storage/compressed_tile.cpp
poojanilangekar: Will change it.

src/storage/compressed_tile.cpp
poojanilangekar: Will change it.

src/storage/compressed_tile.cpp
wangxiaoyou11: @poojanilangekar I mean where do you use `min_diff` and `max_diff` in your code? It seems that you just calculate those two values but not use it?

rohit-cmu: We need to calculate these values to ascertain whether we can compress this column or not and if so, to what data type we can compress this. This is the method for compression as suggested even in the comment of the PR. This knowledge also gives us insight about the skew of the data. Later we might add support for mostly encoding for certain outliers in the data set. If there is any better way to ascertain this information without changing the underlying algorithm, kindly let us know!

vanryan: It is not about the strategy or the algorithm. Yes, `min_diff` and `max_diff` are calculated here, however not used anywhere in the program afterwards.

test/storage/compression_test.cpp
rohit-cmu: fixed

test/storage/compression_test.cpp
rohit-cmu: fixed

test/storage/compression_test.cpp
rohit-cmu: @poojanilangekar has fixed this

PR #619, opened 2017-04-12
Stats and Cost Model

AllisonWang:
This PR adds statistics collection, storage, selectivity calculation and single table cost model. It uses two ways to collect statistics: probabilistic data structure and sampling.  All modified source files are located under `src/optimizer/stats`, `src/includes/optimizer/stats` and corresponding tests are in `test/optimizer`. 

### Probabilistic data structure

- Count-min sketch & Top_k - calculate most common values and their frequencies
- HyperLogLog (included as a third library) - estimate number of distinct values
- Online histogram building 

Those stats collectors are created by `ColumnStatsCollector` which represents all relevant statistics for the column. `TableStatsCollector` contains one or more `ColumnStatsCollector` and holds statistics for a table. 

### Sampling

The class `TupleSampler` can take in a `DataTable` pointer and perform random sampling to get the expected amount of tuple samples.

### Store table in catalog 

The class `StatsStorage` provides the following functionalities:
1. Provide an interface for analyzing all tables' statistics: `CollectStatsForAllTables`
2. Provide interfaces for adding stats into Catalog.
3. Provide interfaces for adding samples tables into `samples_db`.
4. Provide interfaces for retrieving stats and samples from Catalog and `samples_db`.

The stats table's schema is designed as:
* `database_id`: INTEGER, NOT_NULL
* `table_id`: INTEGER, NOT_NULL
* `column_id`: INTEGER, NOT_NULL
* `num_row`: INTEGER, NOT_NULL
* `cardinality`: DECIMAL, NOT_NULL
* `frac_null`: DECIMAL, NOT_NULL
* `most_common_vals`: VARCHAR
* `most_common_freqs`: DECIMAL
* `histogram_bounds`: VARCHAR
(Since peloton doesn't support ARRAY type now, we temporarily convert the double array to a string and store it as VARCHAR)

Corresponding unit tests are added for `StatsStorage` and `TupleSampler` functions.

Some interfaces in `StatsStorage` are not implemented for now. They are: `GetColumnStatsByID`, `GetTupleSamples`, `GetColumnSamples`.

### Analyze
We implemented `ANALYZE` command, which is the only way of invoking statistics collection without specifically calling functions in `StatsStorage`. It can be changed to auto-execute when the number of tuples changed in a table is over a certain percentage.

### Selectivity Estimation
We implemented all selectivity estimation for compare operators except LIKE, IN and IS DISTINCT FROM. Right now all selectivity estimations are using collected statistics. In the future, we should also try using sampling for selectivity estimation.

### Cost Model (Single Table)
Cost model for single table operations is implemented in `cost.h` and `cost.cpp`.  It needs to be tested with new optimizer and adjust accordingly.


test/optimizer/column_stats_test.cpp
```diff
@@ -0,0 +1,118 @@
...
+  EXPECT_EQ(colstats.GetCardinality(), 3);
+  for (int i = 0; i < 100; i++) {
+    type::Value v = type::ValueFactory::GetDecimalValue(5.1525 + i);
+    colstats.AddValue(v);
+    colstats.AddValue(v);
+    colstats.AddValue(v);
+  }
+  std::vector<std::pair<type::Value, double>> valfreq =
+      colstats.GetCommonValueAndFrequency();
+  // EXPECT_EQ(valfreq.size(), 10);
```
rohit-cmu: Should this be commented out or does it change the correctness of the test ?

rohit-cmu: Also in general, if there could be more comments about what each unit test is testing, it would help

test/optimizer/column_stats_test.cpp
```diff
@@ -0,0 +1,118 @@
...
+  EXPECT_EQ(colstats.GetHistogramBound().size(),
+            0);  // No histogram for categorical data
+  std::vector<ValueFrequencyPair> valfreq =
+      colstats.GetCommonValueAndFrequency();
+  // EXPECT_EQ(valfreq.size(), 2);
+  // PrintCommonValueAndFrequency(colstats.GetCommonValueAndFrequency());
+}
+
+// All stats collectors should work with trivial case
+TEST_F(ColumnStatsTests, TrivialValueTest) {
```
rohit-cmu: I think this unit test case can be removed later on. For now, this may be fine from a code development and testing perspective.

test/optimizer/column_stats_test.cpp
```diff
@@ -0,0 +1,118 @@
...
+    colstats.AddValue(v);
+  }
+  EXPECT_EQ(colstats.GetCardinality(), 2);
+  EXPECT_EQ(colstats.GetFracNull(), 0);
+  EXPECT_EQ(colstats.GetHistogramBound().size(),
+            0);  // No histogram for categorical data
+  std::vector<ValueFrequencyPair> valfreq =
+      colstats.GetCommonValueAndFrequency();
+  // EXPECT_EQ(valfreq.size(), 2);
+  // PrintCommonValueAndFrequency(colstats.GetCommonValueAndFrequency());
```
rohit-cmu: Dead code ?

test/optimizer/histogram_test.cpp
```diff
@@ -0,0 +1,24 @@
...
+class HistogramTests : public PelotonTest {};
+
+TEST_F(HistogramTests, SimpleHistogramTest) {
+  Histogram h{5};
+  for (int i = 1; i <= 10; i++) {
+    h.Update(i);
+  }
+  h.Print();
+  h.Uniform(4);
+}
```
rohit-cmu: Instead of printing values in the test case, there should be EXPECT_EQ statements which check expected values. While the printing may be done initially while writing the test case, these should be later replaced with EXPECT statements.

test/optimizer/tuple_sampler_test.cpp
```diff
@@ -0,0 +1,58 @@
...
+class TupleSamplerTests : public PelotonTest {};
+
+TEST_F(TupleSamplerTests, SampleCountTest) {
+  const int tuple_count = 100;
+
+  // Create a table
+  auto &txn_manager = concurrency::TransactionManagerFactory::GetInstance();
+  auto txn = txn_manager.BeginTransaction();
+  std::unique_ptr<storage::DataTable> data_table(
+      TestingExecutorUtil::CreateTable(tuple_count, false));
```
rohit-cmu: the parameter for CreateTable is `tuples_per_tilegroup_count`.

src/optimizer/stats/tuple_sampler.cpp
```diff
@@ -0,0 +1,126 @@
...
+        new storage::Tuple(tuple_schema, true));
+
+    LOG_TRACE("tuple_group_offset = %lu, tuple_offset = %lu",
+              rand_tilegroup_offset, rand_tuple_offset);
+    if (!GetTupleInTileGroup(tile_group, rand_tuple_offset, tuple)) {
+      continue;
+    }
+    LOG_TRACE("Add sampled tuple: %s", tuple->GetInfo().c_str());
+    sampled_tuples.push_back(std::move(tuple));
+    sampled_count++;
```
rohit-cmu: Why is there a need for doing this ? The while loop in line 44 could be written as `while (sampled_count.size() < target_sample_count)` . size() will likely be inlined by the compiler to directly access the private size field on the vector<T> class.

src/optimizer/stats/tuple_sampler.cpp
```diff
@@ -0,0 +1,126 @@
...
+    storage::Tuple tile_tuple(&schema, tile_tuple_location);
+
+    for (oid_t tile_column_itr = 0; tile_column_itr < tile_column_count;
+         tile_column_itr++) {
+      type::Value val = (tile_tuple.GetValue(tile_column_itr));
+      tuple->SetValue(tuple_column_itr, val, pool_.get());
+      tuple_column_itr++;
+    }
+  }
+  LOG_TRACE("offset %lu, Tuple info: %s", tuple_offset,
```
rohit-cmu: Why can't `TileGroup::CopyTuple` be used here ? A lot of the code looks identical. Is there something that can't be achieved with that ? If not could you try writing wrappers around it or some other way to reduce the same code being in 2 places in the peloton repository ?

src/include/optimizer/stats/stats_storage.h
```diff
@@ -0,0 +1,127 @@
...
+      storage::DataTable *data_table,
+      std::vector<std::unique_ptr<storage::Tuple>> &sampled_tuples);
+
+  void GetTupleSamples(oid_t database_id, oid_t table_id,
+                       std::vector<storage::Tuple> &tuple_samples);
+
+  void GetColumnSamples(oid_t database_id, oid_t table_id, oid_t column_id,
+                        std::vector<type::Value> &column_samples);
+
+  std::string GenerateSamplesTableName(oid_t db_id, oid_t table_id) {
```
rohit-cmu: Maybe make it  inline ?

src/optimizer/stats/stats_storage.cpp
```diff
@@ -0,0 +1,318 @@
...
+ * GetColumnSamples - Query column samples by db_id, table_id and column_id.
+ * TODO: Implement this function.
+ */
+void StatsStorage::GetColumnSamples(
+    UNUSED_ATTRIBUTE oid_t database_id, UNUSED_ATTRIBUTE oid_t table_id,
+    UNUSED_ATTRIBUTE oid_t column_id,
+    UNUSED_ATTRIBUTE std::vector<type::Value> &column_samples) {}
+
+} /* namespace optimizer */
+} /* namespace peloton */
```
rohit-cmu: In general this is well written, however do you feel are there going to be more columns added to the stats table at a later stage? If so, then probably making this code more modular especially functions like GetColumnStatsTuple and InitializeStatsSchema might help. Else this seems fine.

src/optimizer/stats/tuple_sampler.cpp
wangxiaoyou11: `TileGroup::CopyTuple` is used to copy a tuple into the tilegroup at a given `tuple_slot_id`. But here I want to copy a tuple out from the tilegroup.

rohit-cmu: In that case, maybe this functionality should be added to the tile_group code and not this file.

wangxiaoyou11: Yes. I agree with that.

test/optimizer/column_stats_test.cpp
vanryan: Right

test/optimizer/column_stats_test.cpp
vanryan: Dead code it is.

src/include/optimizer/stats/histogram.h
```diff
@@ -0,0 +1,298 @@
...
+ public:
+
+ /*
+  * Represent a bin of a historgram.
+  */
+  class Bin {
+   public:
+
+    double p;
+    double m;
```
Lst0621: I guess p stands for position and m stands for something like weight. I am not sure if it is a convention. If not, maybe it would be better to have some comments describing p and m.

Does m always take int value? Seems m will often be multiplied by p, so save it as double will not need further casting. Sometimes m will be added to another m, this may be easier for int.  

Maybe p and m can be private. Seems they are only modified by the helper functions like MergeWith().

src/include/optimizer/stats/histogram.h
```diff
@@ -0,0 +1,298 @@
...
+ * - Add mean, variance,
+ * - Add thresholds to keep histogram bound constant
+ */
+class Histogram {
+ public:
+
+  class Bin;
+
+  const uint8_t max_bins;   // For performance it shouldn't be greater than 255
+  std::vector<Bin> bins;
```
Lst0621: So what is needed is some bins with unique p value, and are always sorted according to the p value. 

Operations are inserting, iterating and merging two bins( modify one and delete another). Will std::set also work? Seems easy to just call insert() for set without manually deciding the place to insert. Not sure which is more efficient.

src/optimizer/stats/table_stats.cpp
```diff
@@ -0,0 +1,65 @@
...
+void TableStats::InitColumnStats() {
+  oid_t database_id = table_->GetDatabaseOid();
+  oid_t table_id = table_->GetOid();
+  for (oid_t column_id = 0; column_id < column_count_; column_id++) {
+    std::unique_ptr<ColumnStats> colstats(new ColumnStats(
+        database_id, table_id, column_id, schema_->GetType(column_id)));
+    column_stats_.push_back(std::move(colstats));
+  }
+}
+
```
vanryan: Yeah. However, as you can see when running `cmake`, `Release CXX flags :   -O3` and the O3 optimization level will inline functions whenever the compiler deems worthy, so frankly there is no need for us to write any *inline* functions in Peloton in this case.

vittvolt:
@AllisonWang Hi I'm in the constraint team, and we tend to throw exceptions when constraint check failed (in datatable.cpp, insert_executor.cpp and plan_executor.cpp). Yet this is not handled in stats_test. Any idea how to quickly solve the problem? Thanks

src/executor/analyze_executor.cpp
```diff
@@ -0,0 +1,69 @@
...
+      executor_context_(executor_context) {}
+
+bool AnalyzeExecutor::DInit() {
+  LOG_TRACE("Initializing analyze executor...");
+  LOG_TRACE("Analyze executor initialized!");
+  return true;
+}
+
+bool AnalyzeExecutor::DExecute() {
+  LOG_TRACE("Executing Analyze...");
```
rohit-cmu: I don't see a need for this LOG statement

src/executor/analyze_executor.cpp
```diff
@@ -0,0 +1,69 @@
...
+
+AnalyzeExecutor::AnalyzeExecutor(const planner::AbstractPlan *node,
+                                 ExecutorContext *executor_context)
+    : AbstractExecutor(node, executor_context),
+      executor_context_(executor_context) {}
+
+bool AnalyzeExecutor::DInit() {
+  LOG_TRACE("Initializing analyze executor...");
+  LOG_TRACE("Analyze executor initialized!");
+  return true;
```
rohit-cmu: I don't see a need for this LOG statement. Additionally I don't see the purpose of this function. It always returns true and prints 2 LOG statements.

src/include/executor/analyze_executor.h
```diff
@@ -0,0 +1,48 @@
...
+
+ protected:
+  bool DInit();
+
+  bool DExecute();
+
+ private:
+  ExecutorContext* executor_context_;
+};
+
```
rohit-cmu: A couple of new lines could be removed I think. - Line 32, 40.

Rest is good

src/include/optimizer/stats/column_stats_collector.h
```diff
@@ -0,0 +1,88 @@
...
+// ColumnStatsCollector
+//===--------------------------------------------------------------------===//
+class ColumnStatsCollector {
+ public:
+  /* Default parameters for probabilistic stats collector */
+  int hll_precision = 8;
+  double cmsketch_eps = 0.01;
+  double cmsketch_gamma = 0.01;
+  uint8_t max_bins = 100;
+  uint8_t top_k = 10;
```
rohit-cmu: These variables should be private

src/include/optimizer/stats/cost.h
```diff
@@ -0,0 +1,165 @@
...
+   * Cost of LIMIT = limit * tuple_cost
+   */
+  static double LimitCost(
+    const std::shared_ptr<TableStats>& input_stats, size_t limit,
+    std::shared_ptr<TableStats>& output_stats);
+
+  /*
+   * Cost of ORDER BY.
+   * TODO: implement me!
+   */
```
rohit-cmu: If this is not implemented shouldn't it just return default cost ?

src/include/optimizer/stats/histogram.h
```diff
@@ -0,0 +1,305 @@
...
+
+    inline bool operator >(const Bin &bin) const {
+      return p > bin.p;
+    }
+
+    inline bool operator !=(const Bin &bin) const {
+      return p != bin.p;
+    }
+
+    void Print() {
```
rohit-cmu: inline function ?

src/include/optimizer/stats/histogram.h
```diff
@@ -0,0 +1,305 @@
...
+      LOG_INFO("Bin: p=[%f],m=[%f]", p, m);
+    }
+  };
+};
+
+} /* namespace optimizer */
+} /* namespace peloton */
```
rohit-cmu: Although the file api design is clean and easy to understand, I did not understand why int and double are being used instead of the peloton datatypes that are there. Is there any specific reason why the standard c++ data types are being used instead ?

src/include/optimizer/stats/selectivity.h
```diff
@@ -0,0 +1,95 @@
...
+#include "catalog/column_catalog.h"
+#include "catalog/catalog.h"
+#include "optimizer/stats/table_stats.h"
+#include "optimizer/stats/value_condition.h"
+#include "optimizer/stats/column_stats.h"
+
+namespace peloton {
+namespace optimizer {
+
+static constexpr double DEFAULT_SELECTIVITY = 0.5;
```
rohit-cmu: Any reason why this is declared outside the scope of the class ?

src/include/optimizer/stats/stats_util.h
```diff
@@ -0,0 +1,47 @@
...
+#include "type/value.h"
+#include "type/value_factory.h"
+
+namespace peloton {
+namespace optimizer {
+
+class StatsUtil {
+public:
+  // Convert numeric peloton value type to primitive value.
+  // Return NaN if value is not numeric.
```
rohit-cmu: This looks like a generic function and not specific to StatsUtil. In that case wouldn't it make more sense to make this function a part of the value / type.cpp files ? Since all related such functionalities are present in those files/directories

src/include/optimizer/stats/value_condition.h
```diff
@@ -0,0 +1,50 @@
...
+//===--------------------------------------------------------------------===//
+class ValueCondition {
+public:
+  oid_t column_id;
+  std::string column_name;
+  ExpressionType type;
+  type::Value value;
+
+  // Only with id.
+  ValueCondition(oid_t column_id, ExpressionType type, const type::Value& value)
```
rohit-cmu: why not define column_name="" as a default parameter to combine these 2 functions into 1. So if you recieve a column name you use that else it is "".

src/optimizer/stats/cost.cpp
```diff
@@ -0,0 +1,209 @@
...
+  PL_ASSERT(rhs != nullptr);
+  PL_ASSERT(num_rows > 0);
+
+  size_t num_tuples = 1;
+  double sel1 = lhs->num_rows / static_cast<double>(num_rows);
+  double sel2 = rhs->num_rows / static_cast<double>(num_rows);
+  LOG_TRACE("Conjunction sel1[%f] sel2[%f]", sel1, sel2);
+  switch (type) {
+    case ExpressionType::CONJUNCTION_AND:
+      // (sel1 * sel2) * num_rows
```
rohit-cmu: Dead Code in line 71 and 75

src/optimizer/stats/cost.cpp
```diff
@@ -0,0 +1,209 @@
...
+  for (oid_t column : columns) {
+    double cardinality = input_stats->GetCardinality(column);
+    max_cardinality = std::max(max_cardinality, cardinality);
+    rows *= cardinality;
+  }
+  return static_cast<size_t>(rows + max_cardinality / 2);
+}
+
+} /* namespace optimizer */
+} /* namespace peloton */
```
rohit-cmu: Nice work. :1st_place_medal:

src/optimizer/stats/table_stats.cpp
```diff
@@ -0,0 +1,161 @@
...
+  } else {
+    num_rows = col_stats_ptrs[0]->num_rows;
+  }
+}
+
+void TableStats::UpdateNumRows(size_t new_num_rows) { num_rows = new_num_rows; }
+
+bool TableStats::HasIndex(const std::string column_name) {
+  auto column_stats = GetColumnStats(column_name);
+  if (column_stats == nullptr) {
```
rohit-cmu: shouldn't this be a PL_ASSERT instead of returning false. There is no way of distinguishing whether the column_stats was null or it did not have an index, since in both cases you return false.

src/optimizer/stats/table_stats.cpp
```diff
@@ -0,0 +1,161 @@
...
+    return false;
+  }
+  return column_stats->has_index;
+}
+
+bool TableStats::HasPrimaryIndex(const std::string column_name) {
```
rohit-cmu: shouldn't these functions be moved to table_stats.h

src/optimizer/stats/table_stats.cpp
```diff
@@ -0,0 +1,161 @@
...
+bool TableStats::RemoveColumnStats(const oid_t column_id) {
+  if (column_id >= col_stats_list_.size()) {
+    return false;
+  }
+  col_stats_list_.erase(col_stats_list_.begin() + column_id);
+  return true;
+}
+
+}  // namespace optimizer
+}  // namespace peloton
```
rohit-cmu: A lot of functions here I thought could be moved to the table_stats.h file.

src/optimizer/stats/table_stats_collector.cpp
```diff
@@ -0,0 +1,90 @@
...
+  InitColumnStatsCollectors();
+
+  size_t tile_group_count = table_->GetTileGroupCount();
+  // Collect stats for all tile groups.
+  for (size_t offset = 0; offset < tile_group_count; offset++) {
+    std::shared_ptr<storage::TileGroup> tile_group =
+        table_->GetTileGroup(offset);
+    storage::TileGroupHeader* tile_group_header = tile_group->GetHeader();
+    oid_t tuple_count = tile_group->GetAllocatedTupleCount();
+    active_tuple_count_ += tile_group_header->GetActiveTupleCount();
```
rohit-cmu: If this function is called multiple times through the analyze SQL query, then the active_tuple_count_ will just keep incrementing with this GetActiveTupleCount() value. Isn't this wrong in that case then ?

test/optimizer/stats_storage_test.cpp
```diff
@@ -0,0 +1,279 @@
...
+ * 2. Compare the column stats values with the ground truth.
+ */
+void VerifyAndPrintColumnStats(storage::DataTable *data_table,
+                               int expect_tuple_count) {
+  // Check the tuple count in the 'pg_column_stats' catalog.
+  // auto column_stats_catalog =
+  // catalog::ColumnStatsCatalog::GetInstance(nullptr);
+  StatsStorage *stats_storage = StatsStorage::GetInstance();
+
+  // Verify both the internal data and the stats got by interface.
```
poojanilangekar: Dead code?

test/optimizer/stats_storage_test.cpp
```diff
@@ -0,0 +1,279 @@
...
+  EXPECT_NE(column_stats_ptr, nullptr);
+
+  EXPECT_EQ(column_stats_ptr->num_rows, num_row_1);
+  EXPECT_EQ(column_stats_ptr->cardinality, cardinality_1);
+  EXPECT_EQ(column_stats_ptr->frac_null, frac_null_1);
+
+  EXPECT_EQ(column_stats_ptr->column_name, column_name_1);
+}
+
+// TEST_F(StatsStorageTests, AnalyzeStatsForAllTablesTest) {
```
poojanilangekar: Dead code / Missing code?

test/optimizer/top_k_elements_test.cpp
```diff
@@ -0,0 +1,164 @@
...
+  top_k_elements.Add("Million", 1000000);
+
+  top_k_elements.Add(std::string{"Cowboy"}, 2333);
+  top_k_elements.Add(std::string{"Bebop"}, 2334);
+  top_k_elements.Add(std::string{"RayCharles"}, 2335);
+  int i;
+  for (i = 0; i < 30; ++i) {
+    top_k_elements.Add(i, i);
+  }
+  // top_k_elements.PrintTopKQueueOrderedMaxFirst();
```
poojanilangekar: Dead code.

test/optimizer/cost_test.cpp
```diff
@@ -0,0 +1,122 @@
...
+using namespace optimizer;
+
+const int N_ROW = 10000;
+
+class CostTests : public PelotonTest {};
+
+// tablename: test
+// database name: DEFAULT_DB_NAME
+void CreateAndLoadTable() {
+  TestingSQLUtil::ExecuteSQLQuery(
```
rohit-cmu: Is there a specific reason why you have defined your own Create and Load Table instead of using the CreateandPopulateTable() from the TestingExecutorUtil. The table it creates has 2 integer columns, 1 varchar and 1 decimal column.

rohit-cmu:
In general, I feel the code is modular to add the cost for multiple other functionalities without rewriting a lot of other code which is great.
As @poojanilangekar mentioned I also felt that at certain places using polymorphism/ overriding / templates / default variables extra code in some places can be eliminated. Additionally at some places there were existing functionalities of Peloton which could have been used better. In some other places a lot of variables should be made private/protected. 

Overall, the code is well structured and cleanly written.

src/optimizer/stats/table_stats_collector.cpp
wangxiaoyou11: Each table has multiple tile groups. We get the active tuple count of each tile group, and add them together to get the total number of active tuples as the `active_tuple_count_`.

test/optimizer/cost_test.cpp
wangxiaoyou11: This is because we want to have the control over the tuples we want to insert, so that we can know the basic ground truth of the costs and build some `EXPECT` statement to test our cost functions.

src/executor/plan_executor.cpp
```diff
@@ -376,6 +376,11 @@ executor::AbstractExecutor *BuildExecutorTree(
       child_executor = new executor::DropExecutor(plan, executor_context);
       break;
 
+    case PlanNodeType::ANALYZE:
+      LOG_TRACE("Adding Analyze Executer");
```
wangxiaoyou11: Just find the typo was copied from other log traces in the same switch block, haha. We will fix others too.
See:
https://github.com/cmu-db/peloton/blob/master/src/executor/plan_executor.cpp#L303
https://github.com/cmu-db/peloton/blob/master/src/executor/plan_executor.cpp#L308

src/include/optimizer/stats/column_stats.h
```diff
@@ -0,0 +1,99 @@
...
+        column_id(column_id),
+        column_name(column_name),
+        has_index(has_index),
+        num_rows(num_rows),
+        cardinality(cardinality),
+        frac_null(frac_null),
+        most_common_vals(most_common_vals),
+        most_common_freqs(most_common_freqs),
+        histogram_bounds(histogram_bounds),
+        is_basetable{true} {}
```
wangxiaoyou11: Actually, the `ColumnStats` is just a wrapper of all column stats and we just want to use it like a `struct`. We wan to make these variables public so that we can directly access the variables instead of calling functions.

src/include/optimizer/stats/tuple_sampler.h
```diff
@@ -0,0 +1,52 @@
...
+//
+//===----------------------------------------------------------------------===//
+
+#pragma once
+
+#include "type/types.h"
+#include "type/ephemeral_pool.h"
+
+namespace peloton {
+namespace storage {
```
wangxiaoyou11: I just need the declaration of these classes in this header file. If I don't use the full object, It will not be a good practice to include the header files of these classes here, which may result in "a include loop". This strategy is called "forward declaration" in C++.
Check the following links for reference:
http://stackoverflow.com/questions/8632430/about-class-declaration-in-header-file-c
http://jatinganhotra.com/blog/2012/11/25/forward-class-declaration-in-c-plus-plus/

src/include/optimizer/stats/stats_storage.h
```diff
@@ -0,0 +1,141 @@
...
+
+#include <sstream>
+
+#include "common/macros.h"
+#include "type/types.h"
+#include "type/value_factory.h"
+
+namespace peloton {
+
+namespace storage {
```
wangxiaoyou11: https://github.com/cmu-db/peloton/pull/619#discussion_r116358872

src/optimizer/stats/table_stats.cpp
wangxiaoyou11: Thanks for pointing out that. The default value of `has_index` is false here. If `column_stats` is null, I will just return default value to the cost model. I've defined the macro in the `table_stats.h` for clarification.

src/optimizer/stats/cost.cpp
wangxiaoyou11: It's a comment for how we calculate `num_tuples` for conjunction expression instead of dead code.

src/executor/analyze_executor.cpp
AllisonWang: @rohit-cmu Right, this is added just for consistency with other executor implementations (`create_executor`, `drop_executor` etc). A LOG_TRACE won't hurt in this case.

src/include/optimizer/stats/column_stats_collector.h
AllisonWang: Giving those variables public access is for quick access of approximation algorithm's precisions and error bound outside of this class. I guess giving them getter and setter also works but make them public fits more of the design purpose (right now they are not used but we still want to expose those information for public access)

src/include/optimizer/stats/cost.h
AllisonWang: This is not yet implemented lol

src/include/optimizer/stats/histogram.h
AllisonWang: @rohit-cmu Double are used here because only numerics can be used to build the histogram bounds (for now). Theoretically, as long as the value can be compared and can handle arithmetic operations, it should work without converting to C++ data types (maybe in the future we can use template). But I am not sure if it will impact the performance since we will have many intermediate value object creations and primitive type are always faster when during calculations instead of using value type's own methods. Not sure we can test this in the future.

src/include/optimizer/stats/selectivity.h
AllisonWang: @rohit-cmu To make the `static constexpr` compile (It doesn't compile when defined inside of the class)

src/include/optimizer/stats/stats_util.h
AllisonWang: @rohit-cmu Good idea! Will transfer this function.

src/include/optimizer/stats/value_condition.h
AllisonWang: @rohit-cmu sounds good!

src/include/optimizer/stats/hyperloglog.h
```diff
@@ -0,0 +1,92 @@
...
+  const int precision_;
+  const int register_count_;
+  libcount::HLL* hll_;
+
+  // Compute hash for given value using Murmur3.
+  uint64_t Hash(const type::Value& value) {
+    uint64_t hash[2];
+    if (value.CheckInteger())
+    {
+      int raw_value = value.GetAs<int>();
```
AllisonWang: @poojanilangekar You are right. This will be replaced with more accurate type checking and casting :)

src/optimizer/stats/stats_storage.cpp
```diff
@@ -0,0 +1,295 @@
...
+              table_name.c_str());
+    return ResultType::FAILURE;
+  }
+  std::unique_ptr<TableStatsCollector> table_stats_collector(
+      new TableStatsCollector(table));
+  table_stats_collector->CollectColumnStats();
+  InsertOrUpdateTableStats(table, table_stats_collector.get(), txn);
+  return ResultType::SUCCESS;
+}
+
```
wangxiaoyou11: It's a TODO. Just added comment for it.

src/optimizer/stats/table_stats.cpp
wangxiaoyou11: @rohit-cmu I think it's always a good practice in C++ to separate function declarations and function implementations in different files.

AllisonWang:
@vittvolt Could you elaborate on which part we should change?

src/include/optimizer/stats/column_stats_collector.h
apavlo: Make them inline getter/setter funcs.

apavlo:
This PR is great. We should try to merge this in as soon as possible.

src/optimizer/stats/selectivity.cpp
```diff
@@ -0,0 +1,180 @@
...
+  // TODO: make sure condition uses column id. check if column name is not
+  // empty.
+  std::shared_ptr<ColumnStats> column_stats =
+      table_stats->GetColumnStats(condition.column_id);
+  // Return default selectivity if no column stats for given column_id
+  if (column_stats == nullptr) {
+    return DEFAULT_SELECTIVITY;
+  }
+  // Use histogram to estimate selectivity
+  std::vector<double> histogram = column_stats->histogram_bounds;
```
hanfei1991: Why not consider the row number of every buckets ?

apavlo:
I merged this by hand.

PR #622, opened 2017-04-12
UDF Code Review 1

NasrinJaleel93:
Currently we support parsing of C and PGSQL UDFs. 

This PR adds parsing support for PGSQL and C UDFs. The code currently handles conversion of PostgreSQL's parse tree into Peloton's parse tree. 

**Newly added Files:**
1. /peloton/src/include/parser/create_function_statement.h : This file defines the CreateFunctionStatement struct which defines the structure in which we capture PostgreSQL's parse tree.

**Modified Files:**
1. /peloton/src/parser/postgresparser.cpp : Added a case in the switch statement inside NodeTransform() to handle CreateFunctionStmt. Added CreateFunctionTransform() to actually do the conversion of PostgreSQL's parse tree object(CreateFunctionStmt) to Peloton's parse tree object(CreateFunctionstatement).
2. peloton/test/parser/postgresparser_test.cpp  : Added 2 tests to check the validity of the parsed object for C UDF as well as PGSQL UDF.
Prashasthi • 16 mins


prashasthip:
This PR adds parsing support for PGSQL and C UDFs. The code currently handles conversion of PostgreSQL's parse tree into Peloton's parse tree. 

**Newly added Files:**
1. /peloton/src/include/parser/create_function_statement.h : This file defines the CreateFunctionStatement struct which defines the structure in which we capture PostgreSQL's parse tree.

**Modified Files:**
1. /peloton/src/parser/postgresparser.cpp : Added a case in the switch statement inside NodeTransform() to handle CreateFunctionStmt. Added CreateFunctionTransform() to actually do the conversion of PostgreSQL's parse tree object(CreateFunctionStmt) to Peloton's parse tree object(CreateFunctionstatement).
2. peloton/test/parser/postgresparser.cpp  : Added 2 tests to check the validity of the parsed object for C UDF as well as PGSQL UDF.

src/include/parser/create_function_statement.h
```diff
@@ -0,0 +1,136 @@
...
+  enum ASclause {
+    EXECUTABLE=0,
+    QUERY_STRING=1	  	
+  };
+
+  CreateFunctionStatement() :SQLStatement(StatementType::CREATE_FUNC){
+  }; 
+
+  virtual ~CreateFunctionStatement() {
+    //insert delete code for attributes later
```
xinzhang1234: I am little confused about why use virtual destructor here? Since I have not seen any child inherited this function, from my point of view, there is no need to use virtual here.

src/include/parser/create_function_statement.h
```diff
@@ -0,0 +1,136 @@
...
+
+#include "expression/abstract_expression.h"
+#include "type/types.h"
+#include "common/sql_node_visitor.h"
+#include "parser/sql_statement.h"
+
+namespace peloton {
+namespace parser {
+
+struct Parameter {
```
xinzhang1234: It is better to use class instead of struct here because there are so many functions and variables in this struct. Use class you can use private and public but for struct all will be public here.

src/include/parser/create_function_statement.h
```diff
@@ -0,0 +1,136 @@
...
+      case BOOL:
+      case BOOLEAN:
+        return type::Type::BOOLEAN;
+        break;
+        // add other types as necessary 
+    }
+  }	
+};
+
+struct ReturnType : Parameter {
```
xinzhang1234: Also class is better than struct since class is more common for inheritance.

src/include/parser/create_function_statement.h
```diff
@@ -0,0 +1,136 @@
...
+
+  virtual void Accept(SqlNodeVisitor* v) const override {
+    v->Visit(this);
+  }
+
+  PLType language;
+  ASclause as_type;
+  std::vector<std::string> function_body;
+  ReturnType *return_type;
+  std::vector<FuncParameter*>* func_parameters;
```
xinzhang1234: why use pointer and pointer for vector here since vector is already a pointer here？A little bit confused here.

src/parser/postgresparser.cpp
```diff
@@ -800,6 +800,143 @@ parser::SQLStatement* PostgresParser::CreateTransform(CreateStmt* root) {
...
+                    ->val.str);
+  parser::ReturnType* result = nullptr;
+
+  // Transform parameter type
+  if ((strcmp(name, "int") == 0) || (strcmp(name, "int4") == 0)) {
+    data_type = FuncParameter::DataType::INT;
+  } 
+  //*******See which of the below are needed later***********
+  // I think text/varchar may be used for the pg_catalog
+  else if (strcmp(name, "varchar") == 0) {
```
xinzhang1234: This code part is same to the part in FunctionParameterTransform, it is better to use a small common function here.

src/parser/postgresparser.cpp
```diff
@@ -800,6 +800,143 @@ parser::SQLStatement* PostgresParser::CreateTransform(CreateStmt* root) {
...
+ 
+  ReturnType* ret_temp =
+          ReturnTypeTransform(reinterpret_cast<TypeName*>(root->returnType));
+  result->return_type = ret_temp;
+
+  // Assuming only one function name can be passed for now.
+  char* name = (reinterpret_cast<value*>(root->funcname->tail->data.ptr_value)
+                    ->val.str);
+  std::string func_name_string(name);
+  result->function_name = func_name_string; //We may have to change this to another function for handlign char * to string
```
xinzhang1234: A little bit confused about why use another func_name? it is better to merge this one into one line code: result->function_name = name

src/parser/postgresparser.cpp
```diff
@@ -800,6 +800,143 @@ parser::SQLStatement* PostgresParser::CreateTransform(CreateStmt* root) {
...
+        auto query_string = reinterpret_cast<value*>(cell2->data.ptr_value)->val.str;
+        // LOG_DEBUG("%s\n",query_string);
+        std::string new_func_body(query_string);
+        result->function_body.push_back(new_func_body);
+      }
+      result->set_as_type();
+    }
+    else if(strcmp(def_elem->defname, "language") == 0) {
+      auto lang = reinterpret_cast<value*>(def_elem->arg)->val.str;
+      if ((strcmp(lang, "plpgsql") == 0)) {
```
xinzhang1234: It is better to use enum to compare because compare string takes longer time.

src/include/parser/create_function_statement.h
apavlo: You should be using smart pointers too.

src/parser/postgresparser.cpp
apavlo: Use `enum class` too

http://www.cprogramming.com/c++11/c++11-nullptr-strongly-typed-enum-class.html

src/include/parser/create_function_statement.h
prashasthip: We kept it consistent with the other ParseNode classes present in Peloton currently, which use struct.

src/include/parser/create_function_statement.h
```diff
@@ -0,0 +1,136 @@
...
+
+  enum FuncParamMode {
+    FUNC_PARAM_IN = 'i',		/* input only */
+    FUNC_PARAM_OUT = 'o',		/* output only */
+    FUNC_PARAM_INOUT = 'b',		/* both */
+    FUNC_PARAM_VARIADIC = 'v',	/* variadic (always input) */
+    FUNC_PARAM_TABLE = 't'	
+  };
+
+  enum DataType {
```
prashasthip: Yes, we will be refactoring this code to use type::Type.

src/parser/postgresparser.cpp
```diff
@@ -800,6 +800,143 @@ parser::SQLStatement* PostgresParser::CreateTransform(CreateStmt* root) {
   return reinterpret_cast<parser::SQLStatement*>(result);
 }
 
+// This helper function takes in a Postgres FunctionParameter object and transforms
+// it into a Peloton FunctionParameter object
+parser::FuncParameter* PostgresParser::FunctionParameterTransform(FunctionParameter* root) {
+  parser::FuncParameter::DataType data_type;
+  TypeName* type_name = root->argType;
+  char* name = (reinterpret_cast<value*>(type_name->names->tail->data.ptr_value)
```
prashasthip: screen shot 2017-04-16 at 6 29 35 pm

src/include/parser/create_function_statement.h
NasrinJaleel93: @xinzhang1234 we might consider changing this. A lot of the CreateFunctionStatement has been kept very similar to CreateFunction. @apavlo we will incorporate this change, thanks!

src/parser/postgresparser.cpp
prashasthip: We only need the type name.

src/parser/postgresparser.cpp
NasrinJaleel93: @xinzhang1234 Will do. Wasn't aware the = operator for string was overloaded to handle char* types too.

test/parser/postgresparser_test.cpp
```diff
@@ -724,5 +724,23 @@ TEST_F(PostgresParserTests, CreateDbTest) {
   delete stmt_list;
 }
 
+TEST_F(PostgresParserTests, CreatePGSQLFuncTest) {
+  std::string query = "CREATE OR REPLACE FUNCTION increment(i integer) RETURNS integer AS $$   BEGIN  RET  j;  END; $$ LANGUAGE plpgsql;";
+
+  auto parser = parser::PostgresParser::GetInstance();
+  auto stmt_list = parser.BuildParseTree(query).release();
+  EXPECT_TRUE(stmt_list->is_valid);
```
prashasthip: Yup. 
We will be adding more tests.

src/parser/postgresparser.cpp
NasrinJaleel93: The def_elem->arg, returns a string. Is there a way to make this an enum compare still? If you are talking about an enum with string values, I don't understand how it would save time. (I might be misunderstanding you here @xinzhang1234 )

src/include/parser/create_function_statement.h
```diff
@@ -0,0 +1,136 @@
...
+      case BOOLEAN:
+        return type::Type::BOOLEAN;
+        break;
+        // add other types as necessary 
+    }
+  }	
+};
+
+struct ReturnType : Parameter { 
+  ReturnType(DataType type):Parameter(type){};
```
NasrinJaleel93: It is the return type of the UDF function

test/parser/postgresparser_test.cpp
NasrinJaleel93: Yes, we haven't completed the registering of the function yet. Currently working on this part.

NasrinJaleel93: The PR we submitted creates a Peloton parsenode from a Postgres parsenode.

test/parser/postgresparser_test.cpp
```diff
@@ -724,5 +724,23 @@ TEST_F(PostgresParserTests, CreateDbTest) {
   delete stmt_list;
 }
 
+TEST_F(PostgresParserTests, CreatePGSQLFuncTest) {
+  std::string query = "CREATE OR REPLACE FUNCTION increment(i integer) RETURNS integer AS $$   BEGIN  RET  j;  END; $$ LANGUAGE plpgsql;";
```
NasrinJaleel93: Just a sample UDF. We just support a preliminary parse as of now, that does not validate the syntax completely. This level validation only happens when the function is actually executed.

src/include/parser/create_function_statement.h
opensky2: On the other hand, all the other statement types are implemented as structs, so the benefits of a class, I think, are overridden by the consistency of a struct.

src/parser/postgresparser.cpp
```diff
@@ -800,6 +800,143 @@ parser::SQLStatement* PostgresParser::CreateTransform(CreateStmt* root) {
...
+                    ->val.str);
+  parser::FuncParameter* result = nullptr;
+
+  // Transform parameter type
+  if ((strcmp(name, "int") == 0) || (strcmp(name, "int4") == 0)) {
+    data_type = FuncParameter::DataType::INT;
+  } 
+  //*******See which of the below are needed later***********
+  // I think text/varchar may be used for the pg_catalog
```
opensky2: This comment should be clarified or removed.

src/parser/postgresparser.cpp
```diff
@@ -800,6 +800,143 @@ parser::SQLStatement* PostgresParser::CreateTransform(CreateStmt* root) {
...
+    data_type = FuncParameter::DataType::SMALLINT;
+  } else if (strcmp(name, "text") == 0) {
+    data_type = FuncParameter::DataType::TEXT;
+  } else if (strcmp(name, "tinyint") == 0) {
+    data_type = FuncParameter::DataType::TINYINT;
+  } else if(strcmp(name, "bool") == 0) {
+     data_type = FuncParameter::DataType::BOOL;
+  } else {
+    LOG_ERROR("Column DataType %s not supported yet...\n", name);
+    throw NotImplementedException("...");
```
opensky2: I think it would be more useful if the Exception message returned the same information as is logged in the ERROR stream.

src/parser/postgresparser.cpp
```diff
@@ -800,6 +800,143 @@ parser::SQLStatement* PostgresParser::CreateTransform(CreateStmt* root) {
...
+  // Transform Varchar parameter name
+  result = new ReturnType(data_type);
+
+  return result;
+}
+
+// This function takes in a Postgres CreateFunctionStmt parsenode
+// and transfers into a Peloton CreateFunctionStatement parsenode.
+// Please refer to parser/parsenode.h for the definition of
+// CreateFunctionStmt parsenodes.
```
opensky2: The file is "parser/parsenodes.h" (with an 's'), but that's a minor quibble with an otherwise useful comment.

test/parser/postgresparser_test.cpp
```diff
@@ -724,5 +724,23 @@ TEST_F(PostgresParserTests, CreateDbTest) {
...
+TEST_F(PostgresParserTests, CreateCFuncTest) {
+  std::string query = "CREATE FUNCTION c_overpaid(integer, integer) RETURNS boolean AS 'DIRECTORY/funcs', 'c_overpaid' LANGUAGE C STRICT;";
+  auto parser = parser::PostgresParser::GetInstance();
+  auto stmt_list = parser.BuildParseTree(query).release();
+  EXPECT_TRUE(stmt_list->is_valid);
+  delete stmt_list;
+}
+
+
 }  // End test namespace
```
opensky2: Other useful tests might be:
1) check if we can handle CREATE FUNCTION statements when there already exists a function with that name
2) implement more complex functions to ensure the parser can handle specific tables and attributes

src/tcop/tcop.cpp
```diff
@@ -142,7 +142,7 @@ ResultType TrafficCop::ExecuteStatement(
     std::vector<FieldInfo> &tuple_descriptor, int &rows_changed,
     std::string &error_message,
     const size_t thread_id UNUSED_ATTRIBUTE) {
-  LOG_TRACE("Received %s", query.c_str());
+  LOG_DEBUG("Received %s", query.c_str());
```
opensky2: Since this is the only change in an otherwise untouched file, I don't know that it's appropriate?  It may be useful for your own debugging, but LOG_TRACE is used throughout the file, and in an essential function like ExectuteStatement, it is likely worthwhile to keep it.  One compromise might be to include both a LOG_TRACE and a LOG_DEBUG statement.

prashasthip: Yes, we had changed this for our debugging purposes. Will change this back.

src/include/parser/create_function_statement.h
```diff
@@ -0,0 +1,136 @@
...
+};
+
+//move it to types.h
+enum PLType {
+   PL_PGSQL=0,
+   PL_C=1
+};
+
+
+//might want to change it to char* instead of string
```
opensky2: What are the pros and cons of changing instances of std::string in this struct to char*?  I think its easier in general to work with C++ strings and convert it with c_str() if necessary.

prashasthip: The thought process behind it was to keep it consistent with the other header files. But we might not change it.

PR #624, opened 2017-04-13
Add support for IS DISTINCT FROM and add COMPARE_DISTINCT_FROM type

haojin2:
This PR is for adding support for IS DISTINCT FROM in the parse tree transformer. It includes:
- Add checking for "kind" of A_Expr parsenodes
- Add COMPARE_DISTINCT_FROM type and corresponding evaluate function
- Fix support for not equal operators
- Added missing exceptions for not supported things


src/include/type/types.h
```diff
@@ -180,12 +180,14 @@ enum class ExpressionType {
...
   COMPARE_IN = 19,
+  // IS DISTINCT FROM operator
+  COMPARE_DISTINCT_FROM = 20,
 
   // -----------------------------
   // Conjunction Operators
   // -----------------------------
-  CONJUNCTION_AND = 20,
-  CONJUNCTION_OR = 21,
+  CONJUNCTION_AND = 50,
```
apavlo: The ExpressionTypes need to be in sorted order here.

src/parser/postgresparser.cpp
```diff
@@ -194,7 +194,10 @@ parser::TableRef* PostgresParser::FromTransform(List* root) {
               RangeSubselectTransform(reinterpret_cast<RangeSubselect*>(node)));
           break;
         }
-        default: { LOG_ERROR("From Type %d not supported yet...", node->type); }
+        default: {
+          LOG_ERROR("From Type %d not supported yet...", node->type);
+          throw NotImplementedException(" ");
```
apavlo: I would put an error message in here.

src/parser/postgresparser.cpp
```diff
@@ -219,7 +222,10 @@ parser::TableRef* PostgresParser::FromTransform(List* root) {
       result = RangeSubselectTransform(reinterpret_cast<RangeSubselect*>(node));
       break;
     }
-    default: { LOG_ERROR("From Type %d not supported yet...", node->type); }
+    default: {
+      LOG_ERROR("From Type %d not supported yet...", node->type);
+      throw NotImplementedException(" ");
```
apavlo: Same with these and all the other `NotImplementedException` invocations.

src/parser/postgresparser.cpp
haojin2: The error message is printed using the LOG_ERROR instead since we want to show the type of the node that is causing the error, formatting is not supported in the Exception message.

src/include/type/types.h
haojin2: Okay

src/parser/postgresparser.cpp
haojin2: All other exceptions also have the error messages printed using LOG_ERROR above them.

apavlo: Use `StringUtil::Format()`

https://github.com/cmu-db/peloton/blob/master/src/type/types.cpp#L150

haojin2: Okay

src/parser/postgresparser.cpp
```diff
@@ -600,17 +609,23 @@ expression::AbstractExpression* PostgresParser::AExprTransform(A_Expr* root) {
     default: {
       LOG_ERROR("Right expr of type %d not supported yet...\n",
                 root->rexpr->type);
-      return nullptr;
+      throw NotImplementedException(" ");
```
seojungmin: You will need "delete(left_expr);" before throwing an exception here.

haojin2: Thanks for pointing that out!

haojin2:
Rebased, it should be ready for merge once it passes all checks.

Issue #561, opened 2017-03-03
Support IN in where clause


malin1993ml:
```sql
postgres=# create table a(id int, value varchar);
create table a(id int, value varchar) 0
postgres=# insert into a values(1, 'hi');
insert into a values(1, 'hi') 1
postgres=# select * from a;
 id | value 
----+-------
  1 | hi
(1 row)

postgres=# select * from a where id in (1);
Error parsing SQL statement
```

poojanilangekar:
Duplicate of #568

Issue #552, opened 2017-02-27
Support "is null" and "is not null" in where clause


malin1993ml:
Right now we cannot evaluate whether a value is (not) NULL is where clause:
```sql
postgres=# create table a(id int, value varchar);
create table a(id int, value varchar) 0
postgres=# insert into a values(1, 'hi');
insert into a values(1, 'hi') 1
postgres=# select * from a where value is null;
Error parsing SQL statement
postgres=# select * from a where value is not null;
Error parsing SQL statement
postgres=# 
```

apavlo:
Merged in #842

PR #628, opened 2017-04-14
Added support for IN, is null, and is not null in where clause

YuanLuo1:
The pull request is for added support for IN, is null and is not null in where clause, details are:

- Insert null is solved by switching to the new parser.
- Support IN in where clause now only support one value in the list. This is to solve the issue #561.
- Added support to is null and is not null in where clause. However, when tuple field type id is greater than DECIMAL the query will fail. This is to solve the issue #552.
```
enum TypeId {
    INVALID = 0,
    PARAMETER_OFFSET,
    BOOLEAN,
    TINYINT,
    SMALLINT,
    INTEGER,
    BIGINT,
    DECIMAL,
    TIMESTAMP,
    DATE,
    VARCHAR,
    VARBINARY,
    ARRAY,
    UDT
  };
``` 
- Added 4 unit test for all the new features above, add all the tests passed.
```
Test  #82: new_feature_test ..................   Passed   85.38 sec
```


test/sql/newFeatures2.sql
```diff
@@ -0,0 +1,5 @@
+create table a(id int, value varchar);
```
dvanaken: Also rename this file to something like null2.sql.

test/sql/new_feature_test.cpp
```diff
@@ -0,0 +1,198 @@
+//===----------------------------------------------------------------------===//
```
dvanaken: Can you separate the tests for these two new features put them into different files (e.g., null_sql_test.cpp & in_sql_test.cpp)? Make sure you update the headers too.

test/sql/new_feature_test.cpp
```diff
@@ -0,0 +1,198 @@
...
+              "Query: Iinsert into a values(4, 4)");
+
+      TestingSQLUtil::ExecuteSQLQuery("insert into a values(1, 1);");
+      TestingSQLUtil::ExecuteSQLQuery("insert into a values(2, null);");
+      TestingSQLUtil::ExecuteSQLQuery("insert into a values(3, null);");
+      TestingSQLUtil::ExecuteSQLQuery("insert into a values(4, 4);");
+
+      LOG_INFO("Tuple inserted!");
+    }
+      TEST_F(NewFeatureTest, WhereInTest) {
```
dvanaken: Do we now support IN with multiple values (or when the value is a set)? How about when the value is an expression? If so, these test cases should be added.

src/parser/postgresparser.cpp
```diff
@@ -1183,18 +1285,27 @@ parser::SQLStatementList* PostgresParser::ParseSQLString(
...
 }
 
-parser::SQLStatementList* PostgresParser::ParseSQLString(
-    const std::string& sql) {
-  return ParseSQLString(sql.c_str());
-}
+    parser::SQLStatementList* PostgresParser::ParseSQLString(
+            const std::string& sql) {
+      std::string str = (std::string&) sql;
+      std::cout << str << std::endl;
```
dvanaken: Remove this.

Issue #631, opened 2017-04-18
Add parser support DEFAULT and REFERENCES constraint


vittvolt:
We are working on the database constraint and there are two major issues currently:

1. When inserting a tuple with the number of values given smaller than the column count, the database gives an insertion failure and exits.

2. REFERENCES keyword (and maybe ON DELETE, ON UPDATE) for foreign key does not seem to be supported.

Issue #626, opened 2017-04-13
Add TEXT type


apavlo:
The "UDF" team (@NasrinJaleel93, @prashasthip, +1) need to store the function source in the new catalog (#592).  Postgres catalog table for UDFs ([pg_proc](https://www.postgresql.org/docs/current/static/catalog-pg-proc.html)) stores this field as a `TEXT` type (`prosrc`), which we currently do not support in Peloton.

AFAIK, the `TEXT` type is just a `VARCHAR` without a length. The Postgres docs says that the maxlength of `TEXT` is 1GB. So we could just make it an alias to `VARCHAR(1000000000)`. We would need to add support for this in the parser and the convert it in the planner/optimizer.

Issue #625, opened 2017-04-13
Add support for CHECK in CREATE TABLE


apavlo:
The "Constraints" team (@vittvolt, @xinzhang1234, +1) have added support for `CHECK` constraints in the `DataTable`. They currently can only test this if they programmatically create the tables. This is because `CREATE TABLE (...)` does not support the `CHECK` keyword in column definitions.

The steps to complete this task are as follows:

1. Add support for extracting the `CHECK` metadata in the `PostgresParser`. 
2. Modify the planner/optimizer to add this metadata to the `CreatePlan` object. We will need to convert this metadata into the necessary `AbstractExpression` tree.
3. Add support in the `CreateExecutor` to install this constraint into the `DataTable`.
4. Make sure that we store this constraint information in the new catalog (#592)

Here is the simple table that we want to support:

```sql
CREATE TABLE foo (
  id INT CHECK (id > 0)
);
```

luochenUmich:
@vittvolt @xinzhang1234  I am modifying parser to support this. However, I need to know where to store this parsed data. I looked at your Constraint PR and didn't see you guys changing the create_statement. CreateStatement is the place where this parsed data should be stored. Probably you guys should add a vector<Constraint> member variable in ColumnDefinition. Let me know when you guys add that. Once that's merged, I can start adding parser support.

Issue #578, opened 2017-03-23
Insert crashes when the number of inputs does not match the column count


seojungmin:
drop table if exists foo2;
create table foo2(id1 integer, id2 varchar);
insert into foo2 values(1, 'hi');
insert into foo2 values(2, 'ha');
insert into foo2 values(3);
select * from foo2;

This script crashes with

peloton: /home/jungmin/w-tony/peloton-w-codegen/src/planner/insert_plan.cpp:65: peloton::planner::InsertPlan::InsertPlan(peloton::storage::DataTable*, std::vector<char*>*, std::vector<std::vector<peloton::expression::AbstractExpression*>*>*): Assertion `(values->size() == table_schema->GetColumnCount())' failed.

PR #633, opened 2017-04-21
Parser Fix

hzxa21:
This PR contains several fixes and adds support for several features in the parser according to the open issues:

- Support DEFAULT, REFERENCE and CHECK constraints in CREATE TABLE  (#631, #625 )
- Fix assertion failure in insert executor (#631, #578)
- Make TEXT type an alias to VARCHAR(1000000000) (#626)
Now Peloton can support the following CREATE TABLE statement in the parser:
``` sql
CREATE TABLE TEST (
  a int DEFAULT 1+2,
  b int REFERENCES table2 (bb) ON UPDATE CASCADE,
  c varchar(32) REFERENCES table3 (cc) MATCH FULL ON DELETE SET NULL,
  d int CHECK (d+1 > 0),
  e TEXT
);
```
- Support parsing FunctionExpression `SELECT add(1,a) FROM TEST WHERE FUN(b) > 2`

**_For the constraint team_**  (@vittvolt, @xinzhang1234, +1):
Some members are added in the create_statement.h to store the parsing result for the constraints and you may want to refactor the class if needed. For INSERT with DEFAULT keyword, e.g. `INSERT INTO TEST VALUES (DEFAULT, 1, 2, 3)`, the parser just put nullptr in the value list, you should check it to deal with this case. See https://github.com/hzxa21/peloton/commit/579806548300ca31495b5076845b190b9cf0e8c2#diff-557f49deefef560d7e7b9c3c85259f5fR980

**_For the UDF team_**  (@NasrinJaleel93, @prashasthip, +1):
Now the parser treats UDF and built-in functions as the same and all the information are available in FunctionExpression. https://github.com/cmu-db/peloton/blob/master/src/include/expression/function_expression.h

### DIFF
--- a/src/parser/postgresparser.cpp
+++ b/src/parser/postgresparser.cpp
@@ -973,6 +973,12 @@ PostgresParser::ValueListsTransform(List* root) {
         cur_result->push_back(ParamRefTransform((ParamRef*)expr));
       else if (expr->type == T_A_Const)
         cur_result->push_back(ConstTransform((A_Const*)expr));
+      else if (expr->type == T_SetToDefault){
+        // TODO handle default type
+        // add corresponding expression for
+        // default to cur_result
+        cur_result->push_back(nullptr); 
+      }
     }
     result->push_back(cur_result);
   }
--- a/src/parser/postgresparser.cpp
+++ b/src/parser/postgresparser.cpp
@@ -460,8 +460,9 @@ expression::AbstractExpression* PostgresParser::FuncCallTransform(
         result = new expression::AggregateExpression(
             agg_fun_type, root->agg_distinct, child);
       } else {
-      throw NotImplementedException(
-        "Aggregation over multiple columns not supported yet...\n");
+        throw NotImplementedException(
+            "Aggregation over multiple columns not supported yet...\n");
+      }
     }
   }
   return result;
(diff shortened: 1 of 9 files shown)


src/parser/postgresparser.cpp
```diff
@@ -405,41 +405,65 @@ expression::AbstractExpression* PostgresParser::ConstTransform(A_Const* root) {
...
-        throw NotImplementedException(
-          "Function within Aggregate is not supported yet\n");
+        children.push_back(ColumnRefTransform(reinterpret_cast<ColumnRef *>(expr_node)));
+      } else if (expr_node->type == T_FuncCall){
+        children.push_back(FuncCallTransform(reinterpret_cast<FuncCall*>(expr_node)));
       }
+      else {
+        std::ostringstream oss;
+        oss<<"Type "<<expr_node->type<<"is not supported in function call";
+        throw NotImplementedException(oss.str());
```
haojin2: I guess you can combine the above three lines into "throw NotImplementedException(StringUtil::Format("Type %d is not supported in function call\n", expr_node->type));"

hzxa21: True. Thanks.

apavlo:
@hzxa21 Rebase and then I can merge.

PR #635, opened 2017-04-21
remove dependency on logging component

yingjunwu:
Copy executor should not dependent on logging. 

@malin1993ml Please help review this PR. Thanks!


malin1993ml:
Looks good, thanks! This is a small PR and should be isolated to what everybody else is doing. So I'll just merge this one.

PR #640, opened 2017-04-22
Optimizer Updates

hzxa21:
This PR contains a batch of updates for the new optimizer before our team starts to support multi-way join and includes features that are required for other teams.

### Single Table Stuff
- Support projection on top of aggregation function:
`SELECT MIN(b + c) * SUM(a - 2) + b * c FROM test GROUP BY b,c`
- Support SELECT without FROM (UDF team needs this) and MOD operator:
`SELECT 18 % 4, 2 / 3 * 8 - 1, NOT 1>3`
- Support sort and hash aggregator for GROUP BY queries.
- Support both index scan and sequential scan for table scan.
- Avoid additional HashPlan if DISTINCT and GROUP BY have the same columns.
- Add PropertyLimit and distinguish normal group expression with enforced group expression to avoid unnecessary enumerations.
- Refactor optimizer_sql_test.
- Modify seq_sacn_executor to support nested loop join on top of sequential scans.

### Multi-way Join
- Support extracting join predicates from WHERE clause.
- Support multi-way inner join implemented as nested loop join or hash join.
```
SELECT test.a, test.b, test1.b, test2.c 
FROM test2 
JOIN test ON test.b = test2.b
JOIN test1 ON test2.c = test1.c;

SELECT test.a, test1.b FROM test, test1
WHERE test.a = test1.a AND test1.b = 22;
```
- Support Cartesian products and joining same physical table with different alias.
`SELECT A.b, B.b FROM test1 as A, test1 as B`


src/executor/seq_scan_executor.cpp
```diff
@@ -219,5 +224,73 @@ bool SeqScanExecutor::DExecute() {
...
+  std::vector<oid_t> predicate_column_ids;
+
+  PL_ASSERT(column_ids.size() <= column_ids_.size());
+
+  // columns_ids is the column id
+  // in the join executor, should
+  // convert to the column id in the
+  // seq scan executor
+  for (auto column_id : column_ids) {
+    predicate_column_ids.push_back(column_ids[column_id]);
```
Michael-Tieying-Zhang: Do you mean` column_ids_` rather than `column_ids` ?

src/executor/seq_scan_executor.cpp
```diff
@@ -219,5 +224,73 @@ bool SeqScanExecutor::DExecute() {
...
+  }
+
+  expression::AbstractExpression *new_predicate =
+      values.size() != 0 ? ColumnsValuesToExpr(predicate_column_ids, values, 0)
+                         : nullptr;
+
+  // Currently a hack that prevent memory leak
+  // we should eventually make prediate_ a unique_ptr
+  new_predicate_.reset(new_predicate);
+  predicate_ = new_predicate;
```
Michael-Tieying-Zhang: What if `predicate_` is not nullptr ? In that case, `new_predicate` should be added to `predicate_` as a conjunct expression.

src/executor/seq_scan_executor.cpp
```diff
@@ -219,5 +224,73 @@ bool SeqScanExecutor::DExecute() {
...
+      ->SetValueType(target_table_->GetSchema()->GetColumn(column_id).GetType());
+  reinterpret_cast<expression::TupleValueExpression *>(lexpr)
+      ->SetValueIdx(column_id);
+
+  expression::AbstractExpression *rexpr =
+      new expression::ConstantValueExpression(value);
+
+  expression::AbstractExpression *root_expr =
+      new expression::ComparisonExpression(
+          ExpressionType::COMPARE_EQUAL, lexpr, rexpr);
```
Michael-Tieying-Zhang: How about the other expression types besides `COMPARE_EQUAL`?

src/executor/seq_scan_executor.cpp
chenboy: Yes. I will fix it.

src/executor/seq_scan_executor.cpp
chenboy: I Will fix it.

src/executor/seq_scan_executor.cpp
chenboy: I think the current interface for UpdatePredicate only seems to support equality comparison

hzxa21:
join_test has been modified to test nested loop join with sequential scan.

src/executor/seq_scan_executor.cpp
```diff
@@ -219,5 +226,81 @@ bool SeqScanExecutor::DExecute() {
...
+  reinterpret_cast<expression::TupleValueExpression *>(lexpr)->SetValueType(
+      target_table_->GetSchema()->GetColumn(column_id).GetType());
+  reinterpret_cast<expression::TupleValueExpression *>(lexpr)
+      ->SetValueIdx(column_id);
+
+  expression::AbstractExpression *rexpr =
+      new expression::ConstantValueExpression(value);
+
+  expression::AbstractExpression *root_expr =
+      new expression::ComparisonExpression(ExpressionType::COMPARE_EQUAL, lexpr,
```
malin1993ml: Why it's always a COMPARE_EQUAL here? Can't we have greater than or less than in the predicate?

src/include/optimizer/group.h
```diff
@@ -34,7 +34,7 @@ class Group {
   Group(GroupID id);
   void add_item(Operator op);
 
-  void AddExpression(std::shared_ptr<GroupExpression> expr);
+  void AddExpression(std::shared_ptr<GroupExpression> expr, bool enforced);
```
malin1993ml: Please explain the meaning of the arguments.

src/include/optimizer/operators.h
```diff
@@ -141,17 +142,6 @@ class LogicalGroupBy : public OperatorNode<LogicalGroupBy> {
 };
```
malin1993ml: I know this is not written this time, but do we still want to have the group by columns and having expression stored in the LogicalGroupBy operator instead of store them as physical properties? Have we discussed that? If so, I seem forget the reason for this...

src/include/optimizer/child_property_generator.h
```diff
@@ -52,6 +54,11 @@ class ChildPropertyGenerator : public OperatorVisitor {
   void Visit(const PhysicalAggregate *) override;
 
  private:
+  /***** Helper functions *****/
```
malin1993ml: Actually what do those two functions do? Please add some comments.

src/executor/seq_scan_executor.cpp
chenboy: I think this function should only be called in a nested loop join executor before scanning the right table to find the matched tuples for a tuple from the left table. Because the current design of the nested loop join plan node only support joins with equal sign, and the index scan version of UpdatePredicate also only support joins with equal sign, so I think we can only support COMPARE_EQUAL here.

src/include/optimizer/operators.h
chenboy: We have came to the conclusion that if we need to enumerate the implementation for a component, e.g. hash group by v.s. sort group by, index scan v.s. seq scan, then we may want treat it as a operator. If the implementation is fixed for a component, e.g. order by, then we will store it as a property.

src/executor/seq_scan_executor.cpp
domino-succ: Index scan only supports equal, that true. But there is TODO in index scan code to support other types. If you can add other types, it would be great.

src/include/optimizer/operators.h
luochenUmich: I think Lin is trying to say can we take out group by columns and having expression and store them in a property like PropertyGroupby. Currently, all the properties are describing some characteristics of output of the operator. I feel like PropertyGroupby is very specific to group by operator and not general enough to be taken out.

src/executor/seq_scan_executor.cpp
luochenUmich: LLVM will eventually replace that code. I am wondering if it's worth the time to fix it right now.

src/include/optimizer/operators.h
malin1993ml: I see. Yes, that's what I meant. And that explanation makes sense.

src/executor/seq_scan_executor.cpp
malin1993ml: I see. I think it's okay to leave it for now, but it would be good to at least leave some explanation or notice here.

src/include/optimizer/child_property_generator.h
luochenUmich: SeqScan and IndexScan have very similar behavior when deriving the child properties. Thus, we just take them out as a helper function to avoid duplicate code. The newest commit should have all the comments

src/optimizer/cost_and_stats_calculator.cpp
```diff
@@ -32,9 +32,21 @@ void CostAndStatsCalculator::CalculateCostAndStats(
...
+}
+
+void CostAndStatsCalculator::Visit(const PhysicalSeqScan *) {
+  // TODO: Replace with more accurate cost
+  output_stats_.reset(new Stats(nullptr));
+  output_cost_ = 2;
+};
+void CostAndStatsCalculator::Visit(const PhysicalIndexScan *) {
+  // TODO: Simple cost function
+  // indexSearchable ? SeqScan : IndexScan
```
malin1993ml: Yes. I think we should check whether the query is really index searchable or not here. If not, then the cost of this operator should be higher than SeqScan, instead of "1". Otherwise we'll choose the wrong plan even with simple benchmarking. Is it hard to add that logic into this PR?

src/optimizer/util.cpp
```diff
@@ -11,9 +11,190 @@
...
+      expr_types.push_back(expression->GetExpressionType());
+
+      // Potential memory leak in line 664:
+      // 24 bytes in 1 blocks are definitely lost in loss record 1 of 3
+      // malloc (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
+      // peloton::do_allocation(unsigned long, bool) (allocator.cpp:27)
+      // operator new(unsigned long) (allocator.cpp:40)
+      // peloton::type::IntegerValue::Copy() const (numeric_value.cpp:1288)
+      // peloton::expression::ConstantValueExpression::GetValue() const
+      // (constant_value_expression.h:40)
```
malin1993ml: Do this potential memory leak still exists?

malin1993ml:
@chenboy Sounds good. But I think the plan we decided yesterday is that we wait for the cogen PR #612 to be merged in first, and then you rebase on top of that, then I'll merge your PR.

@pmenon Is #612 (actually I suggested that we close that one and reopen a new one because of duplicated commit messages) going to be merged soon?

luochenUmich:
@malin1993ml We only need that PR(#612) when we are adding hash join. For this PR, we don't have any code for join so it should be safe to merge this PR first.

malin1993ml:
@pmenon what do you think? Do you think we can merge this one first or we should wait for the fsanatize PR and cogen PR to be merged?

pmenon:
The fsanitize PR (#646) has been merged. I just create a clean codegen PR (#648). If it passes the build, it can be merged. Then this PR should be good to go.

malin1993ml:
@pmenon Sounds good. Just one thing, the master brunch failed after #646 has been merged:
https://travis-ci.org/cmu-db/peloton

Do you know why?

pmenon:
Hmmm, the JDBC test is failing in the BWTree with a heap-buffer overflow. This was happening on GCC 4.9, but it looks like it's also in GCC 5. 

The failure is spurious. @wangziqi2013 is aware of the problem, but I'll take a second look while he finishes his submission.

hzxa21:
@pmenon I have just added the AggregationTranslator in codegen and modify AggregateExpression to deal with queries like "SELECT SUM(a) + b * MAX(c) from test GROUP BY b". Could you help to do a quick review?

Also, I notice that when the child plan of Group By is Projection, e.g. `SELECT b+c FROM test GROUP BY b+c`, there is SEGFAULT in the codegen. Should we use the interpreted version instead for these queries?

pmenon:
I'm not sure what the purpose of AggregateExpression is over OperatorExpression. But, it depends on how you're constructing the aggregation plan node. In your first example, you should have two aggregates (for the sum and max), and an output projection that computes the final value per group.

The second query *should* be working, where does it error out?

hzxa21:
For the first example, the optimizer already fed two aggregations to the aggregate_plan. But I think we cannot simply use the TargetList in the output projection without adding AggregateTranlator to derive the value produced by the group by aggregator, because the projection will traverse the expression tree until it reaches TupleValueExpression and we need to stop traversing down if we meet the AggregateExpression and get back the value according to the AttributeInfo as what TupleValueExpression does. I am not very sure about the execution logics in the llvm engine.
So feel free to correct me if I am wrong.

For the second example, here is the debug log and the backtrace:
```
2017-04-30 12:24:43 [test/sql/optimizer_sql_test.cpp:78:TestUtil] DEBUG - Running Query "SELECT b + c FROM test GROUP BY b + c"
2017-04-30 12:24:43 [test/sql/testing_sql_util.cpp:81:ExecuteSQLQueryWithOptimizer] DEBUG - Plan Type: AGGREGATE_V2
  AggregatePlan
NumChildren: 1
Plan Type: PROJECTION
    Target List: < DEST_column_id , expression >
Dest Col id: 0
Expr: 
        Expression ::  expression type = OPERATOR_PLUS, value type = INTEGER,

DirectMap List: < NEW_col_id , <tuple_idx , OLD_col_id>  > 
<1, <0, 1> >
<2, <0, 0> >

NumChildren: 1
Plan Type: SEQSCAN
      SeqScan
NumChildren: 0

2017-04-30 12:24:43 [src/planner/abstract_scan_plan.cpp:52:PerformBinding] DEBUG - Attribute 'test.b' (0) binds to AI 0x8b0530
2017-04-30 12:24:43 [src/planner/abstract_scan_plan.cpp:52:PerformBinding] DEBUG - Attribute 'test.c' (1) binds to AI 0x8b0560
2017-04-30 12:24:43 [src/planner/project_info.cpp:124:PerformRebinding] DEBUG - Direct: Dest col 1 is bound to AI 0x8b0560
2017-04-30 12:24:43 [src/planner/project_info.cpp:124:PerformRebinding] DEBUG - Direct: Dest col 2 is bound to AI 0x8b0530
2017-04-30 12:24:43 [src/expression/tuple_value_expression.cpp:88:PerformBinding] DEBUG - TVE Column ID 0.0 binds to AI 0x8b0530 (b)
2017-04-30 12:24:43 [src/expression/tuple_value_expression.cpp:88:PerformBinding] DEBUG - TVE Column ID 0.1 binds to AI 0x8b0560 (c)
2017-04-30 12:24:43 [src/planner/project_info.cpp:137:PerformRebinding] DEBUG - Target: Dest col 0 is bound to AI 0x880b78
2017-04-30 12:24:43 [src/planner/aggregate_plan.cpp:31:PerformBinding] DEBUG - Grouping col 0 binds to AI 0x880b78
2017-04-30 12:24:43 [src/expression/tuple_value_expression.cpp:88:PerformBinding] DEBUG - TVE Column ID 0.2 binds to AI 0x8b0530 (b)
2017-04-30 12:24:43 [src/expression/tuple_value_expression.cpp:88:PerformBinding] DEBUG - TVE Column ID 0.1 binds to AI 0x8b0560 (c)
2017-04-30 12:24:43 [src/planner/project_info.cpp:137:PerformRebinding] DEBUG - Target: Dest col 0 is bound to AI 0x7ddab8
2017-04-30 12:24:43 [src/codegen/hash_group_by_translator.cpp:38:HashGroupByTranslator] DEBUG - Constructing HashGroupByTranslator ...
2017-04-30 12:24:43 [src/codegen/table_scan_translator.cpp:35:TableScanTranslator] DEBUG - Constructing TableScanTranslator ...
2017-04-30 12:24:43 [src/codegen/table_scan_translator.cpp:57:TableScanTranslator] DEBUG - Finished constructing TableScanTranslator ...
2017-04-30 12:24:43 [src/codegen/compilation_context.cpp:78:GeneratePlan] DEBUG - Main pipeline: HashGroupBy
2017-04-30 12:24:43 [src/codegen/projection_translator.cpp:34:Produce] DEBUG - Projection starting to produce tuples ...
2017-04-30 12:24:43 [src/codegen/table_scan_translator.cpp:65:Produce] DEBUG - TableScan on [33554542] starting to produce tuples ...
2017-04-30 12:24:43 [src/codegen/table_scan_translator.cpp:162:SetupRowBatch] DEBUG - Adding attribute 'test.b' (0x8b0530) into row batch
2017-04-30 12:24:43 [src/codegen/table_scan_translator.cpp:162:SetupRowBatch] DEBUG - Adding attribute 'test.c' (0x8b0560) into row batch
2017-04-30 12:24:43 [src/codegen/projection_translator.cpp:41:Consume] DEBUG - Projection consuming results ...
2017-04-30 12:24:43 [src/codegen/hash_group_by_translator.cpp:214:Consume] DEBUG - HashGroupBy consuming results ...

Program received signal SIGSEGV, Segmentation fault.
0x00007ffff6b377e0 in llvm::Value::getContext() const () from /home/patrick/CLionProjects/peloton/build/lib/libpeloton-d.so.0.0.5
(gdb) bt
#0  0x00007ffff6b377e0 in llvm::Value::getContext() const () from /home/patrick/CLionProjects/peloton/build/lib/libpeloton-d.so.0.0.5
#1  0x00007ffff6afbb76 in llvm::StoreInst::StoreInst(llvm::Value*, llvm::Value*, bool, unsigned int, llvm::AtomicOrdering, llvm::SynchronizationScope, llvm::Instruction*) ()
   from /home/patrick/CLionProjects/peloton/build/lib/libpeloton-d.so.0.0.5
#2  0x00007ffff6afbc23 in llvm::StoreInst::StoreInst(llvm::Value*, llvm::Value*, bool, unsigned int, llvm::Instruction*) () from /home/patrick/CLionProjects/peloton/build/lib/libpeloton-d.so.0.0.5
#3  0x00007ffff66a268e in llvm::IRBuilder<true, llvm::ConstantFolder, llvm::IRBuilderDefaultInserter<true> >::CreateStore (this=0x8b0090, Val=0x0, Ptr=0x7df938, isVolatile=false)
    at /usr/include/llvm-3.7/llvm/IR/IRBuilder.h:986
#4  0x00007ffff66a1731 in peloton::codegen::CompactStorage::StoreValues (this=0x7e9bc8, codegen=..., ptr=0x7df7f0, to_store=std::vector of length 1, capacity 1 = {...})
    at /home/patrick/CLionProjects/peloton/src/codegen/compact_storage.cpp:98
#5  0x00007ffff66c5c90 in peloton::codegen::OAHashTable::TranslateProbing(peloton::codegen::CodeGen&, llvm::Value*, llvm::Value*, std::vector<peloton::codegen::Value, std::allocator<peloton::codegen::Value> > const&, std::function<void (llvm::Value*)>, std::function<void (llvm::Value*)>, bool, bool, bool) const (this=0x7e9bc0, codegen=..., hash_table=0x7dd368, hash=0x0, 
    key=std::vector of length 1, capacity 1 = {...}, key_found=..., key_not_found=..., process_value=true, process_only_one_value=true, create_key_if_missing=true)
    at /home/patrick/CLionProjects/peloton/src/codegen/oa_hash_table.cpp:382
#6  0x00007ffff66c62f8 in peloton::codegen::OAHashTable::ProbeOrInsert (this=0x7e9bc0, codegen=..., ht_ptr=0x7dd368, hash=0x0, key=std::vector of length 1, capacity 1 = {...}, probe_callback=..., 
    insert_callback=...) at /home/patrick/CLionProjects/peloton/src/codegen/oa_hash_table.cpp:424
#7  0x00007ffff66b68ab in peloton::codegen::HashGroupByTranslator::Consume (this=0x7e9b60, row=...) at /home/patrick/CLionProjects/peloton/src/codegen/hash_group_by_translator.cpp:243
#8  0x00007ffff66abfd9 in peloton::codegen::ConsumerContext::Consume (this=0x7fffffffc7b0, row=...) at /home/patrick/CLionProjects/peloton/src/codegen/consumer_context.cpp:59

```

pmenon:
You should be able to use an expression (OperatorExpression) in the aggregate plan's output projection target list to calculate the final value. The attribute binding will work itself out as long as the column IDs are hooked up correctly.

The latter is a bug. I can probably send you a PR that fixes it.

src/include/expression/aggregate_expression.h
```diff
@@ -100,9 +120,11 @@ class AggregateExpression : public AbstractExpression {
 
  protected:
   AggregateExpression(const AggregateExpression& other)
-      : AbstractExpression(other) {}
+      : AbstractExpression(other), value_idx_(other.value_idx_) {}
 
  private:
+  int value_idx_ = -1;
+  const planner::AttributeInfo *ai_;
```
camellyx: initialize pointer for ai_

src/optimizer/child_property_generator.cpp
```diff
@@ -89,253 +149,281 @@ void ChildPropertyGenerator::Visit(const PhysicalScan *) {
...
-        expression::ExpressionUtil::GetTupleValueExprs(columns, expr.get());
+  if (!has_star) {
+    // All base columns in join condition needs to be generated by the child
+    expression::ExpressionUtil::GetTupleValueExprs(child_cols, join_cond);
+
+    // Decide which base columns should be provided by which child based on
+    // the table alias.
+    PL_ASSERT(child_groups_.size() == 2);
+    auto &left_table_alias = child_groups_[0]->GetTableAliases();
+    auto &right_table_alias = child_groups_[1]->GetTableAliases();
```
ChTimTsubasa: unused variable ‘right_table_alias’ generated when compiled in Release mode.

src/optimizer/optimizer.cpp
```diff
@@ -196,9 +222,18 @@ unique_ptr<planner::AbstractPlan> Optimizer::ChooseBestPlan(
...
-        child_groups[i], required_input_props[i], &child_expr_map));
-    children_expr_map.push_back(move(child_expr_map));
+    auto child_plan = ChooseBestPlan(child_groups[i], required_input_props[i],
+                                     &child_expr_map);
+    if (child_plan) {
+      children_plans.push_back(move(child_plan));
+      children_expr_map.push_back(move(child_expr_map));
+    }
+  }
+  LOG_INFO("Children map: %ld", children_expr_map.size());
```
ChTimTsubasa: This may need to be LOG_TRACE.

src/optimizer/optimizer.cpp
```diff
@@ -196,9 +222,18 @@ unique_ptr<planner::AbstractPlan> Optimizer::ChooseBestPlan(
...
+  for (auto map : children_expr_map) {
+    for (auto iter : map) {
+      LOG_INFO("%d", iter.second);
```
ChTimTsubasa: This may need to be LOG_TRACE too

apavlo: Fixed.

src/optimizer/optimizer.cpp
apavlo: Fixed.

src/optimizer/child_property_generator.cpp
apavlo: Fixed.

apavlo:
Ugh I keep screwing this up by using Github's editor. I will fix this.

apavlo:
@hzxa21 @ChTimTsubasa Can you try my fix on your local machine? It works for me but it's failing on Travis.

Also, for some reason it doesn't link on Ubuntu 14.04 with gcc4.8 *but* it does work if it's built using Jenkins.

hzxa21:
@apavlo Works on my machine and all tests are passed.

ChTimTsubasa:
It passes on my machine too.

apavlo:
OK I think that I will just disable this one failing query for now so that we can proceed with merging. I will write an issue up so that we revisit it later.

apavlo:
@hzxa21 I think something more sinister is going on here. Different queries are failing at different times. Let's see if this passes.

apavlo:
This was merged by hand.

Issue #638, opened 2017-04-21
Implement Sqrt operator


hzxa21:
"|/" operator to compute the sqrt root of the value is not supported for now although we already have the Sqrt functions implemented for different value types.

``` sql
postgres=# create table test(a decimal(5,2));
create table test(a decimal(5,2)) 0
postgres=# insert into test values(2.5);
insert into test values(2.5) 1
postgres=# select sqrt(a) from test;
function sqrt not found.
postgres=# select |/ a from test;
No ExpressionType conversion from string '|/'
```

See
https://www.postgresql.org/docs/9.4/static/functions-math.html
https://github.com/cmu-db/peloton/blob/7a3483e137f454602bda918d3932604f1f34bf77/src/type/decimal_type.cpp#L182
https://github.com/cmu-db/peloton/blob/7a3483e137f454602bda918d3932604f1f34bf77/src/type/integer_type.cpp#L161

mrmiywj:
Hi, is this issue available for in-coming CMU student? I am a MS student in INI 17 fall and find this project interesting. Could I pick this up?

apavlo:
@mrmiywj Yes! Go for it. Let me know if you need more guidance.

Issue #632, opened 2017-04-19
Move built-in SQL functions to pg_proc


apavlo:
Right now we have a dirty hash map in the catalog that we use to map function names to `ExpressionType`.

https://github.com/cmu-db/peloton/blob/master/src/include/type/types.h#L247

Then the binder stuffs the pointer to the function inside of a `FunctionExpression`. This is a big hack.

https://youtu.be/btWsg4sCRJI?t=12s

We need to do what postgres does and move all of the built-in functions into the `pg_proc` catalog table.

1. Create [pg_language](https://www.postgresql.org/docs/current/static/catalog-pg-language.html) table and add an "internal" type.
2. Add the ability in the binder to look-up a function name from `pg_proc`. Check whether the language type (`prolang`) is "internal". If it is, then stuff the pointer inside of the `FunctionExpression`.
3. Add code in the catalog boostrap methods to populate the `pg_proc` table with all of the built-in functions.
4. Remove all the hardcoded built-in function names from `ExpressionType`.
5. We should also move all of the built-in functions out of the "expression" directory:
https://github.com/cmu-db/peloton/blob/master/src/include/expression/date_functions.h
https://github.com/cmu-db/peloton/blob/master/src/include/expression/string_functions.h

apavlo:
This has been fixed for a while. Closing.

PR #641, opened 2017-04-25
add support for integer and decimal but it failed when adding them me…

mrmiywj:
Add Sqrt support for integer and decimal type.
But here are some errors.

### DIFF
--- a/src/catalog/catalog.cpp
+++ b/src/catalog/catalog.cpp
@@ -22,6 +22,7 @@
 #include "common/macros.h"
 #include "expression/date_functions.h"
 #include "expression/string_functions.h"
+#include "expression/decimal_functions.h"
 #include "index/index_factory.h"
 #include "util/string_util.h"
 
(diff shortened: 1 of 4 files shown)


mrmiywj:
Hi, Prof.Pavlo @apavlo 
I just write a proto implementation of this feature #638. But there were some problems.

The code works, but only when one of Integer or Decimal type code added. The reason is because that the `FunctionName-FunctionData` defined [here](https://github.com/cmu-db/peloton/blob/master/src/include/catalog/catalog.h#L209) is a 1 by 1 map. Since the integer type and decimal type's sqrt function has the same name. So it will cause duplicated keys.

I have two potential solutions:
1. Change map key from `function_name` to `function_name + argument_type`.
2. Change map<Function_name, function_data> to map<Function_name, map<argument_type, function_data>. 

But I found it maybe hard to get function's argument type in `function_expression.h` when executing functions. Thus we cannot use `argument_type` as the key to find function_data.

Could you give some advice on this issue?

mrmiywj:
BTW, this is just a prototype implement for discussion. Stuffs like tests will be added later.

apavlo:
@mrmiywj Thanks for looking into this. See #632. We need to move all of the built-in functions (like sqrt) into the catalog. Then we can add the argument types and then do proper matching on them at runtime. If you can get this working for integers just for now, and then leave comments about what we need to do fix it later. We can merge what you have for now and then work on fixing #632 this summer.

If you are definitely coming to CMU this fall, send me an email so I know who you are.

mrmiywj:
Implementation of sqrt function for decimal type.

Problems to fix:
On the base of current ad-hoc hack of function name to function data map, the DB does not allow two functions for two different types with same name. So we need #632 to clear the dirty hack.

Issue #644, opened 2017-04-26
DEFAULT keyword causes Seg Fault and other feature requests


vittvolt:
Hi,

I'm in the constraint team and here are some problems we found:

- (**URGENT**)When trying to integrate with the newly added parser feature in #633 , I found that the query parsing gives a seg fault for basic case like CREATE TABLE test (id INT PRIMARY KEY, a INT DEFAULT 1). It passed the postgresparser_test because only the parse tree was built. During execution of the postgresparser it gives a seg fault some where around line 582 in postgresparser.cpp during a reinterpret_cast. 

- This one is not urgent, but it stalls us from creating proper constraints for multi columns. As mentioned in #625 , some updates have been made in create_statement.h. Yet I observe in postgresparser.cpp that when a table is created, each constraint is associated with a single column (a column bool attributes indicating whether it is part of primary key / unique). When executing the create statement the executor checks the attributes of each column to create the primary key constraint. This does not work for constraints that span multiple columns. Maybe adding a vector in create_statement.h with included column numbers for each constraint is a good way for us to handle, but I'm not sure what is a suggested solution.

Some more observations (again not urgent issue):
- Just out of curiosity In postgresparser.cpp from line 834 to line 860 in the if else clause, the code snippet is never touched during execution, and I wonder why it seems to handle primary key and foreign key constraints.

Please make any comments if you think I have missed something, thanks!

hzxa21:
The first problem is solved in #645.
As for the second problem, I agree what you have proposed. The current layout for the create_statement.h is not a good design. You guys should refactor that as said in the comment of #625 and then we can add parser support for that.

F.Y.I
Table constraints are handled here: https://github.com/cmu-db/peloton/blob/master/src/parser/postgresparser.cpp#L827
Column constraints are handled here: https://github.com/cmu-db/peloton/blob/master/src/parser/postgresparser.cpp#L752

vittvolt:
@luochenUmich @hzxa21  The seg fault still appears after executing 

**CREATE TABLE test (id INT PRIMARY KEY, a INT DEFAULT 10);
INSERT INTO test VALUES (2, DEFAULT);**

Would it be possible that you guys can add it to the postgresparser unit test ?

vittvolt:
@hzxa21

poojanilangekar:
Fixed in #645

PR #645, opened 2017-04-27
Minor fix for DEFAULT constraint and refactor postgresparser

hzxa21:
This PR adds the missing part for parsing constant DEFAULT constraint #644 and refactor postgresparser by adding ExprTransform for all expression nodes for better modularity.

### DIFF
--- a/src/parser/postgresparser.cpp
+++ b/src/parser/postgresparser.cpp
@@ -288,36 +288,23 @@ parser::GroupByDescription* PostgresParser::GroupByTransform(List* group,
   result->columns = new std::vector<expression::AbstractExpression*>();
   for (auto cell = group->head; cell != nullptr; cell = cell->next) {
     Node* temp = reinterpret_cast<Node*>(cell->data.ptr_value);
-    switch (temp->type) {
-      case T_ColumnRef:
-        result->columns->push_back(ColumnRefTransform((ColumnRef*)temp));
-        break;
-      case T_A_Expr:
-        result->columns->push_back(AExprTransform((A_Expr*)temp));
-        break;
-      default: {
-        throw NotImplementedException(StringUtil::Format(
-          "Group By type %d not supported...", temp->type));
-      }
+    try {
+      result->columns->push_back(ExprTransform(temp));
+    }
+    catch(NotImplementedException e) {
+      throw NotImplementedException(
+          StringUtil::Format("Exception thrown in group by expr:\n%s", e.what()));
     }
   }
 
   // having clauses not implemented yet, depends on AExprTransform
   if (having != nullptr) {
-    switch (having->type) {
-      case T_A_Expr: {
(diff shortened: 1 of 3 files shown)


src/parser/postgresparser.cpp
```diff
@@ -288,36 +288,23 @@ parser::GroupByDescription* PostgresParser::GroupByTransform(List* group,
   result->columns = new std::vector<expression::AbstractExpression*>();
   for (auto cell = group->head; cell != nullptr; cell = cell->next) {
     Node* temp = reinterpret_cast<Node*>(cell->data.ptr_value);
-    switch (temp->type) {
-      case T_ColumnRef:
```
luochenUmich: Here, group by columns should only be either a tuple value expression or complex expressions like a + b. In your new code, other expressions will also be parsed and passed to the optimizer. Then, it becomes the optimizer's responsibility to throw the errors. But originally, the errors will come out of parser. Same problems for other places.

src/parser/postgresparser.cpp
```diff
@@ -288,36 +288,23 @@ parser::GroupByDescription* PostgresParser::GroupByTransform(List* group,
...
-      default: {
-        delete result;
-        throw NotImplementedException(StringUtil::Format(
-          "HAVING of type %d not supported yet...", having->type));
-      }
+    try {
+      result->having = ExprTransform(having);
+    }
+    catch(NotImplementedException e) {
+      throw NotImplementedException(
```
luochenUmich: Forget delete result before throwing exception?

src/parser/postgresparser.cpp
```diff
@@ -1271,7 +1194,7 @@ parser::SQLStatementList* PostgresParser::ParseSQLString(
   }
 
   // DEBUG only. Comment this out in release mode
-//  print_pg_parse_tree(result.tree);
+  print_pg_parse_tree(result.tree);
```
luochenUmich: Don't forget to comment this out

src/parser/postgresparser.cpp
hzxa21: You are right. GroupByTransform and OrderByTransform should not call the ExprTransform directly but in other cases it should be fine. Note that group by and order by can have constant expressions to directly indicate the column positions, which we may overlook in the new optimizer.

src/parser/postgresparser.cpp
hzxa21: Good catch.

src/parser/postgresparser.cpp
hzxa21: By the way, postgres documentation says that any expression is valid in group by column as long as it is formed from the input columns. I think only ParamExpressions is invalid in this case. See https://www.postgresql.org/docs/8.0/static/sql-select.html. I am also surprised that weird queries like `select a>1 from a group by a>1` also work in postgres. I guess the rewriter has some rules to eliminate group by.

PR #647, opened 2017-04-27
Wire Protocol Fix

hanli32:
# **What We Fixed**

- Add [`LibeventServer::StartServer()`](https://github.com/hanli32/peloton/blob/master/src/wire/libevent_server.cpp#L98) and [`LibeventServer::CloseServer()`](https://github.com/hanli32/peloton/blob/master/src/wire/libevent_server.cpp#L151) to control server's start and terminate explicitly.

- Add [`LibeventMasterThread::CloseConnection()`](https://github.com/hanli32/peloton/blob/master/src/wire/libevent_thread.cpp#L150) so that master thread can break [`event_base_loop`](https://github.com/hanli32/peloton/blob/master/src/wire/libevent_thread.cpp#L73) in worker thread. Therefore libevent server can exit normally.

- Improve destructors in libevent server, thread, socket to fix memory leak problem when closing server.

- Fix all statements are executed in single statement transaction bug from libpqxx test.

- The packet manager now can handle parse messages without parameter types.

# **Testing**
Currently 2 unit tests are added to test the wire protocol
- SimpleQueryTest (https://github.com/hanli32/peloton/blob/master/test/wire/simple_query_test.cpp)
- PrepareStatementTest (https://github.com/hanli32/peloton/blob/master/test/wire/prepare_stmt_test.cpp)


apavlo:
Rebase this and then we can merge it. Nice job.

src/include/wire/libevent_thread.h
```diff
@@ -64,14 +69,29 @@ class LibeventThread {
...
+  void SetThreadIsClosed(bool is_closed) { this->is_closed = is_closed; }
+
+  int GetThreadSockFd() { return sock_fd; }
+
+  void SetThreadSockFd(int fd) { this->sock_fd = fd; }
 };
 
 class LibeventWorkerThread : public LibeventThread {
- private:
+ public:
```
apavlo: Why was this made to public?

src/wire/libevent_callbacks.cpp
```diff
@@ -211,5 +213,36 @@ void StateMachine(LibeventSocket *conn) {
   }
 }
 
+/**
+ * Stop signal handling
+ */
+void Signal_Callback(UNUSED_ATTRIBUTE evutil_socket_t fd,
```
apavlo: Instead of polluting the global namespace, can you make these static functions inside of a utility class?

hanli32:
@apavlo Hi Andy, we have already tweaked the code according to your reviews. Could you help review our modifications and then we will rebase it. There are many conflicts need to resolve so we want to make sure all changes have been approved before the rebase. Thanks.

haojin2:
Hi, would you guys please take a look at the NetworkMessageType in types.h? I found that ERROR_RESPONSE and EXECUTE_COMMAND cast to the same character, which is a bit confusing when I'm debugging the WorkerPool PR. Thanks and have a great day!

hanli32:
@haojin2  Hi, the ErrorResponse packet is sent from server to the client and the ExecuteCommand packet is sent from the client to the server so it is fine for these two packets to have the same char tag.
Here is a great [slide](https://www.pgcon.org/2014/schedule/attachments/330_postgres-for-the-wire.pdf) which describes PostgreSQL wire protocol packets.

haojin2:
That actually does have some problem, right now if I change the order of the two definitions "EXECUTE_COMMAND = 'E'" and "ERROR_RESPONSE = 'E',", I get diffferent msg_type for some packet.

hanli32:
It seems very weird to me, could you give me a specific example?

haojin2:
I met this problem during the load of oltp benchmark and one of the packets should be EXECUTE_COMMAND sent from the benchmark loader, but when I'm debugging I see a ERROR_RESPONSE for the msg_type of the packet.

hanli32:
I see. But I think it should be fine since EXECUTE_COMMAND and ERROR_RESPONSE have the same values this packet can either be an Execute message or an Error response. The [packet manager](https://github.com/cmu-db/peloton/blob/master/src/wire/packet_manager.cpp#L907) can only process EXECUTE_COMMAND so the packet will be treated as an Execute message.

haojin2:
I understand that it may not interfere with the work of packet manager for the time being, but it still causes confusion during debugging, I wonder if it is possible for us to change one of the char to 'e' so that we can distinguish them easily?

apavlo:
@haojin2 I think it has to be 'E'. That's defined in the Postgres wire protocol. It's not something that we invented.

haojin2:
Okay then, I'll just treat that as EXECUTE_COMMAND.

CHUYilei:
The new failing reason `The command "./script/installation/packages.sh" failed and exited with 100 during ` is confusing. Restart the Travis build?

apavlo:
I restarted it again. Not sure why you are getting this error if you haven't rebased with the new LLVM engine.

apavlo:
I think there is something you're missing from the current master. Can you rebase?

CHUYilei:
Ok. We are rebasing. Need some time due to too many conflicts

apavlo:
I will merge when this passes.

apavlo:
@malin1993ml Are you able to merge this PR? My view still says that there is a conflict? I'm not sure whether I am having the same problem as before.

PR #649, opened 2017-04-29
Database Compression 

rohit-cmu:
## Files Added

### Other Existing Files Modified

## Strategy for Compressing

* Once a tile gets full (all slots occupied by tuples), the CompressTile is created.
* Currently we can compress SMALLINT and INTEGER, BIGINT, DECIMAL. TINYINT is already 1 byte so we do not compress it.
* We also support deduplication for VARCHAR field
* We are yet to add support for decimal values.
* This median is also stored as the metadata to later retrieve the original value

## Testing

* This test inserts 25 tuples
* Each tuple is of the form (i, i*100) where i belongs to (0,25)
* Since each tile group contains 1 tile and 10 tuples per tile, there are 3 tile groups formed.
* The first 2 tile groups are full and get compressed.
* The third tile group has 5 slots vacant and is still not full and is uncompressed.

### Compression Size Test:
* The integers get compressed to TINYINT making the tuple sizes now (1+ 1+ 8+ 8) = 18 bytes


vanryan:
I feel it is not a very good practice to include irrelevant (other people's) commits and changed files into one's own pull request... It takes a lot of energy to pick out theme-related files to review and merge conflict might occur if being merged into master...

.gitignore
```diff
@@ -313,3 +313,6 @@ third_party/gflags/
 
 # Test output file logs to ignore
 stats_log
+
+test.txt
```
vanryan: This should not be included into the pull request to the master, right... cuz it is for own specific testing use.

.travis.yml
```diff
@@ -16,6 +17,7 @@ matrix:
           addons:
               apt:
                 sources:
+                    - llvm-toolchain-precise-3.7
```
vanryan: I guess this change was from llvm codegen

CodeReview#1_README.md
```diff
@@ -0,0 +1,53 @@
+## Files Added
```
vanryan: Why is this file included in this pull request...

wangxiaoyou11:
I'm surprised by your PR size with 300+ commits. It's really hard to review your PR because there are so many duplicate commits with `cmu-db:master`. I can hardly find what you have changed among those 272 files.
I think you should `rebase` the `cmu-db:master` instead of doing a `merge`. Could you please make this PR clean first?

rohit-cmu:
@vanryan @wangxiaoyou11 We are fixing this. New Pull request should be available soon!

poojanilangekar:
@vanryan @wangxiaoyou11 

Link to the new Pull Request : https://github.com/cmu-db/peloton/pull/668

Sorry for the inconvenience!

PR #650, opened 2017-05-01
Triggers Support

nblintao:
# Major Updates
- Support execute **create** and **drop** trigger statements
  - Update data structure for statement to support trigger
  - Convert Postgres parsing tree for Peloton statement for trigger
  - Update data structure for plan to support trigger
  - Convert state to plan for trigger
- Data structures for Trigger and TriggerList and related operations
- Put trigger information into the catalog and support updates *(Not all trigger predicates are supported because of the complexity in serialization and deserialization. We currently support binary operator for common types. e.g. `NEW.income > OLD.income`, `NEW.income > 3`. This could be improved by implementing serialization and deserialization of each kind of node in expression tree.)*
- Update DataTable with TriggerList information
- Evaluate the predicates in the trigger
- Invoke UDFs and send contexts *(It takes a log of all data needed for a UDF call (function name, arguments, old/new tuples, etc.), but does not really call UDFs. One concern is that UDF is not supported in the master branch currently. Another concern is that currently UDF is mainly designed for read-only operations without SQL statements, but mostly, functions invoked by a trigger need to apply SQL statements on databases. Hope `ExecCallTriggerFunc` could be truly implemented after these problems are resolved.)*
- Trigger operations for 12 kinds of triggers (before/after, insert/delete/update, row/statement)


src/catalog/trigger_catalog.cpp
```diff
@@ -0,0 +1,302 @@
...
+                      "trigger_type          INT NOT NULL, "
+                      "fire_condition        VARCHAR, "
+                      "function_name         VARCHAR, "
+                      "function_arguments    VARCHAR, "
+                      "time_stamp            INT NOT NULL);",
+                      txn) {
+  // Add secondary index here if necessary
+  Catalog::GetInstance()->CreateIndex(
+      CATALOG_DATABASE_NAME, TRIGGER_CATALOG_NAME,
+      {"database_oid", "table_oid", "trigger_type"}, TRIGGER_CATALOG_NAME "_skey0",
```
hanli32: magic names like "_skey0","_skey1","_skey2" should have an explanation if they appear for the first time. Otherwise maintainers cannot understand  the naming schema.

src/catalog/trigger_catalog.cpp
```diff
@@ -0,0 +1,302 @@
...
+    DatabaseCatalog::GetInstance()->GetDatabaseOid(database_name, txn);
+  if (database_oid == INVALID_OID) {
+    LOG_TRACE("Cannot find database  %s!", database_name.c_str());
+    return ResultType::FAILURE;
+  }
+
+  oid_t table_oid =
+    TableCatalog::GetInstance()->GetTableOid(table_name, database_oid, txn);
+  if (table_oid == INVALID_OID) {
+    LOG_TRACE("Cannot find table %s!", table_name.c_str());
```
hanli32: Is LOG_ERROR more appropriate for failure situation?

src/catalog/trigger_catalog.cpp
```diff
@@ -0,0 +1,302 @@
...
+    LOG_INFO("size of the result tiles = %lu", result_tiles->size());
+  }
+
+  // create the trigger list
+  commands::TriggerList* newTriggerList = new commands::TriggerList();
+  if (result_tiles != nullptr) {
+    for (unsigned int i = 0; i < result_tiles->size(); i++) {
+      size_t tuple_count = (*result_tiles)[i]->GetTupleCount();
+      for (size_t j = 0; j < tuple_count; j++) {
+        LOG_INFO("trigger_name is %s", (*result_tiles)[i]->GetValue(j, 0).ToString().c_str());
```
hanli32: Is it necessary to have these LOG_INFO in final version of code? Will it have too much print out? Also, is LOG_DEBUG better?

src/catalog/trigger_catalog.cpp
```diff
@@ -0,0 +1,302 @@
...
+                      "fire_condition        VARCHAR, "
+                      "function_name         VARCHAR, "
+                      "function_arguments    VARCHAR, "
+                      "time_stamp            INT NOT NULL);",
+                      txn) {
+  // Add secondary index here if necessary
+  Catalog::GetInstance()->CreateIndex(
+      CATALOG_DATABASE_NAME, TRIGGER_CATALOG_NAME,
+      {"database_oid", "table_oid", "trigger_type"}, TRIGGER_CATALOG_NAME "_skey0",
+      false, IndexType::BWTREE, txn);
```
hanli32: Currently you use IndexType::BWTREE, but what if in the future you want to use other index types? Is it possible to make this choice more flexible?

src/executor/create_executor.cpp
```diff
@@ -83,6 +86,69 @@ bool CreateExecutor::DExecute() {
...
+    LOG_INFO("2");
+    // ask target table to update its trigger list variable
+    storage::DataTable *target_table =
+        catalog::Catalog::GetInstance()->GetTableWithName(database_name,
+                                                          table_name);
+    target_table->UpdateTriggerListFromCatalog(current_txn);
+
+
+
+    // // debug:
```
hanli32: Commented out code

src/executor/insert_executor.cpp
```diff
@@ -185,6 +218,28 @@ bool InsertExecutor::DExecute() {
                 target_table->GetTupleCount());
 
       executor_context_->num_processed += 1;  // insert one
+
+
+      // // debug code below
+      // oid_t database_oid = target_table->GetDatabaseOid();
```
CHUYilei: commented code

src/executor/create_executor.cpp
```diff
@@ -83,6 +86,69 @@ bool CreateExecutor::DExecute() {
...
+    // target_table->AddTrigger(newTrigger);
+
+
+    // durable trigger: insert the information of this trigger in the trigger catalog table
+    oid_t database_oid = catalog::DatabaseCatalog::GetInstance()->GetDatabaseOid(database_name, current_txn);
+    oid_t table_oid = catalog::TableCatalog::GetInstance()->GetTableOid(table_name, database_oid, current_txn);
+    auto time_since_epoch = std::chrono::system_clock::now().time_since_epoch();
+    auto time_stamp = std::chrono::duration_cast<std::chrono::seconds>(
+                        time_since_epoch).count();
+    LOG_INFO("1");
```
hanli32: Better to delete these debug info.

src/executor/create_executor.cpp
```diff
@@ -83,6 +86,69 @@ bool CreateExecutor::DExecute() {
...
+    LOG_INFO("enter CreateType::TRIGGER");
+    std::string database_name = node.GetDatabaseName();
+    std::string table_name = node.GetTableName();
+    std::string trigger_name = node.GetTriggerName();
+    // std::unique_ptr<catalog::Schema> schema(node.GetSchema());
+
+    commands::Trigger newTrigger(node);
+
+    // // new approach: add trigger to the data_table instance directly
+    // storage::DataTable *target_table =
```
CHUYilei: Commented code. Is this new approach for future maintainers? If so, TODO is needed.

src/executor/create_executor.cpp
```diff
@@ -83,6 +86,69 @@ bool CreateExecutor::DExecute() {
...
+    // ask target table to update its trigger list variable
+    storage::DataTable *target_table =
+        catalog::Catalog::GetInstance()->GetTableWithName(database_name,
+                                                          table_name);
+    target_table->UpdateTriggerListFromCatalog(current_txn);
+
+
+
+    // // debug:
+    // auto trigger_list = catalog::TriggerCatalog::GetInstance()->GetTriggersByType(database_oid, table_oid,
```
CHUYilei: Commented code. If used for debug, you can consider using IFDEF flag.

src/parser/postgresparser.cpp
```diff
@@ -625,6 +626,27 @@ expression::AbstractExpression* PostgresParser::WhereTransform(Node* root) {
...
+// parsenode and transfers it into Peloton AbstractExpression.
+expression::AbstractExpression* PostgresParser::WhenTransform(Node* root) {
+  if (root == nullptr) {
+    return nullptr;
+  }
+  expression::AbstractExpression* result = nullptr;
+  switch (root->type) {
+    case T_A_Expr: {
+      result = AExprTransform(reinterpret_cast<A_Expr*>(root));
+      break;
```
CHUYilei: Break after return?

src/parser/postgresparser.cpp
```diff
@@ -625,6 +626,27 @@ expression::AbstractExpression* PostgresParser::WhereTransform(Node* root) {
...
+    }
+    case T_BoolExpr: {
+      result = BoolExprTransform(reinterpret_cast<BoolExpr*>(root));
+      break;
```
CHUYilei: Break after return?

src/catalog/trigger_catalog.cpp
```diff
@@ -0,0 +1,302 @@
...
+
+  // create the trigger list
+  commands::TriggerList* newTriggerList = new commands::TriggerList();
+  if (result_tiles != nullptr) {
+    for (unsigned int i = 0; i < result_tiles->size(); i++) {
+      size_t tuple_count = (*result_tiles)[i]->GetTupleCount();
+      for (size_t j = 0; j < tuple_count; j++) {
+        LOG_INFO("trigger_name is %s", (*result_tiles)[i]->GetValue(j, 0).ToString().c_str());
+        LOG_INFO("FIRE_CONDITION is %s", (*result_tiles)[i]->GetValue(j, 1).ToString().c_str());
+        LOG_INFO("FUNCTION_NAME is %s", (*result_tiles)[i]->GetValue(j, 2).ToString().c_str());
```
CHUYilei: Are these LOG_INFO for debug use only?

src/parser/postgresparser.cpp
nblintao: Sorry,  I didn't get it. Could you please explain more? Thanks~

src/catalog/trigger_catalog.cpp
```diff
@@ -0,0 +1,318 @@
...
+  auto val0 = type::ValueFactory::GetIntegerValue(oid_++ | TRIGGER_OID_MASK);
+  auto val1 = type::ValueFactory::GetVarcharValue(trigger_name, pool);
+  auto val2 = type::ValueFactory::GetIntegerValue(database_oid);
+  auto val3 = type::ValueFactory::GetIntegerValue(table_oid);
+  auto val4 = type::ValueFactory::GetIntegerValue(trigger_type);
+  auto val5 = type::ValueFactory::GetVarcharValue(fire_condition, pool);
+  auto val6 = type::ValueFactory::GetVarcharValue(function_name, pool);
+  auto val7 = type::ValueFactory::GetVarcharValue(function_arguments, pool);
+  auto val8 = type::ValueFactory::GetIntegerValue(time_stamp);
+
```
franklinzou: Variable names can be changed to express its represented meaning better.

src/executor/create_executor.cpp
```diff
@@ -83,6 +86,69 @@ bool CreateExecutor::DExecute() {
...
+
+    // if (current_txn->GetResult() == ResultType::SUCCESS) {
+    //   LOG_TRACE("Creating trigger succeeded!");
+    // } else if (current_txn->GetResult() == ResultType::FAILURE) {
+    //   LOG_TRACE("Creating trigger failed!");
+    // } else {
+    //   LOG_TRACE("Result is: %s", ResultTypeToString(
+    //     current_txn->GetResult()).c_str());
+    // }
+
```
franklinzou: This large portion of deleted code can be removed.

test/executor/create_test.cpp
```diff
@@ -71,5 +80,231 @@ TEST_F(CreateTests, CreatingTable) {
...
+
+  if (when) {
+    delete when;
+  }
+
+  if (stmt_list) {
+    delete stmt_list;
+  }
+
+  // TODO: test for creating a trigger without "when"
```
franklinzou: 'TODO' should either be deleted or performed before final release.

test/parser/postgresparser_test.cpp
```diff
@@ -874,6 +877,95 @@ TEST_F(PostgresParserTests, DataTypeTest) {
...
+  // timing
+  EXPECT_TRUE(TRIGGER_FOR_BEFORE(create_trigger_stmt->trigger_type));
+  EXPECT_FALSE(TRIGGER_FOR_AFTER(create_trigger_stmt->trigger_type));
+  EXPECT_FALSE(TRIGGER_FOR_INSTEAD(create_trigger_stmt->trigger_type));
+  // event
+  EXPECT_TRUE(TRIGGER_FOR_UPDATE(create_trigger_stmt->trigger_type));
+  EXPECT_FALSE(TRIGGER_FOR_INSERT(create_trigger_stmt->trigger_type));
+  EXPECT_FALSE(TRIGGER_FOR_DELETE(create_trigger_stmt->trigger_type));
+  EXPECT_FALSE(TRIGGER_FOR_TRUNCATE(create_trigger_stmt->trigger_type));
+
```
franklinzou: Comments in this function are kind of obscure and should be made more intuitive.

src/catalog/trigger_catalog.cpp
nblintao: Thanks for this suggestion. I definitely agree. But we are following the name conventions in other catalog settings like `index_catalog.cpp` and `column_catalog.cpp` etc..

src/catalog/trigger_catalog.cpp
nblintao: Thanks for pointing it out. I don't know much about the difference between DEBUG and TRACE before. After learning from https://softwareengineering.stackexchange.com/questions/279690/why-does-the-trace-level-exist-and-when-should-i-use-it-rather-than-debug, I find the current ways is interestingly correct. Trace is used for finer-gain debug. So in each small step, it uses TRACE, while at the end, it uses DEBUG. I am really glad that you point it out.

src/catalog/trigger_catalog.cpp
nblintao: Good point. But I find it is needed to specify the type of index here. FYI, there is a comment about the type of catalog index in `catalog.cpp`: `This should be hash index rather than tree index?? (but postgres use btree!!)`

src/parser/postgresparser.cpp
nblintao: Sorry, I didn't get it. Could you please explain more? Thanks~

test/executor/create_test.cpp
nblintao: Thanks. I am adding a new test case for this.

nblintao: Fixed

src/catalog/trigger_catalog.cpp
nblintao: I agree. Fixed. Thx!

src/executor/insert_executor.cpp
huangminchn: Thanks for your suggestion. Dead codes have been removed.

src/executor/create_executor.cpp
huangminchn: Actually the approach mentioned in the comment was not right. Please check our correct approach in our latest codes. Useless comments have been removed.

src/catalog/trigger_catalog.cpp
huangminchn: Thanks for pointing it out. We are following the naming style of other catalogs. "val0" means the value for the first column.

src/executor/create_executor.cpp
huangminchn: Thanks for your suggestion. Deleted code have been removed.

test/parser/postgresparser_test.cpp
nblintao: Fixed. Thanks!

nblintao:
LGTM. We are ready to merge:) @apavlo
Such a wonderful experience, thank you all!

src/catalog/trigger_catalog.cpp
apavlo: I think that this is fine for now.

src/catalog/trigger_catalog.cpp
apavlo: Yes, these should be LOG_DEBUG

nblintao: Thanks. This has been fixed in earlier commits.

apavlo:
This PR is great. We should try to merge this in as soon as possible.

nblintao:
@apavlo Thanks!

apavlo:
@nblintao Can you rebase?

nblintao:
@apavlo Done. Thanks!

nblintao:
@apavlo Rebased again successfully

PR #653, opened 2017-05-05
Multi thread support for SCAN (LLVM)

tq5124:
The main focus of review should be `Scan` and `HashJoin`.

Related files:

* `barrier.cpp`, provide barrier for multi threads. This object is shared between threads.
* `buffering_consurer.cpp`, modify the output vector to a lock free array.
* `compilation_context.cpp`, instead creating single plan function, we create a inner plan function that do all the previous work and a new plan function which submit inner plan to a thread pool.
* `hash_join_translator.cpp`, instead of using a single shared hash table, each thread use their own hash table and merge together at the end.
* `multi_thread_context.cpp`, each thread owns their own multi thread context, which include all the information the thread need.
* `query_thread_pool.cpp`, a thread pool designed for plan function. 
* `table.cpp`, split the table using multi thread context. Each thread is only responsible for their own part of the table.
* `transaction_runtime.cpp`, remove read check because it calls function outside of codeine and is not thread safe.
* `oa_hash_table.cpp`, the hash table that is used in hash join.


phisiart:
Hi!

To rebase, you could do this (make sure you create a backup copy before doing this!!!):

In your branch (`multi`), find the hash of **the commit right before your first commit**. That commit is "Synced master", the hash is "696e5d68d4f902738e9b76ce83f81d7ce3d733f4".

Make sure you have the `cmudb:master` branch on your machine.

Replay all your commits on top of master:

```
git rebase -p --onto master 696e5d68d4f902738e9b76ce83f81d7ce3d733f4 multi
```

Now `multi` is rebased on top of `cmudb:master`.

src/codegen/barrier.cpp
```diff
@@ -0,0 +1,39 @@
...
+void Barrier::InitInstance(Barrier *ins, uint64_t n_workers)
+{
+    ins->SetBarrier(new boost::barrier(n_workers));
+    ins->SetWorkerCount(n_workers);
+    ins->InitGlobalHashTableMergeLock();
+}
+
+void Barrier::MasterWait()
+{
+    while(n_workers_ > 0);
```
MatrixCW: might better considering using thread sleep/yield instead of spinning over variables

src/codegen/buffering_consumer.cpp
```diff
@@ -0,0 +1,196 @@
...
+  BufferingState *buffer_state = reinterpret_cast<BufferingState *>(state);
+  buffer_state->output->Append(WrappedTuple(vals, num_vals));
+}
+
+// Get a proxy to BufferingConsumer::BufferTuple(...)
+llvm::Function *BufferingConsumer::_BufferTupleProxy::GetFunction(
+    CodeGen &codegen) {
+  const std::string &fn_name =
+#ifdef __APPLE__
+      "_ZN7peloton7codegen17BufferingConsumer11BufferTupleEPcPNS_4type5ValueEj";
```
MatrixCW: if peloton does not support OSX, then better leave out this mingled function name instead of use the same name as Linux, or give a simple string to explicit mark unsupported platform and let the application fail.

src/codegen/transaction_runtime.cpp
```diff
@@ -0,0 +1,66 @@
...
+  for (uint32_t i = tid_start; i < tid_end; i++) {
+    // Perform the visibility check
+    auto visibility = txn_manager.IsVisible(&txn, tile_group_header, i);
+
+    // Update the output position
+    selection_vector[out_idx] = i;
+    out_idx += (visibility == VisibilityType::OK);
+  }
+
+//  uint32_t tile_group_idx = tile_group.GetTileGroupId();
```
MatrixCW: these commented out code should be deleted

src/codegen/oa_hash_table_proxy.cpp
```diff
@@ -0,0 +1,275 @@
...
+}
+
+//===----------------------------------------------------------------------===//
+// INIT PROXY
+//===----------------------------------------------------------------------===//
+
+const std::string &OAHashTableProxy::_Init::GetFunctionName() {
+  static const std::string kInitFnName =
+#ifdef __APPLE__
+      "_ZN7peloton7codegen5utils11OAHashTable4InitEmmm";
```
MatrixCW: same as the one in buffering_coosumer.cpp, if peloton does not support OSX, then no need for such macro, similar issue in subsequent lines of this file.

src/codegen/compilation_context.cpp
```diff
@@ -0,0 +1,298 @@
...
+    codegen_.CallFunc(MultiThreadContextProxy::InitInstanceFunction(codegen_), {multi_thread_context, thread_id, thread_count, barrier});
+    codegen_.CallFunc(QueryThreadPoolProxy::GetSubmitQueryTaskFunction(codegen_, &runtime_state), {query_thread_pool, runtime_state_ptr, multi_thread_context, inner_func});
+
+    // Move to next thread id in loop.
+    thread_id = codegen_->CreateAdd(thread_id, codegen_.Const64(1));
+    loop.LoopEnd(codegen_->CreateICmpULT(thread_id, thread_count), {thread_id});
+  }
+
+  // Wait for all worker threads to finish
+  codegen_.CallFunc(BarrierProxy::GetMasterWaitFunction(codegen_),{barrier});
```
phisiart: Here's the expensive call to `MasterWait()`.

src/codegen/barrier.cpp
phisiart: Agreed. It seems that in compilation_context.cpp, you are calling `MasterWait()` from the main thread. Notice that this is an expensive function. I would consider using a semaphore or a condition variable.

src/include/codegen/barrier.h
```diff
@@ -0,0 +1,80 @@
...
+
+    inline void SetWorkerCount(uint64_t n_workers)
+    {
+        n_workers_ = n_workers;
+    }
+
+    inline void MergeToGlobalHashTable(utils::OAHashTable *global_ht, utils::OAHashTable *local_ht)
+    {
+        bool expect = false;
+        while (!global_hash_table_merge_lock_.compare_exchange_strong(expect, true))
```
phisiart: An uninteresting detail: You could use `atomic::exchange()` here.

src/include/codegen/barrier.h
```diff
@@ -0,0 +1,80 @@
...
+#include "codegen/codegen.h"
+#include "codegen/utils/oa_hash_table.h"
+
+namespace peloton {
+namespace codegen {
+
+class Barrier
+{
+
+public:
```
phisiart: I suddenly found that the code style is not consistent with other files. Don't forget to re-format it when you have time later!

src/codegen/barrier.cpp
```diff
@@ -0,0 +1,39 @@
...
+//===----------------------------------------------------------------------===//
+
+#include "common/logger.h"
+#include "codegen/barrier.h"
+#include "codegen/utils/oa_hash_table.h"
+
+namespace peloton {
+namespace codegen {
+
+void Barrier::InitInstance(Barrier *ins, uint64_t n_workers)
```
phisiart: Interesting C++ detail: Note that you create a `Barrier` on the stack (on LLVM level) by first allocate memory on the stack, and then call this `InitInstance()` function; and delete a `Barrier` by calling `Destroy()`. You might want to change these 2 functions into using placement new/delete, then you could write normal constructor/destructor.

src/codegen/barrier.cpp
```diff
@@ -0,0 +1,39 @@
...
+{
+    ins->SetBarrier(new boost::barrier(n_workers));
```
avahuang1: Did you tried valgrind? Will this cause a memory leak? Will it be better if you use a smart ptr?

src/include/codegen/barrier.h
```diff
@@ -0,0 +1,79 @@
...
+    {
+        n_workers_ = n_workers;
+    }
+
+    inline void MergeToGlobalHashTable(utils::OAHashTable *global_ht, utils::OAHashTable *local_ht)
+    {
+        bool expect = false;
+        while (!global_hash_table_merge_lock_.compare_exchange_strong(expect, true))
+        {
+          expect = false;
```
bechob: compare_exchange_strong() won't change the value of expect, there's no need to reset the value of expect here.

src/include/codegen/barrier.h
```diff
@@ -0,0 +1,79 @@
...
+        }
+        global_ht->Merge(local_ht);
+        global_hash_table_merge_lock_.store(false);
+    }
```
bechob: global_hash_table_merge_lock_ works as a spinning lock here and can be inefficient if contention is high. Can it be implemented by calculating the offset first and merge at the same time like we learned from parallel join algorithm lecture?

src/include/codegen/barrier.h
avahuang1: Agree. This whole expect and reset thing is trivial.

src/include/codegen/barrier.h
avahuang1: > It actually will....

bechob: Sorry for the mistake!

src/codegen/barrier.cpp
avahuang1: I'm so concerned about all the pointers flying around. From my point of view, "instantiate", "set", and "destroy" are all very dangerous - you might forget to destroy, or set barrier after the barrier is set. This causes trouble in debugging, too. But again, I believe you have your own reason in designing all these functions.

src/codegen/buffering_consumer.cpp
```diff
@@ -173,5 +183,15 @@ void BufferingConsumer::ConsumeResult(ConsumerContext &ctx,
   codegen.CallFunc(_BufferTupleProxy::GetFunction(codegen), args);
 }
 
+const std::vector<WrappedTuple> &BufferingConsumer::GetOutputTuples() {
+  size_t result_size = tuples_.GetSize();
+  tuples_vector_.reserve(result_size);
+  tuples_vector_.clear();
+  for (size_t i = 0; i < result_size; ++i) {
+    tuples_vector_.push_back(tuples_.Find(i));
```
bechob: I'm concerning about the use of LockFreeArray container here since the lock-free array implementation uses only the atomic offset, it does not prevent other threads deleting items from it after you call GetSize(), therefore the Find() call can get invalid read. Better use the FindValid() call because that it checks whether the read is valid for you and it will return the invalid value after it finds no valid items left. However, it can also cause duplicated reads because the result_size you use may larger than the actual one.

src/include/codegen/barrier.h
```diff
@@ -0,0 +1,79 @@
...
+
+private:
+    Barrier(): bar_(nullptr) {}
+
+    inline void InitGlobalHashTableMergeLock()
+    {
+      global_hash_table_merge_lock_.store(false);
+    }
+
+    boost::barrier *bar_;
```
bechob: I don't understand why booset::barrier* bar and n_workers_ are both here because either of it can function as a barrier. Just like there are two wait functions MasterWait and BarrierWait here with separate implementation. Can you explain more on the usage here? Otherwise, using this barrier library can be error-prone.

tq5124:
Going to Create a new PR because rebase is done after PR is created, which cause files change looks mess

PR #656, opened 2017-05-05
Extending Query Compilation

phisiart:
Extending LLVM Query Compilation to support INSERT, UPDATE, and DELETE.

## Caching
There's a LRU cache for compiled query, and a custom physical plan comparator in order to build the map.

## Parameterization
There are 2 kinds of parameterization: MemSQL-style and (partial) Postgres-style parameterizations.
- MemSQL-style Parameterization is to parameterize all constant values.
- Postgres-style is the normal prepared statements in SQL. Since peloton's parameter expressions are not typed, we cannot fully support this.

## DELETE
Supported. For each tuple passed by the scan, we call a wrapper function which performs the deletion.

## INSERT
Supported both raw tuple insertion and copying from another table.
- For raw tuple insertion, the raw tuple is put into a contiguous memory buffer and read from the LLVM side.
- For copying, for each tuple passed by the scan, we materialize the tuple in LLVM, and call a wrapper function which performs the insertion.

## UPDATE
Supported. For each tuple passed by the scan, we transform the tuple in LLVM, and call a wrapper function which inserts the new version (or updates the tuple in-place).

See [this](https://github.com/phisiart/peloton-p3/wiki) for some development documents.


src/codegen/compilation_context.cpp
```diff
@@ -141,7 +186,7 @@ llvm::Function *CompilationContext::GenerateInitFunction() {
       init_fn_name,
       codegen_.VoidType(),
       {{"runtimeState", runtime_state.FinalizeType(codegen_)->getPointerTo()}}};
-
+        llvm::Value *GetInt8ParamPtr();
```
tq5124: Redundant line?

src/codegen/delete/delete_translator.cpp
```diff
@@ -0,0 +1,198 @@
...
+ *
+ * This simply treats one row at a time, since we delete tuples one at a time.
+ *
+ * @param context The parent consumers.
+ * @param batch A reference of the code-gen'ed "current batch" value.
+ */
+void DeleteTranslator::Consume(ConsumerContext &context,
+                               RowBatch &batch) const {
+  batch.Iterate(context.GetCodeGen(),
+                [&](RowBatch::Row &row) { this->Consume(context, row); });
```
tq5124: Can do like hash_join_translator [here](https://github.com/pmenon/peloton-1/blob/codegen-v1/src/codegen/hash_join_translator.cpp#L153), use OperatorTranslator#Consume.

src/codegen/insert/insert_scan_translator.cpp
```diff
@@ -0,0 +1,151 @@
...
+//  RuntimeState::StateID tuple_ptr_state_id = runtime_state.RegisterState(
+//      "insert_tuple_ptr",
+//      codegen.Int8Type()->getPointerTo()
+//  );
+//  RuntimeState::StateID tuple_data_ptr_state_id = runtime_state.RegisterState(
+//      "insert_tuple_data_ptr",
+//      codegen.Int8Type()->getPointerTo()
+//  );
+//  runtime_state.LoadStatePtr(codegen, tuple_ptr_state_id);
+//  runtime_state.LoadStatePtr(codegen, tuple_data_ptr_state_id);
```
tq5124: Please remove comments.

src/codegen/insert/insert_scan_translator.cpp
```diff
@@ -0,0 +1,151 @@
...
+
+  // the child of delete executor will be a scan. call produce function
```
tq5124: This is insert translator right?

src/codegen/insert/insert_tuples_translator.cpp
```diff
@@ -0,0 +1,99 @@
...
+                        codegen.Const32(table->GetOid())});
+
+  llvm::Value *char_ptr = codegen.CallFunc(
+          PrimitiveValueProxy::_GetVarcharVal::GetFunction(codegen),
+          {context.GetCharPtrParamPtr(), codegen.Const64(this->tuples_offset_)});
+  llvm::Value *char_len = codegen.CallFunc(
+          PrimitiveValueProxy::_GetVarcharLen::GetFunction(codegen),
+          {context.GetCharLenParamPtr(), codegen.Const64(this->tuples_offset_)});
+
+  codegen.CallFunc(
```
lihao98722: Function calls on the critical path. Will llvm code generation be a more efficient way? By translating the interpreted function you are calling to llvm? Otherwise we lose the benefits of code generation.

avahuang1: Indeed. But for acquiring lock and inserting structures like "tuple" rather than primitive types, code specialization is extremely hard.

src/codegen/delete/delete_translator.cpp
```diff
@@ -0,0 +1,198 @@
...
+  if (table_ptr == nullptr) {
+    storage::DataTable *table = delete_plan_.GetTable();
+
+    table_ptr = codegen.CallFunc(
+        CatalogProxy::_GetTableWithOid::GetFunction(codegen),
+        {GetCatalogPtr(), codegen.Const32(table->GetDatabaseOid()),
+         codegen.Const32(table->GetOid())});
+  }
+
+  codegen.CallFunc(_DeleteWrapper::GetFunction(codegen),
```
phisiart: It is true that the inner loop of `DELETE` is just a giant function. However, the reasoning here is that `DELETE`'s job is to talk to MVCC & Index stuff, which requires function calls anyway (unless we try to optimize MVCC & Index with LLVM also!), so we just wrap them up into a giant function. We actually have performed a timing test: by removing this giant function call (so `DELETE` doesn't actually delete anything) we reduce the running time of `DELETE` from ~260ms to ~230ms. There must be something else that's slowing this down, which we might have to do things like profiling to figure out.

src/codegen/insert/insert_helpers.cpp
```diff
@@ -0,0 +1,86 @@
...
+      &concurrency::TransactionManagerFactory::GetInstance();
+
+  // Insert tuple into the table.
+  // Tuple will be copied.
+  ItemPointer *index_entry_ptr = nullptr;
+  ItemPointer location = table->InsertTuple(tuple, txn, &index_entry_ptr);
+
+  // It is possible that some concurrent transactions have inserted the same
+  // tuple. In this case, abort the transaction.
+  if (location.block == INVALID_OID) {
```
homelesscmu: I think they already check for this in [InsertTuple](https://github.com/cmu-db/peloton/blob/master/src/storage/data_table.cpp#L277) function ..

src/codegen/insert/insert_scan_translator.cpp
```diff
@@ -0,0 +1,151 @@
...
+#include "codegen/data_table_proxy.h"
+#include "planner/abstract_scan_plan.h"
+#include "storage/data_table.h"
+#include "codegen/insert/insert_scan_translator.h"
+#include "codegen/raw_tuple/raw_tuple_ref.h"
+
+namespace peloton {
+namespace codegen {
+
+InsertScanTranslator::InsertScanTranslator(const planner::InsertPlan &insert_plan,
```
homelesscmu: I understand this insert translator assumes it has a child of scan .... but the class name is a bit confusing at first glance

src/codegen/insert/insert_scan_translator.cpp
```diff
@@ -0,0 +1,151 @@
...
+  auto &codegen = this->GetCodeGen();
+
+  // Retrieve all the attribute infos.
+  std::vector<const planner::AttributeInfo *> ais;
+  auto scan = static_cast<const planner::AbstractScan *>(
+      this->insert_plan_.GetChild(0));
+  scan->GetAttributes(ais);
+
+  // Prepare to materialize the tuple.
+  RawTupleRef raw_tuple_ref(
```
homelesscmu: question: are you reusing the same tuple but keep updating its data?

test/codegen/insert_translator_test.cpp
```diff
@@ -0,0 +1,170 @@
...
+  // We collect the results of the query into an in-memory buffer
+  codegen::BufferingConsumer buffer{{0, 1}, context};
+
+  // COMPILE and execute
+  CompileAndExecute(*insert_plan, buffer,
+                    reinterpret_cast<char *>(buffer.GetState()));
+
+  size_t num_tuples = table->GetTupleCount();
+  LOG_DEBUG("After Insert: #tuples in table = %zu", num_tuples);
+  ASSERT_EQ(num_tuples, 1);
```
homelesscmu: maybe you should also check if the tuple's data looks right?

src/codegen/primitive_value_proxy.cpp
```diff
@@ -0,0 +1,325 @@
...
+//
+// primitive_value_proxy.cpp
+//
+// Identification: src/codegen/primitive_value_proxy.cpp
+//
+// Copyright (c) 2015-17, Carnegie Mellon University Database Group
+//
+//===----------------------------------------------------------------------===//
+
+#include "codegen/primitive_value_proxy.h"
```
homelesscmu: is it possible to refactor it to be less verbose?

src/codegen/projection_translator.cpp
```diff
@@ -63,7 +66,7 @@ void ProjectionTranslator::PrepareProjection(
   }
 }
 
-void ProjectionTranslator::AddNonTrivialAttributes(
+/* static */ void ProjectionTranslator::AddNonTrivialAttributes(
```
homelesscmu: please clean up dead code...

test/codegen/benchmark_delete_translator_test.cpp
```diff
@@ -0,0 +1,474 @@
...
+  timer.Stop();
+  LOG_INFO("executor empty table finished in: %.2f ms\n", timer.GetDuration());
+  EXPECT_EQ(99999, getCurrentTableSize());
+
+  reloadTable();
+}
+
+TEST_F(DeleteTranslatorTest, DeleteHalfRecords) {
+  //
+  // DELETE FROM table where a % 20 == 0; half records will be deleted
```
homelesscmu: To vary the selectivity, maybe you can use a >= (?) so the predicate tree is simpler (though LLVM bakes the predicates in anyway)

src/codegen/insert/insert_scan_translator.cpp
phisiart: Yes, this is re-using the same rendezvous tuple to reduce re-allocation.

test/codegen/benchmark_delete_translator_test.cpp
phisiart: Actually, this is intended because we want tuples that pass this predicate to evenly spread across the table.

src/codegen/primitive_value_proxy.cpp
phisiart: You're right. We could actually turn all these into direct LLVM code. Will put that into TODO list.

src/codegen/insert/insert_helpers.cpp
phisiart: It looks like InsertTuple() doesn't report transaction failure, so this is actually needed.

test/codegen/insert_translator_test.cpp
phisiart: You're right.

src/codegen/delete/delete_translator.cpp
```diff
@@ -0,0 +1,198 @@
+//===----------------------------------------------------------------------===//
+//
+//                         Peloton
+//
+// transaction_manager.h
```
phisiart: Doing a clean up today ...

src/codegen/insert/insert_scan_translator.cpp
phisiart: Doing a clean up today ...

src/codegen/projection_translator.cpp
phisiart: This is actually a comment, since C++ wouldn't allow us to write static in the function definition here.

phisiart:
Just done a rebase.

src/codegen/case_translator.cpp
```diff
@@ -58,8 +58,6 @@ codegen::Value CaseTranslator::DeriveValue(CodeGen &codegen,
     }
     when.EndIf(merge_bb);
   }
-  // Jump to the merging block from the internal If merging block
-  codegen->CreateBr(merge_bb);
```
seojungmin: Add these lines back in, or please explain why these are removed.

src/codegen/case_translator.cpp
```diff
@@ -68,6 +66,9 @@ codegen::Value CaseTranslator::DeriveValue(CodeGen &codegen,
       Type::GetNullValue(codegen, ret.GetType());
   branch_vals.emplace_back(default_ret, codegen->GetInsertBlock());
 
+  // Jump to the merging block from the internal If merging block
+  codegen->CreateBr(merge_bb);
```
seojungmin: These should be removed and the previous lines should be added back in.

src/codegen/data_table_proxy.cpp
```diff
@@ -38,17 +39,40 @@ llvm::Type* DataTableProxy::GetType(CodeGen& codegen) {
...
+  // The function hasn't been registered, let's do it now
+  llvm::Type* table_type = DataTableProxy::GetType(codegen);
+  llvm::FunctionType* fn_type = llvm::FunctionType::get(
+      codegen.Int64Type(), {table_type->getPointerTo()}, false);
+  return codegen.RegisterFunction(fn_name, fn_type);
+}
+
+const std::string& DataTableProxy::_GetSchema::GetFunctionName() {
   // TODO: FIX ME
   static const std::string kGetTileGroupCount =
```
seojungmin: It would be nicer if the variable name is chosen correctly.

src/codegen/data_table_proxy.cpp
```diff
@@ -38,17 +39,40 @@ llvm::Type* DataTableProxy::GetType(CodeGen& codegen) {
 }
 
 const std::string& DataTableProxy::_GetTileGroupCount::GetFunctionName() {
+  static const std::string kGetTileGroupCount =
+      "_ZNK7peloton7storage9DataTable17GetTileGroupCountEv";
```
seojungmin: ifdef __APPLE__ is missing.  Although this macro is not used at the moment, it would be a good practice to keep it as all the codegen code currently have these.  Consistency is much better if we work together on a large source code base.

src/codegen/insert/insert_helpers.cpp
```diff
@@ -0,0 +1,84 @@
...
+  }
+
+  txn_mgr->PerformInsert(txn, location, index_entry_ptr);
+
+  return true;
+}
+
+void InsertHelpers::InsertValue(concurrency::Transaction *txn,
+                                storage::DataTable *table,
+                                char *value, size_t num_tuples) {
```
seojungmin: _num_tuples_ seems a total size of the tuples to me.  Can we change the variable name?

src/codegen/insert/insert_helpers_proxy.cpp
```diff
@@ -0,0 +1,163 @@
...
+          codegen.Int8Type()->getPointerTo(),                 // tuple
+      },
+      false);
+
+  return codegen.RegisterFunction(fn_name, fn_type);
+}
+
+const std::string &InsertHelpersProxy::_InsertValue::GetFunctionName() {
+  static const std::string kInsertValueFnName =
+      "_ZN7peloton7codegen13InsertHelpers11InsertValueEPNS_11concurrency11TransactionEPNS_7storage9DataTableEPcm";
```
seojungmin: Formatting is required

src/codegen/insert/insert_helpers_proxy.cpp
```diff
@@ -0,0 +1,163 @@
...
+      {
+          SchemaProxy::GetType(codegen)->getPointerTo()
+      },
+      false);
+
+  return codegen.RegisterFunction(fn_name, fn_type);
+}
+
+const std::string &InsertHelpersProxy::_GetTupleData::GetFunctionName() {
+  static const std::string kInsertValueFnName =
```
seojungmin: We would need these variable name changes throughout the code.

src/codegen/primitive_value_proxy.cpp
```diff
@@ -0,0 +1,325 @@
...
+}
+
+//===----------------------------------------------------------------------===//
+// PEEK TIMESTAMP
+//===----------------------------------------------------------------------===//
+
+const std::string &PrimitiveValueProxy::_GetTimestamp::GetFunctionName() {
+  static const std::string kGetTimestampFnName =
+#ifdef __APPLE__
+        "_ZN7peloton7codegen13ValuesRuntime13outputVarcharEPcjS2_j";
```
seojungmin: The name mangling does not look correct.

src/include/storage/abstract_table.h
```diff
@@ -96,7 +96,7 @@ class AbstractTable : public Printable {
 
   void SetSchema(catalog::Schema *given_schema) { schema = given_schema; }
 
-  catalog::Schema *GetSchema() const { return (schema); }
+  catalog::Schema* __attribute__ ((noinline)) __attribute__((optimize("O0"))) GetSchema() const { return (schema); }
```
seojungmin: Please leave a comment why this was necessary.

src/codegen/case_translator.cpp
phisiart: The original implementation will cause an LLVM compilation error if we change `DerivedValue()` for constant VARCHAR (I just added a dummy function call, but I guess a load instruction might trigger the same error). I made a branch containing only that change as a demo: https://github.com/phisiart/peloton-p3/tree/case. The failing test case is "case_sql_test".

I'm not sure whether the modification here is correct - though it passed tests. Could you take a look at the error above and figure out what's the correct way of handling this?

seojungmin: After all, the error in the given branch is caused by the codegen.CallPrintf() function calls you added in the DeriveValue().  This is not dummy, but it actually builds and generates a function call.   While doing so, these CallPrintf() code generation seems to complicate the optimization of the case LLVM IR code, which removes a bitcast instruction which should be there.

I recommend to remove the codegen.CallPrintf() call in the constant translator, rather than changing this code.  Meanwhile, I will have a look why this CallPrintf() removes the bitcast operation.

src/codegen/primitive_value_proxy.cpp
avahuang1: The system does not support OSX anyway..

seojungmin: I understand your argument, and appreciate for bringing it up.  I think it would be good to think that we write code with supporting OSX in mind. Although it is broken now, we supported OSX last year, and we are still expecting to fix it in very near future. 

The best solution to this is to apply the correct name mangling for OSX, but for the time being we choose to apply the same mangled name as a NON-APPLE one, throughout the codegen code. I think it would be a good practice to keep this strategy. Just be kind to the engineers who might be working on it in near future.  You could choose the best solution, though.

The more important problem here is the inconsistencies in handling these name mangling's throughout this PR: some have no APPLE macro, some others have APPLE macro with the same name as non-apple's, and the others have APPLE macro with some other names.  I think this is not what you want to do as an engineer in this kind of big project.

phisiart:
I'll do make other changes. Then I'll address all the reviews.

apavlo:
After a discussion with @pmenon and @seojungmin, we decided that it is going to be too much work to merge this PR. We may use it as a reference in the future. Thanks for all your work!

PR #662, opened 2017-05-07
Added data structure for unique constraints in create_statement.h

vittvolt:
## Modifications:
- This commit adds the data structure for multi-column unique constraint in create_statement.h.

- This aligns with the current foreign key constraints. It will be needed by the parser team to pass column names of each unique constraint.

- The parser could also parse a multi-column unique constraint as a ColumnDefinition and put it in the columns list, just like the foreign key constraint.

### DIFF
--- a/src/include/parser/create_statement.h
+++ b/src/include/parser/create_statement.h
@@ -31,6 +31,8 @@ struct ColumnDefinition {
     PRIMARY,
     FOREIGN,
 
+    MULTIUNIQUE,
+
     CHAR,
     INT,
     INTEGER,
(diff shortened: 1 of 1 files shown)


apavlo:
@mengranwo @camellyx @AngLi-Leon Can you guys check this and merge it in?

mengranwo:
Sure, I'll start to look at it today

src/include/parser/create_statement.h
```diff
@@ -174,6 +181,8 @@ struct ColumnDefinition {
   std::vector<char*>* foreign_key_source = nullptr;
   std::vector<char*>* foreign_key_sink = nullptr;
 
+  std::vector<char *>* multi_unique_cols = nullptr;
```
mengranwo: Just wondering, why not use smart pointer like std::unique_ptr in the functions, so that deletion will be automatically handled.

src/include/parser/create_statement.h
```diff
@@ -174,6 +181,8 @@ struct ColumnDefinition {
...
+
   char* foreign_key_table_name = nullptr;
   FKConstrActionType foreign_key_delete_action;
```
camellyx: foreign_key_delete_action, foreign_key_update_action, foreign_key_match_type are not initialized in the constructor.

apavlo:
Merged by hand in #749

Issue #663, opened 2017-05-08
select clause cause crushes 


lishuomountain:
After installing peloton, I follow the instructios in Terminal wiki (https://github.com/cmu-db/peloton/wiki/Terminal). When I execute "select * from foo;", the peloton crashes with the following info:

postgres=# select id from foo;
server closed the connection unexpectedly
	This probably means the server terminated abnormally
	before or while processing the request.
The connection to the server was lost. Attempting reset: Failed.

and 

peloton: /home/lsh/Downloads/llvm-3.7.0/lib/IR/Instructions.cpp:957: llvm::LoadInst::LoadInst(llvm::Type*, llvm::Value*, const char*, bool, llvm::Instruction*): Assertion `Ty == cast<PointerType>(Ptr->getType())->getElementType()' failed.
Aborted (core dumped)

There is no llvm 3.7 in the repositories of ubuntu 14.04, so I installed llvm 3.7.0 from source files.

Ubuntu 14.04, gcc 4.8.5

apavlo:
@lishuomountain Can you also provide the schema and data that you were testing?

You can install llvm-3.7 from packages on 14.04. We will update the installation instructions.

pmenon:
I just tested this on master with GCC 4.8.5, 4.9.4 and 5.4.1 on Ubuntu 14.04 and it all works.

@lishuomountain please use the llvm-3.7 packages for ubuntu 14.04. Also, attach the output from the peloton process.

lishuomountain:
It's OK. Thank you!

PR #664, opened 2017-05-08
Fixed packages script install custom repo for llvm-3.7 on Ubuntu 14.04

apavlo:
This is for #663 

https://www.youtube.com/watch?v=XMe0PTYDnXw

### DIFF
--- a/script/installation/packages.sh
+++ b/script/installation/packages.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/bash -x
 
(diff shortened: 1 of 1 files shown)


crd477:
Is there a reason for the dependency on LLVM 3.7 specifically?  As you've already clearly noticed, it adds a minor complication to the build system and binary versions of that package are not supplied in the official upstream repositories for several supported platforms.  I expect that the situation will get even worse over time.

PR #668, opened 2017-05-11
Database Compression

poojanilangekar:
## Files Added

## Other Existing Files Modified

* [transaction_level_gc_manager.cpp](https://github.com/rohit-cmu/compression-peloton/blob/master/src/gc/transaction_level_gc_manager.cpp)

## Slides 
* [Click Here](https://www.dropbox.com/s/4z5hocerp77kq9k/Final%20Presentation%20-%20Database%20Compression.pdf?dl=0)

## Strategy for Compressing

* Once a tile gets full (all slots occupied by tuples), the CompressTile can be created using the CompressTable function
* Currently we can compress SMALLINT and INTEGER and BIGINT. TINYINT is already 1 byte so we do not compress it.

## Modifications done to GC
* While adding a tuple to the recycle map,the garbage collector ensures that the tuple does not belong to a compressed tile group.
* If it belongs to a compressed tile group, it skips this tuple to ensure that it is never returned when the ReturnFreeSlot function is called.

### Tile Compaction
* Tile Compaction can be implemented in the GC, once the GetActiveTupleCount function is fixed. The strategy would be to reset the tiles inside the tile group. The notion is that the executor would only examine a tile group if it has non zero active tuple count.

## Testing

* [VARCHAR Deduplication Test](https://github.com/rohit-cmu/compression-peloton/blob/master/test/sql/compression_dictionary_sql_test.cpp)

* This test inserts 2500 tuples (Unit Tests for Integers and Decimals are included)
* The first 2 tile groups are full (since the max capacity is 1000 tuples per tile group) and get compressed.
* The third tile group has 500 slots vacant and is still not full and is uncompressed.

### Compression Size Test:
* The integers get compressed to TINYINT making the tuple sizes now (1+ 1+ 2+ 8) = 12 bytes

### VARCHAR Reduplication Tests
* Unit tests to check reduplication by creating 5 distinct VARCHAR [foo, bar, hello, world] and duplicating entries for each of them
* The CompressTable function de-duplicates the data and stores the value as SMALLEST.


CodeReview#1_README.md
```diff
@@ -29,7 +40,7 @@ Compression Correctness Test:
 * Thus we now have compressed and uncompressed data.
 * We now perform a SELECT * on this and expect to correctly recieve the true value of the compressed data and uncompressed data.
 
-Compression Size Test:
+### Compression Size Test:
 * This test inserts 100 tuples
 * Each tuple is of the form [Integer, Integer, Decimal, Varchar]
 * Thus the tuple length is (4+ 4+ 8+ 8) = 24 bytes.
```
poojanilangekar: @rohit-cmu  Need to Remove this file

src/storage/compressed_tile.cpp
```diff
@@ -0,0 +1,588 @@
...
+      new_value = dictionary.find(word);
+    }
+    modified_raw[i] = new_value;
+  }
+
+  LOG_TRACE("number of uniq words: %d", counter);
+  if ((oid_t)counter == num_tuples) {
+    // no duplicate
+    LOG_TRACE("All words are unique");
+    modified_values.clear();
```
vanryan: It seems this is unnecessary, modified_values is not modified up to this point anyways.

src/storage/compressed_tile.cpp
```diff
@@ -0,0 +1,588 @@
...
+    return deserializedValue;
+  }
+
+  type::Type::TypeId compressed_type = GetCompressedType(column_id);
+
+  PL_ASSERT(compressed_type != type::Type::INVALID);
+
+  type::Value original = GetUncompressedValue(column_id, deserializedValue);
+
+  return original;
```
vanryan: We can just return GetUncompressedValue(column_id, deserializedValue).

src/storage/compressed_tile.cpp
```diff
@@ -0,0 +1,588 @@
...
+void CompressedTile::SetValueFast(const type::Value &value,
+                                  const oid_t tuple_offset,
+                                  const size_t column_offset,
+                                  const bool is_inlined,
+                                  const size_t column_length) {
+  oid_t column_id = GetColumnFromOffset(column_offset);
+
+  PL_ASSERT(column_id < column_count);
+
+  if (IsCompressed()) {
```
vanryan: Why not just write in one single if

src/storage/tile.cpp
```diff
@@ -204,8 +201,7 @@ Tile *Tile::CopyTile(BackendType backend_type) {
       // Copy the column over to the new tile group
       for (oid_t tuple_itr = 0; tuple_itr < allocated_tuple_count;
            tuple_itr++) {
-        type::Value val =
-            (new_tile->GetValue(tuple_itr, uninlined_col_offset));
+        type::Value val = (new_tile->GetValue(tuple_itr, uninlined_col_offset));
```
vanryan: The outer parentheses are unnecessary.

src/storage/tile.cpp
```diff
@@ -243,6 +239,8 @@ const std::string Tile::GetInfo() const {
   return os.str();
 }
 
+void Tile::CompressTile(Tile *tile) { (void)tile; }
```
AllisonWang: You can use UNUSED_ATTRIBUTE to disable the warning. `void Tile::CompressTile(UNUSED_ATTRIBUTE Tile *tile)`

src/tcop/tcop.cpp
```diff
@@ -307,7 +307,7 @@ std::shared_ptr<Statement> TrafficCop::PrepareStatement(
 
 #ifdef LOG_DEBUG_ENABLED
     if (statement->GetPlanTree().get() != nullptr) {
-      LOG_DEBUG("Statement Prepared: %s", statement->GetInfo().c_str());
+      LOG_TRACE("Statement Prepared: %s", statement->GetInfo().c_str());
```
AllisonWang: why?

src/storage/data_table.cpp
```diff
@@ -1023,6 +1024,22 @@ storage::TileGroup *DataTable::TransformTileGroup(
...
 }
 
+// Compress all the TileGroups of the DataTable.
+// Currently called manually.
+// TODO: Call it while restructuring TileGroups
+void DataTable::CompressTable() {
+  oid_t tilegroups_size = tile_groups_.GetSize();
+  for (oid_t i = 0; i < tilegroups_size; i++) {
+    storage::TileGroup *tile_group =
+        GetTileGroupById(tile_groups_.Find(i)).get();
```
AllisonWang: why not just use the smart pointer?

src/storage/data_table.cpp
```diff
@@ -1023,6 +1024,22 @@ storage::TileGroup *DataTable::TransformTileGroup(
...
+
+    if (!(tile_group->GetCompressionStatus())) {
```
AllisonWang: check `tile_group != nullptr` before dereferencing?

src/storage/data_table.cpp
```diff
@@ -171,14 +170,23 @@ ItemPointer DataTable::GetEmptyTupleSlot(const storage::Tuple *tuple) {
...
   auto free_item_pointer = gc_manager.ReturnFreeSlot(this->table_oid);
+  bool valid_slot = true;
   if (free_item_pointer.IsNull() == false) {
     // when inserting a tuple
     if (tuple != nullptr) {
       auto tile_group =
           catalog::Manager::GetInstance().GetTileGroup(free_item_pointer.block);
-      tile_group->CopyTuple(tuple, free_item_pointer.offset);
+
+      if (tile_group->GetCompressionStatus() == true) {
```
AllisonWang: `== true` is an anti-pattern. I think in this case `GetCompressionStatus()` could be changed to something more descriptive, like `isCompressed()`.

src/storage/data_table.cpp
```diff
@@ -171,14 +170,23 @@ ItemPointer DataTable::GetEmptyTupleSlot(const storage::Tuple *tuple) {
   // check if there are recycled tuple slots
   auto &gc_manager = gc::GCManagerFactory::GetInstance();
   auto free_item_pointer = gc_manager.ReturnFreeSlot(this->table_oid);
+  bool valid_slot = true;
   if (free_item_pointer.IsNull() == false) {
```
AllisonWang: This should also be changed...

src/storage/tile_group.cpp
```diff
@@ -314,6 +315,37 @@ oid_t TileGroup::InsertTupleFromCheckpoint(oid_t tuple_slot_id,
...
+  oid_t next_tuple_slot = tile_group_header->GetCurrentNextTupleSlot();
+
+  if (next_tuple_slot < num_tuple_slots) {
+    LOG_TRACE("TileGroup is currently not full. Skipping compression.");
+    return;
+  }
+
+  for (oid_t i = 0; i < num_tiles; i++) {
+    LOG_TRACE("Compressing Tile %d in Tile Group", i);
+    Tile *old_tile = tiles[i].get();
```
AllisonWang: Why not use `std::shared_ptr<Tile> old_tile`?

src/include/storage/compressed_tile.h
```diff
@@ -0,0 +1,165 @@
...
+
+ protected:
+  //===--------------------------------------------------------------------===//
+  // Data members
+  //===--------------------------------------------------------------------===//
+
+  bool is_compressed;
+  oid_t compressed_columns_count;
+  int tuple_count;
+
```
wangxiaoyou11: I would suggest to add comments for these maps, such as what the key and value are.

src/storage/tile_group.cpp
```diff
@@ -314,6 +315,37 @@ oid_t TileGroup::InsertTupleFromCheckpoint(oid_t tuple_slot_id,
...
+    std::shared_ptr<CompressedTile> new_tile(new CompressedTile(
+        backend_type, old_tile->GetHeader(), *(old_tile->GetSchema()),
+        old_tile->GetTileGroup(), old_tile->GetAllocatedTupleCount()));
+
+    new_tile->CompressTile(old_tile);
+
+    if (new_tile->IsCompressed()) {
+      tiles[i].reset();
+      tiles[i] = new_tile;
```
AllisonWang: There are multiple ways to release a shared_ptr. It can be done through going out of scope, assigning another shared_ptr or calling `reset`. Here I think calling `reset` is unnecessary in both if and else branch.

src/storage/compressed_tile.cpp
```diff
@@ -0,0 +1,588 @@
...
+
+@brief : The following function checks if the tile is compressed. If yes, return
+the uncompressed value, else return the deserializedValue.
+
+@return : Actual Value(uncompressed) of the tuple at tuple offset.
+*/
+
+type::Value CompressedTile::GetValue(const oid_t tuple_offset,
+                                     const oid_t column_id) {
+  type::Value deserializedValue = Tile::GetValue(tuple_offset, column_id);
```
AllisonWang: `deserializedValue` should be `deserialized_value` to be consistent.

src/storage/compressed_tile.cpp
```diff
@@ -0,0 +1,588 @@
...
+                               TileGroup *tile_group, int tuple_count)
+    : Tile(backend_type, tile_header, tuple_schema, tile_group, tuple_count) {
+  is_compressed = false;  // tile is currently uncompressed.
+  compressed_columns_count = 0;
+}
+
+CompressedTile::~CompressedTile() {}
+
+// CompressTile
+// @brief: This function identifies the columns that can be compressed and
```
wangxiaoyou11: Just one suggestion: Could you unify the comments format in this file? Sometimes you use `//@breif` and sometimes you use `/*@brief`.

src/storage/compressed_tile.cpp
```diff
@@ -0,0 +1,588 @@
...
+*/
+
+type::Value CompressedTile::GetMaxExponentLength(Tile *tile, oid_t column_id) {
+  oid_t num_tuples = tile->GetAllocatedTupleCount();
+  auto tile_schema = tile->GetSchema();
+  bool is_inlined = tile_schema->IsInlined(column_id);
+  size_t column_offset = tile_schema->GetOffset(column_id);
+  auto column_type = tile_schema->GetType(column_id);
+
+  type::Value decimal_value;
```
wangxiaoyou11: The definition can be put in the for loop. (line 292)

src/storage/compressed_tile.cpp
```diff
@@ -0,0 +1,588 @@
...
+  auto allocated_tuple_count = tile->GetAllocatedTupleCount();
+  std::vector<type::Value> new_column_values(allocated_tuple_count);
+  std::vector<std::vector<type::Value>> new_columns(column_count);
+  std::vector<catalog::Column> columns;
+  type::Value max_exponent_count;
+  std::vector<type::Value> column_values;
+  auto tile_schema = tile->GetSchema();
+  auto compression_type = type::Type::TINYINT;
+
+  for (oid_t i = 0; i < column_count; i++) {
```
wangxiaoyou11: I'd suggest changing all `i` to some meaningful variable name. For example, you can change`i` to `column_id` here.

src/storage/compressed_tile.cpp
```diff
@@ -0,0 +1,588 @@
...
+//           1. INTEGER Types other than TINYINT - Find the median and store
+//              it as the base_value, store all the remaining values in types
+//              of smaller sizes.
+//           2. DECIMAL - Same as INTEGER Types, additionally store the
+//              maximum exponent.
+//           3. VARCHAR - Deduplicate the VARCHAR columns using a Cuckoo Hash.
+//        For each compressed column, store the metadata for retrieving the
+//        original values.
+
+void CompressedTile::CompressTile(Tile *tile) {
```
wangxiaoyou11: This function is too long. I would break it down and add some private helper function to make it more readable.

src/storage/compressed_tile.cpp
```diff
@@ -0,0 +1,588 @@
...
+@brief : The following function is used to compress all columns of a datatable
+         except VARCHAR. It sorts the column and sets the median value as the
+base value for the delta encoding. The remainig values are stored as offsets
+from this base value.
+
+@return :The modified values of the column i.e. the offsets of the values of
+         the columns in a vector.
+*/
+
+std::vector<type::Value> CompressedTile::CompressColumn(
```
wangxiaoyou11: Maybe change the name to `CompressNumericColumn()` ?

src/storage/compressed_tile.cpp
```diff
@@ -0,0 +1,588 @@
...
+
+void CompressedTile::CompressTile(Tile *tile) {
+  auto allocated_tuple_count = tile->GetAllocatedTupleCount();
+  std::vector<type::Value> new_column_values(allocated_tuple_count);
+  std::vector<std::vector<type::Value>> new_columns(column_count);
+  std::vector<catalog::Column> columns;
+  type::Value max_exponent_count;
+  std::vector<type::Value> column_values;
+  auto tile_schema = tile->GetSchema();
+  auto compression_type = type::Type::TINYINT;
```
wangxiaoyou11: From my understanding, the `compression_type` is the output of the `CompressColumn` function. So I think you don't need to assign the value here.

src/tcop/tcop.cpp
poojanilangekar: Compression can be called only when the TileGroup is full. So it requires atleast 1001 tuples. LOG_DEBUG generates a large log file.

src/include/storage/compressed_tile.h
rohit-cmu: Done

src/storage/compressed_tile.cpp
rohit-cmu: Done

poojanilangekar:
We have addressed all the comments from Code Review 2.

apavlo:
We should have a discussion about whether we want to merge this.

Nice job guys.

apavlo:
We decided that it is going to be too much work to merge this PR. We may use it as a reference in the future. Let's talk about plans for next semester when you get back. Thanks for all your work!

PR #657, opened 2017-05-05
Fixes for #651 and #640

pmenon:
Fixes for #651 and #640.

PR #671, opened 2017-05-15
Multi thread support for CodeGen SeqScan and HashJoin

tq5124:
The main focus of review should be `Scan` and `HashJoin`.

Related src files:

* `barrier.cpp`, provide barrier for multi threads. This object is shared between threads.
* `buffering_consurer.cpp`, modify the output vector to a lock free array.
* `compilation_context.cpp`, instead creating single plan function, we create a inner plan function that do all the previous work and a new plan function which submit inner plan to a thread pool.
* `hash_join_translator.cpp`, instead of using a single shared hash table, each thread use their own hash table and merge together at the end.
* `multi_thread_context.cpp`, each thread owns their own multi thread context, which include all the information the thread need.
* `query_thread_pool.cpp`, a thread pool designed for plan function. 
* `table.cpp`, split the table using multi thread context. Each thread is only responsible for their own part of the table.
* `transaction_runtime.cpp`, remove read check because it calls function outside of codeine and is not thread safe.
* `oa_hash_table.cpp`, the hash table that is used in hash join. 

Related test files:

* `table_scan_translator_test.cpp`
* `benchmark_scan_test.cpp`
* `hash_join_translator_test.cpp`
* `benchmark_join_test.cpp`
* `oa_hash_table_test.cpp`


phisiart:
This is a new PR. Our reviews are in https://github.com/cmu-db/peloton/pull/653.

tq5124:
Cannot pass the `jdbc_validator` test. We tried to rebase with the peloton master branch, and also cherry pick [these two commits](https://github.com/cmu-db/peloton/pull/657/commits) which should fix this bug, but both not work.

phisiart:
From the log, I suspect this is due to the BwTree bug.

apavlo:
After a discussion with @pmenon and @seojungmin, we decided that it is going to be too much work to merge this PR. We may use it as a reference in the future. Thanks for all your work!

Issue #643, opened 2017-04-26
The lifecycle of a planner::AbstractPlan


phisiart:
We are the LLVM team. This issue is blocking us from integrating plan caching into the system (our caching system already passes standalone tests). Since this requires API changes which might affect other teams, we leave this issue here and wait until the end of the semester. Changes should happen then.

## The Cache system

Our cache is a singleton object which boils down to a:
```cpp
std::map<std::unique_ptr<planner::AbstractPlan>, // physical plan
         std::unique_ptr<codegen::Query>>        // compiled query
```

(We've implemented a plan comparator.)

The use of `std::unique_ptr`'s is reasonable - the cache **owns** the cached plan and compiled query. We could also use `std::shared_ptr` here. The key is that a plan in the cache should not be freed.

When the user wants to access a plan in the cache, we would provide a `const planner::AbstractPlan *` (a raw pointer doesn't give away any ownership - the cache still owns the plan).

When a new `planner::AbstractPlan` comes, we check whether a structurally equivalent plan is already in the cache. If there isn't, then we store the plan in the cache.

Now we have to carefully manage the lifecycle of a plan. There are basically 3 ways we come up with:

- The user provides a `std::unique_ptr<planner::AbstractPlan>`.
We **move** that plan into the cache. The user no longer owns the plan.

- The user provides a `std::shared_ptr<planner::AbstractPlan>`.
We **increase the refcnt**. The cache has a shared ownership of the plan.

- The user provides a `const planner::AbstractPlan *`.
  In this way the user isn't giving us any ownership. [We never use non-owning raw pointers.](https://youtu.be/xnqTKD8uD64?t=12m37s) We make a copy of the plan and store the copy into the cache. Like this:
  ```cpp
  const planner::AbstractPlan *plan = ...; // user gives us this
  std::unique_ptr<planner::AbstractPlan> copy = plan->Copy();
  std::unique_ptr<codegen::Query> compiled_query = ...;
  Insert(std::move(copy), std::move(compiled_query));
  ```

We believe that the first and second methods are better.

There are also technical issues with regards to the second approach: currently the `Copy()` method for a lot of kinds of plans are incorrectly implemented (some are empty).

## Ownership Problem

We have currently adopted the first approach (moving to the second is also easy), and the cache is passing standalone test cases. However, the current peloton system is blocking us from integrating the cache. Here is why.

[...]

PR #675, opened 2017-05-19
Change Plan Executor API into accepting a shared_ptr.

phisiart:
This PR changes `PlanExecutor::ExecutePlan()` into accepting a `std::shared_ptr<planner::AbstractPlan>` instead of a `const planner::AbstractPlan *`.

This is the minimal change to allow a cache to hold ownership of the plan, which solves https://github.com/cmu-db/peloton/issues/643.

### DIFF
--- a/src/executor/plan_executor.cpp
+++ b/src/executor/plan_executor.cpp
@@ -129,11 +129,11 @@ ExecuteResult PlanExecutor::ExecutePlan(
 
     result.clear();
 
-    // Bind: casting const should be removed with later refactoring executor
-    planner::AbstractPlan *planp = plan.get();
+    // Perform binding
     planner::BindingContext context;
-    planp->PerformBinding(context);
+    plan->PerformBinding(context);
 
+    // Prepare output buffer
     std::vector<oid_t> columns;
     plan->GetOutputColumns(columns);
     codegen::BufferingConsumer consumer{columns, context};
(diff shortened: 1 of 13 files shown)


src/executor/plan_executor.cpp
```diff
@@ -129,7 +130,7 @@ ExecuteResult PlanExecutor::ExecutePlan(const planner::AbstractPlan *plan,
     result.clear();
 
     // Bind: casting const should be removed with later refactoring executor
-    planner::AbstractPlan *planp = const_cast<planner::AbstractPlan *>(plan);
+    planner::AbstractPlan *planp = plan.get();
```
seojungmin: We can remove this _planp_ pointer and directly use the _plan_ pointer now.  The previous casting was only necessary to remove _const_.  And it would be nice if you could also change the comment accordingly.

phisiart:
Oops. Looks like the weird BwTree error pops up randomly.

src/executor/plan_executor.cpp
phisiart: Done.

apavlo:
What happened with this PR? Do we still want to merge it?

seojungmin:
This PR was looking good to me.  As @phisiart has explained well in #643 and `const` keyword for `planner::AbstractPlan` has no longer been effective for a while, the approach chosen in this PR is a way to go for now.  This is required for later caching plans. If @phisiart  does not have time to rebase, then I will rebase this and make a new PR early next week.

seojungmin:
This PR is rebased with the current source code at #751. I am closing this PR.

PR #682, opened 2017-05-25
DELETE in codegen

seojungmin:
This is a PR for DELETE for codegen. Most of the works here are done by replicating the works completed in #656, and tidying up for a new PR.


apavlo:
@pmenon I looked this over. It looks reasonable but I am not familiar with this part of the system yet.

src/codegen/delete_translator.cpp
```diff
@@ -0,0 +1,99 @@
...
+  }
+
+  auto tile_group_id = batch.GetTileGroupID();
+
+  tile_group_ = table_.GetTileGroup(codegen, table_ptr_, tile_group_id);
+
+  batch.Iterate(context.GetCodeGen(),
+                [&](RowBatch::Row &row) { Consume(context, row); });
+}
+
```
pmenon: You can get rid of this batch consume method since the delete happens tuple-at-a-time.

src/include/codegen/delete_translator.h
```diff
@@ -0,0 +1,57 @@
...
+
+  void Produce() const override;
+
+  void Consume(ConsumerContext &, RowBatch &) const override;
+  void Consume(ConsumerContext &, RowBatch::Row &) const override;
+
+ private:
+  mutable llvm::Value *table_ptr_;
+  mutable llvm::Value *tile_group_;
+  static bool delete_wrapper(int64_t tile_group_id, uint32_t tuple_id,
```
pmenon: It doesn't look like this method is needed.

test/codegen/delete_translator_test.cpp
```diff
@@ -0,0 +1,222 @@
...
+//
+//===----------------------------------------------------------------------===//
+
+class DeleteTranslatorTest : public PelotonCodeGenTest {
+ public:
+  DeleteTranslatorTest() : PelotonCodeGenTest(), num_rows_to_insert(64) {
+    LoadTestTable(TestTableId(), num_rows_to_insert);
+  }
+
+  size_t getCurrentTableSize() {
```
pmenon: I think you replace this method with calls to the underlying DataTable.

test/codegen/delete_translator_test.cpp
```diff
@@ -0,0 +1,222 @@
...
+    scan.PerformBinding(context);
+
+    codegen::BufferingConsumer buffer{{0, 1}, context};
+    CompileAndExecute(scan, buffer, reinterpret_cast<char*>(buffer.GetState()));
+    return buffer.GetOutputTuples().size();
+  }
+
+  // delete all entries in table and then repopulate it.
+  void reloadTable() {
+    std::unique_ptr<planner::DeletePlan> delete_plan{
```
pmenon: You can remove this method and just load the test table at the start of each test, because this contains functionality that's under test.

test/include/codegen/codegen_test_util.h
```diff
@@ -67,7 +67,7 @@ class PelotonCodeGenTest : public PelotonTest {
   // Compile and execute the given plan
   codegen::QueryCompiler::CompileStats CompileAndExecute(
       const planner::AbstractPlan &plan, codegen::QueryResultConsumer &consumer,
-      char *consumer_state);
+      char *consumer_state, std::vector<type::Value> *params = nullptr);
```
pmenon: Looks like this is from parameterization. You can undo this.

test/sql/optimizer_sql_test.cpp
```diff
@@ -144,7 +144,7 @@ TEST_F(OptimizerSQLTests, SelectOrderByTest) {
   EXPECT_EQ("3", TestingSQLUtil::GetResultValueAsString(result, 1));
 
   // Something wrong with column property.
-  query = "SELECT * from test order by a + c";
+  query = "SELECT a, b, c from test order by a + c";
```
pmenon: :)

src/codegen/row_batch.cpp
```diff
@@ -196,6 +196,18 @@ RowBatch::RowBatch(CompilationContext &ctx, llvm::Value *tid_start,
...
+      tid_end_(tid_end),
+      num_rows_(nullptr),
+      selection_vector_(selection_vector),
+      filtered_(filtered) {}
+
+RowBatch::RowBatch(CompilationContext &ctx, llvm::Value *tile_group_id,
+                   llvm::Value *tid_start, llvm::Value *tid_end,
+                   Vector &selection_vector, bool filtered)
+    : context_(ctx),
+      tile_group_id_(tile_group_id),
```
pmenon: Change this so that the second constructor (without the tile_group_id) calls the first one (that does). Ensures initialization is consistent.

src/codegen/row_batch.cpp
```diff
@@ -345,6 +357,12 @@ llvm::Value *RowBatch::GetNumTotalRows(CodeGen &codegen) {
   return num_rows_;
 }
 
+llvm::Value *RowBatch::GetTileGroupID() const {
+  PL_ASSERT(tile_group_id_ != nullptr ||
+                "This row batch doesn not have tile_group_id");
+  return this->tile_group_id_;
```
pmenon: Remove assertion.

src/codegen/table.cpp
```diff
@@ -91,11 +91,11 @@ void Table::DoGenerateScan(CodeGen &codegen, llvm::Value *table_ptr,
 
     // Generate the scan cover over the given tile group
     if (vector_size > 1) {
-      tile_group_.GenerateVectorizedTidScan(
-          codegen, tile_group_ptr, column_layouts, vector_size, consumer);
+      tile_group_.GenerateVectorizedTidScan(codegen, tile_group_idx,
```
pmenon: Can we propagate the ID through the ScanConsumer callback hooks [here](https://github.com/cmu-db/peloton/blob/master/src/include/codegen/scan_consumer.h#L36)?

src/codegen/delete_translator.cpp
seojungmin: This batch consume method is called when there is a scan plan attached to it as a child as far as I know, so removing this entirely would not be correct for now.  We also have a test case in the delete_translator_test.  If you have meant something else, please let me know.

pmenon: There is a default implementation of batch consume in the parent class. If we remove the propagation of the tile group ID, this method isn't necessary.

apavlo:
@seojungmin Is this ready to merge?

seojungmin:
@apavlo If the last commit passes the travis test, I think it is ready to merge.

@pmenon @apavlo Thanks for the review! I have applied all the comments.  We could refactor more on this with the following codegen PRs, if there is any more.

seojungmin:
Could somebody, who has authority, re-run this travis test? The last travis test passes all the unit tests, but still produces the random failure.  This is typically resolved by re-running the travis.

PR #683, opened 2017-05-30
Fix incorrect comment

spiritlcx:


ChTimTsubasa:
I checked the log of travis-ci and found the failing is due to 13 - optimizer_sql_test (Failed).

I double checked the log of pulling #683 and found his commit failed due to the same reason. 

Since the only change in that commit is one line of comment (https://github.com/cmu-db/peloton/pull/683/files#diff-4a370c6459a678e6cce4e24439d6e452L43), I am wondering if the optimizer_sql_test is actually caused by the bugs in original code instead of mine?

PR #684, opened 2017-06-01
Fix issue semi-colon bug.

### LINKED ISSUE #681: Remove Semi-Colon Hack
This issue supplants #422.

We currently cannot support queries that have a semi-colon in them. The reason is because the `PacketManager::ExecQueryMessage` splits every incoming query by semi-colons:

https://github.com/cmu-db/peloton/blob/master/src/wire/packet_manager.cpp#L302

This was originally meant to support queries with multiple statements in them:

```sql
SELECT 1; SELECT 2;
```

But this will cause problems if there is a semi-colon **in** the query itself:

```sql
SELECT * FROM foo WHERE col = 'hello;world';
```

-----------------------

We need to refactor the `PacketManager` and `TrafficCop` to be able to handle multi-statement queries. The key thing is that the new SQL parser will return multiple parse trees if there are multiple statements. So we don't need to parse semi-colons manually anywhere in our own code.

Here is a rough outline of what I think we need to do:

1. The first thing to change is obviously to remove the split in `PacketManager::ExecQueryMessage()`. 

   https://github.com/cmu-db/peloton/blob/master/src/wire/packet_manager.cpp#L302

   Then remove the **for** loop in this function. It should not be iterating over multiple queries because it does not know that there are multiple queries.

   https://github.com/cmu-db/peloton/blob/master/src/wire/packet_manager.cpp#L316

2. Refactor `TrafficCop::ExecuteStatement()` to return back an **std::vector** of query results. This is instead of passing the output result buffers by reference as function arguments. Combine the following four items into a object (you can use a name like `StatementResult`):

   https://github.com/cmu-db/peloton/blob/master/src/wire/packet_manager.cpp#L319

   You'll also want to add the `ResultType` status code that `TrafficCop::ExecuteStatement()` currently returns as the function result into the `StatementResult` object.

3. Here's where things get tricky. `TrafficCop::ExecuteStatement()` calls `TrafficCop::PrepareStatement()`.

   https://github.com/cmu-db/peloton/blob/master/src/tcop/tcop.cpp#L149

   This function is where the `TrafficCop` passes the query to the SQL parser:

   https://github.com/cmu-db/peloton/blob/master/src/tcop/tcop.cpp#L286

   I don't know what happens here if there are multiple statements in the query (@haojin2?). The **for** loop at the URL below looks like we can handle it right now:

   https://github.com/cmu-db/peloton/blob/master/src/tcop/tcop.cpp#L299

   But I don't know why there could be multiple statements but then it looks like we are still assuming in the rest of the code that there is only one. So this means that we are going to need to change `TrafficCop::PrepareStatement()` to return multiple `Statement` objects instead of just one.

   Hopefully these changes are localized and we don't have to make major changes in other parts of the code.

ChTimTsubasa:
Fix #681 by deleting the split-query statement and for loop, since the query has already been parsed and split by PostgreSQL client.

### DIFF
--- a/src/wire/packet_manager.cpp
+++ b/src/wire/packet_manager.cpp
@@ -459,7 +459,7 @@ void PacketManager::ExecQueryMessage(InputPacket *pkt, const size_t thread_id) {
 
     CompleteCommand(query_type, rows_affected);
 
-    //Empty Query when the input is like " ;"
+  // handle the case when the string is "  ;"
   } else {
     SendEmptyQueryResponse();
   }
--- a/src/wire/packet_manager.cpp
+++ b/src/wire/packet_manager.cpp
@@ -445,7 +445,7 @@ void PacketManager::ExecQueryMessage(InputPacket *pkt, const size_t thread_id) {
       // check status
       if (status == ResultType::FAILURE) {
         SendErrorResponse(
-            {{NetworkMessageType::HUMAN_READABLE_ERROR, error_message}});
+          {{NetworkMessageType::HUMAN_READABLE_ERROR, error_message}});
         SendReadyForQuery(NetworkTransactionStateType::IDLE);
         return;
       }
@@ -458,11 +458,10 @@ void PacketManager::ExecQueryMessage(InputPacket *pkt, const size_t thread_id) {
     SendDataRows(result, tuple_descriptor.size(), rows_affected);
 
     CompleteCommand(query_type, rows_affected);
-  //Empty Query
+
+    //Empty Query when the input is like " ;"
   } else {
     SendEmptyQueryResponse();
-    SendReadyForQuery(NetworkTransactionStateType::IDLE);
-    return;
   }
 
   // PAVLO: 2017-01-15
--- a/src/wire/packet_manager.cpp
+++ b/src/wire/packet_manager.cpp
@@ -332,140 +332,137 @@ bool PacketManager::HardcodedExecuteFilter(std::string query_type) {
 }
 
 // The Simple Query Protocol
+// Fix mis-split bug: Previously, this function assumes there are multiple
+// queries in the string and split it by ';', which would cause one containing
+// ';' being split into multiple queries.
+// However, the multi-statement queries has been split by the psql client and
+// there is no need to split the query again.
 void PacketManager::ExecQueryMessage(InputPacket *pkt, const size_t thread_id) {
-  std::string q_str;
-  PacketGetString(pkt, pkt->len, q_str);
+  std::string query;
+  PacketGetString(pkt, pkt->len, query);
 
-  std::vector<std::string> queries;
-  boost::split(queries, q_str, boost::is_any_of(";"));
-
-  if (queries.size() == 0) {
-    SendEmptyQueryResponse();
-    SendReadyForQuery(txn_state_);
-    return;
+  // pop out the last character if it is ';'
+  if (query.back() == ';') {
+    query.pop_back();
   }
+  boost::trim(query);
 
-  if (queries.size() == 1 && queries.at(0).empty()) {
-    SendEmptyQueryResponse();
-    SendReadyForQuery(txn_state_);
-    return;
-  }
+  if (!query.empty()) {
+    std::vector<StatementResult> result;
+    std::vector<FieldInfo> tuple_descriptor;
+    std::string error_message;
+    int rows_affected = 0;
+    std::string query_type;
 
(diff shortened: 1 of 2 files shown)


ChTimTsubasa:
A unit test failure occurred in 13-optimizer_sql_test in travis-ci. Specifically: 
13: [0;31m[  FAILED  ] [mOptimizerSQLTests.SelectOrderByTest
13: [0;31m[  FAILED  ] [mOptimizerSQLTests.SelectDistinctTest

I retested it on my own computer but they passed. I have done 
`make check -j4`  
and  
`./test/optimizer_sql_test "--gtest_color=yes" "--gtest_output=xml:./build/test/optimizer_sql_test.xml"`

I also reviewed the failure log in #683 and found has the same failure, while it only changed a line of comment. So is it possible that this of failure is caused by test environment or test code?

apavlo:
@yingjunwu @haoxianghua I think this problem is the same as #655. It's failing in your PR #585

PauloAmora:
Is this reproductible locally?

ChTimTsubasa:
@PauloAmora No, it passed the test locally.

PauloAmora:
I think I found an problem with the tests, unrelated to the PR, will open an issue.

seojungmin:
@ChTimTsubasa @PauloAmora The test fails due to the nondeterministic column ordering with the output of "SELECT *".   AFAIK, since "SELECT *" does not guarantee the order of the column output, and it was supposed to use the column names explicitly instead of *.  This does not fail in a local test, but it does in travis and this is something you could investigate.  For a correct fix, you can refer to the last 2 commits in #682.

apavlo:
@seojungmin Good find. Let's merge your PR in now. This might be the major problem that we needed to overcome before we can merge all the other PRs.

yingjunwu:
Yes I agree with @seojungmin. This is a non-deterministic bug and is independent from the "tuple deletion" bug I am currently working on.

src/wire/packet_manager.cpp
```diff
@@ -294,60 +294,41 @@ bool PacketManager::HardcodedExecuteFilter(std::string query_type) {
...
+    // execute the query using tcop
+    auto status = traffic_cop_->ExecuteStatement(
+        query, result, tuple_descriptor, rows_affected, error_message,
+        thread_id);
+
+    // check status
+    if (status == ResultType::FAILURE) {
+      SendErrorResponse(
+          {{NetworkMessageType::HUMAN_READABLE_ERROR, error_message}});
+    }
```
seojungmin: I think we probably need to have `else` here and put the logic for sending the correct response into this `else` block, so that after sending the error response it does not send the result.

ChTimTsubasa: I am not the author of this part of the code but your suggestion sounds reasonable. I tested it with `else` on my machine and it passed all test case.

src/wire/packet_manager.cpp
```diff
@@ -294,45 +294,26 @@ bool PacketManager::HardcodedExecuteFilter(std::string query_type) {
 }
 
 // The Simple Query Protocol
+// Fix mis-split bug
```
apavlo: Anybody reading the code isn't going to know what you are talking about. Please provide a better description

src/wire/packet_manager.cpp
```diff
@@ -294,45 +294,26 @@ bool PacketManager::HardcodedExecuteFilter(std::string query_type) {
...
-      // check status
-      if (status == ResultType::FAILURE) {
-        SendErrorResponse(
-            {{NetworkMessageType::HUMAN_READABLE_ERROR, error_message}});
-        break;
-      }
-
+  std::string query;
+  PacketGetString(pkt, pkt->len, query);
+  if (!query.empty()) {
```
apavlo: Add a comment here that explains when the query could be empty.

src/wire/packet_manager.cpp
ChTimTsubasa: Added description on this bug

src/wire/packet_manager.cpp
ChTimTsubasa: Added some pre-process on the query so that it could detect empty query.

apavlo:
@ChTimTsubasa I am going to stop trying fix conflicts directly in Github. Can you rollback c3a58ecb0bdba960615f16bc4a6510c4a808ea3f and then rebase your PR on top of the master?

Sorry about this.

ChTimTsubasa:
No problem. I will do that.

PR #687, opened 2017-06-05
Clean up codegen deletion

pmenon:
This PR adds a helper class to perform tuple deletion in the codegen component. An instance of this class is created for every delete query.  The instance is initialized in the query plan's init() function with the transaction and table. A call to Delete() is generated in the primary loop for all qualifying tuples.  This is much cleaner than before.


pmenon:
@apavlo You can merge this in.

PR #694, opened 2017-06-11
keyword case insensitive issue

dengyuchenkit:
1. The query_type is still of string type. Convert it to upper case to match the keywords('BEGIN',  'END', 'ROLLBACK') 
2. Originally the query_type was generated by splitting the query_string using blank characters.(https://github.com/cmu-db/peloton/blob/master/src/common/statement.cpp#L31) A special case is that when the query is 'BEGIN;', it takes query_type as 'BEGIN;' instead of 'BEGIN'. In this case, remove the ending semi-colon.

### DIFF
--- a/src/common/statement.cpp
+++ b/src/common/statement.cpp
@@ -29,6 +29,10 @@ void Statement::ParseQueryType(const std::string& query_string,
                                std::string& query_type) {
   std::stringstream stream(query_string);
   stream >> query_type;
+	if (query_type.back() == ';') {
+		query_type = query_type.substr(0, query_type.length() - 1);
+	}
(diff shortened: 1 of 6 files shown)


apavlo:
Does this work with "BEGIN ;"?

dengyuchenkit:
Yes. If the ending character is a semicolon, it removes the semicolon. Otherwise, it retrieves the query_type by splitting the query_string using blank characters as before.
test

apavlo:
@dengyuchenkit Rebase this PR and I will merge.

dengyuchenkit:
Okay, I will resolve the conflicts soon

PR #695, opened 2017-06-19
Codegen INSERT

seojungmin:
This is a PR for INSERT with Codegen.  The initial code has come from #656, and the code has been refactored and clarified for a clean PR:

- The original two INSERT translators are merged into one INSERT translator.
- Most of the newly introduced classes are also merged into existing classes or are made compact.
- A helper class like that in #687 has also been introduced to clean up the code in the translator.

including some more miscellaneous code clean up along the way.


src/codegen/proxy/tuple_proxxy.cpp
```diff
@@ -0,0 +1,21 @@
...
+//===----------------------------------------------------------------------===//
+
+#include "codegen/proxy/tuple_proxy.h"
+
+namespace peloton {
+namespace codegen {
+
+DEFINE_TYPE(Tuple, "storage::Tupele", MEMBER(opaque));
+
+}  // namespace codegen
```
pmenon: Fix this file's name.

src/codegen/proxy/tuple_proxxy.cpp
```diff
@@ -0,0 +1,21 @@
...
+// Copyright (c) 2015-2017, Carnegie Mellon University Database Group
+//
+//===----------------------------------------------------------------------===//
+
+#include "codegen/proxy/tuple_proxy.h"
+
+namespace peloton {
+namespace codegen {
+
+DEFINE_TYPE(Tuple, "storage::Tupele", MEMBER(opaque));
```
pmenon: Fix spelling of the type's name.

src/codegen/tuple.cpp
```diff
@@ -0,0 +1,60 @@
...
+#include "catalog/schema.h"
+#include "codegen/proxy/data_table_proxy.h"
+#include "codegen/proxy/tuple_runtime_proxy.h"
+#include "codegen/type/sql_type.h"
+#include "storage/data_table.h"
+
+namespace peloton {
+namespace codegen {
+
+Tuple::Tuple(storage::DataTable &table): table_(table) {}
```
pmenon: Change the tuple constructor to accept the `catalog::Schema` instead of the data table.

src/codegen/tuple.cpp
```diff
@@ -0,0 +1,60 @@
...
+        auto val_ptr = codegen->CreateBitCast(ptr, codegen.CharPtrType());
+        codegen.Call(TupleRuntimeProxy::CreateVarlen,
+                     {v.GetValue(), v.GetLength(), val_ptr, pool});
+        break;
+      }
+      default: {
+        auto val_ptr = codegen->CreateBitCast(ptr, val_type->getPointerTo());
+        codegen->CreateStore(v.GetValue(), val_ptr);
+        break;
+      }
```
pmenon: 1. You need to handle NULLs here.
2. Change the function to take in a `std::vector<codegen::Value>` rather than pulling our values here. The signature will be `(CodeGen &codegen, llvm::Value *tuple_ptr, std::vector<codegen::Value> &col_vals, llvm::Value *pool)`.
3. With the signature change, this feels a lot like `CompactStorage`. Can you add something similar to handle serialization to DataTable? Something like `TableStorage` that performs serialization?

seojungmin:
@pmenon Thanks for the comments.

This is now ready to be merged.  More review comments are also welcome.

PR #696, opened 2017-06-19
Codegen cache

seojungmin:
This is a PR for caching the compiled objects in codegen, so that if the same query is requested, the codegen checks whether or not the plan has already been compiled by comparing each item in the two plans.  This is another PR which is cherry picked from #656, and is based on the PR #675.

This cache could be enhanced further in the future by mixing eviction policies, configuring the cache size, manually pinning some queries in the cache, making them persistent in the cache, etc. In terms of software engineering point of view, the plan comparator could be embedded with plan classes rather than having it separately in codegen.

### DIFF
--- a/src/codegen/plan_comparator.cpp
+++ b/src/codegen/plan_comparator.cpp
@@ -1,0 +1,554 @@
+//===----------------------------------------------------------------------===//
+//
+//                         Peloton
+//
+// plan_comparator.cpp
+//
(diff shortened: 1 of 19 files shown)


seojungmin:
The plan comparison scheme in this PR may not work for some corner cases. The concept is fine, but more careful review on the plan comparison is required.  I am closing the PR, and resubmit the PR once it is more ready.

PR #697, opened 2017-06-20
Fixed NULL handling in codegen

pmenon:
This PR fixes NULL handling in the codegen engine. `codegen::Value` has a NULL-bit indicating if the value is NULL. This is in contrast to how NULL-checking is done in `type::Value`, which relies on special values. Additionally, this PR separates the notion of a `type::SqlType`, a SQL data type (i.e., bigint) that is static in the DBMS, from a **runtime** `type::Type` that captures a SQL data type, NULL-ability, and other type-specific information (e.g., the precision and scale of decimals). All `codegen::Value`s have a `codegen::type::Type`. Having both a data type and NULL-ability information allows us to generate specialized code. For example, NULL-ability information allows us to avoid generating NULL-checking code altogether when materializing columns, performing comparisons, or executing binary functions - this is unavoidable in `type::Value`.

Part of this PR introduces a new `TypeSystem` class that is essentially a Postgres operator table. Every `SqlType` has a `TypeSystem` that captures all the operations that a SQL type supports. Any of these can be overridden by modifying the catalog.

This isn't fully ready to be merged. I just want eyes on it now because it's large.


src/codegen/code_context.cpp
```diff
@@ -18,6 +18,9 @@
 #include "llvm/Support/raw_os_ostream.h"
 #include "llvm/Support/TargetSelect.h"
 #include "llvm/Transforms/Scalar.h"
+#if LLVM_VERSION_GE(3,9)
+#include "llvm/Transforms/Scalar/GVN.h"
```
seojungmin: 👍

pmenon:
@seojungmin You can review this PR now.

@apavlo Once the reviews/changes are done, this should be merged first as it's the largest.

src/codegen/values_runtime_proxy.cpp
```diff
@@ -19,6 +19,36 @@ namespace peloton {
 namespace codegen {
 
 //===----------------------------------------------------------------------===//
+// OUTPUT BOOL
+//===----------------------------------------------------------------------===//
+
+// Get the symbol name for ValuesRuntime::OutputBool()
+const std::string &ValuesRuntimeProxy::_OutputBool::GetFunctionName() {
+  static const std::string kOutputTinyIntFnName =
```
seojungmin: The name does not matter on how the system executes, but it would be neat if we align this with the function name, since we have been aligning it.

src/codegen/aggregation.cpp
```diff
@@ -170,82 +159,156 @@ void Aggregation::CreateInitialValues(
...
+      next = curr.Add(codegen, delta);
+      break;
+    }
+    case ExpressionType::AGGREGATE_COUNT_STAR: {
+      auto curr = storage_.GetValueSkipNull(codegen, storage_space,
+                                            aggregate_info.storage_index);
+      auto delta = Value{type::BigInt::Instance(), codegen.Const64(1)};
+      next = curr.Add(codegen, delta);
+      break;
+    }
```
pmenon: Good catch.

src/codegen/aggregation.cpp
```diff
@@ -170,82 +159,156 @@ void Aggregation::CreateInitialValues(
...
+                                null_bitmap);
+              break;
+            }
+            case ExpressionType::AGGREGATE_COUNT: {
+              codegen::Value one{type::BigInt::Instance(), codegen.Const64(1)};
+              storage_.SetValue(codegen, storage_space,
+                                aggregate_info.storage_index, one, null_bitmap);
+              break;
+            }
+            default: { break; }
```
pmenon: COUNT(*) is never NULL-able and will always be handled in `DoAdvanceValue()`. I've simplified the code a little in the latest push to clear this up.

src/codegen/type/varbinary_type.cpp
```diff
@@ -0,0 +1,218 @@
...
+// Unary operators
+static std::vector<TypeSystem::UnaryOpInfo> kUnaryOperatorTable = {};
+
+// Binary operations
+static std::vector<TypeSystem::BinaryOpInfo> kBinaryOperatorTable = {};
+
+}  // anonymous namespace
+
+//===----------------------------------------------------------------------===//
+// TINYINT TYPE CONFIGURATION
```
pmenon: Will do.

src/codegen/values_runtime_proxy.cpp
pmenon: Will do.

pmenon:
Addressed all comments. If the latest build passes, this can be merged.

pmenon:
@apavlo This can be merged.

PR #698, opened 2017-06-20
Codegen UPDATE

seojungmin:
This PR is for processing UPDATE within Codegen. The initial code has come from #656, and much of the code has been refactored and modified for a clean PR, e.g. the global runtime states introduced in the previous PR are all removed and localized, and a helper class like that in #687 has been introduced to accommodate many of the changes.  

`PerformUpdate()` that contains the original executor logic has been touched for a clear code representation.  Although it is more readable now, it is not still in its best condition.  There is also serialization/deserialization for the target/direct_map lists while being passed through the translator, and this may need some attention to be improved near future.

In addition, a bug on handling VARCHAR in codegen has been fixed, which is not related to UPDATE.   The original peloton value system has been including null terminators in its VARCHAR's length value, but that of codegen has been changed to behave in the same way.

### DIFF
--- a/script/testing/sqlite_trace/trace-replay.py
+++ b/script/testing/sqlite_trace/trace-replay.py
@@ -6,6 +6,7 @@ import sys
 import socket
 import time
 import psycopg2
+from optparse import OptionParser
 from config import *
 from subprocess import Popen
 
@@ -103,11 +104,13 @@ def drop_all_tables(conn):
 if __name__ == "__main__":
     global peloton_path, output_dir, peloton_log_file
 
+
     work_path = os.path.abspath(".")
     output_dir = os.path.abspath(output_dir)
     sql_file = os.path.abspath(os.path.join(output_dir, "test.sql"))
     config = os.path.abspath(os.path.join(output_dir, "config"))
     peloton_log_file = os.path.abspath(os.path.join(output_dir, peloton_log_file))
+
 
     if not os.path.exists(output_dir):
         os.mkdir(output_dir)
@@ -116,6 +119,11 @@ if __name__ == "__main__":
     peloton_log = open(peloton_log_file, "w")
     peloton = None
 
+    parser = OptionParser()
+    parser.add_option("--config", dest="config", help="config file path for the testing framework", default=config)
+    parser.add_option("--out", dest="out", help="Output path for JUnit XML files", default=output_dir)
+    (options, args) = parser.parse_args()
+
     try:
         # Open a connection to postgres so that we can drop the tables
         conn = psycopg2.connect(
@@ -123,17 +131,17 @@ if __name__ == "__main__":
                 user = pg_username,
                 password = pg_password,
                 port = 5433)
-      
+
         # Extract the SQL from the trace files
         for path in traces:
             print "="*80
             print "test: ", path
             print "="*80
             sys.stdout.flush()
-            
+
             # Always drop the tables in the database first
             drop_all_tables(conn)
-            
+
             # Convert the trace file into a format that we can handle
             extract_sql(path, sql_file)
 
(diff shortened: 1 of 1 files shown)


seojungmin:
The Coveralls program seemed to behave a little bit abnormally.  It reported that the coverage went down about 11%, which was unlikely considering the change that had been made in this PR.  @pmenon helped me to rerun the travis, and it seems to report the coverage correctly.

seojungmin:
I will rebase this, now.

seojungmin:
This PR can be merged. Review comments are always welcome.

pmenon:
@seojungmin Can you provide high-level points on how this works? Also add comments in the code describing each component.

seojungmin:
@pmenon Here are some high-level points. I added comments on the last commit.  

This PR introduces two new modules, and also changes roles of some previous modules:

[New]
- _Update translator_ builds the information required for update and hand it over to the updater.
- _Updater_ is a helper for the update execution, and deals with updating with and without primary key.  It also calls TransactionRuntime to find out(or to acquire) the ownership of the tuple. Depending on the ownership it has different execution strategies.  The execution strategy is the same as before, but the code is clean up.

[Role Changes]
- _TransactionRuntime_ used to do all the execution work, but I moved these roles to Deleter and I only left helper APIs for acquiring ownership.
- _Deleter_ used to be only a bridge to TransactionRuntime, but it now has all the execution logic inside, which is consistent with Updater.

Since he update execution logic was so huge and complex, these separations and changes were inevitable.   After all, the translators do whatever codegen can do, and pass over to Updater/Inserter/Deleter which actually do the execution jobs that cannot be handled by codegen.

src/codegen/operator/update_translator.cpp
```diff
@@ -0,0 +1,157 @@
...
+  auto &context = GetCompilationContext();
+  const auto *project_info = update_plan_.GetProjectInfo();
+  auto target_list = project_info->GetTargetList();
+  auto direct_map_list = project_info->GetDirectMapList();
+  uint32_t column_num =
+      static_cast<uint32_t>(target_list.size() + direct_map_list.size());
+  auto *scan =
+      static_cast<const planner::AbstractScan *>(update_plan_.GetChild(0));
+  std::vector<const planner::AttributeInfo *> ais;
+  scan->GetAttributes(ais);
```
pmenon: Shouldn't this be done in `UpdatePlan`? At this point, the plans should already know all attribute information.

src/codegen/updater.cpp
```diff
@@ -0,0 +1,173 @@
...
+#include "storage/tile.h"
+#include "storage/tuple.h"
+#include "type/abstract_pool.h"
+#include "type/types.h"
+#include "type/value.h"
+
+namespace peloton {
+namespace codegen {
+
+void Updater::Init(concurrency::Transaction *txn, storage::DataTable *table,
```
pmenon: Modify the initialization by removing `Transaction`, and add `ExecutorContext`. This way, you get both the memory pool and transaction handle at initialization time. You can then remove `GetPool()` and remove `ExecutorContext` from `Update()`

src/codegen/updater.cpp
```diff
@@ -0,0 +1,173 @@
...
+                   Target *target_vector, uint32_t target_vector_size) {
+  PL_ASSERT(txn != nullptr && table != nullptr && target_vector != nullptr);
+  txn_ = txn;
+  table_ = table;
+  // Target list is kept since it is required at a new version update
+  target_list_.reset(
+      new TargetList(target_vector, target_vector + target_vector_size));
```
pmenon: Does `target_list_` have to be a unique_ptr?

src/include/codegen/operator/update_translator.h
```diff
@@ -0,0 +1,73 @@
...
+  // No state finalization
+  void TearDownState() override;
+
+  // Get the stringified name of this translator
+  std::string GetName() const override { return "Update"; }
+
+ private:
+  void SetTargetValue(llvm::Value *target_val_vec, llvm::Value *target_id,
+                      type::Type type, llvm::Value *value,
+                      llvm::Value *length) const;
```
pmenon: Not sure this is implemented.

src/include/codegen/operator/update_translator.h
```diff
@@ -0,0 +1,73 @@
...
+
+ private:
+  // Plan
+  const planner::UpdatePlan &update_plan_;
+
+  // Target table
+  storage::DataTable *target_table_;
+
+  // Boolean value representing whether it is a primary key update or not
+  bool update_primary_key_;
```
pmenon: Not sure if this is used.

src/include/codegen/transaction_runtime.h
```diff
@@ -13,6 +13,8 @@
 #pragma once
 
 #include <cstdint>
+#include "type/types.h"
+#include "common/item_pointer.h"
```
pmenon: Is this needed?

Issue #523, opened 2017-01-30
Remove unnecessary null pointer checks


elfring:
[Extra null pointer checks are not needed in functions](https://isocpp.org/wiki/faq/freestore-mgmt#delete-handles-null "Do I need to check for null before delete p?") like [the destructor for the class “UpdateClause”](https://github.com/cmu-db/peloton/blob/6b154b312e57aa10b573ce596ed95c3b51c566a2/src/include/parser/update_statement.h#L32 "Destructor for UpdateClause").

apavlo:
Can you send a PR?

elfring:
How do you think about to wrap any pointer data members with the template class "[std::unique_ptr](https://en.wikipedia.org/wiki/Smart_pointer#unique_ptr "Description for the usage of smart pointers")"?

apavlo:
@elfring Yes, we should be using `unique_ptr` for these classes.

apavlo:
We still haven't fixed this. We still should do this.

poojanilangekar:
@apavlo I looked through all the files where there were nullptr checks, there are no check just before a delete. It is usually in an if statement where it combines other function calls. So I am closing this issue.

elfring:
[How do you think about](https://github.com/cmu-db/peloton/blob/8a41d7c53a1543d546733a4a67a4e22f75fec64e/src/catalog/system_catalogs.cpp#L70 "Update candidate: Destructor for SystemCatalogs") to [use a development tool like “clang-tidy” for corresponding source code adjustments](https://clang.llvm.org/extra/clang-tidy/checks/readability-delete-null-pointer.html "Deletion of unnecessary null pointer checks")?

PR #702, opened 2017-06-24
Remove unnecessary null pointer checks #523

lixupeng:
The unnecessary null pointer checks mainly exist in include/parser. I just simply removed those if-statements.

### DIFF
--- a/src/binder/bind_node_visitor.cpp
+++ b/src/binder/bind_node_visitor.cpp
@@ -36,8 +36,8 @@ void BindNodeVisitor::Visit(const parser::SelectStatement *node) {
   if (node->order != nullptr) node->order->Accept(this);
   if (node->limit != nullptr) node->limit->Accept(this);
   if (node->group_by != nullptr) node->group_by->Accept(this);
-  for (auto select_element = node->select_list->begin(); select_element != node->select_list->end(); ++select_element) {
-    (*select_element)->Accept(this);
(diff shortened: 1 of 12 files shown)


seojungmin:
This is good. I wonder whether we could use smart pointers here, more specifically unique pointers, even to get rid of delete's too.

apavlo:
@lixupeng - @seojungmin is correct here. You should switch these to use smart pointers.

lixupeng:
Yes, I'm trying to switch all pointers used in parser to smart pointers. But it occurs a number of compiling errors need to be fixed. I will fix them later. Further, SelectStatement is incomplete so SelectStatement* cannot be replaced by unique_ptr<SelectStatement>.

pmenon:
@lixupeng This is because your first loop is trying to make a copy of a unique_ptr, which is not allowed by definition. You should iterate using a reference to elements, like so:

```
vector<unique_ptr<T>> vec;
for (auto &item : vec) ... 
```

Or, if you don't intend to modify the elements:

`for (const auto &item : vec) ... `

lixupeng:
@pmenon Yeah, your solution is cool, much prettier than mine. And how about those situations that a function should recieve a pointer as argument but we have unique_ptr now?  We should use unique_ptr.get() or modify the definition of that function to receive a unique_ptr& recursively?

Issue #715, opened 2017-07-03
release mode crashed


yingjunwu:
Peloton does not work when built with release mode.

simply build the system using `cmake .., make -j`. Then use `psql` to issue some queries like:

```
create table foo(a int)
select * from foo
```

The system will crash, complaining that 
```
LLVM ERROR: Program used external function '_ZNK7peloton7storage9TileGroup14GetTileGroupIdEv' which could not be resolved!
terminate called without an active exception
```

pmenon:
Looks like name mangling changes in release mode. I knew this would come back to bite us. I'll take a look.

pmenon:
So, I can't reproduce this on my machine. What GCC version are you using? 

`storage::TileGroup::GetTileGroupId` is defined in the header. In release mode, I suspect it's getting inlined into all call-sites, doesn't make it into the binary, and, hence, isn't resolvable by LLVM. This doesn't occur in GCC 5.4, but may happen in 4.8 or 6.

yingjunwu:
In fact, I reproduced it in a machine with GCC 5.4 (an NUS machine) and one with GCC 4.8 (CMUDB dev2). Can you try dev2 or any other machine?

Issue #705, opened 2017-06-28
New Optimizer Crashes in TPC-C StockLevel Join Query


ChTimTsubasa:
The new optimizer Seg_Fault when doing TPC-C testing.

The problem only happens when there are 3 or more than 3 warehouses, and is unrelated to the number of terminals. Specifically, it is caused by this query:"https://github.com/oltpbenchmark/oltpbench/blob/master/src/com/oltpbenchmark/benchmarks/tpcc/procedures/StockLevel.java#L42".

@hzxa21 suggested that the problem is caused by mixing up multi-table join predicates with single-table predicates in hash join. 

**Peloton Log**
See the attatchment. The set up was to create the tables and load data. After that, we typed the error-causing query.

[peloton error message in debug mode.txt](https://github.com/cmu-db/peloton/files/1109366/peloton.error.message.in.debug.mode.txt)
[peloton error message in release mode.txt](https://github.com/cmu-db/peloton/files/1109364/peloton.error.message.in.release.mode.txt)
[peloton error message in trace mode.txt](https://github.com/cmu-db/peloton/files/1109365/peloton.error.message.in.trace.mode.txt)

**GDB Backtrace:**
#0  0x00007ffff69d3a63 in peloton::executor::AbstractJoinExecutor::BuildOutputLogicalTile(peloton::executor::LogicalTile*, peloton::executor::LogicalTile*) ()
   from /home/tianyic1/peloton_640/build/lib/libpeloton.so.0.0.5
#1  0x00007ffff69bc81f in peloton::executor::HashJoinExecutor::DExecute() ()
   from /home/tianyic1/peloton_640/build/lib/libpeloton.so.0.0.5
#2  0x00007ffff69d19ce in peloton::executor::AggregateExecutor::DExecute() ()
   from /home/tianyic1/peloton_640/build/lib/libpeloton.so.0.0.5

apavlo:
Two comments about this issue:

1. The DBMS should execute the join query in TPC-C StockLevel using an index nested loop join. Using a hash join is an overkill and will give us slow benchmark results.

2. Your example is crashing in the old execution engine (i.e., `HashJoinExecutor`). What happens when you run this query in @pmenon's LLVM engine?

ChTimTsubasa:
@apavlo I have just fetch the current version of Peloton and ran the test. Is this the one with new LLVM engine?
[...]

PR #717, opened 2017-07-03
Fix the name-mangling issue on release mode

seojungmin:
This fixes the function name mangling in #715, which is also partially also mentioned in #705.  This issue seems to occur on some GCC versions, but it can be reproducible on dev1(5.4.0-6ubuntu1, 16.04.4), dev2(4.8.4-2ubuntu1, 14.04.3) and dev3(4.8.4-2ubuntu1, 14.04.3).


apavlo:
I will merge when this passes

PR #718, opened 2017-07-03
Travis build several configs

PauloAmora:
This is the first try for supporting some compilers on Travis.


apavlo:
Once we get this working we should squash the commits into a single commit.

PauloAmora:
Since this can get dirty, I'll just submit another PR once finished. I'll make sure to add the coveralls cmake change as well.

PR #720, opened 2017-07-03
check the length before inserting data into VARCHAR or VARBINARY columns

### LINKED ISSUE #636: DBMS does not enforce VARCHAR length
maxlen specified in varchar(maxlen) seems not work:

```
postgres=# *create* table T(a varchar(3));

>create table T(a varchar(3)) 0

postgres=# *insert* into T values ('abcdefg');

>insert into T values ('abcdefg') 1

postgres=# *select* * from T;

>   a
>---------
>abcdefg
>(1 row)
```

lixupeng:
fix the bug #636. I add a check in planner phase. When generating a InsertPlan,  one should call tuple->SetValue for each element to be inserted. So I add the check in Tuple::SetValue.

### DIFF
--- a/src/storage/tuple.cpp
+++ b/src/storage/tuple.cpp
@@ -62,7 +62,7 @@ void Tuple::SetValue(const oid_t column_offset, const type::Value &value,
     if ((type == type::TypeId::VARCHAR || type == type::TypeId::VARBINARY)
         && (column_length != 0 && value.GetLength() != type::PELOTON_VALUE_NULL)
         && value.GetLength() > column_length) {
-      throw peloton::ValueOutOfRangeException(value.GetData(), type, column_length);
+      throw peloton::ValueOutOfRangeException(type, column_length);
     }
     value.SerializeTo(value_location, is_inlined, data_pool);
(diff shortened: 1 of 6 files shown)


src/storage/tuple.cpp
```diff
@@ -59,6 +59,10 @@ void Tuple::SetValue(const oid_t column_offset, const type::Value &value,
 
   // Skip casting if type is same
   if (type == value.GetTypeId()) {
+    if ((type == type::Type::VARCHAR || type == type::Type::VARBINARY)
+        && value.GetLength() > column_length) {
+      throw peloton::IncompatibleTypeException(type, " data is too long");
```
apavlo: I don't think we want to throw `IncompatibleTypeException` here. Is there another exception type we can throw instead? What does Postgres do?

lixupeng: Postgres reports an ERRCODE_STRING_DATA_RIGHT_TRUNCATION in such situation. Should we add a new Exception for this? 
Is ValueOutOfRangeException or CastException more suitable here?

Issue #680, opened 2017-05-24
SEG FAULT when typing wrong 'SELECT'


ChTimTsubasa:
The program crashes and gives no other information in DEBUG mode when typing SELECT query without specifying the tuple that is queried for. 

The information printed on the console is:
2017-05-24 17:35:43 [src/wire/libevent_callbacks.cpp:46:WorkerHandleNewConn] DEBUG - Creating new socket fd:27
2017-05-24 17:35:43 [src/wire/libevent_server.cpp:53:CreateNewConn] INFO  - create new connection: id = 27
2017-05-24 17:35:43 [src/wire/packet_manager.cpp:65:PacketManager] DEBUG - Registered new PacketManager [count=2]
Segmentation fault (core dumped)

Those crash-causing queries examples including but (may) not be limited to: "SELECT;", "SELECT FROM <TABLE_NAME>;", etc.

Other invalid query strings such as "CREATE;”, " INSERT;" can be recognized and works well.

haojin2:
Would you please rerun it with gdb and provide the backtrace for the segfault?

ChTimTsubasa:
Thread 4 "peloton" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7fffccfff700 (LWP 3435)]
0x00007ffff6917d11 in peloton::parser::PostgresParser::TargetTransform (
    root=0x0)
    at /home/tim/Documents/project/peloton/src/parser/postgresparser.cpp:496
496	  for (auto cell = root->head; cell != nullptr; cell = cell->next) {
(gdb) backtrace
#0  0x00007ffff6917d11 in peloton::parser::PostgresParser::TargetTransform (
    root=0x0)
    at /home/tim/Documents/project/peloton/src/parser/postgresparser.cpp:496
#1  0x00007ffff691af4b in peloton::parser::PostgresParser::SelectTransform (
    root=0x7fffca438628)
    at /home/tim/Documents/project/peloton/src/parser/postgresparser.cpp:1078
#2  0x00007ffff691b441 in peloton::parser::PostgresParser::NodeTransform (
    stmt=0x7fffca438628)
    at /home/tim/Documents/project/peloton/src/parser/postgresparser.cpp:1142
#3  0x00007ffff691b6a6 in peloton::parser::PostgresParser::ListTransform (
    root=0x7fffca438758)
    at /home/tim/Documents/project/peloton/src/parser/postgresparser.cpp:1198
[...]

PR #727, opened 2017-07-05
Implement parser edge case

ChTimTsubasa:
This PR intents to fix issue #680 that peloton would crash with statements like "SELECT *" and "SELECT;"

The reason is that the Postgres parser does not think the statements mentioned above as wrong statements and the binder and optimizer would continue to execute the wrong parse tree and eventually crash.

The solution is to explicitly check that
1. The target list is not null, which can overcome "SELECT;" 
2. The from clause list is not null **when the target contains any variables 'like * or item column'**, this would overcome "SELECT *;"

A unit test case is added in parser_test to check "SELECT *" and "SELECT;".

There may still be other edge cases in 2 since right now only constant, bool expression and expression type are checked.

### DIFF
--- a/src/optimizer/stats/cost.cpp
+++ b/src/optimizer/stats/cost.cpp
@@ -74,7 +74,7 @@ void Cost::CombineConjunctionStats(const std::shared_ptr<TableStats> &lhs,
       break;
     default:
       LOG_TRACE("Cost model conjunction on expression type %d not supported",
-                type);
+                (int)type);
   }
   if (output_stats != nullptr) {
     output_stats->num_rows = num_tuples;
--- a/src/optimizer/stats/cost.cpp
+++ b/src/optimizer/stats/cost.cpp
@@ -73,8 +73,8 @@ void Cost::CombineConjunctionStats(const std::shared_ptr<TableStats> &lhs,
       num_tuples = static_cast<size_t>((sel1 + sel2 - sel1 * sel2) * num_rows);
       break;
     default:
-      LOG_TRACE("Cost model conjunction on expression type %d not supported",
-                (int)type);
+      LOG_WARN("Cost model conjunction on expression type %s not supported",
+               ExpressionTypeToString(type).c_str());
   }
   if (output_stats != nullptr) {
     output_stats->num_rows = num_tuples;
--- a/src/optimizer/stats/selectivity.cpp
+++ b/src/optimizer/stats/selectivity.cpp
@@ -43,7 +43,7 @@ double Selectivity::ComputeSelectivity(const std::shared_ptr<TableStats> &stats,
       return DistinctFrom(stats, condition);
     default:
       LOG_TRACE("Expression type %d not supported for computing selectivity",
-                type);
+                (int)(condition.type));
       return DEFAULT_SELECTIVITY;
   }
 }
--- a/src/optimizer/stats/selectivity.cpp
+++ b/src/optimizer/stats/selectivity.cpp
@@ -42,8 +42,8 @@ double Selectivity::ComputeSelectivity(const std::shared_ptr<TableStats> &stats,
     case ExpressionType::COMPARE_DISTINCT_FROM:
       return DistinctFrom(stats, condition);
     default:
-      LOG_TRACE("Expression type %d not supported for computing selectivity",
-                (int)(condition.type));
+      LOG_WARN("Expression type %s not supported for computing selectivity",
+                ExpressionTypeToString(condition.type).c_str());
       return DEFAULT_SELECTIVITY;
   }
 }
--- a/src/parser/postgresparser.cpp
+++ b/src/parser/postgresparser.cpp
@@ -185,24 +185,24 @@ parser::TableRef* PostgresParser::RangeSubselectTransform(
 }
 
 // Get in a target list and check if is with variables
-bool IsTargetListWithVar(List* target_list) {
+bool IsTargetListWithVariable(List* target_list) {
   // The only valid situation of a null from list is that all targets are constant
-  LOG_DEBUG("size is : %d", target_list->length);
-  print_pg_parse_tree(target_list);
-
   for (auto cell = target_list->head; cell != nullptr; cell = cell->next) {
-
     ResTarget* target = reinterpret_cast<ResTarget*>(cell->data.ptr_value);
-      LOG_DEBUG("Type: %d, expected %d", target->type, T_A_Const);
+    LOG_TRACE("Type: %d", target->type);
 
-      switch (target->val->type) {
-        case T_A_Const:
-        case T_A_Expr:
-        case T_BoolExpr:
-          continue;
-        default:
-          return true;
-      }
+    // Bypass the target nodes with type:
+    // constant("SELECT 1;"), expression ("SELECT 1 + 1"),
+    // and boolean ("SELECT 1!=2;");
+    // TODO: We may want to see if there are more types to check.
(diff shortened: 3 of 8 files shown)


src/optimizer/stats/cost.cpp
```diff
@@ -74,7 +74,7 @@ void Cost::CombineConjunctionStats(const std::shared_ptr<TableStats> &lhs,
       break;
     default:
       LOG_TRACE("Cost model conjunction on expression type %d not supported",
-                type);
+                (int)type);
```
apavlo: Change this to `LOG_WARN`.
Also use `ExpressionTypeToString()` instead of printing out the type integer value.

src/optimizer/stats/selectivity.cpp
```diff
@@ -43,7 +43,7 @@ double Selectivity::ComputeSelectivity(const std::shared_ptr<TableStats> &stats,
       return DistinctFrom(stats, condition);
     default:
       LOG_TRACE("Expression type %d not supported for computing selectivity",
-                type);
+                (int)(condition.type));
```
apavlo: Same as my other comment. Switch to `LOG_WARN` and `ExpressionTypeToString()`

src/parser/postgresparser.cpp
```diff
@@ -184,12 +184,50 @@ parser::TableRef* PostgresParser::RangeSubselectTransform(
...
+  /* Statement like 'SELECT *;' cannot detect by postgres parser and would lead to
+   * a null list of from clause*/
+  if (root == nullptr) {
+    auto target_list = select_root->targetList;
+    // The only valid situation of a null 'from list' is that all targets are constant
+    LOG_TRACE("size is : %d", target_list->length);
+    //print_pg_parse_tree(target_list);
+    if (IsTargetListWithVariable(target_list)) {
+      throw ParserException(
+          StringUtil::Format("Error parsing SQL statement"));
```
apavlo: I don't think you need `StringUtil::Format()` here.

src/parser/postgresparser.cpp
```diff
@@ -503,6 +542,13 @@ expression::AbstractExpression* PostgresParser::FuncCallTransform(
 // It checks the type of each target and call the corresponding helpers.
 std::vector<expression::AbstractExpression*>* PostgresParser::TargetTransform(
     List* root) {
+  /* Statement like 'SELECT;' cannot detect by postgres parser and would lead to
```
apavlo: Switch to double slash comments instead of `/* .. */`

src/parser/postgresparser.cpp
```diff
@@ -503,6 +542,13 @@ expression::AbstractExpression* PostgresParser::FuncCallTransform(
...
+   * null list*/
+  if (root == nullptr) {
+      throw ParserException(
+          StringUtil::Format("Error parsing SQL statement"));
```
apavlo: Same comment here about `StringUtil::Format()`

src/parser/postgresparser.cpp
```diff
@@ -1228,8 +1274,13 @@ parser::SQLStatementList* PostgresParser::ListTransform(List* root) {
     return nullptr;
   }
   LOG_TRACE("%d statements in total\n", (root->length));
-  for (auto cell = root->head; cell != nullptr; cell = cell->next) {
-    result->AddStatement(NodeTransform((Node*)cell->data.ptr_value));
+  try {
+    for (auto cell = root->head; cell != nullptr; cell = cell->next) {
+      result->AddStatement(NodeTransform((Node*)cell->data.ptr_value));
+    }
+  } catch (ParserException e) {
```
apavlo: Do we really want to catch the exception here? The error message is lost.

test/parser/parser_test.cpp
```diff
@@ -412,6 +412,17 @@ TEST_F(ParserTests, CopyTest) {
...
+// Test that the wrong queries can be detected.
+TEST_F(ParserTests, WrongQueryTest) {
+  std::vector<std::string> queries;
+  queries.push_back("SELECT;");
+  queries.push_back("SELECT *;");
+  // Parsing
+  for (auto query : queries) {
+    UNUSED_ATTRIBUTE parser::SQLStatementList* result =
+        parser::PostgresParser::ParseSQLString(query.c_str());
+    EXPECT_EQ(result->is_valid, false);
```
apavlo: I think that you may want to check for `ParserException` here instead.

ChTimTsubasa:
One of the four travis jobs is failed. 
The test failed when running the script ../script/testing/psql/psql_test.sh. The line of error causing code is 
at 
`peloton: /home/travis/build/cmu-db/peloton/src/index/bwtree_index.cpp:106: bool peloton::index::BWTreeIndex<KeyType, ValueType, KeyComparator, KeyEqualityChecker, KeyHashFunc, ValueEqualityChecker, ValueHashFunc>::CondInsertEntry(const peloton::storage::Tuple*, peloton::ItemPointer*, std::function<bool(const void*)>) [with KeyType = peloton::index::CompactIntsKey<1ul>; ValueType = peloton::ItemPointer*; KeyComparator = peloton::index::CompactIntsComparator<1ul>; KeyEqualityChecker = peloton::index::CompactIntsEqualityChecker<1ul>; KeyHashFunc = peloton::index::CompactIntsHasher<1ul>; ValueEqualityChecker = peloton::ItemPointerComparator; ValueHashFunc = peloton::ItemPointerHashFunc]: Assertion `ret == true' failed.`


The full log is as following:
```
travis_time:start:0c6582f0
[0K$ bash ../script/testing/psql/psql_test.sh
Exception Type :: Catalog
Message :: Table foo is not found
peloton: /home/travis/build/cmu-db/peloton/src/index/bwtree_index.cpp:106: bool peloton::index::BWTreeIndex<KeyType, ValueType, KeyComparator, KeyEqualityChecker, KeyHashFunc, ValueEqualityChecker, ValueHashFunc>::CondInsertEntry(const peloton::storage::Tuple*, peloton::ItemPointer*, std::function<bool(const void*)>) [with KeyType = peloton::index::CompactIntsKey<1ul>; ValueType = peloton::ItemPointer*; KeyComparator = peloton::index::CompactIntsComparator<1ul>; KeyEqualityChecker = peloton::index::CompactIntsEqualityChecker<1ul>; KeyHashFunc = peloton::index::CompactIntsHasher<1ul>; ValueEqualityChecker = peloton::ItemPointerComparator; ValueHashFunc = peloton::ItemPointerHashFunc]: Assertion `ret == true' failed.
../script/testing/psql/psql_test.sh: line 59: 15717 Aborted                 (core dumped) bin/peloton -port $PELOTON_PORT > /dev/null
***FAIL***
```

ChTimTsubasa:
@apavlo 
The line in the tests causing error was in an insert statement:
https://github.com/cmu-db/peloton/blob/master/script/testing/dml/basic2.sql#L48
Which caused peloton not passing the insertion:
https://github.com/cmu-db/peloton/blob/master/src/index/bwtree_index.cpp#L106

I tested again using the current master and it still failed on my computer. I am afaid this error is not related to this PR.

ChTimTsubasa:
@apavlo 

Although this time the PR has passed, all I did was only rebase on top of current master. There are still bugs in BW tree, as mentioned in #730. Other PRs could still possibly fail due to the same problem.

I have tested manually with the basic1.sql.
Peloton still crashed when inserted the new tuples at 
https://github.com/cmu-db/peloton/blob/master/script/testing/dml/basic1.sql#L51
This happens both for this PR and current master.



I

PR #729, opened 2017-07-07
Beginning work on CatalogStorageManager

PauloAmora:
This PR starts to separate the catalog logic between catalog stuff (catalog.cpp) and physical data stuff (catalog_storage_manager.cpp). It's mostly moving things to another class and adapting the rest of the system to work along.

- `TestingTransactionUtil` uses the catalog database by default, so catalog must be initialized, even if it's not used (refactor candidate)?

### DIFF
--- a/src/binder/binder_context.cpp
+++ b/src/binder/binder_context.cpp
@@ -12,6 +12,7 @@
 
 #include "parser/table_ref.h"
 #include "binder/binder_context.h"
+#include "catalog/catalog_storage_manager.h"
 #include "catalog/catalog.h"
 
 namespace peloton {
@@ -50,7 +51,7 @@ bool BinderContext::GetColumnPosTuple(
     type::TypeId& value_type) {
   auto db_id = std::get<0>(table_id_tuple);
   auto table_id = std::get<1>(table_id_tuple);
-  auto schema = catalog::Catalog::GetInstance()
+  auto schema = catalog::CatalogStorageManager::GetInstance()
                     ->GetTableWithOid(db_id, table_id)
                     ->GetSchema();
   auto col_pos = schema->GetColumnID(col_name);
--- a/src/binder/binder_context.cpp
+++ b/src/binder/binder_context.cpp
@@ -47,18 +47,22 @@ void BinderContext::AddTable(const std::string db_name,
 
 bool BinderContext::GetColumnPosTuple(
     std::string& col_name, std::tuple<oid_t, oid_t>& table_id_tuple,
-    std::tuple<oid_t, oid_t, oid_t>& col_pos_tuple,
-    type::TypeId& value_type) {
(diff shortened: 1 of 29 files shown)


src/catalog/catalog_storage_manager.cpp
```diff
@@ -0,0 +1,117 @@
...
+CatalogStorageManager::CatalogStorageManager() {}
+
+//===--------------------------------------------------------------------===//
+// GET WITH OID - DIRECTLY GET FROM STORAGE LAYER
+//===--------------------------------------------------------------------===//
+
+/* Find a database using its oid from storage layer,
+ * throw exception if not exists
+ * */
+storage::Database *CatalogStorageManager::GetDatabaseWithOid(
```
mengranwo: Do you mind double checking that whenever other methods invoke Catalog Storage Manager's methods (like GetDatabaseWithOid()), it's always surrounded by try-catch block?

src/binder/binder_context.cpp
```diff
@@ -50,7 +51,7 @@ bool BinderContext::GetColumnPosTuple(
     type::TypeId& value_type) {
   auto db_id = std::get<0>(table_id_tuple);
   auto table_id = std::get<1>(table_id_tuple);
-  auto schema = catalog::Catalog::GetInstance()
+  auto schema = catalog::CatalogStorageManager::GetInstance()
```
mengranwo: @apavlo I remembered we had this discussion before. If HaoXiang is able to start the transaction before bind & optimize phase, we should read catalog table(in this case pg_attribute) to retrieve related information instead of going to storage level and get memory object?

apavlo: Yes that's correct. 

I don't think that we should call this the "CatalogStorageManager".  It should just be the "StorageManager"

PauloAmora: There is a `StorageManager` under storage that handles very low-level storage stuff, such as cpu instructions for various storage media. I'm not sure if this code is actually used, but it is there.

apavlo: Rename current `StorageManager` to `BackendManager`.

PauloAmora: Should I also move the renamed `CatalogStorageManager` to the `storage` namespace, or leave it in catalog, as it is?

src/catalog/catalog_storage_manager.cpp
PauloAmora: Almost everywhere was good, I added a couple of try-catch blocks in the next commit, just making sure it's passing locally.

src/binder/binder_context.cpp
apavlo: It should be in the `storage` namespace.

PauloAmora:
@apavlo It was coveralls' fault, after all, I restarted the build and it is passing.

PR #733, opened 2017-07-07
Transaction handler

haoxianghua:
This pull request is to solve multi-statements in a transaction issue. In the previous implement, each query will implicitly begin a new transaction handler and implicitly aborts or commits itself. Compared to previous one, after you 'BEGIN' a transaction and before you 'COMMIT' that transaction, queries that you input will have same transaction handler. In addition, after 'BEGIN' a transaction, when a query in a transaction aborted, the whole transaction will be aborted and unless you input 'COMMIT' or 'ROLLBACK' in the terminal, any following queries cannot be executed.

The main update is in tcop.cpp. Transaction begins when traffic cop firstly receive a statement(TrafficCop::ExecuteStatement) and I push this transaction handler(concurrency::Transaction *) into a stack(tcop_txn_state) in order to pass required transaction handler when prepare statement(PrepareStatement), execute statement(ExecuteStatementPlan), commit transaction and so on.

Another important change is I declare a member variable (concurrency:: Transaction *) in SqlNodeVisitor. It is convenient to pass transaction handler in binder and insertQueryTree.

I also change some code in catalog.cpp, for example, GetTableWithName. If you cannot find that table, you won't abort that query on catalog level and only throw an exception on that level.

I also modified 10 test cases. In some cases, I replaced simple optimizer by optimizer and I changed other cases to apply to new version tcop.cpp.


haoxianghua:
E: Failed to fetch http://ppa.launchpad.net/boost-latest/ppa/ubuntu/dists/trusty/main/binary-amd64/Packages  404  Not Found
E: Failed to fetch http://ppa.launchpad.net/boost-latest/ppa/ubuntu/dists/trusty/main/binary-i386/Packages  404  Not Found
E: Some index files failed to download. They have been ignored, or old ones used instead
Can you give me some hints how to solve this issue?

plus:
When I run 'CXX=g++-5 PELOTON_BUILD_TYPE=Release COVERALLS=Off', 3 test cases cannot pass, but these test cases can pass other 3 config. 

Thank you very much!!!
@apavlo @yingjunwu @pmenon

src/binder/binder_context.cpp
```diff
@@ -18,9 +18,10 @@
 namespace peloton {
 namespace binder {
 
-void BinderContext::AddTable(const parser::TableRef* table_ref) {
+void BinderContext::AddTable(const parser::TableRef* table_ref,
+                             concurrency::Transaction* consistentTxn) {
   storage::DataTable* table = catalog::Catalog::GetInstance()->GetTableWithName(
-      table_ref->GetDatabaseName(), table_ref->GetTableName());
+      table_ref->GetDatabaseName(), table_ref->GetTableName(), consistentTxn);
```
apavlo: What does "consistent txn" mean? Why isn't just "txn"?

src/binder/binder_context.cpp
```diff
@@ -33,9 +34,10 @@ void BinderContext::AddTable(const parser::TableRef* table_ref) {
 }
 
 void BinderContext::AddTable(const std::string db_name,
-                             const std::string table_name) {
-  storage::DataTable* table =
-      catalog::Catalog::GetInstance()->GetTableWithName(db_name, table_name);
+                             const std::string table_name,
+                             concurrency::Transaction* consistentTxn) {
```
apavlo: Same comment as before. Why isn't it just "txn"?

src/catalog/catalog.cpp
```diff
@@ -748,6 +748,7 @@ storage::DataTable *Catalog::GetTableWithName(const std::string &database_name,
 
   if (database_oid == INVALID_OID) {
     txn_manager.AbortTransaction(txn);  // Implicitly abort txn
+
```
apavlo: Unnecessary whitespace.

src/catalog/catalog.cpp
```diff
@@ -756,17 +757,17 @@ storage::DataTable *Catalog::GetTableWithName(const std::string &database_name,
       TableCatalog::GetInstance()->GetTableOid(table_name, database_oid, txn);
 
   if (table_oid == INVALID_OID) {
-    txn_manager.AbortTransaction(txn);  // Implicitly abort txn
+//    txn_manager.AbortTransaction(txn);  // Implicitly abort txn
```
apavlo: Remove dead code.

src/concurrency/timestamp_ordering_transaction_manager.cpp
```diff
@@ -29,7 +29,7 @@ Spinlock *TimestampOrderingTransactionManager::GetSpinlockField(
     const storage::TileGroupHeader *const tile_group_header,
     const oid_t &tuple_id) {
   return (Spinlock *)(tile_group_header->GetReservedFieldRef(tuple_id) +
-                      LOCK_OFFSET);
```
apavlo: This looks like you just made formatting changes to this file. What logic did you actually change?

src/include/planner/delete_plan.h
```diff
@@ -64,7 +65,9 @@ class DeletePlan : public AbstractPlan {
   }
 
  private:
-  void BuildInitialDeletePlan(parser::DeleteStatement *delete_statemenet);
+  void BuildInitialDeletePlan(
+      parser::DeleteStatement *delete_statemenet,
+      concurrency::Transaction *consistentTxn = nullptr);
```
apavlo: When will the txn be null?

src/include/planner/drop_plan.h
```diff
@@ -32,10 +33,12 @@ class DropPlan : public AbstractPlan {
   DropPlan() = delete;
 
   explicit DropPlan(storage::DataTable *table);
+  // # 623
+  explicit DropPlan(std::string name,
+                    concurrency::Transaction *consistentTxn = nullptr);
```
apavlo: Again, when will the txn be null? Why is that okay?

src/include/tcop/tcop.h
```diff
@@ -44,6 +44,10 @@ class TrafficCop {
   // static singleton method used by tests
   static TrafficCop &GetInstance();
 
+  // # 623
+  //  concurrency::Transaction *consistentTxn;
+  bool single_statement_txn;
```
apavlo: Is this a public variable?

src/optimizer/optimizer.cpp
```diff
@@ -81,22 +82,21 @@ shared_ptr<planner::AbstractPlan> Optimizer::BuildPelotonPlanTree(
...
   bool is_ddl_stmt;
   auto ddl_plan = HandleDDLStatement(parse_tree, is_ddl_stmt);
   if (is_ddl_stmt) {
     return move(ddl_plan);
   }
-
   // Run binder
+  LOG_TRACE("RUN BINDER IN OPTIMIZER");
   auto bind_node_visitor = make_shared<binder::BindNodeVisitor>();
+  (*bind_node_visitor).consistentTxn = consistentTxn;
```
apavlo: This casting seems wrong/unnecessary.

src/optimizer/query_to_operator_transformer.cpp
```diff
@@ -45,18 +45,17 @@ void QueryToOperatorTransformer::Visit(const parser::SelectStatement *op) {
 
   if (op->where_clause != nullptr) {
     SingleTablePredicates where_predicates;
-    util::ExtractPredicates(op->where_clause, where_predicates, join_predicates_);
```
apavlo: This looks like just formatting changes. What logic did you change in this file?

src/tcop/tcop.cpp
```diff
@@ -80,35 +81,32 @@ TrafficCop::TcopTxnState &TrafficCop::GetCurrentTxnState() {
...
+    // # 623
     return result;
   } else {
     // otherwise, the txn has already been aborted
-    return ResultType::ABORTED;
+    LOG_TRACE("final tcop_txn_state_ size() %lu", tcop_txn_state_.size());
+    auto txn = curr_state.first;
+    auto &txn_manager = concurrency::TransactionManagerFactory::GetInstance();
+    auto result = txn_manager.AbortTransaction(txn);
+    // # 623
```
apavlo: Nobody is going to know what this means when looking at the code. Can you write a better comment?

src/tcop/tcop.cpp
```diff
@@ -80,35 +81,32 @@ TrafficCop::TcopTxnState &TrafficCop::GetCurrentTxnState() {
...
+    return result;
+    //      return ResultType::ABORTED;
```
apavlo: Remove dead code.

src/tcop/tcop.cpp
```diff
@@ -133,6 +134,255 @@ ResultType TrafficCop::ExecuteStatement(
...
+//  LOG_DEBUG("Result Type %d", tcop_txn_state_.top().second);
+
+  try {
+    switch (statement->GetQueryType()) {
+      case QueryType::QUERY_BEGIN:
+        return ResultType::SUCCESS;
+      case QueryType::QUERY_COMMIT:
+        return CommitQueryHelper();
+      case QueryType::QUERY_ROLLBACK:
+      LOG_TRACE("abort!!!");
```
apavlo: Remove unhelpful log message.

src/tcop/tcop.cpp
```diff
@@ -133,6 +134,255 @@ ResultType TrafficCop::ExecuteStatement(
...
+    txn = txn_manager.BeginTransaction(thread_id);
+    single_statement_txn = true;
+  } else {
+    // get ptr to current active txn
+    txn = curr_state.first;
+  }
+
+//  auto &curr_state = GetCurrentTxnState();
+//  // if tcop_txn_state is empty, <first, second> first is nullptr
+//  txn = curr_state.first;
```
apavlo: Remove dead code.

test/binder/binder_test.cpp
```diff
@@ -42,13 +42,17 @@ void SetupTables() {
 
   auto& parser = parser::PostgresParser::GetInstance();
   auto& traffic_cop = tcop::TrafficCop::GetInstance();
+
   optimizer::Optimizer optimizer;
 
-  txn = txn_manager.BeginTransaction();
```
apavlo: Remove dead code.

test/executor/delete_test.cpp
```diff
@@ -47,8 +47,16 @@ void ShowTable(std::string database_name, std::string table_name) {
   executor::ExecuteResult status;
   std::vector<type::Value> params;
   std::vector<StatementResult> result;
+
   optimizer::Optimizer optimizer;
+
+//  optimizer::SimpleOptimizer optimizer;
```
apavlo: Why is this line still here? That was deleted in the latest version of the master branch.

test/executor/delete_test.cpp
```diff
@@ -71,7 +80,12 @@ TEST_F(DeleteTests, VariousOperations) {
   catalog::Catalog::GetInstance()->CreateDatabase(DEFAULT_DB_NAME, txn);
   LOG_INFO("Bootstrapping completed!");
 
-  optimizer::Optimizer optimizer;
+
+//  optimizer::SimpleOptimizer optimizer;
```
apavlo: Same comment here as above.

test/executor/update_test.cpp
```diff
@@ -218,12 +227,15 @@ TEST_F(UpdateTests, UpdatingOld) {
   LOG_INFO("Statement executed. Result: %s",
            ResultTypeToString(status.m_result).c_str());
   LOG_INFO("Tuple inserted!");
-  txn_manager.CommitTransaction(txn);
+  traffic_cop.CommitQueryHelper();
+//  txn_manager.CommitTransaction(txn);
```
apavlo: Remove dead code.

test/executor/update_test.cpp
```diff
@@ -218,12 +227,15 @@ TEST_F(UpdateTests, UpdatingOld) {
...
 
   LOG_INFO("%s", table->GetInfo().c_str());
 
   // Now Updating end-to-end
   txn = txn_manager.BeginTransaction();
+  traffic_cop.tcop_txn_state_.emplace(txn, ResultType::SUCCESS);
```
apavlo: This looks like a public variable. That's bad.

test/optimizer/optimizer_test.cpp
```diff
@@ -117,10 +123,12 @@ TEST_F(OptimizerTests, HashJoinTest) {
   LOG_INFO("Statement executed. Result: %s",
            ResultTypeToString(status.m_result).c_str());
   LOG_INFO("Tuple inserted to table_a!");
-  txn_manager.CommitTransaction(txn);
+//  txn_manager.CommitTransaction(txn);
 
   // Inserting a tuple to table_b
   txn = txn_manager.BeginTransaction();
+  traffic_cop.tcop_txn_state_.emplace(txn, ResultType::SUCCESS);
+  traffic_cop.single_statement_txn = true;
```
apavlo: More public class members.

test/sql/drop_sql_test.cpp
```diff
@@ -64,18 +62,17 @@ TEST_F(DropSQLTests, DropTableTest) {
...
 
   // Check the table does not exist
-  try {
-    table = catalog::Catalog::GetInstance()->GetTableWithName(DEFAULT_DB_NAME,
-                                                              "test");
-  } catch (CatalogException &e) {
-    table = nullptr;
-  }
-  EXPECT_EQ(table, nullptr);
+//  try {
```
apavlo: Remove dead code.

test/sql/optimizer_sql_test.cpp
```diff
@@ -124,21 +127,27 @@ class OptimizerSQLTests : public PelotonTest {
...
+
   // Something wrong with column property.
   string query = "SELECT b from test order by c";
-   
+
+  // # 623
   // check for plan node type
   auto select_plan =
       TestingSQLUtil::GeneratePlanWithOptimizer(optimizer, query);
+  LOG_INFO("haoxianghua");
```
apavlo: Remove your name.

test/sql/testing_sql_util.cpp
```diff
@@ -13,6 +13,10 @@
 #include "common/logger.h"
 #include "executor/plan_executor.h"
 #include "optimizer/rule.h"
+
+
```
apavlo: Why the extra spaces?

test/sql/testing_sql_util.cpp
```diff
@@ -54,13 +58,33 @@ ResultType TestingSQLUtil::ExecuteSQLQuery(
...
+  traffic_cop_.optimizer_.reset(new optimizer::Optimizer());
+  LOG_INFO("Tcop_txn_size: %lu", traffic_cop_.tcop_txn_state_.size());
   auto status = traffic_cop_.ExecuteStatement(query, result, tuple_descriptor,
                                               rows_changed, error_message);
   LOG_INFO("Statement executed. Result: %s",
            ResultTypeToString(status).c_str());
   return status;
 }
 
+// 623 Execute a SQL query end-to-end
```
apavlo: Nobody is going to know what "623" means.

test/sql/testing_sql_util.cpp
```diff
@@ -69,6 +93,12 @@ ResultType TestingSQLUtil::ExecuteSQLQueryWithOptimizer(
     std::string &error_message) {
   auto &peloton_parser = parser::PostgresParser::GetInstance();
   std::vector<type::Value> params;
+  // # 623
+  auto& txn_manager = concurrency::TransactionManagerFactory::GetInstance();
+  auto txn = txn_manager.BeginTransaction();
+  traffic_cop_.tcop_txn_state_.emplace(txn, ResultType::SUCCESS);
+  traffic_cop_.single_statement_txn = true;
+  optimizer->consistentTxn = txn;
```
apavlo: This looks like a public data member. Also is this thread-safe?

src/binder/bind_node_visitor.cpp
```diff
@@ -59,7 +58,8 @@ void BindNodeVisitor::Visit(const parser::TableRef *node) {
   }
   // Single table
   else {
-    context_->AddTable(node);
+    LOG_DEBUG("Before Add Table!");
```
pmenon: I think this can be removed.

src/binder/bind_node_visitor.cpp
```diff
@@ -92,7 +92,8 @@ void BindNodeVisitor::Visit(const parser::UpdateStatement *node) {
 void BindNodeVisitor::Visit(const parser::DeleteStatement *node) {
   context_ = std::make_shared<BinderContext>();
 
-  context_->AddTable(node->GetDatabaseName(), node->GetTableName());
+  context_->AddTable(node->GetDatabaseName(), node->GetTableName(),
+                     consistentTxn);
```
pmenon: We don't use camel-case for variable names. Rename to `consistent_txn`

src/catalog/catalog.cpp
```diff
@@ -756,17 +757,17 @@ storage::DataTable *Catalog::GetTableWithName(const std::string &database_name,
...
   }
 
   if (single_statement_txn) {
     txn_manager.CommitTransaction(txn);
   }
-
   return storage_manager->GetTableWithOid(database_oid, table_oid);
-}
 
+}
```
pmenon: Unnecessary whitespace.

src/concurrency/timestamp_ordering_transaction_manager.cpp
```diff
@@ -243,13 +232,13 @@ bool TimestampOrderingTransactionManager::PerformRead(
...
       if (IsOwner(current_txn, tile_group_header, tuple_id) == false) {
-
         // Acquire ownership if we haven't
         if (IsOwnable(current_txn, tile_group_header, tuple_id) == false) {
           // Cannot own
           return false;
         }
-        if (AcquireOwnership(current_txn, tile_group_header, tuple_id) == false) {
+        if (AcquireOwnership(current_txn, tile_group_header, tuple_id) ==
+            false) {
```
pmenon: Unnecessary whitespace.

src/concurrency/timestamp_ordering_transaction_manager.cpp
```diff
@@ -1058,8 +1035,8 @@ ResultType TimestampOrderingTransactionManager::AbortTransaction(
 
         if (old_prev.IsNull() == false) {
           auto old_prev_tile_group_header = catalog::Manager::GetInstance()
-                                                .GetTileGroup(old_prev.block)
-                                                ->GetHeader();
+              .GetTileGroup(old_prev.block)
+              ->GetHeader();
           old_prev_tile_group_header->SetNextItemPointer(
               old_prev.offset, ItemPointer(tile_group_id, tuple_slot));
         }
```
pmenon: It looks like the changes to this file are mostly manual formatting. To be consistent, run the formatter on this file.

src/include/binder/binder_context.h
```diff
@@ -13,6 +13,7 @@
 #include <unordered_map>
 #include <string>
 #include "type/types.h"
+#include "concurrency/transaction.h"
```
pmenon: Forward declare the transaction class and remove the include, since the definition isn't needed in the header.

src/include/common/sql_node_visitor.h
```diff
@@ -12,6 +12,8 @@
 
 #pragma once
 
+#include "concurrency/transaction.h"
```
pmenon: Ditto, forward declare.

src/include/common/sql_node_visitor.h
```diff
@@ -88,6 +90,7 @@ class SqlNodeVisitor {
   virtual void Visit(expression::ParameterValueExpression *expr);
   virtual void Visit(expression::StarExpression *expr);
   virtual void Visit(expression::TupleValueExpression *expr);
+  concurrency::Transaction *consistentTxn = nullptr;
```
pmenon: Rename using snake-case.

src/include/optimizer/abstract_optimizer.h
```diff
@@ -43,6 +44,9 @@ class AbstractOptimizer {
       const std::unique_ptr<parser::SQLStatementList> &parse_tree) = 0;
 
   virtual void Reset(){};
+
+  // # 623
```
pmenon: Remove comment. Rename variable using snake-case.

src/include/planner/analyze_plan.h
```diff
@@ -15,6 +15,7 @@
 #include "planner/abstract_plan.h"
 
 #include <vector>
+#include "concurrency/transaction.h"
```
pmenon: Forward declare transaction.

src/include/planner/analyze_plan.h
```diff
@@ -37,19 +38,22 @@ class AnalyzePlan : public AbstractPlan {
 
   explicit AnalyzePlan(storage::DataTable *table);
 
-  explicit AnalyzePlan(std::string table_name);
+  explicit AnalyzePlan(std::string table_name,
+                       concurrency::Transaction *consistentTxn = nullptr);
```
pmenon: Rename all occurrences to snake-cases.

src/include/tcop/tcop.h
pmenon: Also remove comment.

src/include/tcop/tcop.h
```diff
@@ -73,8 +77,38 @@ class TrafficCop {
...
+      const std::vector<int> &result_format, const size_t thread_id = 0);
+
+  // InitBindPrepStmt - Prepare and bind a query from a query string
+  std::shared_ptr<Statement> PrepareStatementJDBC(const std::string &statement_name,
+                                              const std::string &query_string,
                                               std::string &error_message);
 
+
+
+
```
pmenon: Why exactly is there a separate set of methods in the traffic cop for JDBC?

src/include/tcop/tcop.h
```diff
@@ -86,27 +120,41 @@ class TrafficCop {
 
   int BindParameters(std::vector<std::pair<int, std::string>> &parameters,
                      Statement **stmt, std::string &error_message);
+  // for test, these should be private
+  typedef std::pair<concurrency::Transaction *, ResultType> TcopTxnState;
+  std::stack<TcopTxnState> tcop_txn_state_;
+  ResultType CommitQueryHelper();
+  std::unique_ptr<optimizer::AbstractOptimizer> optimizer_;
```
pmenon: Move these back to private.

src/wire/packet_manager.cpp
```diff
@@ -1029,42 +1030,42 @@ bool PacketManager::ProcessPacket(InputPacket *pkt, const size_t thread_id) {
   // we see a SYNC
   switch (pkt->msg_type) {
     case NetworkMessageType::SIMPLE_QUERY_COMMAND: {
-      LOG_TRACE("SIMPLE_QUERY_COMMAND");
+      LOG_DEBUG("SIMPLE_QUERY_COMMAND");
```
pmenon: Can you change these back to trace?

yingjunwu:
@haoxianghua I will pass through the code again once you have addressed all the raised comments.

src/binder/binder_context.cpp
haoxianghua: I use consistentTxn because I want to express this variables to keep txn handle consistent from "BEGIN" to "COMMIT"

yingjunwu: I think it would be better if you could change `consistentTxn` to `consistent_txn`. Just check the naming convention.

src/concurrency/timestamp_ordering_transaction_manager.cpp
haoxianghua: I didn't change this file

src/include/planner/delete_plan.h
haoxianghua: It shouldn't be null in psql part. But for JDBC part, it needs it to be null. I didn't change the logic of JDBC part.

src/include/tcop/tcop.h
haoxianghua: This is private variable, I set it to public in order to pass test file.

src/optimizer/optimizer.cpp
haoxianghua: You are right, I change it to bind_node_visitor->consistentTxn

src/optimizer/query_to_operator_transformer.cpp
haoxianghua: I didn't change logic in this file

src/include/tcop/tcop.h
apavlo: Do not create public class variables.

src/binder/binder_context.cpp
apavlo: I don't think it should be called "consistent" anything. It doesn't make sense. It seems to imply that other txns are not consistent. It's confusing.

Change it to be just "txn"

src/optimizer/query_to_operator_transformer.cpp
apavlo: Don't commit changes to files that you didn't modify.

src/include/tcop/tcop.h
yingjunwu: also, member variables should be named as `single_statement_txn_`, rather than `single_statement_txn`.

src/include/tcop/tcop.h
haoxianghua: That's a good question. I hacked code based on psql and people told me jdbc has solved transaction handle. So I separate some methods in traffic cop.

src/include/binder/binder_context.h
haoxianghua: AddTable in this file need consistent_txn as input

test/sql/testing_sql_util.cpp
haoxianghua: I think it is thread-safe. one worker thread has several connections, one connection has one tcop and one tcop has one transaction.

yingjunwu:
After talking to @haoxianghua, we find there exists some bugs in JDBC. So let's have a discussion in tmr's group meeting. @apavlo

apavlo:
@haoxianghua Run TPC-C on this PR to make sure that it doesn't crash and then we can merge this.

test/sql/optimizer_sql_test.cpp
```diff
@@ -124,17 +127,19 @@ class OptimizerSQLTests : public PelotonTest {
...
-  
+
   // Something wrong with column property.
   string query = "SELECT b from test order by c";
-   
+
   // check for plan node type
   auto select_plan =
       TestingSQLUtil::GeneratePlanWithOptimizer(optimizer, query);
+  delete optimizer->txn;
```
pmenon: Why is this here?

test/sql/optimizer_sql_test.cpp
```diff
@@ -165,12 +170,12 @@ TEST_F(OptimizerSQLTests, SimpleSelectTest) {
...
   // Testing predicate
   TestUtil("SELECT c, b from test where a=1", {"333", "22"}, false);
-  
+
   // Something wrong with column property.
   query = "SELECT a, b, c from test order by a + c";
 
   // check for plan node type
-  select_plan = TestingSQLUtil::GeneratePlanWithOptimizer(optimizer, query);
+//  select_plan = TestingSQLUtil::GeneratePlanWithOptimizer(optimizer, query);
```
pmenon: Why are these tests commented out?

src/include/planner/drop_plan.h
```diff
@@ -32,10 +33,12 @@ class DropPlan : public AbstractPlan {
   DropPlan() = delete;
 
   explicit DropPlan(storage::DataTable *table);
+  // # 623
```
pmenon: Remove comment.

test/binder/binder_test.cpp
```diff
@@ -71,10 +75,14 @@ TEST_F(BinderCorrectnessTest, SelectStatementTest) {
   // Test regular table name
   LOG_INFO("Parsing sql query");
   unique_ptr<binder::BindNodeVisitor> binder(new binder::BindNodeVisitor());
+  // # 623
```
pmenon: Remove comment.

test/binder/binder_test.cpp
```diff
@@ -140,7 +148,10 @@ TEST_F(BinderCorrectnessTest, SelectStatementTest) {
 
   // Check alias ambiguous
   LOG_INFO("Checking duplicate alias and table name.");
-  binder.reset(new binder::BindNodeVisitor());
+  delete binder->txn;
+  // # 623
```
pmenon: Remove comment. Also, why is there a raw delete here?

test/binder/binder_test.cpp
```diff
@@ -152,7 +163,10 @@ TEST_F(BinderCorrectnessTest, SelectStatementTest) {
   }
 
   // Test select from different table instances from the same physical schema
-  binder.reset(new binder::BindNodeVisitor());
+  delete binder->txn;
+  // # 623
```
pmenon: Remove all these '623' comments. People won't know what this refers to.

src/include/optimizer/abstract_optimizer.h
```diff
@@ -43,6 +44,8 @@ class AbstractOptimizer {
       const std::unique_ptr<parser::SQLStatementList> &parse_tree) = 0;
 
   virtual void Reset(){};
+
+  concurrency::Transaction *txn = nullptr;
```
pmenon: 1. Move this to private
2. Initialize in constructor
3. Remove "transaction.h" include and forward declare `concurrency::Transaction`

pmenon: 1. Move this to private.
2. Initialize in constructor.
3. Don't include transaction header, instead forward declare it here.

src/include/optimizer/query_to_operator_transformer.h
```diff
@@ -52,6 +53,7 @@ class QueryToOperatorTransformer : public SqlNodeVisitor {
   void Visit(const parser::UpdateStatement *op) override;
   void Visit(const parser::CopyStatement *op) override;
   void Visit(const parser::AnalyzeStatement *op) override;
+  concurrency::Transaction *txn = nullptr;
```
pmenon: 1. Move to private.
2. Initialize in constructor.
3. Forward declare `Transaction`.

src/include/planner/abstract_plan.h
```diff
@@ -26,6 +26,7 @@
 #include "type/serializer.h"
 #include "type/types.h"
 #include "type/value.h"
+#include "concurrency/transaction.h"
```
pmenon: This isn't used in this file.

src/include/planner/drop_plan.h
pmenon: Remove comment.

src/include/planner/update_plan.h
```diff
@@ -79,7 +79,8 @@ class UpdatePlan : public AbstractPlan {
  private:
   // Initialize private members and construct colum_ids given a UpdateStatement.
   void BuildInitialUpdatePlan(const parser::UpdateStatement *parse_tree,
-                              std::vector<oid_t> &columns);
+                              std::vector<oid_t> &columns,
+                              concurrency::Transaction *txn = nullptr);
```
pmenon: Can transaction be NULL? Should it?

src/optimizer/optimizer.cpp
```diff
@@ -81,22 +82,21 @@ shared_ptr<planner::AbstractPlan> Optimizer::BuildPelotonPlanTree(
   unique_ptr<planner::AbstractPlan> child_plan = nullptr;
 
   auto parse_tree = parse_tree_list->GetStatements().at(0);
-
+//  LOG_TRACE("HANDLE DDL IN OPTIMIZER");
```
pmenon: Remove commented out code.

test/binder/binder_test.cpp
pmenon: Remove '623'. People will not know what you're referring to.

test/binder/binder_test.cpp
```diff
@@ -152,7 +163,10 @@ TEST_F(BinderCorrectnessTest, SelectStatementTest) {
   }
 
   // Test select from different table instances from the same physical schema
-  binder.reset(new binder::BindNodeVisitor());
+  delete binder->txn;
```
pmenon: Why are you performing a raw delete here?

test/binder/binder_test.cpp
```diff
@@ -167,7 +181,10 @@ TEST_F(BinderCorrectnessTest, SelectStatementTest) {
 
   // Test alias and select_list
   LOG_INFO("Checking select_list and table alias binding");
-  binder.reset(new binder::BindNodeVisitor());
+  delete binder->txn;
```
pmenon: Why raw delete here? If this is cleanup, it shouldn't happen here. More importantly, I don't think this is correct.

test/binder/binder_test.cpp
```diff
@@ -178,10 +195,10 @@ TEST_F(BinderCorrectnessTest, SelectStatementTest) {
   tupleExpr =
       (expression::TupleValueExpression*)(selectStmt->select_list->at(1));
   EXPECT_EQ(tupleExpr->GetBoundOid(), make_tuple(db_oid, tableB_oid, 1));
-
+  delete binder->txn;
```
pmenon: Why?

test/binder/binder_test.cpp
```diff
@@ -178,10 +195,10 @@ TEST_F(BinderCorrectnessTest, SelectStatementTest) {
...
   // Delete the test database
-  auto& txn_manager = concurrency::TransactionManagerFactory::GetInstance();
-  auto txn = txn_manager.BeginTransaction();
+//  auto& txn_manager = concurrency::TransactionManagerFactory::GetInstance();
```
pmenon: Remove commented out code.

test/sql/optimizer_sql_test.cpp
haoxianghua: optimizer has member variable txn. If I don't delete it, the txn will be the previous txn, because in this test, optimizer is globe variable.

test/sql/optimizer_sql_test.cpp
haoxianghua: The variable 'select_plan' has never been used in the following part. '//  EXPECT_EQ(select_plan->GetPlanNodeType(), PlanNodeType::ORDERBY);' this test is also commented out by others.

test/binder/binder_test.cpp
haoxianghua: It is similar as 'delete optimizer->txn;'.

src/include/optimizer/abstract_optimizer.h
haoxianghua: I declare it in abstract optimizer because there was simple optimizer before. I have change it as you required but I don't know why you want me initialize in constructor. If someone add new constructor or if there are many constructor, it is kinda trouble than initialize here?

haoxianghua: And I also think txn is optimizer's quality. You should know what txn you optimizer each time. Why I cannot declare it in high level?

src/include/planner/update_plan.h
haoxianghua: This function want to pass the transaction to addTableWithName(), it transaction is null, addTableWIthName will begin a new transaction.

src/include/binder/bind_node_visitor.h
```diff
@@ -57,6 +58,7 @@ class BindNodeVisitor : public SqlNodeVisitor {
   void Visit(expression::CaseExpression *expr) override;
   // void Visit(const expression::ConstantValueExpression *expr) override;
   void Visit(expression::TupleValueExpression *expr) override;
+  concurrency::Transaction *txn = nullptr;
```
apavlo: Why are you declaring this here? Why isn't it passed in as an argument?

src/include/binder/binder_context.h
```diff
@@ -14,6 +14,7 @@
 #include <string>
 
 #include "type/types.h"
+#include "concurrency/transaction.h"
```
apavlo: I agree with @pmenon. You should forward declare the `concurrency::Transaction` class.

src/include/optimizer/optimizer.h
```diff
@@ -52,11 +53,14 @@ class Optimizer : public AbstractOptimizer {
 
   Optimizer();
 
+
   std::shared_ptr<planner::AbstractPlan> BuildPelotonPlanTree(
       const std::unique_ptr<parser::SQLStatementList> &parse_tree) override;
 
   void Reset() override;
 
+  concurrency::Transaction *txn = nullptr;
```
apavlo: I don't think the optimizer should have a txn data member. It should be passed in to the functions.

@hzxa21 What are the implications of this?

src/include/planner/analyze_plan.h
```diff
@@ -37,19 +38,22 @@ class AnalyzePlan : public AbstractPlan {
 
   explicit AnalyzePlan(storage::DataTable *table);
 
-  explicit AnalyzePlan(std::string table_name);
+  explicit AnalyzePlan(std::string table_name,
+                       concurrency::Transaction *txn = nullptr);
```
apavlo: When is the txn going to be null? I thought that the whole point of this PR was so that we explicitly pass the txn pointer to everything?

src/include/planner/delete_plan.h
```diff
@@ -64,7 +65,9 @@ class DeletePlan : public AbstractPlan {
   }
 
  private:
-  void BuildInitialDeletePlan(parser::DeleteStatement *delete_statemenet);
+  void BuildInitialDeletePlan(
+      parser::DeleteStatement *delete_statemenet,
+      concurrency::Transaction *txn = nullptr);
```
apavlo: Same comment as above. When is txn null?

src/include/planner/update_plan.h
apavlo: @pmenon's comment is correct. When will the txn be null? Why will it be null?

src/include/tcop/tcop.h
```diff
@@ -22,7 +22,7 @@
 #include "common/statement.h"
 #include "concurrency/transaction.h"
 #include "executor/plan_executor.h"
-#include "optimizer/abstract_optimizer.h"
```
apavlo: Why did you change this? Do we still need `AbstractOptimizer`?

src/optimizer/optimizer.cpp
```diff
@@ -115,7 +114,7 @@ shared_ptr<planner::AbstractPlan> Optimizer::BuildPelotonPlanTree(
     ExprMap output_expr_map;
     auto best_plan = ChooseBestPlan(root_id, properties, &output_expr_map);
     if (best_plan == nullptr) return nullptr;
-
+
```
apavlo: Unnecessary newline

src/optimizer/optimizer.cpp
```diff
@@ -184,8 +184,8 @@ unique_ptr<planner::AbstractPlan> Optimizer::HandleDDLStatement(
     }
     case StatementType::ANALYZE: {
       LOG_TRACE("Adding Analyze plan...");
-      unique_ptr<planner::AbstractPlan> analyze_plan(
-        new planner::AnalyzePlan((parser::AnalyzeStatement *)tree));
```
apavlo: Use `static_cast`

src/optimizer/optimizer.cpp
```diff
@@ -184,8 +184,8 @@ unique_ptr<planner::AbstractPlan> Optimizer::HandleDDLStatement(
...
+      unique_ptr<planner::AbstractPlan> analyze_plan(new planner::AnalyzePlan(
+          (parser::AnalyzeStatement *)tree, txn));
```
apavlo: Same here. Use `static_cast`

src/optimizer/optimizer.cpp
```diff
@@ -199,13 +199,13 @@ unique_ptr<planner::AbstractPlan> Optimizer::HandleDDLStatement(
...
       is_ddl_stmt = false;
   }
-
   return ddl_plan;
 }
 
 shared_ptr<GroupExpression> Optimizer::InsertQueryTree(
     parser::SQLStatement *tree) {
   QueryToOperatorTransformer converter;
+  converter.txn = txn;
```
apavlo: This is gross. Why wouldn't you just pass it into the constructor?

src/tcop/tcop.cpp
```diff
@@ -36,6 +36,7 @@
 #include <include/parser/postgresparser.h>
 
 
+
```
apavlo: Unnecessary newline.

src/tcop/tcop.cpp
```diff
@@ -133,14 +126,255 @@ ResultType TrafficCop::ExecuteStatement(
...
+  if (status == ResultType::SUCCESS) {
+    LOG_TRACE("Execution succeeded!");
+    tuple_descriptor = statement->GetTupleDescriptor();
+  } else {
+    LOG_TRACE("Execution failed!");
+  }
+
+  return status;
+}
+// Might be removed
```
apavlo: Please explain what you mean by this comment.

src/tcop/tcop.cpp
```diff
@@ -133,14 +126,255 @@ ResultType TrafficCop::ExecuteStatement(
...
+}
+
+executor::ExecuteResult TrafficCop::ExecuteStatementPlan(
+    const planner::AbstractPlan *plan, const std::vector<type::Value> &params,
+    std::vector<StatementResult> &result, const std::vector<int> &result_format,
+    const size_t thread_id UNUSED_ATTRIBUTE) {
+  concurrency::Transaction *txn;
+  bool init_failure = false;
+  executor::ExecuteResult p_status;
+
```
apavlo: Unnecessary newline.

test/binder/binder_test.cpp
```diff
@@ -59,8 +62,9 @@ void SetupTables() {
         statement->GetPlanTree().get(), params, result, result_format);
     LOG_INFO("Table create result: %s",
              ResultTypeToString(status.m_result).c_str());
+    traffic_cop.CommitQueryHelper();
   }
-  txn_manager.CommitTransaction(txn);
+//  txn_manager.CommitTransaction(txn);
```
apavlo: Remove dead code.

test/binder/binder_test.cpp
apavlo: @pmenon is right. This is not how you do this. It should be in the deconstructor.

test/binder/binder_test.cpp
apavlo: @pmenon is correct.

test/binder/binder_test.cpp
```diff
@@ -214,12 +236,12 @@ TEST_F(BinderCorrectnessTest, DeleteStatementTest) {
...
   LOG_INFO("Checking second condition in where clause");
   tupleExpr = (expression::TupleValueExpression*)deleteStmt->expr->GetChild(1)
-                  ->GetChild(0);
+      ->GetChild(0);
   EXPECT_EQ(tupleExpr->GetBoundOid(), make_tuple(db_oid, tableB_oid, 1));
 
   // Delete the test database
-  auto& txn_manager = concurrency::TransactionManagerFactory::GetInstance();
-  auto txn = txn_manager.BeginTransaction();
+//  auto& txn_manager = concurrency::TransactionManagerFactory::GetInstance();
```
apavlo: Remove dead code.

test/executor/create_index_test.cpp
```diff
@@ -126,10 +130,12 @@ TEST_F(CreateIndexTests, CreatingIndex) {
   LOG_INFO("Statement executed. Result: %s",
            ResultTypeToString(status.m_result).c_str());
   LOG_INFO("Tuple inserted!");
-  txn_manager.CommitTransaction(txn);
+  traffic_cop.CommitQueryHelper();
+//  txn_manager.CommitTransaction(txn);
```
apavlo: Remove dead code.

test/executor/create_index_test.cpp
```diff
@@ -153,7 +159,8 @@ TEST_F(CreateIndexTests, CreatingIndex) {
   LOG_INFO("Statement executed. Result: %s",
            ResultTypeToString(status.m_result).c_str());
   LOG_INFO("INDEX CREATED!");
-  txn_manager.CommitTransaction(txn);
+  traffic_cop.CommitQueryHelper();
+//  txn_manager.CommitTransaction(txn);
```
apavlo: Remove dead code.

test/executor/delete_test.cpp
```diff
@@ -71,7 +77,12 @@ TEST_F(DeleteTests, VariousOperations) {
...
+  std::unique_ptr<optimizer::AbstractOptimizer> optimizer;
+  optimizer.reset(new optimizer::Optimizer);
+
```
apavlo: Unnecessary newline.

test/executor/update_test.cpp
```diff
@@ -152,21 +152,25 @@ TEST_F(UpdateTests, UpdatingOld) {
   catalog->CreateDatabase(DEFAULT_DB_NAME, txn);
   LOG_INFO("Bootstrapping completed!");
 
-  optimizer::Optimizer optimizer;
+  std::unique_ptr<optimizer::AbstractOptimizer> optimizer;
+  optimizer.reset(new optimizer::Optimizer);
+
```
apavlo: More unnecessary newlines.

test/executor/update_test.cpp
```diff
@@ -152,21 +152,25 @@ TEST_F(UpdateTests, UpdatingOld) {
...
   auto& traffic_cop = tcop::TrafficCop::GetInstance();
 
+
   // Create a table first
   LOG_INFO("Creating a table...");
-  auto id_column = catalog::Column(type::TypeId::INTEGER,
-                                   type::Type::GetTypeSize(type::TypeId::INTEGER),
+  auto id_column = catalog::Column(type::INTEGER,
+                                   type::Type::GetTypeSize(type::INTEGER),
```
apavlo: This seems incorrect. All of the code uses `type::TypeId::INTEGER` but you have `type::INTEGER`. Why did you change this?

test/executor/update_test.cpp
```diff
@@ -152,21 +152,25 @@ TEST_F(UpdateTests, UpdatingOld) {
...
                                    "dept_id", true);
   catalog::Constraint constraint(ConstraintType::PRIMARY, "con_primary");
   id_column.AddConstraint(constraint);
   auto manager_id_column = catalog::Column(
-      type::TypeId::INTEGER, type::Type::GetTypeSize(type::TypeId::INTEGER),
+      type::INTEGER, type::Type::GetTypeSize(type::INTEGER),
       "manager_id", true);
   auto name_column =
-      catalog::Column(type::TypeId::VARCHAR, 32, "dept_name", false);
+      catalog::Column(type::VARCHAR, 32, "dept_name", false);
```
apavlo: Same comment here for `type::VARCHAR`.

test/include/sql/testing_sql_util.h
```diff
@@ -40,28 +40,34 @@ class TestingSQLUtil {
...
+                                    int &rows_affected);
+
 
   // Execute a SQL query end-to-end with the specific optimizer
   // Note: right now this is not executed in the context of a transaction, we
   // may want to pass a transaction pointer here if that API is exposed after
   // the refactor by Siddharth
   static ResultType ExecuteSQLQueryWithOptimizer(
-      std::unique_ptr<optimizer::AbstractOptimizer> &optimizer,
+      std::unique_ptr<optimizer::Optimizer> &optimizer,
```
apavlo: Why did you change this from `AbstractOptimizer`?

test/sql/drop_sql_test.cpp
```diff
@@ -50,11 +50,9 @@ TEST_F(DropSQLTests, DropTableTest) {
 
   // Insert and query from that table
   TestingSQLUtil::ExecuteSQLQuery("INSERT INTO test VALUES (1, 10);", result,
-                                  tuple_descriptor, rows_affected,
-                                  error_message);
```
apavlo: Why did you remove `error_message`?

test/sql/optimizer_sql_test.cpp
apavlo: This is incorrect. This is not how to do this in C++.

test/sql/optimizer_sql_test.cpp
apavlo: @haoxianghua What is this test doing? Should we be using it?

test/sql/testing_sql_util.cpp
```diff
@@ -61,14 +61,34 @@ ResultType TestingSQLUtil::ExecuteSQLQuery(
...
+  LOG_INFO("Statement executed. Result: %s",
+           ResultTypeToString(status).c_str());
+
+  return status;
+}
+
 // Execute a SQL query end-to-end with the specific optimizer
 ResultType TestingSQLUtil::ExecuteSQLQueryWithOptimizer(
-    std::unique_ptr<optimizer::AbstractOptimizer> &optimizer,
+    std::unique_ptr<optimizer::Optimizer> &optimizer,
```
apavlo: Why did you change this from `AbstractOptimizer`?

src/concurrency/timestamp_ordering_transaction_manager.cpp
yingjunwu: I think the former formatting scheme is better. Maybe you could change it back.

src/concurrency/timestamp_ordering_transaction_manager.cpp
```diff
@@ -39,17 +39,17 @@ cid_t TimestampOrderingTransactionManager::GetLastReaderCommitId(
     const storage::TileGroupHeader *const tile_group_header,
     const oid_t &tuple_id) {
   return *(cid_t *)(tile_group_header->GetReservedFieldRef(tuple_id) +
-                    LAST_READER_OFFSET);
+      LAST_READER_OFFSET);
```
yingjunwu: Same. The former formatting scheme is probably better.

src/concurrency/timestamp_ordering_transaction_manager.cpp
```diff
@@ -39,17 +39,17 @@ cid_t TimestampOrderingTransactionManager::GetLastReaderCommitId(
...
     const storage::TileGroupHeader *const tile_group_header,
-    const oid_t &tuple_id, 
-    const cid_t &current_cid, 
+    const oid_t &tuple_id,
+    const cid_t &current_cid,
     const bool is_owner) {
   // get the pointer to the last_reader_cid field.
   cid_t *ts_ptr = (cid_t *)(tile_group_header->GetReservedFieldRef(tuple_id) +
-                            LAST_READER_OFFSET);
+      LAST_READER_OFFSET);
```
yingjunwu: again...

src/concurrency/timestamp_ordering_transaction_manager.cpp
yingjunwu: It seems that all the changes are about formatting. I suggest reverting all these changes so that we can trace the person who really changes the logic of the code.

src/include/binder/bind_node_visitor.h
yingjunwu: @apavlo is right. Please write down the reason why you declare it as member variable. Also, please try to make member variables private.

src/include/binder/binder_context.h
```diff
@@ -37,10 +38,12 @@ class BinderContext {
   BinderContext() { upper_context = nullptr; }
 
   // Update the table alias map given a table reference (in the from clause)
-  void AddTable(const parser::TableRef* table_ref);
+  void AddTable(const parser::TableRef* table_ref,
+                concurrency::Transaction* txn = nullptr);
```
yingjunwu: I am not sure whether it is a good practice to set the default value as `nullptr`. Developers who are not familiar with the transaction logic may simply ignore passing the `txn` parameter, which subsequentally leads to the transaction bugs we found previously.

src/include/binder/binder_context.h
```diff
@@ -37,10 +38,12 @@ class BinderContext {
...
 
   // Update the table alias map given a table reference (in the from clause)
-  void AddTable(const std::string db_name, const std::string table_name);
+  void AddTable(const std::string db_name, const std::string table_name,
```
yingjunwu: same here. We should not give default value to `txn` parameter.

src/include/common/sql_node_visitor.h
```diff
@@ -75,7 +75,7 @@ class SqlNodeVisitor {
   virtual void Visit(const parser::TransactionStatement *) {}
   virtual void Visit(const parser::UpdateStatement *) {}
   virtual void Visit(const parser::CopyStatement *) {}
-  virtual void Visit(const parser::AnalyzeStatement *) {};
+  virtual void Visit(const parser::AnalyzeStatement *){};
```
yingjunwu: Revert this change pls.

src/include/planner/analyze_plan.h
yingjunwu: Yes. Same comments as my previous one. We should not allow default parameter value.

src/include/planner/delete_plan.h
yingjunwu: agreed here.

src/include/planner/drop_plan.h
```diff
@@ -33,9 +34,11 @@ class DropPlan : public AbstractPlan {
 
   explicit DropPlan(storage::DataTable *table);
 
-  explicit DropPlan(std::string name);
+  explicit DropPlan(std::string name,
```
yingjunwu: same issue.

src/include/tcop/tcop.h
```diff
@@ -73,8 +73,42 @@ class TrafficCop {
   // InitBindPrepStmt - Prepare and bind a query from a query string
   std::shared_ptr<Statement> PrepareStatement(const std::string &statement_name,
                                               const std::string &query_string,
+                                              std::string &error_message,
+                                              const size_t thread_id = 0);
+
+  // PortalExec - Execute query string
+  ResultType ExecuteStatementJDBC(const std::string &query,
```
yingjunwu: probably rename it as `ExecuteStatementExtended` or something else? The `ExecuteStatement` can also be invoked by jdbc.

src/include/tcop/tcop.h
```diff
@@ -87,24 +121,32 @@ class TrafficCop {
   int BindParameters(std::vector<std::pair<int, std::string>> &parameters,
                      Statement **stmt, std::string &error_message);
 
+  void SetTcopTxnState(concurrency::Transaction * txn) {
+    tcop_txn_state_.emplace(txn, ResultType::SUCCESS);
+  }
+
+  ResultType CommitQueryHelper();
+
+  bool prepared_ = false;
```
yingjunwu: Please make it private. Also, why should we have this variable? Please give comments here.

src/optimizer/optimizer.cpp
```diff
@@ -83,22 +84,20 @@ shared_ptr<planner::AbstractPlan> Optimizer::BuildPelotonPlanTree(
   unique_ptr<planner::AbstractPlan> child_plan = nullptr;
 
   auto parse_tree = parse_tree_list->GetStatements().at(0);
-
+//  LOG_TRACE("HANDLE DDL IN OPTIMIZER");
   // Handle ddl statement
   bool is_ddl_stmt;
```
yingjunwu: what's the initial value of `is_ddl_stmt`?

src/include/binder/bind_node_visitor.h
haoxianghua: First, it is thread-safe. Second, in bind_node_visitor, it defines many Visit function and there are only two Visit function use txn, however, the Visit functions is recursive, so pass in as an argument will change binder_node_visitor a lot for example, use if..else.. at beginning of every Visit function to see if this node need txn. Compared with pass in as an argument, member variable won't change code so much and brief.

src/include/binder/binder_context.h
haoxianghua: I need to pass txn in AddTable because in this function, getTableWithName needs txn. And this is the most front level I can find to declare the concurrency::Transaction.

src/include/optimizer/optimizer.h
haoxianghua: I can modify txn to argument in optimizer. And please give me a reason why do you hate member variable so much?

src/include/planner/delete_plan.h
haoxianghua: I also agree here and plus there are also redundant function in analyze_plan

src/include/tcop/tcop.h
haoxianghua: I don't know if we still need AbstractOptimizer. On my opinion, we use AbstractOptimizer because we can easily change optimizer from simple to not simple optimizer mode. And in this part, pmenon doesn't allow me declare txn in abstract_optimizer, but I still need use txn, so I can only change it to optimizer.

src/include/planner/analyze_plan.h
haoxianghua: You are right! Changed it.

src/optimizer/optimizer.cpp
haoxianghua: I am sorry, because there are too many newlines in peloton and I don't know which is necessary and which is not. After I debug and remove the log statement, sometimes I just left newline

src/include/binder/binder_context.h
malin1993ml: @haoxianghua Forward declaration means that you type the declaration of the class at the top of this .h file by yourself, and include the real declaration of the class in the .cpp file. This will significantly reduce coupling and compilation time of the code base. Just as how we declare TableRef here:
https://github.com/cmu-db/peloton/blob/master/src/include/binder/binder_context.h#L21

You can google "forward declaration c++" for more details.

test/executor/update_test.cpp
haoxianghua: I also don't know, once I rebase it and merge a new type.h and the specified define is in src/include/type/type_id.h

test/sql/testing_sql_util.cpp
haoxianghua: I also don't know who change optimizer_sql_test...

src/tcop/tcop.cpp
```diff
@@ -133,14 +126,255 @@ ResultType TrafficCop::ExecuteStatement(
...
 
-  // Prepare the statement
   std::string unnamed_statement = "unnamed";
   auto statement = PrepareStatement(unnamed_statement, query, error_message);
+  // After prepareStatement, statements that might be aborted in prepareStatement will be filtered out.
+  // for example, create a table but that table exists; drop a table but that table has been dropped
+  if (statement.get() == nullptr) {
+    if (single_statement_txn_) {
+      LOG_TRACE("SINGLE ABORT!");
+      rows_changed = 0;
```
haoxianghua: I set it to zero because some test use rows_changed and I cannot pass that test

src/tcop/tcop.cpp
```diff
@@ -133,14 +126,255 @@ ResultType TrafficCop::ExecuteStatement(
...
+    single_statement_txn_ = false;
+    // multi-statment txn has been aborted, just block it
+    if (tcop_txn_state_.top().second == ResultType::ABORTED) {
+      return nullptr;
+    }
+  } else {
+    // --BEGIN
+    // --single-statement transaction
+    // ----prepareStatement abort
+    if (query_string == "BEGIN") {  // only begin a new transaction
```
haoxianghua: Yes, you are right. When I rewrite packet_manager, I will solve this issue. And I also didn't consider lowercase and uppercase.

src/tcop/tcop.cpp
seojungmin: The test cases should be modified to see whether the transaction is aborted or not, rather than checking rows_changed to be zero when aborted.  This fix is a hack and so incorrect, since I have seen cases where they do not set rows_changed zero when they fail.  I recommend to remove this, and fix the test cases.

src/tcop/tcop.cpp
seojungmin: The packet manager is not supposed to parse the packet message content, and so it will not be capable to handle this.  That is my understanding of the Peloton to-be layered architecture, and let me know if I am wrong.   For this, I wonder whether you could do these transaction-related check-ups and settings after building the parse tree at line 327.  That will remove first this string comparison, and second let all the query content parsing done inside the parser.

src/tcop/tcop.cpp
haoxianghua: I totally agree with you. Some test cases are not unified, and I also want to change it. But now my urgent assignment is to rewrite packet_manager and I must modify many functions from packet_manager and tcop which means I need to change many test cases again.

test/sql/optimizer_sql_test.cpp
```diff
@@ -316,22 +321,22 @@ TEST_F(OptimizerSQLTests, DDLSqlTest) {
...
-    EXPECT_TRUE(false);
-  } catch (Exception& e) {
-    LOG_INFO("Correct! Exception(%s) catched", e.what());
-  }
+//  try {
+//    catalog::Catalog::GetInstance()->GetTableWithName(DEFAULT_DB_NAME, "test2");
+//    EXPECT_TRUE(false);
+//  } catch (Exception& e) {
+//    LOG_INFO("Correct! Exception(%s) catched", e.what());
+//  }
```
haoxianghua: This issue is really tricky. In previous, if you use gettablewithname and don't pass txn in it. It will begin a transaction and end that transaction in it. However, as I want to keep txn consistent, I end the transaction outside the gettablewithname in tcop when I receive 'COMMIT'. But test file doesn't care, it just arbitrates to call gettablewithname and do not pass any txn.

src/tcop/tcop.cpp
haoxianghua: Thanks for your suggestion. I find out I wrote redundant 'query == BEGIN' code. Actually, when a statement is newed, its construct has checked up query's type.

src/catalog/catalog.cpp
```diff
@@ -749,7 +749,7 @@ storage::DataTable *Catalog::GetTableWithName(const std::string &database_name,
       DatabaseCatalog::GetInstance()->GetDatabaseOid(database_name, txn);
 
   if (database_oid == INVALID_OID) {
-    txn_manager.AbortTransaction(txn);  // Implicitly abort txn
+//    txn_manager.AbortTransaction(txn);  // Implicitly abort txn
```
apavlo: Remove dead code.

src/planner/update_plan.cpp
```diff
@@ -118,7 +119,7 @@ void UpdatePlan::BuildInitialUpdatePlan(
 UpdatePlan::UpdatePlan(const parser::UpdateStatement *parse_tree)
     : update_primary_key_(false) {
   std::vector<oid_t> column_ids;
-  BuildInitialUpdatePlan(parse_tree, column_ids);
+  BuildInitialUpdatePlan(parse_tree, column_ids, nullptr);
```
apavlo: Why is this null? I would add a comment to explain what's going on and why you're not passing in `txn`.

src/planner/update_plan.cpp
```diff
@@ -143,7 +144,7 @@ UpdatePlan::UpdatePlan(const parser::UpdateStatement *parse_tree,
                        std::vector<type::Value> &values, oid_t &index_id)
     : update_primary_key_(false) {
   std::vector<oid_t> column_ids;
-  BuildInitialUpdatePlan(parse_tree, column_ids);
+  BuildInitialUpdatePlan(parse_tree, column_ids, nullptr);
```
apavlo: Again, why is this null?

test/binder/binder_test.cpp
```diff
@@ -152,7 +161,10 @@ TEST_F(BinderCorrectnessTest, SelectStatementTest) {
   }
 
   // Test select from different table instances from the same physical schema
-  binder.reset(new binder::BindNodeVisitor());
+  txn_manager.CommitTransaction(txn);
+
+  txn = txn_manager.BeginTransaction();
+  binder->SetTxn(txn);
```
apavlo: Can the binder ever run without a txn? If it always needs it, then we should make it a required argument to its constructor.

test/executor/copy_test.cpp
```diff
@@ -43,18 +43,20 @@ TEST_F(CopyTests, Copying) {
   catalog->CreateDatabase("emp_db", txn);
   txn_manager.CommitTransaction(txn);
 
-  optimizer::Optimizer optimizer;
+  std::unique_ptr<optimizer::AbstractOptimizer> optimizer;
```
apavlo: Switch this initialization to be one line:

`std::unique_ptr<optimizer::AbstractOptimizer> optimizer(new optimizer::Optimizer);`

test/executor/copy_test.cpp
```diff
@@ -90,7 +92,8 @@ TEST_F(CopyTests, Copying) {
               ResultTypeToString(status.m_result).c_str());
   }
   LOG_INFO("Tuples inserted!");
-  txn_manager.CommitTransaction(txn);
+  traffic_cop.CommitQueryHelper();
+//  txn_manager.CommitTransaction(txn);
```
apavlo: Removed dead code.

test/executor/update_test.cpp
```diff
@@ -152,9 +152,9 @@ TEST_F(UpdateTests, UpdatingOld) {
   catalog->CreateDatabase(DEFAULT_DB_NAME, txn);
   LOG_INFO("Bootstrapping completed!");
 
-  optimizer::Optimizer optimizer;
+  std::unique_ptr<optimizer::AbstractOptimizer> optimizer;
```
apavlo: See comment above. Switch this to be the one-liner.

test/optimizer/old_optimizer_test.cpp
```diff
@@ -136,10 +140,10 @@ TEST_F(OldOptimizerTests, UpdateDelWithIndexScanTest) {
   LOG_TRACE("Updating a tuple...");
   LOG_TRACE(
       "Query: UPDATE department_table SET dept_name = 'CS' WHERE student_id = "
-      "52");
+          "52");
   update_stmt = peloton_parser.BuildParseTree(
       "UPDATE department_table SET dept_name = 'CS' WHERE student_id = 52");
-  auto update_plan = optimizer.BuildPelotonPlanTree(update_stmt);
+  auto update_plan = optimizer.BuildPelotonPlanTree(update_stmt, nullptr);
```
apavlo: Why is null okay here?

test/optimizer/operator_transformer_test.cpp
```diff
@@ -60,7 +60,7 @@ class OperatorTransformerTests : public PelotonTest {
         "SELECT %s FROM %s", true_predicates.c_str(), table_names.c_str());
     auto parsed_stmt = peloton_parser.BuildParseTree(ref_query);
     auto ref_stmt = parsed_stmt->GetStatement(0);
-    binder::BindNodeVisitor binder;
+    binder::BindNodeVisitor binder(nullptr);
```
apavlo: Same question. Why is the txn allowed to be null?

src/binder/bind_node_visitor.cpp
```diff
@@ -18,7 +18,10 @@
 namespace peloton {
 namespace binder {
 
-BindNodeVisitor::BindNodeVisitor() { context_ = nullptr; }
+BindNodeVisitor::BindNodeVisitor(concurrency::Transaction *txn) {
```
pmenon: Use C++11 initializer lists for member initialization [like so](http://en.cppreference.com/w/cpp/language/initializer_listl). What you have isn't wrong, but it is inconsistent with the rest of the code.

src/include/optimizer/optimizer.h
```diff
@@ -19,6 +19,7 @@
 #include "optimizer/memo.h"
 #include "optimizer/property_set.h"
 #include "optimizer/rule.h"
+#include "concurrency/transaction.h"
```
pmenon: Is this needed in the header file or can it be forward declared?

src/include/optimizer/query_to_operator_transformer.h
```diff
@@ -62,6 +66,8 @@ class QueryToOperatorTransformer : public SqlNodeVisitor {
   type::TypeId output_type;
   int output_size;
   bool output_inlined;
+  concurrency::Transaction *txn = nullptr;
```
pmenon: Remove initialization here.

src/include/planner/analyze_plan.h
```diff
@@ -13,6 +13,7 @@
 #pragma once
 
 #include "planner/abstract_plan.h"
+#include "concurrency/transaction.h"
```
pmenon: Is this needed or can it be forward declared?

src/include/planner/analyze_plan.h
```diff
@@ -58,9 +62,9 @@ class AnalyzePlan : public AbstractPlan {
   }
 
  private:
-  storage::DataTable* target_table_ = nullptr;
+  storage::DataTable *target_table_ = nullptr;
```
pmenon: While you're at it, can you remove this class body initialization? Again, it isn't wrong, but it's inconsistent.

src/include/planner/delete_plan.h
```diff
@@ -16,6 +16,7 @@
 #include "type/types.h"
 #include "parser/table_ref.h"
 #include "planner/abstract_plan.h"
+#include "concurrency/transaction.h"
```
pmenon: Forward declare?

src/optimizer/optimizer.cpp
```diff
@@ -77,30 +77,29 @@ Optimizer::Optimizer() {
...
-  auto ddl_plan = HandleDDLStatement(parse_tree, is_ddl_stmt);
+  auto ddl_plan = HandleDDLStatement(parse_tree, is_ddl_stmt, txn);
   if (is_ddl_stmt) {
     return move(ddl_plan);
   }
-
   // Run binder
-  auto bind_node_visitor = make_shared<binder::BindNodeVisitor>();
+  auto bind_node_visitor = make_shared<binder::BindNodeVisitor>(txn);
+//  bind_node_visitor->txn = txn;
```
pmenon: Remove dead code.

src/optimizer/query_to_operator_transformer.cpp
```diff
@@ -33,6 +33,9 @@ using std::shared_ptr;
 
 namespace peloton {
 namespace optimizer {
+QueryToOperatorTransformer::QueryToOperatorTransformer(concurrency::Transaction *txn) {
```
pmenon: Use initializer list for member initialization.

src/tcop/tcop.cpp
```diff
@@ -238,53 +241,92 @@ executor::ExecuteResult TrafficCop::ExecuteStatementPlan(
...
+    const size_t thread_id UNUSED_ATTRIBUTE) {
   LOG_TRACE("Prepare Statement name: %s", statement_name.c_str());
   LOG_TRACE("Prepare Statement query: %s", query_string.c_str());
 
   std::shared_ptr<Statement> statement(
       new Statement(statement_name, query_string));
+  // transaction
+  // --multi-statements except BEGIN in a transaction
+  // ----not aborted
+  // ----COMMIT, ROLLBACK, and aborted
```
pmenon: Can you elaborate on these comments? They're helpful, but require more explanation.

src/tcop/tcop.cpp
```diff
@@ -238,53 +241,92 @@ executor::ExecuteResult TrafficCop::ExecuteStatementPlan(
...
+  auto &txn_manager = concurrency::TransactionManagerFactory::GetInstance();
+  if (!tcop_txn_state_.empty()) {
+    single_statement_txn_ = false;
+    // multi-statment txn has been aborted, just block it
+    // cannot return nullptr in case that commit cannot be execute
```
pmenon: Not quite sure what this is saying.

src/tcop/tcop.cpp
```diff
@@ -238,53 +241,92 @@ executor::ExecuteResult TrafficCop::ExecuteStatementPlan(
...
+    if (tcop_txn_state_.top().second == ResultType::ABORTED) {
+      return statement;
+    }
+  } else {
+    // --BEGIN
+    // --single-statement transaction
+    // ----prepareStatement abort
```
pmenon: Again, more complete text on what you mean.

src/planner/update_plan.cpp
haoxianghua: same as above

src/planner/update_plan.cpp
haoxianghua: Above BuildInitialUpdatePlan, annotation says "FIXME: Should remove when the simple_optimizer tears down". And I think this function is outdated. Peloton won't use it anymore. So it doesn't matter.

apavlo: Fix the function then. The SimpleOptimizer is gone.

seojungmin: I already spent some time to fix this, and have fixes in my local machine.  I will upload them as a separate PR, today.

src/include/binder/bind_node_visitor.h
```diff
@@ -57,9 +57,13 @@ class BindNodeVisitor : public SqlNodeVisitor {
   void Visit(expression::CaseExpression *expr) override;
   // void Visit(const expression::ConstantValueExpression *expr) override;
   void Visit(expression::TupleValueExpression *expr) override;
+  void SetTxn(concurrency::Transaction *txn) {
+    this->txn = txn;
+  }
 
  private:
   std::shared_ptr<BinderContext> context_;
+  concurrency::Transaction *txn;
```
pmenon: Private members must have an underscore suffix (i.e., this should be `txn_`).

Issue #709, opened 2017-06-30
System Crashes After Aborted Txn


apavlo:
The system will crash if you try to start another txn after the previous txn aborted:

**First create the table**
```
postgres=# CREATE TABLE foo (id INT PRIMARY KEY);
CREATE TABLE foo (id INT PRIMARY KEY) 0
```

**Then start a txn that will attempt to insert the same key twice. The second insert will (correctly) fail, causing the whole txn to abort**
``` 
postgres=# BEGIN;
BEGIN
postgres=# INSERT INTO foo VALUES (1);
INSERT INTO foo VALUES (1) 1
postgres=# SELECT * FROM foo;
 id 
----
  1
(1 row)

postgres=# INSERT INTO foo VALUES (1);
INSERT INTO foo VALUES (1) 0
postgres=# SELECT * FROM foo;
SELECT * FROM foo 0
postgres=# COMMIT;
COMMIT
postgres=# SELECT * FROM foo;
 id 
----
(0 rows)
```

**But now if you try to insert that same key again, the system crashes**
```
postgres=# BEGIN;
BEGIN
postgres=# SELECT * FROM foo;
 id 
----
(0 rows)

postgres=# INSERT INTO foo VALUES (1);
server closed the connection unexpectedly
        This probably means the server terminated abnormally
        before or while processing the request.
The connection to the server was lost. Attempting reset: Failed.
```

Here is the error from the server:
```
peloton: /home/pavlo/Documents/Peloton/Github/peloton-master/src/index/bwtree_index.cpp:106: bool peloton::index::BWTreeIndex<KeyType, ValueType, KeyComparator, KeyEqualityChecker, KeyHashFunc, ValueEqualityChecker, ValueHashFunc>::CondInsertEntry(const peloton::storage::Tuple*, peloton::ItemPointer*, std::function<bool(const void*)>) [with KeyType = peloton::index::CompactIntsKey<1ul>; ValueType = peloton::ItemPointer*; KeyComparator = peloton::index::CompactIntsComparator<1ul>; KeyEqualityChecker = peloton::index::CompactIntsEqualityChecker<1ul>; KeyHashFunc = peloton::index::CompactIntsHasher<1ul>; ValueEqualityChecker = peloton::ItemPointerComparator; ValueHashFunc = peloton::ItemPointerHashFunc]: Assertion `ret == true' failed.
```

yingjunwu:
[...]

PR #742, opened 2017-07-11
Task queue and worker pool infranstructure

ChTimTsubasa:
We reuse most of Matt's code in worker pool while also added:
1. Synchronous task or task batch submission using the C++ native conditional variable so that there won't be race condition with libevent.
2. A wrapper interface that supports requirements from @pmenon.


ChTimTsubasa:
This PR fails at two builds because of #709

Issue #744, opened 2017-07-13
CHAR and VARCHAR seem to require strlen+1


mrks:
When creating a table, it appears as if the size for CHAR and VARCHAR have to be one larger than the actual strings:

```
postgres=# CREATE TABLE TMP(X CHAR(2));
CREATE TABLE
postgres=# INSERT INTO TMP VALUES ('A');
INSERT 0 1
postgres=# INSERT INTO TMP VALUES ('AA');
The value is too long to fit into type VARCHAR(2)
```

We found this to be an issue when running the TPC-C benchmark. After increasing the size of all CHARs in https://github.com/oltpbenchmark/oltpbench/blob/master/src/com/oltpbenchmark/benchmarks/tpcc/ddls/tpcc-peloton-ddl.sql by 1, we were able to run the benchmark.

apavlo:
Fixed in #745

PR #745, opened 2017-07-13
Fix bug when checking VARCHAR length

lixupeng:
Value::GetLength() returns the memory size of a string which is equals to strlen + 1,
While column_length records the limit of strlen.

I also reset the optimizer when Exception occurs.


apavlo:
@lixupeng Does this fix #744? Can you add a test case for this issue? Check whether @PauloAmora already has one written.

lixupeng:
@apavlo Yes, this is for #744, and another bug when ValueOutOfRangeException actually occurs. It should reset optimizer after creating Plan whether Plan is created successfully or not.

PauloAmora:
Do I make a separate PR with the tests ?

apavlo:
Have @lixupeng pull from your branch.

PauloAmora:
Ok, I just uploaded it.

@lixupeng, the file can be found under the testcase_744 branch.

PR #748, opened 2017-07-14
Remove STL headers from coverage

pmenon:
By default, `gcov` analyzes **all** included header files, including STL, boost, and LLVM headers. This is overkill. Moreover, it's been blowing up our builds because the log file size grows beyond 4MB.

This PR adds a flag to exclude library headers when performing code coverage.


apavlo:
Awesome PR. I will merge

apavlo:
@pmenon It looks like coveralls choked on this.

pmenon:
Hmmm, let me take a look.

pmenon:
Looks like the `-r` doesn't because all paths are absolute.

There isn't an easy fix. I propose filtering `gcov` output for anything with "peloton" in it. It's hacky, but seems to do the job - plus, no one looks at `gcov` output in the logs directly anyways.

apavlo:
@pmenon OK sounds reasonable

pmenon:
@apavlo This is good to go.

PR #763, opened 2017-07-23
ConfigurationManager & ConfigCatalog

lixupeng:
# Configurations
There are two major parts in this PR:
- SettingsCatalog
- ConfigurationManager
## SettingsCatalog
I create a new catalog table 'pg_settings' to store configuration information which contains several columns:
- name (varchar) : the name of a parameter
- value (varchar) : the current value of a parameter
- value_type (integer) : the type (type::TypeId) of the value
- description (varchar) : a short description of usage
- min_value (varchar) : the minimum legal value
- max_value (varchar) : the maximum legal value
- default_value (varchar) : default value
- is_mutable (boolean) : be true if the parameter is allowed to modify during the system running
- is_persistent (boolean) : be true if the value should be restored from logs/checkpoints when the system restarts
## ConfigurationManager
This is a wrapper of SettingsCatalog. It provides several APIs that DefineConfig, GetValue and SetValue. It also keeps a map from name to value which is used during bootstrap before catalog being initialized.
## Other Things
- ConfigurationId : An enum class contains members whose names are as same as the names of parameters.
- ConfigurationUtil : A wrapper for GetValue and SetValue. Support three types temporarily (SET_INT/GET_INT, SET_BOOL/GET_BOOL, SET_STRING/GET_STRING).
## How To Use It
When you want to add a new parameter, define it with macros CONFIG_(int, bool, string) in [configuration.h](https://github.com/lixupeng/peloton/blob/pg_config/src/include/configuration/configuration.h)
For example:
`//CONFIG_int(name, description, default_value, is_mutable, is_persistent)`
`CONFIG_int(port, "Peloton port (default: 15721)", 15721, false, false)`
The macro defined in [configuration_macro.h](https://github.com/lixupeng/peloton/blob/pg_config/src/include/configuration/configuration_macro.h) will be replaced by four types of content on demand:
- a gflags definition
- a gflags declaration
- a member in ConfigurationId
- a statement of ConfigurationManager::DefineConfig

To get or set values, use ConfigurationUtil. For example:
`ConfigurationUtil::GET_INT(ConfigurationId::port)`
`ConfigurationUtil::SET_BOOL(ConfigurationId::index_tuner, true)`

### TODO
- Support more types, not only INTEGER, BOOLEAN, STRING
- Support validation check, I leave blank in min_value and max_value temporarily.


apavlo:
1. Please write more descriptive commit messages. "debug" is insufficient.

2. Why is this still modeled after "pg_config"? I thought we said to target "pg_settings"?

src/configuration/configuration_manager.cpp
```diff
@@ -0,0 +1,149 @@
...
+//
+//===----------------------------------------------------------------------===//
+#define __PELOTON_CONFIG__
+
+#include "common/exception.h"
+#include "configuration/configuration_manager.h"
+#include "configuration/configuration.h"
+#include "catalog/config_catalog.h"
+
+uint64_t GET_INT(const std::string& name) {
```
apavlo: Don't pollute the global namespace.
Make these static methods inside of a utility class.

src/configuration/configuration_manager.cpp
```diff
@@ -0,0 +1,149 @@
...
+                                 conf.second.value_type,
+                                 conf.second.desc, "", "", conf.second.default_value,
+                                 conf.second.is_mutable, conf.second.is_persistent,
+                                 pool, txn);
+  }
+  txn_manager.CommitTransaction(txn);
+  catalog_initialized = true;
+}
+
+void ConfigurationManager::PrintConfiguration() {
```
apavlo: Don't create "print" functions. `ConfigurationManager` should extend `Printable` and you should implement a `GetInfo` method.

src/include/catalog/config_catalog.h
```diff
@@ -0,0 +1,93 @@
...
+// config_catalog.h
+//
+// Identification: src/include/catalog/config_catalog.h
+//
+// Copyright (c) 2015-17, Carnegie Mellon University Database Group
+//
+//===----------------------------------------------------------------------===//
+
+//===----------------------------------------------------------------------===//
+// pg_config
```
apavlo: pg_settings

src/include/catalog/config_catalog.h
```diff
@@ -0,0 +1,93 @@
...
+  //===--------------------------------------------------------------------===//
+  // Read-only Related API
+  //===--------------------------------------------------------------------===//
+  std::string GetConfigValue(const std::string &name,
+                             concurrency::Transaction *txn);
+
+  std::string GetDefaultValue(const std::string &name,
+                             concurrency::Transaction *txn);
+
+  enum ColumnId {
```
apavlo: Should this be private too?

apavlo: Switch to `enum class`

src/include/catalog/config_catalog.h
```diff
@@ -0,0 +1,93 @@
...
+    DEFAULT_VALUE = 6,
+    IS_MUTABLE = 7,
+    IS_PERSISTENT = 8,
+    // Add new columns here in creation order
+  };
+
+private:
+  ConfigCatalog(concurrency::Transaction *txn);
+
+  enum IndexId {
```
apavlo: Make this `enum class` too.

src/include/configuration/configuration.h
```diff
@@ -10,9 +10,24 @@
 //
 //===----------------------------------------------------------------------===//
 
-#pragma once
```
apavlo: Why did you remove this?

src/include/configuration/configuration.h
```diff
@@ -10,9 +10,24 @@
...
+// internal use only in
+// configuration/configuration_manager.cpp
+#ifdef __PELOTON_CONFIG__
 
-#include <gflags/gflags.h>
+#include "configuration/configuration_utils.h"
```
apavlo: This file should be called "configuration_util.h" to match the other files in the repository.

src/include/configuration/configuration_utils.h
```diff
@@ -0,0 +1,45 @@
+//===----------------------------------------------------------------------===//
+//
+//                         Peloton
+//
+// configuration_utils.h
```
apavlo: Rename to "configuration_util"

src/include/configuration/configuration_utils.h
```diff
@@ -0,0 +1,45 @@
...
+//
+// Identification: src/include/configuration/configuration_utils.h
+//
+// Copyright (c) 2015-16, Carnegie Mellon University Database Group
```
apavlo: Use the updated header. Run "script/formatting/formatter.py"

src/include/configuration/configuration_manager.h
```diff
@@ -0,0 +1,137 @@
...
+//
+//===----------------------------------------------------------------------===//
+#pragma once
+
+#include <gflags/gflags.h>
+#include "type/types.h"
+#include "type/ephemeral_pool.h"
+#include "common/exception.h"
+#include "catalog/config_catalog.h"
+#include "concurrency/transaction_manager_factory.h"
```
malin1993ml: It seems that you don't need to include config_catalog and transaction_manager_factory here. You can include them in the .cpp file to reduce coupling of the code base. You should also forward declare ephemeral_pool since it's only used with unique_ptr.

src/common/init.cpp
```diff
@@ -46,22 +48,23 @@ void PelotonInit::Initialize() {
   gc::GCManagerFactory::GetInstance().StartGC();
 
   // start index tuner
-  if (FLAGS_index_tuner == true) {
+  if (Config::GET_BOOL("index_tuner")) {
```
apavlo: This is what I don't want to happen. We don't people to pass an arbirtraty string to get a parameter because we have no way of checking this at compile time. Can you define a new `enum class` that has the list of parameters?

src/configuration/configuration_manager.cpp
```diff
@@ -0,0 +1,164 @@
...
+//                         Peloton
+//
+// configuration_manager.cpp
+//
+// Identification: src/configuration/configuration_manager.cpp
+//
+// Copyright (c) 2015-2017, Carnegie Mellon University Database Group
+//
+//===----------------------------------------------------------------------===//
+#define __PELOTON_CONFIG__
```
apavlo: What is this?

src/configuration/configuration_manager.cpp
```diff
@@ -0,0 +1,164 @@
...
+                                 pool, txn);
+  }
+  txn_manager.CommitTransaction(txn);
+  catalog_initialized = true;
+}
+
+const std::string ConfigurationManager::GetInfo() const {
+  std::stringstream str;
+  char buffer[256];
+  sprintf(buffer, "%30s", "//===-------------- PELOTON CONFIGURATION --------------===//");
```
apavlo: Don't use `sprintf`. Use `std::ostringstream`. See the other `GetInfo()` implementations.

src/include/catalog/config_catalog.h
```diff
@@ -0,0 +1,76 @@
...
+#pragma once
+
+#include "catalog/abstract_catalog.h"
+
+#define CONFIG_CATALOG_NAME "pg_settings"
+
+namespace peloton {
+namespace catalog {
+
+class ConfigCatalog : public AbstractCatalog {
```
apavlo: What are we calling this `ConfigCatalog`? It should be called `SettingsCatalog` to match the naming convention of the other catalog objects.

src/include/configuration/configuration.h
```diff
@@ -2,17 +2,33 @@
...
+// Copyright (c) 2015-2017, Carnegie Mellon University Database Group
 //
 //===----------------------------------------------------------------------===//
-
 #pragma once
 
-#include <gflags/gflags.h>
+// internal use only in
+// configuration/configuration_manager.cpp
+#ifdef __PELOTON_CONFIG__
```
apavlo: Again, I don't understand what `__PELOTON_CONFIG__` is used for.

src/include/configuration/configuration.h
```diff
@@ -2,17 +2,33 @@
...
+
+#include "configuration/configuration_util.h"
+
+//===----------------------------------------------------------------------===//
+// Use following macros to define configuration parameters
+//===----------------------------------------------------------------------===//
+// CONFIG_int(name, description, default_value, is_mutable, is_persistent)
+// CONFIG_bool(name, description, default_value, is_mutable, is_persistent)
+// CONFIG_string(name, description, default_value, is_mutable, is_persistent)
```
apavlo: Remove dead code.

src/include/configuration/configuration.h
```diff
@@ -56,45 +90,112 @@ DECLARE_string(certificate_file);
...
+
+// special case: -help
 DECLARE_bool(help);
 
 namespace peloton {
 namespace configuration {
 
-// Print configuration
-void PrintConfiguration();
+void register_parameters() {
```
apavlo: You should put these in a utility class.

src/include/configuration/configuration_manager.h
```diff
@@ -0,0 +1,133 @@
...
+namespace peloton {
+namespace configuration {
+  class Config;
+}}
+using peloton::configuration::Config;
+
+namespace peloton {
+namespace configuration {
+
+class Config {
```
apavlo: Shouldn't this be `ConfigurationUtil`. It's not obvious from just `Config` that it is a helper class.

Also, put it in a separate file. Don't combine it with the ConfigurationManager.

src/include/configuration/configuration_manager.h
```diff
@@ -0,0 +1,133 @@
...
+
+#include "type/types.h"
+#include "common/exception.h"
+#include "common/printable.h"
+
+namespace peloton {
+namespace configuration {
+  class Config;
+}}
+using peloton::configuration::Config;
```
apavlo: Why is this necessary?

src/include/configuration/configuration_manager.h
```diff
@@ -0,0 +1,133 @@
...
+  T GetValue(const std::string &name) {
+    auto param = config.find(name);
+    if (param == config.end()) {
+      throw new ConfigurationException("no such configuration: " + name);
+    }
+    return to_value<T>(param->second.value);
+  }
+
+  template<typename T>
+  void SetValue(const std::string &name, const T &value) {
```
apavlo: We may want to switch this to use our own `Value` type object. Your casting is unsafe.

src/include/configuration/configuration_util.h
```diff
@@ -0,0 +1,45 @@
...
+//
+//===----------------------------------------------------------------------===//
+
+// internal use only in
+// configuration/configuration_manager.cpp
+#ifdef __PELOTON_CONFIG__
+
+#include <gflags/gflags.h>
+
+#define CONFIG_int(name, description, default_value, is_mutable, is_persistent)                   \
```
apavlo: Can you add documentation about what these macros do?

src/common/init.cpp
lixupeng: OK, I'll do that

src/include/configuration/configuration.h
lixupeng: Yes, it is useless now. At the beginning, I tried to unfold the content of this header file in a function defined in configuration_manager.cpp. But gflags doesn't allow me to do so.

lixupeng: I just want to make this header file seem to be independent and easy to use. I don't want to define one parameter twice. If I don't do so, I may need to define a parameter with gflag's DEFINE_* as well as ConfigurationManager::DefineConfig.

pmenon:
This is really cool!  @lixupeng: Can you produce a README that details how this works? 

I'd like to take a pass over the PR once the README is available.

lixupeng:
Hi @pmenon, I write a README. Could you have a look, thank you!

src/include/configuration/configuration_manager.h
```diff
@@ -0,0 +1,77 @@
...
+
+private:
+
+  struct Param {
+    std::string name;
+    void* gflags_ptr;
+    std::string desc;
+    type::Value value, default_value;
+    bool is_mutable, is_persistent;
+    Param() {}
```
pmenon: Do you need an empty-constructor?

src/include/configuration/configuration_manager.h
```diff
@@ -0,0 +1,77 @@
...
+  void DefineConfig(ConfigurationId id, const std::string &name,
+                    void* gflags_ptr, const std::string &description,
+                    const type::Value &value, const type::Value &default_value,
+                    bool is_mutable, bool is_persistent);
+
+  type::Value GetValue(ConfigurationId id);
+
+  void SetValue(ConfigurationId id, const type::Value &value);
+
+  void insert_into_catalog(const Param &param);
```
pmenon: Change name format.

src/include/configuration/configuration_util.h
```diff
@@ -0,0 +1,35 @@
...
+
+class ConfigurationUtil {
+public:
+  static int32_t GET_INT(ConfigurationId id);
+  static bool GET_BOOL(ConfigurationId id);
+  static std::string GET_STRING(ConfigurationId id);
+
+  static void SET_INT(ConfigurationId id, int32_t value);
+  static void SET_BOOL(ConfigurationId id, bool value);
+  static void SET_STRING(ConfigurationId id, const std::string &value);
```
lixupeng: Yes, thanks for suggestion. These functions were macros at the very beginning. I'll fix it.

src/include/configuration/configuration_util.h
```diff
@@ -0,0 +1,35 @@
...
+};
+
+} // End configuration namespace
+} // End peloton namespace
```
lixupeng: ok, I'll do that.

src/include/configuration/configuration_manager.h
lixupeng: Yes, it is needed in definition of std::map<ID, Param>.

lixupeng:
Hi @pmenon , thank you for these valuable comments. 

1) About the name. Andy suggests that we should use 'pg_settings' as the catalog's name. So I'm going to change all names to 'settings'.

2) Yes, nothing will happen if user do updates through PSQL directly. Right now my thought is to use catalog table as a log or a backup. This should be fixed when triggers and callback functions are supported. I had tried to load value from catalog directly, but the problem is, querying in catalog needs to create a transaction which will read settings' values again. An infinite recursion will occur.

3) For now, ConfigurationManager reads the values parsed by GFlags and save them into catalog. When SetValue is called, it will change three things: 1.catalog; 2.FLAGS_*; 3. internal map

src/include/configuration/configuration_manager.h
```diff
@@ -0,0 +1,77 @@
...
+    Param() {}
+    Param(std::string name, void* gflags_ptr, std::string desc,
+          const type::Value &value, const type::Value &default_value,
+          bool is_mutable, bool is_persistent)
+            : name(name), gflags_ptr(gflags_ptr), desc(desc),
+              value(value), default_value(default_value),
+              is_mutable(is_mutable), is_persistent(is_persistent) {}
+  };
+
+  std::map<ConfigurationId, Param> config;
```
seojungmin: Could you describe why you have chosen `std::map` for configuration data structure?  Why not a customized memory reference table?  `std::map` seems to be a balanced tree, and a memory reference will be faster and consistent in performance than a balanced tree.

src/configuration/configuration_manager.cpp
```diff
@@ -0,0 +1,155 @@
...
+  }
+}
+
+void ConfigurationManager::insert_into_catalog(const Param &param) {
+  auto config_catalog = catalog::SettingsCatalog::GetInstance();
+  auto &txn_manager = concurrency::TransactionManagerFactory::GetInstance();
+  auto txn = txn_manager.BeginTransaction();
+  type::AbstractPool *pool = pool_.get();
+  config_catalog->DeleteSetting(param.name, txn);
+  config_catalog->InsertSetting(param.name, param.value.ToString(),
```
seojungmin: Could you also leave a comment why you are not updating the catalog, but delete-then-inserting?

src/include/configuration/configuration_manager.h
pmenon: It's only needed if you use `operator[]`. Remove the default constructor and don't use the `operator[]` to keep the interface clean. Replace searches with `config.find()` and insertions with `config.insert()`, or even better `config.emplace()`.

src/include/configuration/configuration_util.h
```diff
@@ -0,0 +1,35 @@
...
+
+using namespace peloton::configuration;
```
pmenon: Don't do `using namespace`.

src/include/catalog/settings_catalog.h
```diff
@@ -0,0 +1,75 @@
...
+// Copyright (c) 2015-2017, Carnegie Mellon University Database Group
+//
+//===----------------------------------------------------------------------===//
+
+
+#pragma once
+
+#include "catalog/abstract_catalog.h"
+
+#define SETTINGS_CATALOG_NAME "pg_settings"
```
pmenon: This is used exclusively in the implementation, move to CPP file.

pmenon:
@lixupeng:
1. Good
2. So we'll need triggers to support updating cached copies of settings (i.e., in your configuration map).
3. Why do the GFlag values have to be updated? They're read once on startup and never used again.

lixupeng:
@pmenon 
Here's only one place where updating GFlags is needed, which is the 'help' flag. 
In [peloton.cpp](https://github.com/cmu-db/peloton/blob/master/src/main/peloton/peloton.cpp), we need to check if '-h' is set. If so, we need to set FLAGS_help to true so that we can print the help information correctly.
Maybe I should treat it as a special case.

apavlo:
@lixupeng I agree with @pmenon. Let's not modify gflags.

src/settings/settings_util.cpp
```diff
@@ -0,0 +1,62 @@
...
+
+#include "type/value_factory.h"
+#include "settings/settings_util.h"
+#include "settings/settings_manager.h"
+
+// gflags DECLARE
+#define __SETTING_GFLAGS_DECLARE__
+#include "settings/settings_macro.h"
+#include "settings/settings.h"
+#undef __SETTING_GFLAGS_DECLARE__
```
pmenon: Is this needed here?

src/settings/settings_util.cpp
```diff
@@ -0,0 +1,62 @@
...
+
+#define __SETTING_DEFINE__
+#include "settings/settings_macro.h"
```
pmenon: Is this needed here?

src/include/settings/settings_type.h
```diff
@@ -0,0 +1,29 @@
...
+namespace settings {
+
+enum class SettingsId {
+  #include "settings/settings.h"
+};
+
+}  // namespace settings
+}  // namespace peloton
+
+#undef __SETTING_ENUM__
```
pmenon: 1. Should we rename this to `setting_id.h`
2. Can you rename `SettingsId` -> `SettingId`.
3. Can you move the #define/#undefine into the enumeration to make clear the scope? Something like:
```
enum class SettingId {
  #define __SETTING_ENUM__
  #include "settings/settings.h"
  #undef __SETTING_ENUM__
};
```

src/include/settings/settings_util.h
```diff
@@ -0,0 +1,33 @@
...
+//
+// settings_util.h
+//
+// Identification: src/include/settings/settings_util.h
+//
+// Copyright (c) 2015-2017, Carnegie Mellon University Database Group
+//
+//===----------------------------------------------------------------------===//
+
+
```
pmenon: Remove empty line.

src/settings/settings.cpp
```diff
@@ -0,0 +1,46 @@
...
+
+
+#include <gflags/gflags.h>
+#include "type/types.h"
+
+// gflags DEFINE
+#define __SETTING_GFLAGS_DEFINE__
+#include "settings/settings_macro.h"
+#include "settings/settings.h"
+#undef __SETTING_GFLAGS_DEFINE__
```
pmenon: Is this needed?

src/settings/settings_manager.cpp
```diff
@@ -0,0 +1,151 @@
...
+#define __SETTING_DEFINE__
+#include "settings/settings_macro.h"
+#undef __SETTING_DEFINE__
+
+namespace peloton {
+namespace settings {
+
+SettingsManager* SettingsManager::GetInstance() {
+  static std::unique_ptr<SettingsManager> config_manager(
+          new SettingsManager());
```
pmenon: Change to `static SettingsManager config_manager;`

src/settings/settings_manager.cpp
```diff
@@ -0,0 +1,151 @@
...
+//
+// settings_manager.cpp
+//
+// Identification: src/settings/settings_manager.cpp
+//
+// Copyright (c) 2015-2017, Carnegie Mellon University Database Group
+//
+//===----------------------------------------------------------------------===//
+
+
```
pmenon: Remove empty line.

src/include/settings/settings_manager.h
```diff
@@ -0,0 +1,90 @@
...
+          default_value(default_value),
+          is_mutable(is_mutable), is_persistent(is_persistent) {}
+  };
+
+  // internal map
+  struct EnumClassHash {
+    template <typename T> std::size_t operator()(T t)
+    const { return static_cast<std::size_t>(t); }
+  };
+  std::unordered_map<SettingsId, Param, EnumClassHash> settings;
```
pmenon: Add underscore suffix to member variables.

src/include/settings/settings_manager.h
```diff
@@ -0,0 +1,90 @@
...
+
+  std::unique_ptr<type::AbstractPool> pool_;
+
+  bool catalog_initialized;
```
pmenon: Missing underscore suffix.

src/settings/settings_manager.cpp
```diff
@@ -0,0 +1,151 @@
...
+#include "type/value_factory.h"
+#include "catalog/settings_catalog.h"
+#include "concurrency/transaction_manager_factory.h"
+#include "settings/settings_util.h"
+#include "settings/settings_manager.h"
+
+#define __SETTING_GFLAGS_DECLARE__
+#include "settings/settings_macro.h"
+#include "settings/settings.h"
+#undef __SETTING_GFLAGS_DECLARE__
```
pmenon: Do you need gflag declarations here?

src/settings/settings_manager.cpp
```diff
@@ -0,0 +1,151 @@
...
+
+#define __SETTING_DEFINE__
+#include "settings/settings_macro.h"
+#undef __SETTING_DEFINE__
```
pmenon: Move this into constructor.

src/settings/settings_manager.cpp
```diff
@@ -0,0 +1,151 @@
...
+}
+
+SettingsManager::SettingsManager() {
+  catalog_initialized = false;
+  pool_.reset(new type::EphemeralPool());
+  #include "settings/settings.h"
+}
+
+} // namespace settings
+} // namespace peloton
```
pmenon: Two spaces in closing namespace.

src/settings/settings_util.cpp
```diff
@@ -0,0 +1,62 @@
...
+
+void SettingsUtil::SetString(SettingsId id, const std::string &value) {
+  SettingsManager::GetInstance()->SetValue(
+      id, type::ValueFactory::GetVarcharValue(value));
+}
+
+}  // namespace settings
+}  // namespace peloton
+
+#undef __SETTING_DEFINE__
```
pmenon: Needed?

src/catalog/settings_catalog.cpp
```diff
@@ -0,0 +1,140 @@
...
+  Catalog::GetInstance()->CreateIndex(
+          CATALOG_DATABASE_NAME, SETTINGS_CATALOG_NAME,
+          {"name"}, SETTINGS_CATALOG_NAME "_skey0",
+          false, IndexType::BWTREE, txn);
+}
+
+SettingsCatalog::~SettingsCatalog() {}
+
+bool SettingsCatalog::InsertSetting(
+    const std::string &name, const std::string &value,
```
lixupeng: Yes, wrapping them into a struct is better but other catalogs do insert in this way so I don't modify it.

src/include/catalog/settings_catalog.h
```diff
@@ -0,0 +1,73 @@
...
+namespace peloton {
+namespace catalog {
+
+class SettingsCatalog : public AbstractCatalog {
+public:
+  ~SettingsCatalog();
+
+  // Global Singleton
+  static SettingsCatalog *GetInstance(
+          concurrency::Transaction *txn = nullptr);
```
lixupeng: After the first call to GetInstance funciton, we needn't pass a transaction pointer here.

src/settings/settings.cpp
lixupeng: I prefer to put all setting variables in settings.cpp. There are still some hard code variables in settings.cpp such as peloton_layout_mode and peloton_gc_mode. They are not controlled by gflags. Should we also move them into pg_settings?

src/settings/settings_manager.cpp
lixupeng: The script formatter.py generates two empty lines automatically after the header information. Maybe we can reduce it by one?

lixupeng:
@apavlo @pmenon I think it is ready for merge now. Could you have a review and merge it? Thanks!

apavlo:
This looks reasonable to me. There are a bunch of things that I think we're going to need to change  (like that singleton) but we can fix it later.

Issue #756, opened 2017-07-17
Concurrency issue for delete operation


mengranwo:
Let's say I've already created a table foo by executing query **CREATE TABLE foo(a int)** and then insert several tuples. 
```
postgres=# select * from foo;
 a 
---
 1
 2
 3
(3 rows)
```
The first transaction (begin early) delete tuple 1 from table foo, does not commit. And the second transaction, select * from foo using sequential scan. The correct result should return 3 rows instead of 2 since the first transaction has no committed yet, all the delete or insert operations should remain visible only within the same transaction. 
```
TXN 1                            
postgres=# BEGIN;                                                
BEGIN                                                                            
postgres=# delete from foo where A = 1;                          
delete from foo where A = 1 1                                        
postgres=# select * from foo;                                            
 a                                                                                      
---                                                                                      
 2                                                                                       
 3                                                                                        
(2 rows)

TXN 2 (begin after TXN 1)
postgres=# BEGIN;                                                
BEGIN                                                                                                              
postgres=# select * from foo;                                            
 a                                                                                      
---                                                                                      
 2                                                                                       
 3                                                                                        
(2 rows)
```
My preliminary thought after taking a brief look at the logic within delete_executor.cpp is that when deleting a tuple, transaction first inserts a tombstone in the storage level table and also grabs the deleted tuple's write lock(by setting this tuple's transaction-id metadata). If current transaction does not commit, the write lock will not be released, thus other concurrent transactions can't see the deleted tuple.

yingjunwu:
#712 #652 

all these problems are related to the tuple-level visibility. We should carefully design it after fixing #753.

Issue #757, opened 2017-07-17
Another concurrency issue 


seojungmin:
This may be from the same cause in #756.  I just report to help the person who is fixing it.  This time, it makes the server crash.  

With the same `create table` and `insert` in #756, the `foo` table has 3 tuples written: 1, 2 and 3, then, the following was executed.

```
TXN 1                            
postgres_txn1=# BEGIN;                                                
BEGIN                                                                                                             

TXN 2 
postgres_txn2=# BEGIN;                                                
BEGIN

TXN1
postgres_txn1=# delete from foo where a=1;                          
DELETE 1

TXN2
postgres_txn2=# delete from foo where a=3;                          
DELETE 1                                                                                               
postgres_txn2=# rollback;
server closed the connection unexpectedly
	This probably means the server terminated abnormally
	before or while processing the request.
The connection to the server was lost. Attempting reset: Failed.
```

And this is from GDB:

```
(gdb) where
#0  0x00007f75fa21b201 in peloton::AtomicUpdateItemPointer (src_ptr=0x0, 
    value=...) at /home/jungmins/w/compile/src/common/item_pointer.cpp:26
#1  0x00007f75fa4c9e97 in peloton::concurrency::TimestampOrderingTransactionManager::AbortTransaction (
    this=0x7f75fb9e18f0 <peloton::concurrency::TimestampOrderingTransactionManager::GetInstance(peloton::ProtocolType, peloton::IsolationLevelType, peloton::ConflictAvoidanceType)::txn_manager>, current_txn=0x7f75bc000db0)
    at /home/jungmins/w/compile/src/concurrency/timestamp_ordering_transaction_manager.cpp:1048
#2  0x00007f75fa5ea855 in peloton::tcop::TrafficCop::ExecuteStatementPlan (
    this=0x7f75c0000c20, plan=0x7f75bc000ef0, params=
    std::vector of length 0, capacity 0, 
    result=std::vector of length 0, capacity 0, 
    result_format=std::vector of length 0, capacity 0, thread_id=4)
    at /home/jungmins/w/compile/src/tcop/tcop.cpp:257
```

Issue #713, opened 2017-07-02
BwTree crashes when re-inserting deleted key


apavlo:
**I think that this bug is related to the same problem in #709. I don't think that we are cleaning up entries from the indexes properly.**

--------

First create a table with a single column and a primary key:
```
postgres=# CREATE TABLE foo (id INT PRIMARY KEY);
CREATE TABLE foo (id INT PRIMARY KEY) 0
postgres=# INSERT INTO foo VALUES (123);
INSERT INTO foo VALUES (123) 1
postgres=# SELECT * FROM foo;
 id  
-----
 123
(1 row)
```

Next we will delete the tuple from the table and execute a `SELECT` statement to show that it is empty:
```
postgres=# DELETE FROM foo;
DELETE FROM foo 1
postgres=# SELECT * FROM foo;
 id 
----
(0 rows)
```

Now let's try to insert that same tuple back into the table. This should succeed because the table is supposed to be empty:
```
postgres=# INSERT INTO foo VALUES (123);
server closed the connection unexpectedly
        This probably means the server terminated abnormally
        before or while processing the request.
The connection to the server was lost. Attempting reset: Failed.
```

Here is the assert message of the failure:
```
peloton: /home/pavlo/Documents/Peloton/Github/peloton-master/src/index/bwtree_index.cpp:106: \
bool peloton::index::BWTreeIndex<KeyType, ValueType, KeyComparator, KeyEqualityChecker, KeyHashFunc, ValueEqualityChecker, ValueHashFunc>::CondInsertEntry(const peloton::storage::Tuple*, peloton::ItemPointer*, std::function<bool(const void*)>) \
[with KeyType = peloton::index::CompactIntsKey<1ul>; ValueType = peloton::ItemPointer*; KeyComparator = peloton::index::CompactIntsComparator<1ul>; KeyEqualityChecker = peloton::index::CompactIntsEqualityChecker<1ul>; KeyHashFunc = peloton::index::CompactIntsHasher<1ul>; ValueEqualityChecker = peloton::ItemPointerComparator; ValueHashFunc = peloton::ItemPointerHashFunc]: \
Assertion `ret == true' failed.
```

Backtrace is attached.
[backtrace.txt](https://github.com/cmu-db/peloton/files/1117911/backtrace.txt)

yingjunwu:
Yes I agree that this is caused by the same problem as in #709.

yingjunwu:
fixed by #764

pmenon:
[...]

Issue #712, opened 2017-07-02
Txn Manager Crashes on UPDATE without WHERE


apavlo:
The system crashes when you try to execute an `UPDATE` query that does not have a `WHERE` clause in it. This fails regardless whether the codegen feature is enabled. I only tried running this with the new optimizer enabled.
```
postgres=# CREATE TABLE foo (id INT PRIMARY KEY);
CREATE TABLE foo (id INT PRIMARY KEY) 0
postgres=# INSERT INTO foo VALUES (123);
INSERT INTO foo VALUES (123) 1
postgres=# INSERT INTO foo VALUES (456);
INSERT INTO foo VALUES (456) 1
postgres=# UPDATE foo SET id = id + 1;
server closed the connection unexpectedly
        This probably means the server terminated abnormally
        before or while processing the request.
The connection to the server was lost. Attempting reset: Failed.
```

Here is the server's assert message:
```
peloton: /home/pavlo/Documents/Peloton/Github/peloton-master/src/concurrency/timestamp_ordering_transaction_manager.cpp:656: \
virtual void peloton::concurrency::TimestampOrderingTransactionManager::PerformDelete(peloton::concurrency::Transaction*, const peloton::ItemPointer&, const peloton::ItemPointer&): \
Assertion `(GetLastReaderCommitId(tile_group_header, old_location.offset) == current_txn->GetCommitId())' failed.
```

The backtrace is attached.
[backtrace.txt](https://github.com/cmu-db/peloton/files/1117905/backtrace.txt)

PauloAmora:
Wasn't this fixed in #558 ?

apavlo:
@PauloAmora I think that this problem is different. It definitely works correctly if you include the WHERE clause:
```
postgres=# CREATE TABLE foo (id INT PRIMARY KEY);                                                                                                                 
CREATE TABLE foo (id INT PRIMARY KEY) 0                                                                                                                                               
postgres=# INSERT INTO foo VALUES (123);                                                                                 
INSERT INTO foo VALUES (123) 1                                                                                                                                                        
postgres=# INSERT INTO foo VALUES (456);                                                           
INSERT INTO foo VALUES (456) 1                                                                                                                                                        
postgres=# UPDATE foo SET id = id + 1 WHERE id > 0;                                                
UPDATE foo SET id = id + 1 WHERE id > 0 2                                                                                                                                             
postgres=# SELECT * FROM foo;                                                                                                                                                         
[...]

PR #764, opened 2017-07-23
fix visibility bug

yingjunwu:
The insertion/deletion bugs reported recently are mostly related to the design of GC manager. This PR tries to fundamentally solve this problem by redesigning the GC component.

### DIFF
--- a/src/include/concurrency/transaction_manager.h
+++ b/src/include/concurrency/transaction_manager.h
@@ -150,22 +150,9 @@ class TransactionManager {
     return EpochManagerFactory::GetInstance().GetExpiredCid();
   }
 
-  void SetDirtyRange(std::pair<cid_t, cid_t> dirty_range) {
-    this->dirty_range_ = dirty_range;
-  }
-
   IsolationLevelType GetIsolationLevel() {
     return isolation_level_;
   }
-
- protected:
-  inline bool CidIsInDirtyRange(cid_t cid) {
-    return ((cid > dirty_range_.first) & (cid <= dirty_range_.second));
-  }
-  // invisible range after failure and recovery;
-  // first value is exclusive, last value is inclusive
-  std::pair<cid_t, cid_t> dirty_range_ =
-      std::make_pair(INVALID_CID, INVALID_CID);
 
  protected:
   static ProtocolType protocol_;
--- a/test/gc/transaction_level_gc_manager_test.cpp
+++ b/test/gc/transaction_level_gc_manager_test.cpp
@@ -32,49 +32,58 @@ namespace test {
 
 class TransactionLevelGCManagerTests : public PelotonTest {};
 
-void UpdateTuple(storage::DataTable *table, const int update_num, const int total_num) {
+ResultType UpdateTuple(storage::DataTable *table, const int key) {
   srand(15721);
 
   auto &txn_manager = concurrency::TransactionManagerFactory::GetInstance();
   TransactionScheduler scheduler(1, table, &txn_manager);
-  for (int i = 0; i < update_num; i++) {
-    scheduler.Txn(0).Update(rand() % total_num, rand() % 15721);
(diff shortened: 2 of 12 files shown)


apavlo:
This is fantastic. Can you be specific about what bugs you are fixing (i.e., #756, #712, #713, #709)?

yingjunwu:
I can confirm that this PR can fix #709 and #713, but I am not sure whether it can fix #756 and #757. The reason is that we haven't fixed the networking layer bug in the master branch, so I have no idea whether the bug is caused by the networking layer or by the concurrency and gc modules. We need to merge #733 before fixing #756 and #757.

Issue #712 is about how we scan the table. This PR cannot address this issue, but we can have a discussion on whether we need to create a default primary index for every table, and whether the sequential scan should be executed as traversing the leaf node of primary index.

For now, this PR does not address all the issues. We are using logical pointer index management scheme, meaning that we will insert an entry into secondary index only when the corresponding key value has been updated. This brings some issues with the GC design, such as the one reported in #686. We should have a discussion on whether we should switch to physical pointer scheme.

We haven't investigated how to handle unique index.

PauloAmora:
What exactly do you mean by unique index? The behavior when we try to create an unique index on an attribute that has duplicates?

yingjunwu:
@PauloAmora I mean this: https://www.postgresql.org/docs/current/static/indexes-unique.html. Currently we treat unique index the same way as we treat primary index. Not sure whether it is the right way to go.

apavlo:
@yingjunwu This is a great PR. We should have multiple reviewers look it over when you are ready.

We also need test cases to prove that you fixed the different issues.

yingjunwu:
More test cases added. Merging this PR will fix #709 and #713.

jarulraj:
@yingjunwu It would be great if you can add more details based on my comments, and importantly provide a high-level overview GC in a README.

yingjunwu:
Hey @jarulraj, sorry but where I can find your comments?

src/include/concurrency/transaction_manager.h
```diff
@@ -150,24 +150,11 @@ class TransactionManager {
...
 
-  void SetDirtyRange(std::pair<cid_t, cid_t> dirty_range) {
-    this->dirty_range_ = dirty_range;
-  }
-
   IsolationLevelType GetIsolationLevel() {
     return isolation_level_;
   }
 
  protected:
```
yingjunwu: This "dirty range" is used by the WBL. As the logging components are under refactoring, I removed this code for now. \cc @PauloAmora

test/gc/transaction_level_gc_manager_test.cpp
```diff
@@ -194,12 +203,171 @@ TEST_F(TransactionLevelGCManagerTests, GCTest) {
   table.release();
 
   // DROP!
-  TestingExecutorUtil::DeleteDatabase("DATABASE");
+  TestingExecutorUtil::DeleteDatabase("DATABASE0");
```
yingjunwu: Sure, but the `DeleteDatabase()` function is simply to drop the tables.

test/type/types_test.cpp
```diff
@@ -695,5 +695,62 @@ TEST_F(TypesTests, ConflictAvoidanceTypeTest) {
                peloton::Exception);
 }
 
+
+
+TEST_F(TypesTests, RWTypeTest) {
+  std::vector<RWType> list = {
+      RWType::INVALID,
```
yingjunwu: added comments at `types.h`.

yingjunwu:
Design docs can be found here: https://docs.google.com/document/d/1PdixbQ7WV21OnhsRkK3Oo-aq1qOrmQThLSUsfpuJ5sk/edit?usp=sharing

This PR can be merged as it is independent from any other PRs.

PR #765, opened 2017-07-25
First pass at updating Docker builds for eventual Jenkins use

crd477:
This PR includes some minor changes to the Dockerfiles and packages.sh to ease building on supported (or soon-to-be supported platforms).

I've also dropped the ubuntu-precise Dockerfile as that release is no longer supported upstream.


apavlo:
This is dope. Thanks!

PR #771, opened 2017-08-01
Additional build system bug fixes and improvements

crd477:
Some additional changes to fix or improve building, particularly in the Docker containers.  This gets us closer to testing automatically via Docker from Jenkins.


crd477:
Also, I meant to add that if my PR technique is sloppy, please let me know.

src/codegen/code_context.cpp
```diff
@@ -19,7 +19,7 @@
 #include "llvm/Support/raw_os_ostream.h"
 #include "llvm/Support/TargetSelect.h"
 #include "llvm/Transforms/Scalar.h"
-#if LLVM_VERSION_GE(3, 9)
+#if LLVM_VERSION_GE(3, 9) || LLVM_VERSION_GE(4, 0)
```
pmenon: I don't think you need this change since the macro checks for LLVM versions >= 3.9 already.

crd477: You'd think that but not the way the macro is written:
```
#define LLVM_VERSION_GE(major, minor) \
  (LLVM_VERSION_MAJOR > (major) ||    \
   (LLVM_VERSION_MAJOR == (major) && LLVM_VERSION_MINOR >= (minor)))
```

crd477: Oh wait, did this get changed recently?  It was different last time I looked.  Sorry for the noise...

pmenon:
Is it not possible to get LLVM 3.7 on these systems? I'm okay moving to 3.9 - I've tested it up to 4.0 and it all works.  But if we're doing this, we should be consistent on travis and in documentation too.

cmake/Dependencies.cmake
```diff
@@ -48,7 +48,7 @@ list(APPEND Peloton_LINKER_LIBS ${PQXX_LIBRARIES})
 list(APPEND Peloton_LINKER_LIBS "-lssl")
 
 # --[ LLVM 3.7+
-find_package(LLVM 3.7 REQUIRED CONFIG)
+find_package(LLVM REQUIRED CONFIG)
```
pmenon: Keep this to ensure build failure on systems using versions lower than 3.7.

crd477: What is a better version comparison to use?  With just `LLVM 3.7 REQUIRED CONFIG`, I got build failures on LLVM 3.9+ systems.

crd477: Yes, sorry, I was behind 03d20e5 at the time.

src/codegen/code_context.cpp
crd477: Yes, sorry, I was behind 03d20e5 at the time.

cmake/Dependencies.cmake
pmenon: Strange, then can you add a manual version check? Something like:

```
if (${LLVM_PACKAGE_VERSION} VERSION_LESS "3.7")
    message( FATAL_ERROR "LLVM 3.7 or newer is required." )
endif()
```

crd477:
On 2017-08-01 09:19, Prashanth Menon wrote:
> Is it not possible to get LLVM 3.7 on these systems? I'm okay moving to 
> 3.9 - I've tested it up to 4.0 and it all works. But if we're doing 
> this, we should be consistent on travis and in documentation too.
> 

In most cases it is still possible.  My only motivation was that it 
seemed like support for that particular version is getting spotty in 
upstream repositories, e.g., script/installation/packages.sh:66, and 
will probably get spottier over time.

That being said, I don't want to make a mess in your project and if this 
is just causing distraction, I'll cut it out :)  If this direction is OK 
though, I'd be happy to review travis and the documentation for relevant 
changes as well though.

-- 
     -Chad

cmake/Dependencies.cmake
crd477: Yes, that looks like it will work.  I'll do some more testing though.  Thanks!

crd477:
Closing this PR for the time being.  I think it was a bit hasty on my part.  I'm going to try a few different things and then will submit a different PR.

PR #772, opened 2017-08-01
Tidy Up Delete/Update Planner

seojungmin:
This quick PR removes the redundant code in the Update/Delete plans.   

`DeletePlan`'s constructor now only accepts the target table object, and nothing else.  The predicate/truncate information will be obtained from its child `SeqScanPlan`(or `IndexScanPlan` in the future). `UpdatePlan` also has only one constructor, from now on, which receives the target table and ProjectInfo objects. All the related code has also been fixed, such as optimizer, tests, etc.

I separated these from the codegen PR, for a review, and other planners such as `InsertPlan` will also have to be cleaned up.

### DIFF
--- a/src/catalog/abstract_catalog.cpp
+++ b/src/catalog/abstract_catalog.cpp
@@ -111,7 +111,7 @@ bool AbstractCatalog::DeleteWithIndexScan(oid_t index_offset,
       new executor::ExecutorContext(txn));
 
   // Delete node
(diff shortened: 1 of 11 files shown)


seojungmin:
This is ready to be reviewed.

apavlo:
LGTM. @hzxa21 Are you okay with this?

PR #773, opened 2017-08-02
ProcCatalog

lixupeng:
This PR is for issue #632 
1) I create two catalog tables: pg_language and pg_proc
2) I remove function names from ExpressionType and move function relative files to a new directory 'function'
3) I change methods 'Catalog::AddFunction' and 'Catalog::GetFunction' to look up pg_proc directly, instead of func_map in Catalog.
4) GetFunction now requires function name as well as function argument types to look up a function so as to handle those functions which have same name but different argument types.


apavlo:
@lixupeng What's the status of this?

apavlo:
@prashasthip Can you work with @lixupeng on reviewing this PR and getting the code merged? You are going to need this for your capstone project.

lixupeng:
@apavlo Hi Andy, I just arrived at VLDB, and I will update my PR today and tomorrow. By the way, I hope that we merge TriggerPR first, then I can do some more work on ProcPR such as make triggers really execute functions.

src/codegen/expression/function_translator.cpp
```diff
@@ -0,0 +1,103 @@
...
+  // store argument number
+  args.push_back(codegen.Const32((int32_t) child_num));
+
+  // store arguments
+  for (size_t i = 0; i < child_num; ++i) {
+    args.push_back(codegen.Const32(
+        static_cast<int32_t>(func_expr.func_arg_types_[i])));
+    args.push_back(row.DeriveValue(codegen, *func_expr.GetChild(i))
+                       .GetValue());
+  }
```
pmenon: Can you add some comments here on the calling convention?

src/catalog/catalog.cpp
```diff
@@ -803,61 +801,155 @@ Catalog::~Catalog() {
...
+    const std::string &name,
+    const std::vector<type::TypeId> &argument_types) {
+  auto &txn_manager = concurrency::TransactionManagerFactory::GetInstance();
+  auto txn = txn_manager.BeginTransaction();
+  FunctionData result;
+  oid_t prolang = ProcCatalog::GetInstance()->GetProLang(name, argument_types, txn);
+
+  // if the function is "internal", then look up the map in function::BuiltInFunctions
+  // to get the function pointer
+  if (LanguageCatalog::GetInstance()->GetLanguageName(prolang, txn) == "internal") {
```
prashasthip: Why are you doing lookup in the map only if the language is "internal"? It should be returning the function data even in the case of other languages.

lixupeng: Yes, but I think function data should be returned in other ways if the function is in other language. The code only support "internal" language now, and we can add other languages and corresponding handle code in the future.

lixupeng:
@prashasthip Thank you very much! I removed the code for function expression in LLVM and make it ready for review again. Could you please have a look?

src/catalog/catalog.cpp
```diff
@@ -796,61 +796,140 @@ Catalog::~Catalog() {
...
-    type::Value (*func_ptr)(const std::vector<type::Value> &)) {
-  PL_ASSERT(functions_.count(name) == 0);
-  functions_[name] = FunctionData{name, argument_types, return_type, func_ptr};
+/*@brief   Add new built-in function
+ * 1. add the function infomation into pg_proc
+ * 2. register the function pointer in function::BuiltinFunction
+ * @param   name & argument_types   function name and arg types used in SQL
+ * @param   return_type   the return type
+ * @param   prolang       the oid of which language the function is
+ * @param   func_name     the function name in C++ source code (should be unique)
```
prashasthip: Is the func_name value something that you allow the user to define? It looks like you need a unique name to identify the corresponding function_ptr ( to break ties when there are functions with the same name but different argument types).

src/function/functions.cpp
```diff
@@ -0,0 +1,35 @@
...
+//
+//===----------------------------------------------------------------------===//
+
+
+#include "function/functions.h"
+
+namespace peloton {
+namespace function {
+
+std::unordered_map<std::string, BuiltInFuncType>
```
prashasthip: I guess having a map works for now. You will be putting the function pointers into the catalog in the future changes, right?

lixupeng: Andy suggests not to put function pointers into the catalog, just to use a simple switch statement here. I think using a map here is the same as a a switch statement, ans is easier to use.

src/catalog/catalog.cpp
lixupeng: Yes, that's it. It is the user who defines the func_name.

apavlo:
When this passes, I will merge.

PR #780, opened 2017-08-06
Codegen compiled query cache

seojungmin:
This PR brings a new feature, query compilation cache, into Peloton.  This is a rewrite of a 15721 project, but has a new plan comparison mechanism.  Rather than having a separate plan comparator, each plan has its own equality check mechanism.  This lets a developer write the equality comparison and keep the plan comparison in mind, when creating a plan. There are also some minor bug fixes and enhancements.

The PR itself is a work in progress.  Some more works for cache, and verification for plan comparison, are required.


apavlo:
@seojungmin Wow this is impressive.

phisiart:
Let me review this on Friday.

src/codegen/query_cache.cpp
```diff
@@ -0,0 +1,78 @@
+//===----------------------------------------------------------------------===//
```
phisiart: There seems to be a design issue in this PR.

The k-v store in the cache is a `std::map`, which requires a user-defined "less than" comparator, instead of "not equal to".

The reason is that what's inside `std::map` is a balanced binary search tree. It needs a "less than" comparator to perform O(log n) operations. When it wants to determine whether two keys are equal, it just does "!(A < B) && !(B < A)".

The reason why the test cases are passing is that they only test situations when the cache has at most 1 element. But if we test situations where we have more keys, things can go weired.

Consider this example.

```C++
#include <map>
#include <iostream>

using namespace std;

struct Cmp {
    bool operator()(const int &lhs, const int &rhs) const {
        return lhs != rhs;
    }
};

int main() {
    map<int, int, Cmp> m;
    int n = 10;
    for (int i = 0; i < n; ++i) {
        m.emplace(i, i);
    }

    int cnt = 0;
    for (int i = 0; i < n; ++i) {
        cnt += m.count(i);
    }
    std::cout << m.size() << std::endl; // 10
    std::cout << cnt << std::endl; // 5 ???
    return 0;
}
```

If we are to use `std::unordered_map` instead of `std::map`, we would need a hash function as well as equality comparison.

src/include/expression/parameter_value_expression.h
```diff
@@ -44,6 +44,17 @@ class ParameterValueExpression : public AbstractExpression {
 
   virtual void Accept(SqlNodeVisitor *v) override { v->Visit(this); }
 
+  bool operator==(const AbstractExpression &rhs) const override {
+    auto &other = (ParameterValueExpression &)rhs;
```
phisiart:
```C++
if (exp_type_ != rhs.GetExpressionType()) {
  return false;
}
auto &other = static_cast<ParameterValueExpression &>(rhs);
```

src/include/expression/tuple_value_expression.h
```diff
@@ -100,18 +100,24 @@ class TupleValueExpression : public AbstractExpression {
     return new TupleValueExpression(*this);
   }
 
-  virtual bool Equals(AbstractExpression *expr) const override {
-    if (exp_type_ != expr->GetExpressionType()) return false;
-    auto tup_expr = (TupleValueExpression *)expr;
+  virtual bool operator==(const AbstractExpression &rhs) const override {
+    if (exp_type_ != rhs.GetExpressionType())
+      return false;
+    auto &tup_expr = (TupleValueExpression &)rhs;
```
phisiart:
```C++
auto &tup_expr = static_cast<TupleValueExpression &>(rhs);
```

src/include/expression/constant_value_expression.h
```diff
@@ -39,14 +39,20 @@ class ConstantValueExpression : public AbstractExpression {
...
-    if (!alias.empty()) return;
+    if (!alias.empty())
+      return;
     expr_name_ = value_.ToString();
   }
+  
+  virtual bool operator==(const AbstractExpression &rhs) const override {
+    if (exp_type_ != rhs.GetExpressionType())
+      return false;
+    auto &const_expr = (ConstantValueExpression &)rhs;
```
phisiart: auto &const_expr = static_cast<ConstantValueExpression &>(rhs);

src/include/expression/case_expression.h
```diff
@@ -103,6 +103,27 @@ class CaseExpression : public AbstractExpression {
     }
   }
 
+  bool operator==(const AbstractExpression &rhs) const override {
+    auto &other = (expression::CaseExpression &)rhs;
```
phisiart:
```C++
if (exp_type_ != rhs.GetExpressionType()) {
  return false;
}
auto &other = static_cast<expression::CaseExpression &>(rhs);
```

src/planner/aggregate_plan.cpp
```diff
@@ -121,5 +121,66 @@ void AggregatePlan::PerformBinding(BindingContext &binding_context) {
...
+      return false;
+  }
+  return true;
+}
+
+bool AggregatePlan::operator==(AbstractPlan &rhs) const {
+  if (GetPlanNodeType() != rhs.GetPlanNodeType())
+    return false;
+
+  auto &other = reinterpret_cast<planner::AggregatePlan &>(rhs);
```
phisiart:
```C++
auto &other = static_cast<planner::AggregatePlan &>(rhs);
```

src/planner/delete_plan.cpp
```diff
@@ -27,5 +27,19 @@ void DeletePlan::SetParameterValues(std::vector<type::Value> *values) {
   children[0]->SetParameterValues(values);
 }
 
+bool DeletePlan::operator==(AbstractPlan &rhs) const {
+  if (GetPlanNodeType() != rhs.GetPlanNodeType())
+    return false;
+
+  auto &other = reinterpret_cast<planner::DeletePlan &>(rhs);
```
phisiart:
```C++
auto &other = static_cast<planner::DeletePlan &>(rhs);
```

src/planner/hash_join_plan.cpp
```diff
@@ -52,5 +52,59 @@ void HashJoinPlan::HandleSubplanBinding(
   }
 }
 
+bool HashJoinPlan::operator==(AbstractPlan &rhs) const {
+  if (GetPlanNodeType() != rhs.GetPlanNodeType())
+    return false;
+
+  auto &other = reinterpret_cast<planner::HashJoinPlan &>(rhs);
```
phisiart:
```C++
auto &other = static_cast<planner::HashJoinPlan &>(rhs);
```

src/planner/hash_plan.cpp
```diff
@@ -27,5 +27,21 @@ void HashPlan::PerformBinding(BindingContext &binding_context) {
   }
 }
 
+bool HashPlan::operator==(AbstractPlan &rhs) const {
+  if (GetPlanNodeType() != rhs.GetPlanNodeType())
+    return false;
+ 
+  auto &other = reinterpret_cast<planner::HashPlan &>(rhs);
```
phisiart:
```C++
auto &other = static_cast<planner::HashPlan &>(rhs);
```

src/planner/insert_plan.cpp
```diff
@@ -198,5 +198,42 @@ void InsertPlan::SetParameterValues(std::vector<type::Value> *values) {
     }
   }
 }
+
+bool InsertPlan::operator==(AbstractPlan &rhs) const {
+  if (GetPlanNodeType() != rhs.GetPlanNodeType())
+    return false;
+
+  auto &other = reinterpret_cast<planner::InsertPlan &>(rhs);
```
phisiart:
```C++
auto &other = static_cast<planner::InsertPlan &>(rhs);
```

src/planner/order_by_plan.cpp
```diff
@@ -44,5 +44,40 @@ void OrderByPlan::PerformBinding(BindingContext &binding_context) {
   }
 }
 
+bool OrderByPlan::operator==(AbstractPlan &rhs) const {
+  if (GetPlanNodeType() != rhs.GetPlanNodeType())
+    return false;
+
+  auto &other = reinterpret_cast<planner::OrderByPlan &>(rhs);
```
phisiart:
```C++
auto &other = static_cast<planner::OrderByPlan &>(rhs);
```

src/planner/projection_plan.cpp
```diff
@@ -33,5 +34,34 @@ void ProjectionPlan::PerformBinding(BindingContext &context) {
   GetProjectInfo()->PerformRebinding(context, inputs);
 }
 
+bool ProjectionPlan::operator==(AbstractPlan &rhs) const {
+  if (GetPlanNodeType() != rhs.GetPlanNodeType())
+    return false;
+
+  auto &other = reinterpret_cast<planner::ProjectionPlan &>(rhs);
```
phisiart: auto &other = static_cast<planner::ProjectionPlan &>(rhs);

src/planner/seq_scan_plan.cpp
```diff
@@ -265,5 +266,41 @@ void SeqScanPlan::SetParameterValues(std::vector<type::Value> *values) {
   }
 }
 
+bool SeqScanPlan::operator==(AbstractPlan &rhs) const {
+  if (GetPlanNodeType() != rhs.GetPlanNodeType())
+    return false;
+
+  auto &other = reinterpret_cast<planner::SeqScanPlan &>(rhs);
```
phisiart:
```C++
auto &other = static_cast<planner::SeqScanPlan &>(rhs);
```

src/planner/update_plan.cpp
```diff
@@ -43,5 +43,32 @@ void UpdatePlan::SetParameterValues(std::vector<type::Value> *values) {
   children[0]->SetParameterValues(values);
 }
 
+bool UpdatePlan::operator==(AbstractPlan &rhs) const {
+  if (GetPlanNodeType() != rhs.GetPlanNodeType())
+    return false;
+
+  auto &other = reinterpret_cast<planner::UpdatePlan &>(rhs);
```
phisiart:
```C++
auto &other = static_cast<planner::UpdatePlan &>(rhs);
```

src/codegen/query_cache.cpp
seojungmin: I can confirm :-). I admit that I was careless about this. A good catch.

phisiart:
Let me review this later today...

seojungmin:
These are some major query cache features that are not handled by this PR, but can be added later.

  - Persistency - increase the performance when rebooted
  - Intelligent eviction policies - e.g. Pin heavy compilation-required queries which are executed regularly

seojungmin:
#811  includes this codegen cache feature, so I close this one.

PR #781, opened 2017-08-08
Remove namespace 'expression' from ContainerTuple 

seojungmin:
ContainerTuple is in src/common, but was still under namespace expression.  This PR removes namespace expression from ContainerTuple, for the sake of clarification and also of reading the code more effectively.

This is simple and straightforward, so it can be merged after the build passes.


apavlo:
@seojungmin I love the location in the commit message :-)

seojungmin:
@apavlo It was inspired by your commit messages.

seojungmin:
This is now ready.  @apavlo I really recommend it if you get a good deal.  I had one.  It is not big, but upscale.  The valet guy was also cool, and he parked my car on the street to save the overnight parking fee, which was ridiculously expensive.

PR #783, opened 2017-08-08
Removing unused header

PauloAmora:
Removing the unused `mutex` header from `TableFactory`


src/storage/table_factory.cpp
```diff
@@ -19,7 +19,6 @@
 #include "storage/storage_manager.h"
 #include "storage/temp_table.h"
 
-#include <mutex>
```
seojungmin: I think you can remove `#include "index/index.h"` and `#include "common/logger.h"`, too.

PauloAmora: Got it, they were also disposable.

seojungmin:
The PR LGTM.  

However, I think there are two other issues with this file. First, I wonder why this is called a factory.  This does not look like a factory pattern, but just a wrapper. Second, `DropDataTable()` does not seem correct.  The method does not remove table, but frees the object, which will leave a dangling pointer after all.  One should use `DropTableWithOid()` instead of `GetTableWithOid()`+`delete`.  There could be addressed in a separate PR.

apavlo:
@PauloAmora I will merge this when the build passes.

PR #785, opened 2017-08-12
Remove split(), find() functions in the frontend

dengyuchenkit:
This PR removes the usage of functions like split(), find() in the frontend. 

Previously, the parser is called in PrepareStatement() in traffic_cop. In most cases, when protocol_handler receives one query string, it calls PrepareStatement() to parse the query and prepare a statement. But for some query like EXECUTE, it does not need to prepare a statement. So, the frontend will do some preprocessing, something like splitting the query by " ", extracting the first string to determine the query type.

Now, the parser is moved from TrafficCop to ProtocolHandler. So the protocol_handler calls parser to generate a parse tree and determine the query type. Then it gives the parse tree instead of the query string to PrepareStatement to prepare a statement.


apavlo:
@dengyuchenkit What's the status of this? Do you still want to merge this?

dengyuchenkit:
@apavlo I think we can merge it, it's fixed in this pr

src/network/protocol_handler.cpp
```diff
@@ -359,19 +361,20 @@ void ProtocolHandler::ExecQueryMessage(InputPacket *pkt, const size_t thread_id)
...
+    Statement::MapToQueryType(query_type_string, query_type);
 
     switch (query_type) {
       case QueryType::QUERY_PREPARE:
       {
         std::string statement_name;
-        stream >> statement_name;
-        std::size_t pos = query.find("AS");
+        std::vector<std::string> tokens;
+        boost::split(tokens, query, boost::is_any_of("(), "));
```
apavlo: This seems like a hack. Why do we want to do this? Why can't this be done in the parser?

dengyuchenkit: Here it is to get the real statement out of a PREPARE query, such as 'prepare f as insert into foo values($1)'. It will get the statement as 'insert into foo values($1)', and store it in the statement cache with name 'f'. Then it will put the 'f' statement into the parser and build a plan tree. Right now it is done in the protocol handler rather than parser. The parser won't take care of the PREPARE query.
The use of 'split' seems hack. I can go to see how postgres handles PREPARE query.
Can I ask that, do you mean this should be done in the parser instead of protocol handler, or the use of 'split' is too hack?

dengyuchenkit: By testing the parser, it seems parser doesn't parse the PREPARE query with 'prepare' tag. Let me see how it is working and make sure

apavlo: Coordinate with @haojin2 on this. He was in charge of the new parser last semester.

dengyuchenkit: Parser can handle this. Sorry I made a mistake when testing. 
Right now in protocol_handler, there are several places where we use functions like split(), find() to analyze the query, before putting it into the parser, like the way to get query_type and handle PREPARE query and EXECUTE query. Shall we change it now?

apavlo: Yes change it now.

src/network/network_connection.cpp
```diff
@@ -466,6 +467,7 @@ bool NetworkConnection::ProcessStartupPacket(InputPacket* pkt, int32_t proto_ver
       // loop end?
       if (pkt->ptr >= pkt->len) break;
       GetStringToken(pkt, client_.dbname);
+      LOG_DEBUG("%s", client_.dbname.c_str());
```
chenboy: Please remove LOG_DEBUG here

src/network/network_connection.cpp
```diff
@@ -475,6 +477,7 @@ bool NetworkConnection::ProcessStartupPacket(InputPacket* pkt, int32_t proto_ver
       if (pkt->ptr >= pkt->len) break;
       GetStringToken(pkt, value);
       client_.cmdline_options[token] = value;
+      LOG_DEBUG("%s", value.c_str());
```
chenboy: Ditto

src/parser/sql_statement.cpp
```diff
@@ -80,5 +82,61 @@ const std::string SQLStatementList::GetInfo() const {
   return os.str();
 }
 
+QueryType StatementTypeToQueryType(StatementType stmt_type, std::shared_ptr<parser::SQLStatement> sql_stmt) {
+  LOG_INFO("%s", StatementTypeToString(stmt_type).c_str());
+  QueryType query_type = QueryType::QUERY_OTHER;
+  std::map<StatementType, QueryType> type_map {
```
chenboy: Should an unordered_map be enough here?

apavlo:
@chenboy I agree that we should switch to `unique_ptr` where appropriate.

src/parser/sql_statement.cpp
dengyuchenkit: Yes, I will change it to unordered_map!

dengyuchenkit:
Hi, I prepare a new branch that fixes the things mentioned in the review. In this PR I didn't rebase cleanly. I would like to close this PR and then I will submit a new PR about the issue.

PR #793, opened 2017-08-25
Frontend Rewrite (Workable version)

ChTimTsubasa:
This is the first fully-functional version of rewrite front-end.
**What is included:**
1. Worker pool infrastructure;
2. TrafficCop and ProtocolHandler function refactor to support multithreading.
3. ProtocolHandler created at runtime.
4. Separate events in NetworkConnection to support multithreading trigger.
5. Dozens of rename and function relocation.
6. Test cases related to Network and TrafficCop is refactored.

**What remains to be changed, added or discussed.**
1. Check and store database name at the initial packet. Pass the database name when doing the binding.
2. Control of the number of connection. Right now the front end can potentially handle as many connections as possible.
3. Connection scheduling mechanism. We are using round robin right now.
4. The ownership of the request(previously as rpkt) and responses. 
_I personally prefer to put them in ProtocolHandler and make ProtocolHandler looks like directly read from rbuf and write to wbuf. This would simplify the NetworkConnection's logic and give more flexibility to other ProtocolHandler._
5. The ownership of statements and its related data structure like param, results...
6. More test cases required (GMock, Test for JDBC batch insert [though we have tested it manually])

Since this is a major change, we would suggest someone can review and merge it first and we will continue working on the rest remained.


apavlo:
@ChTimTsubasa Do you want somebody to take an initial pass on this for a code review?

ChTimTsubasa:
@apavlo Yes.

src/executor/plan_executor.cpp
```diff
@@ -39,15 +39,16 @@ void CleanExecutorTree(executor::AbstractExecutor *root);
...
     concurrency::Transaction *txn, const std::vector<type::Value> &params,
     std::vector<StatementResult> &result,
-    const std::vector<int> &result_format) {
+    const std::vector<int> &result_format,
+    executor::ExecuteResult &p_status) {
   PL_ASSERT(plan != nullptr && txn != nullptr);
   LOG_TRACE("PlanExecutor Start (Txn ID=%lu)", txxn->GetTransactionId());
 
-  ExecuteResult p_status;
+//  ExecuteResult p_status;
```
hzxa21: Remove useless codes.

src/network/network_connection.cpp
```diff
@@ -21,53 +26,80 @@ void NetworkConnection::Init(short event_flags, NetworkThread *thread,
...
 
     if (result != 0) {
-      LOG_ERROR("Failed to update event");
+      LOG_ERROR("Failed to update network event");
+      PL_ASSERT(false);
+    }
+  }
+
+  if (workpool_event == nullptr) {
+    workpool_event = event_new(thread->GetEventBase(), -1, EV_WRITE|EV_PERSIST,
```
hzxa21: Minor thing: why do we set the EV_WRITE flag if fd=-1 (the event needs to be triggered manually)?

src/network/network_connection.cpp
```diff
@@ -21,53 +26,80 @@ void NetworkConnection::Init(short event_flags, NetworkThread *thread,
...
+    CallbackUtil::EventHandler, this);
+  } else {
+    if (event_del(workpool_event) == -1) {
+      LOG_ERROR("Failed to delete event");
+      PL_ASSERT(false);
+    }
+    auto result = event_assign(workpool_event, thread->GetEventBase(), -1,
+                                EV_WRITE|EV_PERSIST, CallbackUtil::EventHandler, this);
```
hzxa21: Same here.

src/network/network_connection.cpp
dengyuchenkit: Yes, when the worker thread in worker pool finishes execution, it uses event_active to manually trigger the network thread. This event is only triggered manually.

hzxa21: I know. But I mean maybe EV_PERSIST is enough?

apavlo:
@hzxa21 @ChTimTsubasa Do we want to merge this?

ChTimTsubasa:
@apavlo 
Sorry for the delay. I tested it through YCSB and compare with the origin. There is an approximately 8% slowdown in this PR compared to the current branch, which is acceptable since we have neither really applied intraquery parallelism, nor tune the number of threads to use for worker pool or network pool. There is no crash during the test. 

So I believe this PR is good to merge.

Issue #736, opened 2017-07-09
Remove print statements from ParserUtils


apavlo:
Somebody added many of the parser files to the [source_validator.py](https://github.com/cmu-db/peloton/blob/master/script/validators/source_validator.py) script:

https://github.com/cmu-db/peloton/blob/master/script/validators/source_validator.py#L71

The purpose of this script is to help make sure that we don't do anything stupid in our code.

It looks like the biggest offender is ParserUtils:

https://github.com/cmu-db/peloton/blob/master/src/parser/parser_utils.cpp#L198

------------------

Here is what we need to do:

* Remove the parser files from the validator's white list.
* Move all of the functions from 'parser_utils.cpp' into a helper class called `ParserUtils` as static methods.
* All of the "Print" or "Get" methods should return strings and not print output.

vipuldcoder:
Online 132  parser = argparse.ArgumentParser(description='Check for some strings')
So here Ones now what?

apavlo:
@vipuldcoder I don't understand what you are asking?

PR #795, opened 2017-09-02
Fix for Issue #736

haojin2:
This is the fix for issue #736 and contains:
 - Creation of new ParserUtils class
 - Migration of the utils from printing out to building a string for the target statement
 - Addition of utils for Update and Copy statements
 - Removal of related files from validator skipping list
 - Some tests for the parser utils

### DIFF
--- a/src/include/parser/parser_utils.h
+++ b/src/include/parser/parser_utils.h
@@ -19,7 +19,12 @@
 namespace peloton {
 namespace parser {
 
+<<<<<<< 81add772e8e4a9d7c11547d9e025a95b2df65fe4
 class ParserUtils {
+=======
+class ParserUtils
+{
+>>>>>>> Fix for Issue #736
 public:
   static std::string GetSelectStatementInfo(SelectStatement* stmt,
                                             uint num_indent);
@@ -40,8 +45,11 @@ public:
                        const expression::AbstractExpression* expr,
                        uint num_indent);
   static std::string GetTableRefInfo(const TableRef* table, uint num_indent);
+<<<<<<< 81add772e8e4a9d7c11547d9e025a95b2df65fe4
 private:
   static std::string indent(uint num_indent);
+=======
+>>>>>>> Fix for Issue #736
 };
 
 }  // namespace parser
--- a/src/include/parser/parser_utils.h
+++ b/src/include/parser/parser_utils.h
@@ -6,7 +6,7 @@
 //
 // Identification: src/include/parser/parser_utils.h
 //
-// Copyright (c) 2015-16, Carnegie Mellon University Database Group
+// Copyright (c) 2015-17, Carnegie Mellon University Database Group
 //
 //===----------------------------------------------------------------------===//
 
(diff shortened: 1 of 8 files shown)


src/parser/parser_utils.cpp
```diff
@@ -25,55 +25,33 @@ namespace parser {
...
-}
-
-void inprintU(UNUSED_ATTRIBUTE uint64_t val, UNUSED_ATTRIBUTE uint num_indent) {
-  LOG_TRACE("%lu", val);
-}
-
-void PrintTableRefInfo(TableRef* table, UNUSED_ATTRIBUTE uint num_indent) {
+std::string ParserUtils::GetTableRefInfo(const TableRef* table,
+                                         uint num_indent) {
+  std::string output;
```
apavlo: Use `std::ostringstream` instead of appending to a `std::string`. It's more efficient.

src/parser/parser_utils.cpp
```diff
@@ -25,55 +25,33 @@ namespace parser {
 
 std::string indent(uint num_indent) { return std::string(num_indent, '\t'); }
```
apavlo: Don't use tabs. Use spaces.

apavlo: Don't pollute the namespace. Make this a private method in `ParserUtils`

test/parser/parser_utils_test.cpp
```diff
@@ -0,0 +1,160 @@
+//===----------------------------------------------------------------------===//
+//
+//                         Peloton
+//
+// parser_utils_test.cpp
+//
+// Identification: test/parser/parser_utils_test.cpp
+//
+// Copyright (c) 2015-16, Carnegie Mellon University Database Group
```
apavlo: Use latest header.

src/include/parser/parser_utils.h
```diff
@@ -19,13 +19,29 @@
 namespace peloton {
 namespace parser {
 
-void GetSelectStatementInfo(SelectStatement* stmt, uint num_indent);
-void GetInsertStatementInfo(InsertStatement* stmt, uint num_indent);
-void GetCreateStatementInfo(CreateStatement* stmt, uint num_indent);
-void GetDeleteStatementInfo(DeleteStatement* stmt, uint num_indent);
-void GetExpressionInfo(const expression::AbstractExpression* expr,
+class ParserUtils
+{
```
apavlo: Run formatter.

src/parser/parser_utils.cpp
haojin2: Actually this was not written by me, so do you prefer 2 spaces instead?

apavlo: Use `GETINFO_SPACER`:

https://github.com/cmu-db/peloton/blob/master/src/include/util/string_util.h#L25

haojin2: Please take a look at my latest commit, if you like that we can just use that.

apavlo: Good enough.

apavlo:
@haojin2 Rebase so I can merge.

Issue #404, opened 2016-12-16
Create "PelotonMain" object


apavlo:
There is no single object that encapsulates the different objects in the system. Both @malin1993ml and @sid1607 have the problem that they need to maintain stateful objects that are used in different parts of the system but there is no single location to put them. The TrafficCop object is really just a bunch of static methods, so it's not really the right place. 

I propose that  create a new class called `PelotonMain` that is responsible for instantiating the different parts of the system and connecting them together (e.g., thread pools, optimizer, traffic cop, executors). We can make this a singleton object that can be retrieved by any part of the system that needs it (need to think about this, it might be a bad idea).

jessesleeping:
From my experience, for the following components, we should have the orders below:
- epoch manager -> gc manager
- epoch manager -> log manager -> transaction worker threads / networking threads

Currently their initialization order depends on whose GetInstance() method gets called first... I have to manually call the GetInstance() to enforce the order in our micro-benchmarks...

For the implementation, I think it's good to have a class that is in charged of all the initialization logic. May be each component (e.g. epoch_manager, gc_manager, thread_pool) should have a uniform interface of  `Init(...)` and `Start(...)`, where Init() initializes the parameters of the component and Start() spawns threads of that component. The centralized init logic can call Init() and Start() of different components to enforce the order.

PR #796, opened 2017-09-05
PelotonMain

huangminchn:
issue #404 

A singleton object "PelotonMain" is created and it contains all the singleton objects in the system.

1. ThreadPool, NetworkManager, EpochManager and GCManager become member variable in PelotonMain. Whenever these 4 instances are needed, get them from the PelotonMain object instead of calling their static GetInstance function;
2. I didn't make TrafficCop as a member in PelotonMain because TrafficCop is not a global singleton instance (except for unit test); a new TrafficCop instance is created for each network connection;

TODO:
1. add other global singletons to PelotonMain;
2. modify test cases so that they retrieve global singletons from PelotonMain


apavlo:
We need to have a longer discussion about this PR. We need to remove all of the singleton classes entirely and pass around a single `PelotonMain` object everywhere that needs it.

src/peloton_main/peloton_main.cpp
```diff
@@ -1,16 +1,8 @@
-//===----------------------------------------------------------------------===//
 //
-//                         Peloton
+// Created by Min Huang on 8/28/17.
```
apavlo: Don't do this.

apavlo:
@jarulraj @malin1993ml @pmenon @tomasic @hzxa21 @poojanilangekar -- I was thinking that maybe we want to merge this PR as a temporary solution to getting rid of all of the static singleton objects. It would be a first step to get rid of all of the global singleton calls, and then work to change the code to pass around `PelotonMain` handle correctly.

Thoughts?

src/include/peloton_main/peloton_main.h
```diff
@@ -0,0 +1,42 @@
...
+#include "concurrency/epoch_manager.h"
+#include "gc/gc_manager.h"
+
+namespace peloton {
+
+//===--------------------------------------------------------------------===//
+// Global Setup and Teardown
+//===--------------------------------------------------------------------===//
+
+class PelotonMain {
```
seojungmin: Would there be any reason why we could not make this class be just "Peloton"? Peloton sounds better over PelotonMain.

pmenon: Or just `PelotonDB` or `Database`?

poojanilangekar:
@apavlo Are we planning to merge this by the end of this week?

apavlo:
@poojanilangekar No, I think we are going to re-evaluate our options with @pervazea and @tli2 after the new year.

apavlo:
We're going to abandon this PR and revisit it next semester.

PR #801, opened 2017-09-08
Use smart pointers in triggers

pmenon:
I noticed triggers were using raw pointers and `new`. This PR fixes it to use smart pointers. I also formatted some code, updated licenses, and fixed namespace closures.

### DIFF
--- a/src/catalog/multi_constraint.cpp
+++ b/src/catalog/multi_constraint.cpp
@@ -35,5 +35,5 @@ const std::string MultiConstraint::GetInfo() const {
   return os.str();
 }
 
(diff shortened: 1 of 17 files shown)


apavlo:
Looks good to me. Thanks for doing this. I will merge when the build passes.

I don't know why the OSX build is going so slow on Travis.

apavlo:
@pmenon The Travis Mac cluster is down. I can merge this without the Mac build if you have tested it on your laptop?

pmenon:
@apavlo I just kicked off the Mac build and it looks good now 👍🏽

Issue #787, opened 2017-08-13
CREATE INDEX response string returns "CREATE TABLE"


fadgan:
Index creation output tells us as if a table has been created although catalog confirms that the correct actions have been performed under the commands:
```
postgresql=# create table foo3 (id integer);
CREATE TABLE
postgresql=# insert into foo3 values (1);
INSERT 0 1
postgresql=# create index ind on foo3(id);
CREATE TABLE
postgresql=# select * from pg_catalog.pg_index where index_name = 'ind';
 index_oid | index_name | table_oid | index_type | index_constraint | unique_keys | indexed_attributes 
-----------+------------+-----------+------------+------------------+-------------+--------------------
  50331759 | ind        |  33554538 |          1 |                1 | false       | 0 
(1 row)

```

apavlo:
Good find. We will fix.

apavlo:
@sivaprasadsudhir Can you take care of this? The function to modify is here:

https://github.com/cmu-db/peloton/blob/master/src/network/protocol_handler.cpp#L278

You should coordinate with @ChTimTsubasa about this change so that you don't foul up overhaul of the protocol handler code.

apavlo:
Fixed in #805

PR #805, opened 2017-09-19
For #787, to fix the response string for create table/index

sivaprasadsudhir:


apavlo:
@sivaprasadsudhir Fix the conflicts and I will merge.

src/common/statement.cpp
```diff
@@ -51,6 +51,17 @@ void Statement::ParseQueryTypeString(const std::string& query_string,
   boost::to_upper(query_type_string);
 }
 
+void Statement::ParseCreateTypeString(const std::string& query_string,
+                                     std::string& create_type_string) {
+  std::stringstream stream(query_string);
+  std::string query_type_string;
+  stream >> query_type_string;
+  boost::to_upper(query_type_string);
+  PL_ASSERT(query_type_string == "CREATE");
```
seojungmin: Can we remove line 57-59 and put `PL_ASSERT(create_type_string == "CREATE");` after line 62? The line 57-59 look redundant to me.

sivaprasadsudhir: query_type_string should be "CREATE" and create_type_string should be "TABLE" or "INDEX". I can remove the variable query_type_string and read into create_type_string twice. Is that what you mean?

seojungmin: I see. This is going to be ok for now. 👍 Thanks for the explanation.

PR #806, opened 2017-09-20
Fix files that are missing a trailing EOL.  Purely cosmetic change.

crd477:


apavlo:
Why do we want to do this? This seems like a lot of changed files for nothing?

crd477:
On 2017-09-20 07:12, Andy Pavlo wrote:
> Why do we want to do this? This seems like a lot of changed files for 
> nothing?
> 

The intention is merely for consistency and to satisfy editors that 
notice that condition.  You may not want to merge it but I noticed it as 
I was reading some of the files.

-- 
     -Chad

PR #811, opened 2017-09-25
Codegen cache/parameterization

seojungmin:
This PR brings a new feature, parameterization and cache into Peloton. This is a rewrite of a 15721 project.

It introduces two new classes: `expression/Parameter` and `codegen/QueryParameters`.  The former encapsulates parameter information from `ParameterValueExpression` and `ConstantValueExpression`, and the latter keeps all the `Parameter`s related to the query.  Once a query is executed, it extracts all the `Parameter`s and stores them into QueryParameters, which is now handed over to the query module generated by codegen.  Then, the `constant`/`parameter` translators are built to retrieve these values from `QueryParameters`.


seojungmin:
One thing that could be added to this parameterization feature is to build up an LLVM compact storage  when passing the parameters to the compiled query.   Once a query is executed, it builds up its own storage that can be index-referenced and the translator will just access this memory storage to get the required values.  It will transform the function calls into memory accesses.   This implementation will be done separately.

pmenon:
I suggest we either implement the parameter value storage in this PR, or extract out the caching functionality and merge that in separately.

This is because, in its current state, this generates code that makes one function call per constant (or parameter) in any expression, for every row. This is costly. These values should be extracted and cached once.

seojungmin:
@pmenon I think the cache/parameterization could go together. I was being lazy, and I left the issue on #814 hoping that somebody could pick it up.  I thought this would be a good practice for someone who would like a journey with codegen.  Since there is no self-assigned one yet, I will just add the parameter value storage feature on this PR.

pmenon:
@seojungmin Let me know when this is ready for review.

seojungmin:
@pmenon This is ready for review.

pmenon:
I'll take a look at this. Can you provide a high-level overview of what's done, the main classes, and the code flow?

seojungmin:
@pmenon Sure. 

The recent changes after the last review are 1) the special parameter storage that prevents the execution engine from calling functions every time a parameterized value is required, and 2) those changes due to the codegen INSERT merge.

Caching INSERT queries is rewritten by applying the parameter storage feature that I mentioned above.  This parameter storage stores all the values to be inserted.  To achieve this, `InsertPlan` no longer build an tuple object, but puts all the values in a vector.  The traditional interpreter execution engine is also modified for compatibility.  For INSERT, the old engine is still directly called from many places.  We need to garbage collect them later after pre-generating the INSERT compiled object at the table creation time.

Coming back to the parameterization, the major classes related to parameterization are as follow:

1. `QueryParameters` collects the parameter information from each query plan (by `VisitParameters()` calls) and handed over to our codegen compilation/execution engine. The role of the class is the same, but the constant/parameter translators do not call the proxy of this class, but they call the new `ParameterStorage` to retrieve the parametrized value.

2. `ParameterStorage`, which is recently added, holds the parameter values.  This storage is set up at compiled time (See `CompilationContext`).  In other words, the structure of the storage is decided at compilation and the values are set at runtime, which is after runtime state's local state is created.

After all, with this PR,
  - We now cache all the codegen supported queries.
  - We correctly parameterize all the constant/parameterized values including INSERT.

src/include/codegen/query_cache.h
```diff
@@ -0,0 +1,86 @@
...
+class QueryCache : public Singleton<QueryCache> {
+ public:
+  Query* Find(const std::shared_ptr<planner::AbstractPlan> &key);
+
+  void Add(const std::shared_ptr<planner::AbstractPlan> &key,
+           std::unique_ptr<Query> &&val);
+
+  size_t GetSize() { return size_; }
+
+  void SetSize(size_t size) {
```
pmenon: Size and count are a little confusing. Can you rename `size` to capacity to indicate an upper bound?

src/codegen/updater.cpp
```diff
@@ -171,7 +171,7 @@ void Updater::UpdatePK() {
 
 void Updater::TearDown() {
   // Updater object does not destruct its own data structures
-  tile_.reset();
+  if (tile_ != nullptr) tile_.reset();
```
seojungmin: This line will be fixed in the final version of this PR.

seojungmin:
@pmenon Good catches.  Thanks for spending your time on this big PR.  Fixes for the last comments are applied.  I added some fixes such as removing `txn` and modifying `executor_context` type  in Execute() API.  You could take a second pass while I add the planner equality test, which seems to take a little longer.

malin1993ml:
I feel like the high-level overview that Tony has provided above as well as some description of generally how this parameterization works could be something that goes into the design doc or at least some wiki pages? So that other people that start to work on related projects don't need to ask all the details in person or chase information down to random places in PR discussions.

seojungmin:
@malin1993ml I 100% agree. I will add a design doc, once this PR is merged, although it should have been the other way around.

seojungmin:
@pmenon Ready for the next round.  Hash/Equality tests are added.

seojungmin:
I have added the design doc.  Guys, this is ready to review.  Thanks for your time spending on this PR, and I am happy to hear any comment. 

@chenboy @hzxa21 For the non-codegen parts, please look all those other than `codegen` & `executor`, e.g. `expressions`, `planner`, etc.

@prashasthip @luochenUmich For the codegen parts, please look at `codegen` and `executor`.

@pmenon Am I supposed to separate the PR? What was the final decision yesterday?

pmenon:
@seojungmin We will keep this as one PR for now, and have multiple people review it.

src/include/planner/seq_scan_plan.h
```diff
@@ -46,32 +49,45 @@ class SeqScanPlan : public AbstractScan {
...
-  bool SerializeTo(SerializeOutput &output) const ;
-  bool DeserializeFrom(SerializeInput &input);
+  bool SerializeTo(SerializeOutput &output) const override;
+  bool DeserializeFrom(SerializeInput &input) override;
 
   /* For init SerializeOutput */
-  int SerializeSize();
+  int SerializeSize() override;
 
   oid_t GetColumnID(std::string col_name);
```
prashasthip: Make function const

src/planner/aggregate_plan.cpp
```diff
@@ -121,5 +121,131 @@ void AggregatePlan::PerformBinding(BindingContext &binding_context) {
   }
 }
 
+hash_t AggregatePlan::Hash(
+    const std::vector<planner::AggregatePlan::AggTerm> &agg_terms) const {
+  hash_t hash = 0;
+  for (auto &agg_term : agg_terms) {
+    hash = HashUtil::CombineHashes(hash, HashUtil::Hash(&agg_term.aggtype));
+    if (agg_term.expression != nullptr)
+      hash = HashUtil::CombineHashes(hash, agg_term.expression->Hash());
```
prashasthip: Maybe add braces around if-then to make it more readable?

src/codegen/query_cache.cpp
```diff
@@ -0,0 +1,77 @@
...
+    return nullptr;
+  }
+  query_list_.splice(query_list_.begin(), query_list_, it->second);
+  return it->second->second.get();
+}
+
+void QueryCache::Add(const std::shared_ptr<planner::AbstractPlan> &key,
+                     std::unique_ptr<Query> &&val) {
+  query_list_.push_front(make_pair(key, std::move(val)));
+  cache_map_.insert(make_pair(key, query_list_.begin()));
```
hzxa21: I think this function might be called by multiple worker threads at the same time. I may be wrong but It seems like this function is not thread-safe. Will it cause any problem?

seojungmin: A good catch.  A read-write lock around `QueryCache` is added.

src/include/planner/hash_join_plan.h
```diff
@@ -6,7 +6,7 @@
 //
 // Identification: src/include/planner/hash_join_plan.h
 //
-// Copyright (c) 2015-16, Carnegie Mellon University Database Group
+// Copyright (c) 2015-17, Carnegie Mellon University Database Group
```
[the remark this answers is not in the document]
seojungmin: I think we would need a separate script to update this at the same time.

src/expression/abstract_expression.cpp
```diff
@@ -102,12 +102,24 @@ const std::string AbstractExpression::GetInfo() const {
   return os.str();
 }
 
-bool AbstractExpression::Equals(AbstractExpression* expr) const{
-  if (exp_type_ != expr->exp_type_ ||
-      children_.size() != expr->children_.size())
+bool AbstractExpression::AreEqual(const AbstractExpression *expr1,
```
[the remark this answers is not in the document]
seojungmin: This is removed after refactoring the code at the next commit.

src/codegen/query_cache.cpp
```diff
@@ -0,0 +1,77 @@
...
+}
+
+oid_t QueryCache::GetOidFromPlan(const planner::AbstractPlan &plan) {
```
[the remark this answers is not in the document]
seojungmin: Goot catch.

src/expression/abstract_expression.cpp
```diff
@@ -121,5 +133,26 @@ hash_t AbstractExpression::Hash() const {
   return hash;
 }
 
+bool AbstractExpression::ExactlyEquals(const AbstractExpression &other)
```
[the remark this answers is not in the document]
seojungmin: `ExactlyEquals` is different from `==` as the former is used to check that the expressions are exactly the same.  The latter do not check constant/parameter values equality, so they should exist separately.

src/include/codegen/compilation_context.h
```diff
@@ -84,6 +90,14 @@ class CompilationContext {
   // Get a pointer to the executor context instance
   llvm::Value *GetExecutorContextPtr();
 
+  // Get a pointer to the query parameter instance
+  llvm::Value *GetQueryParametersPtr();
+
+  // Get the parameter index to be used to get value, for the given expression
+  size_t GetParameterIdx(const expression::AbstractExpression *expression) {
```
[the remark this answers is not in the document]
seojungmin: A good catch.

src/include/expression/constant_value_expression.h
```diff
@@ -59,12 +97,7 @@ class ConstantValueExpression : public AbstractExpression {
 
   virtual void Accept(SqlNodeVisitor *v) override { v->Visit(this); }
 
-  virtual hash_t Hash() const override {
-    hash_t hash = HashUtil::Hash(&exp_type_);
-    return HashUtil::CombineHashes(hash, value_.Hash());
-  }
-
-  bool IsNullable() const override { return false; }
+  bool IsNullable() const override { return value_.IsNull(); }
```
pmenon: Is this correct? If you're extracting constants out, then NULLness can change between invocations?

src/include/codegen/parameter_cache.h
```diff
@@ -0,0 +1,53 @@
...
+// A parameter cache which stores the parameterized values for codegen runtime
+//===----------------------------------------------------------------------===//
+class ParameterCache {
+ public:
+  // Constructor
+  ParameterCache(const std::vector<expression::Parameter> &parameters) :
+      parameters_(parameters) {}
+
+  // Set the parameter values
+  void StoreValues(CodeGen &codegen, llvm::Value *query_parameters_ptr);
```
pmenon: Change this to `PopulateCache` or something. `StoreValues` doesn't really convey what's going on?

src/include/codegen/query_cache.h
```diff
@@ -0,0 +1,100 @@
...
+  std::list<std::pair<std::shared_ptr<planner::AbstractPlan>,
+                      std::unique_ptr<Query>>> query_list_;
+
+  std::unordered_map<std::shared_ptr<planner::AbstractPlan>,
+           decltype(query_list_.begin()),
+           planner::Hash, planner::Equal> cache_map_;
+
+  RWLock cache_lock_;
+
+  size_t capacity_ = 0;
```
pmenon: Can you move all implementation to the `cpp` file.

src/include/codegen/query.h
```diff
@@ -55,29 +57,48 @@ class Query {
...
+  ParameterCache &GetParameterCache() {
+    return parameters_cache_;
+  }
+
  private:
   friend class QueryCompiler;
 
   // Constructor
-  Query(const planner::AbstractPlan &query_plan);
+  Query(const planner::AbstractPlan &query_plan, QueryParameters &parameters);
```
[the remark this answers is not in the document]
seojungmin: First,I agree that `Query` does not have to (or should not) carry `QueryParameters` and `ParameterCache`. It was something that I missed to refactor after building a working version. I moved them to `CompilationContext`. 

For the second part of your question, I think the separation of those concepts is not much useful for `CompilationContext`.  This is because the parameter cache in the `CompilationContext` should know about the _values_(types, nullability) of the parameters, although the values are not required.  The separation of actual values from the types would clarify this situation.  I will give another thought on this while fixing the null, below, which I thought IsNull() will be incorrect..

Thanks for point these out.

src/include/expression/constant_value_expression.h
seojungmin: This is correct. This only fixes the bug in `IsNullable()`, and not related to parameterization.

The constant values are always considered as non-nullable.  Please see `VisitParameters()` in `constant_value_expression.h` below, which creates parameter information for this constant value expression and specifies its nullability implicitly at `CreateConstParameter()` - default is `true` here. I might specify this explicitly for code readability.

```
  virtual void VisitParameters(std::vector<Parameter> &parameters,
      std::unordered_map<const AbstractExpression *, size_t> &index,
      UNUSED_ATTRIBUTE const std::vector<peloton::type::Value>
          &parameter_values) override {
    // Add a new parameter object for a constant
    parameters.push_back(Parameter::CreateConstParameter(GetValue()));
    index[this] = parameters.size() - 1;
  };
```

src/include/codegen/query_cache.h
seojungmin: Done!

seojungmin:
This should build correctly, and is ready for being merged. The last changes made since the last review are...

- Enhance the API of codegen `Compile()` by passing only the metadata for parameters over (previously we had one class that handles both metadata and data)
- Take optimization for non-nullable constants into account by distinguishing the codegen objects built for non-nullable constants from those for nullable constants (we check this at plan comparison, and whenever we find non-nullable cases, we can benefit from non-nullability codegen optimization)

apavlo:
I will merge this next.

PR #821, opened 2017-10-02
implement cost_and_stats_calculator

bechob:
- Let Stat be the base class of TableStat
- Move some variables and functions inside Cost class to avoid linking problem
- Add cost_and_stats_calculator_test with a simple test case of testing the seqscan

### DIFF
--- a/src/include/optimizer/cost_and_stats_calculator.h
+++ b/src/include/optimizer/cost_and_stats_calculator.h
@@ -35,7 +35,7 @@ class CostAndStatsCalculator : public OperatorVisitor {
       std::vector<double> child_costs);
 
   inline std::shared_ptr<Stats> GetOutputStats() {
-//    return std::move(output_stats_);
+    //    return std::move(output_stats_);
     return output_stats_;
   }
 
@@ -80,5 +80,5 @@ class CostAndStatsCalculator : public OperatorVisitor {
   double output_cost_;
 };
 
-} // namespace optimizer
-} // namespace peloton
+}  // namespace optimizer
+}  // namespace peloton
--- a/src/include/optimizer/cost_and_stats_calculator.h
+++ b/src/include/optimizer/cost_and_stats_calculator.h
@@ -35,7 +35,8 @@ class CostAndStatsCalculator : public OperatorVisitor {
       std::vector<double> child_costs);
 
   inline std::shared_ptr<Stats> GetOutputStats() {
-    return std::move(output_stats_);
+//    return std::move(output_stats_);
+    return output_stats_;
   }
 
   inline double GetOutputCost() { return output_cost_; }
@@ -75,7 +76,7 @@ class CostAndStatsCalculator : public OperatorVisitor {
   std::vector<std::shared_ptr<Stats>> child_stats_;
   std::vector<double> child_costs_;
 
-  std::unique_ptr<Stats> output_stats_;
+  std::shared_ptr<Stats> output_stats_;
   double output_cost_;
 };
 
--- a/src/include/optimizer/stats/cost.h
+++ b/src/include/optimizer/stats/cost.h
@@ -12,36 +12,35 @@
 
 #pragma once
 
-#include <vector>
 #include <cmath>
 #include <memory>
+#include <vector>
 
-#include "table_stats.h"
-#include "value_condition.h"
 #include "common/logger.h"
 #include "common/macros.h"
+#include "table_stats.h"
+#include "value_condition.h"
 
 namespace peloton {
 namespace optimizer {
 // Default cost when cost model cannot compute correct cost.
-	static constexpr double DEFAULT_COST = 1;
+static constexpr double DEFAULT_COST = 1;
 
 // Estimate the cost of processing each row during a query.
-	static constexpr double DEFAULT_TUPLE_COST = 0.01;
+static constexpr double DEFAULT_TUPLE_COST = 0.01;
 
 // Estimate the cost of processing each index entry during an index scan.
-	static constexpr double DEFAULT_INDEX_TUPLE_COST = 0.005;
+static constexpr double DEFAULT_INDEX_TUPLE_COST = 0.005;
 
 // Estimate the cost of processing each operator or function executed during a
 // query.
-	static constexpr double DEFAULT_OPERATOR_COST = 0.0025;
+static constexpr double DEFAULT_OPERATOR_COST = 0.0025;
 
 //===----------------------------------------------------------------------===//
 // Cost
 //===----------------------------------------------------------------------===//
 class Cost {
  private:
-
   // Default cost of sorting n elements
   static double default_sorting_cost(size_t n) { return n * std::log2(n); }
 
@@ -49,13 +48,13 @@ class Cost {
   static double default_index_height(size_t n) { return std::log2(n); }
 
  public:
-	/*
+  /*
    * Cost of seq scan for the whole table
    */
-	static inline double NoConditionSeqScanCost(
-		const std::shared_ptr<TableStats>& input_stats) {
-		return input_stats->num_rows * DEFAULT_TUPLE_COST;
-	}
+  static inline double NoConditionSeqScanCost(
+      const std::shared_ptr<TableStats>& input_stats) {
+    return input_stats->num_rows * DEFAULT_TUPLE_COST;
+  }
   /*
    * Cost of scan for single condition. For scan with multiple conditions,
    * you should use CombineConjunctionStats to combine output_stats.
@@ -126,7 +125,7 @@ class Cost {
    * Note right only first column is taken into consideration.
    */
   static double OrderByCost(const std::shared_ptr<TableStats>& input_stats,
-                            const std::vector<oid_t>& columns,
+                            const std::vector<std::string>& columns,
                             const std::vector<bool>& orders,
                             std::shared_ptr<TableStats>& output_stats);
 
--- a/src/include/optimizer/stats/cost.h
+++ b/src/include/optimizer/stats/cost.h
@@ -24,30 +24,39 @@
 namespace peloton {
 namespace optimizer {
 
-// Default cost when cost model cannot compute correct cost.
-static constexpr double DEFAULT_COST = 1;
-
-// Estimate the cost of processing each row during a query.
-static constexpr double DEFAULT_TUPLE_COST = 0.01;
-
-// Estimate the cost of processing each index entry during an index scan.
-static constexpr double DEFAULT_INDEX_TUPLE_COST = 0.005;
-
-// Estimate the cost of processing each operator or function executed during a
-// query.
-static constexpr double DEFAULT_OPERATOR_COST = 0.0025;
-
-// Default cost of sorting n elements
-  double default_sorting_cost(size_t n) { return n * std::log2(n); }
-
-// Default number of index tuple to access for n elements
-  double default_index_height(size_t n) { return std::log2(n); }
 
 //===----------------------------------------------------------------------===//
 // Cost
 //===----------------------------------------------------------------------===//
 class Cost {
+ private:
+	// Default cost when cost model cannot compute correct cost.
+	static constexpr double DEFAULT_COST = 1;
+
+	// Estimate the cost of processing each row during a query.
+	static constexpr double DEFAULT_TUPLE_COST = 0.01;
+
+	// Estimate the cost of processing each index entry during an index scan.
+	static constexpr double DEFAULT_INDEX_TUPLE_COST = 0.005;
+
+	// Estimate the cost of processing each operator or function executed during a
+	// query.
+	static constexpr double DEFAULT_OPERATOR_COST = 0.0025;
+
+	// Default cost of sorting n elements
+	static double default_sorting_cost(size_t n) { return n * std::log2(n); }
+
+	// Default number of index tuple to access for n elements
+	static double default_index_height(size_t n) { return std::log2(n); }
+
  public:
+	/*
+   * Cost of scan for the whole table
+   */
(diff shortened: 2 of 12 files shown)


src/include/optimizer/stats/table_stats.h
```diff
@@ -24,16 +25,16 @@ namespace optimizer {
 #define DEFAULT_HAS_INDEX false
 
 class ColumnStats;
-
+//class Stats;
```
AllisonWang: Nit

src/optimizer/cost_and_stats_calculator.cpp
```diff
@@ -10,127 +10,306 @@
...
+		//===----------------------------------------------------------------------===//
+		// Generate output stats based on the input stats and the property column
+		std::shared_ptr<TableStats> generateOutputStat(std::shared_ptr<TableStats> input_table_stats,
+																									 const PropertyColumns* columns_prop) {
+			std::vector<std::shared_ptr<ColumnStats>> output_column_stats;
+			for (size_t i = 0; i < columns_prop->GetSize(); i++) {
+				oid_t column_id = (oid_t)((expression::TupleValueExpression *)columns_prop->GetColumn(i).get())->GetColumnId();
+				output_column_stats.push_back(input_table_stats->GetColumnStats(column_id));
+			}
+			return std::shared_ptr<TableStats>(new TableStats(input_table_stats->num_rows, output_column_stats));
```
AllisonWang: make_shared? (We might need to update other places too)

src/optimizer/cost_and_stats_calculator.cpp
```diff
@@ -10,127 +10,306 @@
...
+				ValueCondition rhs_cond(key_column_ids.at(1), expr_types.at(1), values.at(1));
+				//Get the scan_cost
+				output_cost_ = Cost::NoConditionSeqScanCost(table_stats);
+				//Update output stats num_rows
+				updateConjunctionStats(table_stats, lhs_cond, rhs_cond, predicate->GetExpressionType(), output_stats);
+
+			} else {
+
+				// TODO: support multiple conjunctions
+				output_cost_ = 2;
```
AllisonWang: What is the structure looks like if you have more than two predicates? Can we break it down and use the existing functions to compute the cost?

src/optimizer/cost_and_stats_calculator.cpp
```diff
@@ -10,127 +10,306 @@
...
+			}
+
+		};
+
+		void CostAndStatsCalculator::Visit(const PhysicalIndexScan *op) {
```
AllisonWang: Seems like the code here is somewhat repetitive. Can we refactor some of the logic out?

src/optimizer/cost_and_stats_calculator.cpp
```diff
@@ -10,127 +10,306 @@
...
+		};
+		void CostAndStatsCalculator::Visit(const PhysicalHashGroupBy * op) {
+			// TODO: Replace with more accurate cost
+			PL_ASSERT(child_stats_.size() == 1);
+			auto table_stats_ptr = std::dynamic_pointer_cast<TableStats>(child_stats_.at(0));
+			auto output_stats = generateOutputStat(table_stats_ptr, output_properties_
+				->GetPropertyOfType(PropertyType::COLUMNS)->As<PropertyColumns>());
+			std::vector<oid_t> column_ids;
+			for (auto column: op->columns) {
+				oid_t column_id = (oid_t)((expression::TupleValueExpression *)column.get())->GetColumnId();
```
AllisonWang: Why using .get() and type cast into raw pointer here but use std::dynamic_pointer_cast above?

test/optimizer/cost_and_stats_calculator_test.cpp
```diff
@@ -0,0 +1,90 @@
...
+	TestingSQLUtil::ExecuteSQLQuery(
+		"CREATE TABLE test(id INT PRIMARY KEY, name VARCHAR, salary DECIMAL);");
+	for (int i = 1; i <= N_ROW; i++) {
+		std::stringstream ss;
+		ss << "INSERT INTO test VALUES (" << i << ", 'name', 1.1);";
+		TestingSQLUtil::ExecuteSQLQuery(ss.str());
+	}
+}
+
+TEST_F(CostAndStatsCalculatorTests, SeqScanTest) {
```
AllisonWang: Let's add more tests for this later.

src/include/optimizer/stats/cost.h
```diff
@@ -24,30 +24,39 @@
...
 
-// Estimate the cost of processing each row during a query.
-static constexpr double DEFAULT_TUPLE_COST = 0.01;
+//===----------------------------------------------------------------------===//
+// Cost
+//===----------------------------------------------------------------------===//
+class Cost {
+ private:
+	// Default cost when cost model cannot compute correct cost.
+	static constexpr double DEFAULT_COST = 1;
```
AllisonWang: I don't think those constant definitions can be moved inside of the Cost class. This will cause compile errors.

src/include/optimizer/stats/table_stats.h
```diff
@@ -24,16 +25,16 @@ namespace optimizer {
...
-class TableStats {
+class TableStats: public Stats {
  public:
   TableStats() : TableStats(0) {}
 
   TableStats(size_t num_rows)
-      : num_rows(num_rows), col_stats_list_{}, col_name_to_stats_map_{} {}
+      : Stats(nullptr), num_rows(num_rows), col_stats_list_{}, col_name_to_stats_map_{} {}
 
   TableStats(size_t num_rows,
```
AllisonWang: Please also change TableStats::TableStats constructors in `table_stats.cpp`

AllisonWang: Please also change TableStats::TableStats constructors in `table_stats.cpp`

src/optimizer/cost_and_stats_calculator.cpp
bechob: I look into the optimizer's test case and I use this way to get the column id. Can't decide which one is better. Do you have any suggestion?

src/optimizer/cost_and_stats_calculator.cpp
bechob: I think so. But the original code covers the cases when the index is not searchable here which is quite similar to the PhysicalSeqScan. So I just leave this logic here.

src/optimizer/cost_and_stats_calculator.cpp
AllisonWang: let's use dynamic pointer cast

apavlo:
@bechob What is the status of this? There hasn't been any updates in two weeks.

AllisonWang:
@hzxa21 Could you review this PR? Thanks!

src/optimizer/cost_and_stats_calculator.cpp
```diff
@@ -10,127 +10,404 @@
...
+      std::vector<bool> orders;
+      for(size_t i = 0; i < sort_prop->GetSortColumnSize(); i++) {
+        auto expr = sort_prop->GetSortColumn(i);
+        if (expr->GetExpressionType() == ExpressionType::VALUE_TUPLE) {
+          auto column = std::dynamic_pointer_cast<expression::TupleValueExpression>(
+            sort_prop->GetSortColumn(i))->GetColumnName();
+
+          // In case the first column is not a tuple expression
+          if (i != 0 && columns.empty()) {
+            columns.push_back(column);
```
hzxa21: Why do we push_back the same column twice if the first column is not a tuple value expression? Why do we only consider the cases that the first column is not a tuple value expression, not other non-tuple expression columns?

src/include/optimizer/stats/cost.h
```diff
@@ -118,7 +126,7 @@ class Cost {
    * Note right only first column is taken into consideration.
    */
   static double OrderByCost(const std::shared_ptr<TableStats>& input_stats,
-                            const std::vector<oid_t>& columns,
+                            const std::vector<std::string>& columns,
```
hzxa21: Just a reminder here: in the future, the cost model should use the expression object itself to match columns since only the base columns (tuple value expressions) have valid column names. For now it is okay.

src/optimizer/cost_and_stats_calculator.cpp
bechob: Since right now the Cost::OrderByCost uses only the first column and see whether the column is a primary key and the order is ascending.  So to avoid the incorrect calculation of ```order by a+b, a ``` where a is the primary key and a+b is ignored right now, I add a fake column (order is false) first.

hzxa21: I see. Could you also add some comments in the codes for this corner case?

bechob: Sure

src/include/optimizer/stats/cost.h
AllisonWang: @hzxa21  would you mind explaining more what you mean by expression object? For base table, I am guessing expression object is just value expression (table name?) but how about for intermediate tables?

src/include/optimizer/cost_and_stats_calculator.h
```diff
@@ -35,7 +35,8 @@ class CostAndStatsCalculator : public OperatorVisitor {
       std::vector<double> child_costs);
 
   inline std::shared_ptr<Stats> GetOutputStats() {
-    return std::move(output_stats_);
+//    return std::move(output_stats_);
```
AllisonWang: Any reason why we want to change this to shared_ptr?

src/optimizer/cost_and_stats_calculator.cpp
```diff
@@ -10,127 +10,404 @@
...
+
+          auto expr = columns_prop->GetColumn(i);
+          if (expr->GetExpressionType() == ExpressionType::VALUE_TUPLE) {
+            auto column_name = (std::dynamic_pointer_cast<expression::TupleValueExpression>(expr))->GetColumnName();
+            auto column_stat = input_table_stats->GetColumnStats(column_name);
+            if (column_stat != nullptr) {
+              output_column_stats.push_back(column_stat);
+            };
+          } else {
+
```
AllisonWang: else?

src/include/optimizer/stats/cost.h
hzxa21: Each node in the physical operator tree is associated with a PropertyColumns which contains a collection of expressions to specify what the output of the current node should be. This expression object is hashable and comparable so it can be used directly to match both base table and intermediate table.

src/optimizer/cost_and_stats_calculator.cpp
```diff
@@ -10,127 +10,404 @@
...
+      }
+      auto output_table_stats = std::make_shared<TableStats>(input_table_stats->num_rows, output_column_stats);
+      return output_table_stats;
+    }
+
+    // Update output stats num_rows for conjunctions based on predicate
+    double updateMultipleConjuctionStats(const std::shared_ptr<TableStats> &input_stats,
+                                         const expression::AbstractExpression *expr,
+                                         std::shared_ptr<TableStats> &output_stats, bool enable_index) {
+      if (expr->GetChild(0)->GetExpressionType() == ExpressionType::VALUE_TUPLE ||
```
AllisonWang: Can we substitute magic number 0 and 1 to const?

src/optimizer/cost_and_stats_calculator.cpp
```diff
@@ -10,127 +10,404 @@
...
+          expr->GetChild(1)->GetExpressionType() == ExpressionType::VALUE_TUPLE) {
+
+        int right_index = expr->GetChild(0)->GetExpressionType() == ExpressionType::VALUE_TUPLE?1:0;
+        auto left_expr = reinterpret_cast<const expression::TupleValueExpression *>(
+          right_index == 1? expr->GetChild(0):expr->GetChild(1));
+        auto expr_type = expr->GetExpressionType();
+        if (right_index == 0) {
+          if (expr_type == ExpressionType::COMPARE_LESSTHANOREQUALTO) {
```
AllisonWang: How about using switch statement here

src/include/optimizer/cost_and_stats_calculator.h
bechob: Yes, because in the Cost we use shared_ptr for output_stats, if I cast unique_ptr to shared_ptr to call the functions in Cost, there'll be some issues when destroying pointers.

src/optimizer/cost_and_stats_calculator.cpp
```diff
@@ -10,127 +10,404 @@
...
+        stats_storage->GetTableStats(op->table_->GetDatabaseOid(), op->table_->GetOid());
+
+      // No table stats available
+      if (table_stats->GetColumnCount() == 0) {
+        output_stats_.reset(new Stats(nullptr));
+        output_cost_ = 1;
+        return;
+      }
+
+      auto property_ = output_properties_->GetPropertyOfType(PropertyType::COLUMNS)->As<PropertyColumns>();
```
AllisonWang: nit: property_ usually means private class variable. We can just use property here

bechob: OK!

src/optimizer/cost_and_stats_calculator.cpp
AllisonWang: Be sure to add a TODO here.

apavlo:
@bechob @AllisonWang I will merge this after the build passes.

pmenon:
I took a quick look over this and noticed a few files that don't adhere to formatting guidelines. Please run `clang-format` on the files that were changed.

apavlo:
@allisonwang @bechob Can you rebase?

PR #831, opened 2017-10-14
Clean up protocal_handler and traffic cop

haoxianghua:
move global variables from protocal_handler to traffic cop


src/include/traffic_cop/traffic_cop.h
```diff
@@ -115,6 +115,26 @@ class TrafficCop {
 
   bool is_queuing_;
```
apavlo: Can you provide documentation and not make these members public? Add Get/Set methods.

src/include/traffic_cop/traffic_cop.h
```diff
@@ -115,6 +115,26 @@ class TrafficCop {
...
 
+  //TODO: should this stay in traffic_cop?
```
ChTimTsubasa: You may just delete this comment

src/include/traffic_cop/traffic_cop.h
```diff
@@ -115,6 +115,26 @@ class TrafficCop {
...
+  std::shared_ptr<Statement> statement_;
+
+  std::string error_message_;
+
+  //TODO: shoud this stay in traffic_cop?
```
ChTimTsubasa: Similar here and the rest TODOs

src/network/postgres_protocol_handler.cpp
```diff
@@ -316,21 +316,21 @@ ProcessResult PostgresProtocolHandler::ExecQueryMessage(InputPacket *pkt, const
...
-              {{NetworkMessageType::HUMAN_READABLE_ERROR, error_message_}});
+              {{NetworkMessageType::HUMAN_READABLE_ERROR, traffic_cop_->error_message_}});
           SendReadyForQuery(NetworkTransactionStateType::IDLE);
           return ProcessResult::COMPLETE;
         }
 
-        query_type_ = statement_->GetQueryType();
-        query_ = statement_->GetQueryString();
-        std::vector<int> result_format(statement_->GetTupleDescriptor().size(), 0);
+        traffic_cop_->query_type_ = traffic_cop_->statement_->GetQueryType();
```
ChTimTsubasa: If the query type and query can be accessed through the statement object, why can not we directly just use the statement object? That way we can use less field and have a better chance get rid of the inconsistency.

src/network/postgres_protocol_handler.cpp
```diff
@@ -343,13 +343,14 @@ ProcessResult PostgresProtocolHandler::ExecQueryMessage(InputPacket *pkt, const
...
-          statement_->GetPlanTree()->SetParameterValues(&param_values);
+          traffic_cop_->statement_->GetPlanTree()->SetParameterValues(&param_values);
         }
-        param_values_ = param_values;
+        traffic_cop_->param_values_ = param_values;
 
         auto status =
-                traffic_cop_->ExecuteStatement(statement_, param_values_, unnamed, nullptr, result_format_,
-                             results_, rows_affected_, error_message_, thread_id);
+                traffic_cop_->ExecuteStatement(traffic_cop_->statement_, traffic_cop_->param_values_, unnamed, nullptr,
```
ChTimTsubasa: Similar here, if the statement, parameter values can be accessed within traffic_cop, we should change the API of ExecuteStatement and get rid of the arguments passing to the function

src/network/postgres_protocol_handler.cpp
```diff
@@ -362,9 +363,9 @@ ProcessResult PostgresProtocolHandler::ExecQueryMessage(InputPacket *pkt, const
       {
         // prepareStatement
         std::string unnamed_statement = "unnamed";
-        statement_ = traffic_cop_->PrepareStatement(unnamed_statement, query_,
+        traffic_cop_->statement_ = traffic_cop_->PrepareStatement(unnamed_statement, traffic_cop_->query_,
```
ChTimTsubasa: Similar here

src/network/postgres_protocol_handler.cpp
```diff
@@ -373,15 +374,16 @@ ProcessResult PostgresProtocolHandler::ExecQueryMessage(InputPacket *pkt, const
...
         bool unnamed = false;
-        std::vector<int> result_format(statement_->GetTupleDescriptor().size(), 0);
+        std::vector<int> result_format(traffic_cop_->statement_->GetTupleDescriptor().size(), 0);
         result_format_ = result_format;
         // should param_values and result_format be local variable?
         // should results_ be reset when PakcetManager.reset(), why results_ cannot be read?
         auto status =
-            traffic_cop_->ExecuteStatement(statement_, param_values_, unnamed, nullptr, result_format_,
-                                           results_, rows_affected_, error_message_, thread_id);
+            traffic_cop_->ExecuteStatement(traffic_cop_->statement_, traffic_cop_->param_values_, unnamed, nullptr,
```
ChTimTsubasa: Same problem here. Use setter to set the member values first, and then call the function without passing as the argument but rather use that directly inside the function.

src/network/postgres_protocol_handler.cpp
```diff
@@ -878,29 +880,29 @@ ProcessResult PostgresProtocolHandler::ExecExecuteMessage(InputPacket *pkt,
...
-  auto statement_name = statement_->GetStatementName();
+  auto statement_name = traffic_cop_->statement_->GetStatementName();
   bool unnamed = statement_name.empty();
-  param_values_ = portal->GetParameters();
+  traffic_cop_->param_values_ = portal->GetParameters();
 
   auto status = traffic_cop_->ExecuteStatement(
-      statement_, param_values_, unnamed, param_stat, result_format_, results_,
-      rows_affected_, error_message_, thread_id);
+      traffic_cop_->statement_, traffic_cop_->param_values_, unnamed, param_stat, result_format_, traffic_cop_->results_,
```
ChTimTsubasa: Same here

src/network/postgres_protocol_handler.cpp
```diff
@@ -926,18 +928,18 @@ void PostgresProtocolHandler::ExecExecuteMessageGetResult(ResultType status) {
...
+      CompleteCommand(traffic_cop_->statement_->GetQueryTypeString(), query_type, traffic_cop_->rows_affected_);
       return;
     }
   }
 }
 
 void PostgresProtocolHandler::GetResult() {
   traffic_cop_->ExecuteStatementPlanGetResult();
-  auto status = traffic_cop_->ExecuteStatementGetResult(rows_affected_);
+  auto status = traffic_cop_->ExecuteStatementGetResult(traffic_cop_->rows_affected_);
```
ChTimTsubasa: Same here

src/network/postgres_protocol_handler.cpp
```diff
@@ -1177,8 +1179,8 @@ void PostgresProtocolHandler::Reset() {
 
   unnamed_statement_.reset();
   result_format_.clear();
-  results_.clear();
-  param_values_.clear();
+  traffic_cop_->results_.clear();
```
ChTimTsubasa: You might want to encapsulate all the clear into one clear function

apavlo:
@haoxianghua What is the status of this? There hasn't been any updates in two weeks.

src/concurrency/transaction.cpp
```diff
@@ -65,6 +65,11 @@ Transaction::~Transaction() {}
 void Transaction::Init(const size_t thread_id,
                        const IsolationLevelType isolation, const cid_t &read_id,
                        const cid_t &commit_id) {
+  // get current time when Transaction is created
+  struct timeval tv;
+  gettimeofday(&tv, NULL);
```
ChTimTsubasa: Since the timer is a new function, can you submit this to a new PR?

src/include/traffic_cop/traffic_cop.h
```diff
@@ -111,19 +111,87 @@ class TrafficCop {
...
+  std::string query_;
+
 //  struct event* event_;
  private:
 
+  bool is_queuing_;
+
+  std::string error_message_;
+
+  //this should stay in traffic_cop, because ExecutePlanArg will use it
```
ChTimTsubasa: Change the comments to something that explain the variable. Or just discard it.

src/network/postgres_protocol_handler.cpp
```diff
@@ -258,26 +258,24 @@ ProcessResult PostgresProtocolHandler::ExecQueryMessage(InputPacket *pkt, const
...
-    query_type_ = query_type;
+    traffic_cop_->query_ = query;
 
     switch (query_type) {
       case QueryType::QUERY_PREPARE:
       {
         std::string statement_name;
         std::vector<std::string> tokens;
-        boost::split(tokens, query_, boost::is_any_of("(), "));
+        boost::split(tokens, traffic_cop_->query_, boost::is_any_of("(), "));
```
ChTimTsubasa: We might want to have query local variable that deal with all of this and then set the query_ in traffic_cop by setter

src/include/network/postgres_protocol_handler.h
```diff
@@ -198,26 +197,6 @@ class PostgresProtocolHandler: public ProtocolHandler {
   std::unordered_map<std::string, stats::QueryMetric::QueryParamBuf>
       statement_param_types_;
 
-  //TODO: should this stay in traffic_cop?
-  std::shared_ptr<Statement> statement_;
-
-  std::string error_message_;
```
ChTimTsubasa: Error_message should probably stay in protocol handler, since it is protocol related

src/concurrency/transaction.cpp
haoxianghua: I didn't modify so much to add the timestamp. So I just submitted to this PR. Should I open a new PR?

src/include/network/postgres_protocol_handler.h
haoxianghua: I think the Error_message is created when exception threw in prepareStatement and executeStatement in traffic_cop. So I just save it like statement_ in traffic_cop

src/network/postgres_protocol_handler.cpp
haoxianghua: Actually, we can get it from statement_, but first, we should solve hard code part "std::size_t pos = boost::to_upper_copy(query_).find("AS");
        std::string statement_query = query_.substr(pos + 3);
        boost::trim(statement_query);" in postgres_protocal_handler

src/concurrency/transaction.cpp
seojungmin: @haoxianghua Yes, it would be wise to open a new PR for a new feature. What is the purpose of the timestamp?

src/include/network/postgres_protocol_handler.h
ChTimTsubasa: I see,
Then this is good.

src/network/postgres_protocol_handler.cpp
ChTimTsubasa: Then add a TODO comment on the query_ illustrating this. We could fix it afterward

src/concurrency/transaction.cpp
ChTimTsubasa: The timestamp is to make a snapshot of the current time when the txn begins so that when calling now() in build in function, it would return the time when this txn opened.

seojungmin:
This is not an issue of this PR, but when I read traffic_cop.cpp, these would be good TODOs when somebody refactors traffic_cop further.

1. Remove `GetColumnFieldForAggregates()` in traffic_cop. This is not called from outside.
  (Remove anything that is not used.  They can be added when they are required)
2. We have many helpers, such as `BeginQueryHelper()`, etc. in traffic_cop. These are called only once in the code, and I would consider moving them into the caller logic.  A lot of variables used in these helpers are member variables and it is difficult to follow the logic.
3. It is using PostgresValueType in `GetColumnFileForValueType()`.  This should be moved to postgres_protocol_handler or somewhere else.  The traffic cop should not do something special for postgres.  There is also some comment with 'this is a hack'.  Try to remove this hack.

ChTimTsubasa:
@seojungmin Thank you! These are really helpful advice.

For 1, I think it can be done in this PR since it looks trivial.
For 2, it looks to me that such helper can enhance readability. But we can discuss later.
For 3, this is important! It should be fixed but probably not in this PR. We would come up with a more systematic solution that makes tcop built on top of protocol handler.

I am thinking of opening an issue for such kind of advice that pointed out the improvement we neglected.

src/concurrency/transaction.cpp
haoxianghua: Got you!

haoxianghua:
@seojungmin @ChTimTsubasa I agree with you and I will talk with Tianyi and Yucheng. As Tianyi mentioned we would like to makes tcop built on top of protocol handler.

seojungmin:
For 2, it is arguable.  It is fine if the code is reasonably well-structured and easily readable.  

For now, I find it a little difficult on reading the code effortlessly: I have to scroll up and down the code to see what's going on. e.g.

- The names of the functions are not straightforward: What is the difference between ExecuteStatement() and ExecuteStatementPlan()?
- The comments on its header are out-dated: Some descriptions are wrong. Many variables are commented out.  I do not know why some of the functions are there.  
- The function signatures are mixed up: Hard to see when and where I call those functions as a user.

We see this kind of problem many times in the production code, where people just patch over patches without fixing the the main problem.  I think we should not do that, as we have more time. :-)

ChTimTsubasa:
@seojungmin  I agree.

For this PR, the most efficient way to do right now is to rename some of the functions and re-arrange the locations so that the interface functions would be gathered together and the helper functions gather together.

We have already cleaned up the traffic_cop in the summer so that it looks much cleaner than it was before. Previously we have 4 functions that have "ExecuteStatement" terms.

seojungmin:
@ChTimTsubasa Yes. I agree it is much better than better. I remember how hard it was to find out what is going on.  I think you guys are doing a great job, here!

haoxianghua:
@seojungmin @ChTimTsubasa I will solve it by this weekend!

src/network/postgres_protocol_handler.cpp
dengyuchenkit: #874 is dealing with hardcoded problem.

dengyuchenkit:
@seojungmin  Yes, we can rename the functions to reduce confusion. For the difference of ExecuteStatement and ExecuteStatementPlan:
ExecuteStatementPlan is used to execute a statement which needs a plan.
ExecuteStatement is generally to execute a statement. If the statement does not a plan, such as CREATE, COMMIT, ROLLBACK, it calls the helper function to begin/commmit/abort a transaction. Otherwise, it calls ExecuteStatementPlan.

ChTimTsubasa:
@apavlo 
Can we merge this PR so that the plan cache refactor could be done on top on this PR.

apavlo:

Issue #808, opened 2017-09-21
Parser returns success every time after throwing an exception


poojanilangekar:
Once parser returns an error, peloton returns success for all subsequent queries. Here are the debug statements from the peloton instance:

`2017-09-21 18:59:05 [src/parser/postgresparser.cpp:1433:ParseSQLString] ERROR - syntax error at or near "=" at 230

Exception Type :: Parser
Message :: Error parsing SQL statement
2017-09-21 18:59:08 [src/traffic_cop/traffic_cop.cpp:191:ExecuteStatement] DEBUG - rows_changed 0
2017-09-21 18:59:09 [src/traffic_cop/traffic_cop.cpp:191:ExecuteStatement] DEBUG - rows_changed 0
2017-09-21 18:59:10 [src/traffic_cop/traffic_cop.cpp:191:ExecuteStatement] DEBUG - rows_changed 0
2017-09-21 19:05:12 [src/traffic_cop/traffic_cop.cpp:191:ExecuteStatement] DEBUG - rows_changed 0
2017-09-21 19:05:13 [src/traffic_cop/traffic_cop.cpp:191:ExecuteStatement] DEBUG - rows_changed 0
2017-09-21 19:05:19 [src/traffic_cop/traffic_cop.cpp:191:ExecuteStatement] DEBUG - rows_changed 0`

haojin2:
Would you please provide sample queries that triggered this problem?

poojanilangekar:
Hi Hao, I am sorry I missed the notification. 

Here is a sample query which causes the bug:

 UPDATE stop_time AS st SET departure_time_hour = 12, departure_time_minute = 10, estimate_source = 'CMU' FROM stop s WHERE st.agency_id = 5 AND s.agency_id = st.agency_id AND st.stop_id = s.stop_id AND s.stop_code = 'PITT0' AND st.trip_id = 'FOO' ;

Here, the following error is returned:
Invalid table reference s

Add the log sends this error message:
2017-09-30 18:41:44 [src/traffic_cop/traffic_cop.cpp:191:ExecuteStatement] DEBUG - rows_changed 0
Message :: Invalid table reference s

After this point, any query that is passed to peloton does not return an error/invalid result. The log always says:
2017-09-30 18:42:30 [src/traffic_cop/traffic_cop.cpp:191:ExecuteStatement] DEBUG - rows_changed 0
[...]

PR #832, opened 2017-10-14
Handling syntax error in queries

haojin2:
This is the fix for issue #808 and it contains:
- Change the parser to throw ParserException when faced with a syntax error
- Formatted old code
- Added an extra test query with invalid syntax in parser_test.cpp

### DIFF
--- a/src/parser/postgresparser.cpp
+++ b/src/parser/postgresparser.cpp
@@ -1415,10 +1415,11 @@ parser::SQLStatementList* PostgresParser::ParseSQLString(const char* text) {
   auto result = pg_query_parse(text);
   if (result.error) {
     // Parse Error
+    std::string exception_msg = StringUtil::Format(
+        "%s at %d", result.error->message, result.error->cursorpos);
     pg_query_parse_finish(ctx);
     pg_query_free_parse_result(result);
-    throw ParserException(StringUtil::Format("%s at %d", result.error->message,
-                                             result.error->cursorpos));
+    throw ParserException(exception_msg);
   }
 
   // DEBUG only. Comment this out in release mode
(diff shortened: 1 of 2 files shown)


test/parser/parser_test.cpp
```diff
@@ -417,9 +417,13 @@ TEST_F(ParserTests, WrongQueryTest) {
   std::vector<std::string> queries;
   queries.push_back("SELECT;");
```
apavlo: This might actually be valid SQL. It works in Postgres, but not MySQL or SQLite.

src/parser/postgresparser.cpp
```diff
@@ -1408,20 +1415,19 @@ parser::SQLStatementList* PostgresParser::ParseSQLString(const char* text) {
...
     // Parse Error
-    LOG_ERROR("%s at %d\n", result.error->message, result.error->cursorpos);
-    auto new_stmt = new parser::SQLStatementList();
-    new_stmt->is_valid = false;
+    std::string exception_msg = StringUtil::Format(
+        "%s at %d", result.error->message, result.error->cursorpos);
     pg_query_parse_finish(ctx);
     pg_query_free_parse_result(result);
-    return new_stmt;
+    throw ParserException(exception_msg);
```
apavlo: Is this the only major change? The rest is just reformatting?

test/parser/parser_test.cpp
haojin2: Yeah, but if I remember correctly there was an issue about this query before, at that time we added this WrongQueryTest.

src/parser/postgresparser.cpp
haojin2: Yes, I played with that and found that the problem was that we returned a list whose valid flag was false without throwing an exception, so if we just change that we'll be good. I guess why this problem was never discovered was that we never threw queries with wrong syntax on the parser before.

PR #838, opened 2017-10-15
Infrastructure to support functions in codegen

pmenon:
Some initial infrastructure to support functions in codegen. This includes:

1. Added N-Ary builtin functions to type system.
2. Added `codegen::FunctionTranslator`
3. All builtin functions have an associated (internal) `OperatorId`. This is used in codegen to perform function lookup.
4. Added `catalog::LanguageCatalogObject` and `catalog::ProcCatalogObject` to simplify interfacing with `pg_language` and `pg_proc`.
5. Some formatting.

TODO:
1. Fix nary function lookup in type system.


apavlo:
@prashasthip @pmenon I will merge this when it passes. I will do this first before #815

PR #839, opened 2017-10-15
Database Name Binding

ChTimTsubasa:
Previously, the database name is either explicitly given in the query, otherwise, the binder would just bind it to the "default_database" (DEFAULT_DB_NAME).

This PR try to solve this by passing a <default_database_name> string when constructing the optimizer and binder so that the default_database_name would be chosen if the database name is not specified in the query-parsed tree.

The binder_test is modified to test this case.


src/include/optimizer/abstract_optimizer.h
```diff
@@ -45,8 +47,14 @@ class AbstractOptimizer {
...
   virtual std::shared_ptr<planner::AbstractPlan> BuildPelotonPlanTree(
       const std::unique_ptr<parser::SQLStatementList> &parse_tree, concurrency::Transaction *txn) = 0;
+  
+  inline void SetDefaultDatabaseName(std::string default_database_name) {
+      this->default_database_name_ = default_database_name;
+  }
 
   virtual void Reset(){};
+
+  std::string default_database_name_ = DEFAULT_DB_NAME;
```
apavlo: I don't think we should do this. This is just going to lead to everything landing in DEFAULT_DB_NAME. Let's remove this here and have `default_database_name` be passed in to `AbstractOptimizer`'s constructor.

@hzxa21 Thoughts?

src/include/traffic_cop/traffic_cop.h
```diff
@@ -114,10 +114,17 @@ class TrafficCop {
...
+  
+  inline void SetDefaultDatabaseName(std::string default_database_name) {
+    default_database_name_ = default_database_name;
+  }
 
 //  struct event* event_;
  private:
 
+  // Default database name
+  std::string default_database_name_ = DEFAULT_DB_NAME;
```
apavlo: Why do we need to duplicate this here and `AbstractOptimizer`? We should set it in one location so that it will be easier for us to replace this later.

src/include/optimizer/abstract_optimizer.h
hzxa21: The optimizer doesn't need `default_database_name` previously. I think the reason why Tianyi needs to add this into the optimizer is that he changes the `GetDatbaseName()`  function signature in all AST node to be `GetDatbaseName(string default_db_name)` and the optimizer needs to call these functions. However, I think it is not necessary to do that because in the binder, the default db name should already be filled in if the db filed is empty. Correct me if I miss something @ChTimTsubasa.

apavlo: I agree with @hzxa21. It sounds like we do not need to expose this aspect of the API to the optimizer. The optimizer does not need a default DB name.

ChTimTsubasa: @hzxa21 Where should I retrieve the bound database name after binding?

apavlo: Shouldn't it be in the AST node?

ChTimTsubasa: @apavlo I talked with @hzxa21 about this problem. The solution would be to try set the table_info of AST node by passing in the default database name. Previously I just return the database name back without modifying the AST node.

ChTimTsubasa:
@apavlo I have changed the binding API so that database name is bind only at Binder.

@hzxa21 I have removed all the 'const' at Visit(), can you see if that is correct?

src/binder/bind_node_visitor.cpp
```diff
@@ -18,20 +18,28 @@
...
-void BindNodeVisitor::Visit(const parser::SelectStatement *node) {
+void BindNodeVisitor::Visit(parser::SelectStatement *node) {
   // Save the upper level context
   auto pre_context = context_;
   context_ = std::make_shared<BinderContext>();
   context_->upper_context = pre_context;
-  if (node->from_table != nullptr) node->from_table->Accept(this);
+  if (node->from_table != nullptr){
+    node->from_table->Accept(this);
+    node->from_table->TryBindDatabaseName(default_database_name_);
```
hzxa21: One minor thing: is this necessary if we already call TryBindDatabaseName in context_->AddTable()?

ChTimTsubasa: I tested it and it seems that this line is not necessary.
One thing I do not understand is that the from_table is Accepting 'this', which is the current node instead of the from_table itself. How could this bind the from_table?

src/include/traffic_cop/traffic_cop.h
ChTimTsubasa: I have removed the one in AbstractOptimizer.

apavlo:
@ChTimTsubasa I will merge this when the build passes.

PR #840, opened 2017-10-16
Correctly handle NULL strings in codegen

pmenon:
At some point, NULL strings stopped working in codegen. This PR fixes it and more:

1. Fix handling NULL strings in codegen.
2. Added test to checking handling NULL values for all column types.
3. Code formatting to remove god-like header files.


apavlo:
@seojungmin @pmenon I will merge this when the build passes.

PR #853, opened 2017-10-22
fix compiling error of third_party/gmock on Mac

bluealert:
1, fix compiling error of `third_party/gmock` on Mac
2, replace `EXPECT_EQ(false, ...)` with `EXPECT_FALSE(...)`, refer: https://github.com/google/googletest/issues/322\#issuecomment-125645145

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -73,7 +73,7 @@ set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -O0 -ggdb -fno-omit-frame-po
 
 # ---[ Flags
 if(UNIX OR APPLE)
-    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fPIC -Wall -Wextra -Werror -mcx16 -Wno-invalid-offsetof -march=native")
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fPIC -Wall -Wextra -Werror -Wno-sign-compare -mcx16 -Wno-invalid-offsetof -march=native")
     if(${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
         set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-tautological-undefined-compare -Wno-unused-private-field -Wno-braced-scalar-init -Wno-constant-conversion -Wno-potentially-evaluated-expression -Wno-infinite-recursion -D__STDC_CONSTANT_MACROS -D__STDC_LIMIT_MACROS -I/usr/local/opt/openssl/include")
     endif()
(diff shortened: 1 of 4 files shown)


CMakeLists.txt
```diff
@@ -73,7 +73,7 @@ set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -O0 -ggdb -fno-omit-frame-po
 
 # ---[ Flags
 if(UNIX OR APPLE)
-    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fPIC -Wall -Wextra -Werror -mcx16 -Wno-invalid-offsetof -march=native")
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fPIC -Wall -Wextra -Werror -Wno-sign-compare -mcx16 -Wno-invalid-offsetof -march=native")
```
apavlo: We don't want to disable this for all linux/unix machines. Only on Mac. So I think you need to add this under the "Darwin" directive right be below it.

bluealert:
@apavlo Professor, I have tested on Ubuntu 16.04(gcc 5.4.0), `-Wno-sign-compare` is also needed on Ubuntu.

apavlo:
Ubuntu 16.04 gcc5 it was working before you added `-Wno-sign-compare` to disable the warning. You are now disabling the warning on **all** platforms, which is not what we want to do. It should **only** be for Darwin.

bluealert:
@apavlo Professor, I found the broken on Ubuntu 16.04 is because the deletion of `SYSTEM` at https://github.com/cmu-db/peloton/blob/master/test/CMakeLists.txt#L36

But, by the [cmake documentation](https://cmake.org/cmake/help/v3.0/command/include_directories.html), it should not use `SYSTEM` there?

poojanilangekar:
@apavlo Are we planning to merge this? 
AFAIK the build currently works for specific versions of Xcode. @rohit-cmu or @AngLi-Leon can you guys confirm?

rohit-cmu:
Works for the following configuration:
Apple LLVM version 8.0.0 (clang-800.0.42.1)
Xcode 8.1
OS X El Capitan Version 10.11.6

apavlo:
@poojanilangekar I don't think we want to add `-Wno-sign-compare` to our make file.

poojanilangekar:
@apavlo So can we close this PR and the corresponding issue? Since we do build on Mac.

apavlo:
@poojanilangekar Yes, if we're already working on the mac then we probably don't want this @bluealert's suggestion about fixing EXPECT_EQ(false...) to EXPECT_FALSE(..) is a good idea. We should do that as part of the clean-up/

poojanilangekar:
@apavlo Added it to the cleanup list.

PR #855, opened 2017-10-25
replace DropParse with TableParse (typos)

bluealert:


apavlo:
Can you describe what this does?

bluealert:
I think the DropParse in `TableParse(const DropParse &) = delete;` is a typo ? Because I can't find the definition of DropParse in the source code.

apavlo:
@haojin2 Can you confirm that this is the right thing to do? There are a bunch of other places that we reference `DropParse` that we may want to remove:

https://github.com/cmu-db/peloton/blob/master/src/include/parser/table_parse.h#L44
https://github.com/cmu-db/peloton/blob/master/src/include/parser/join_expr_parse.h#L30
https://github.com/cmu-db/peloton/blob/master/src/include/parser/join_expr_parse.h#L142
https://github.com/cmu-db/peloton/blob/master/src/include/parser/parse_node_visitor.h#L20
https://github.com/cmu-db/peloton/blob/master/src/include/parser/parse_node_visitor.h#L45

hzxa21:
@apavlo I just took a quick look here while going over all the PRs and I think these files are no longer used in the current parser. I suspect they are orphans of the old parser.

hzxa21:
@haojin2 Can you take a look at this PR to see whether these files are actually used in the current parser?

haojin2:
Sorry for the late reply on this, none of those is used anymore in the current parser, I guess we can even remove all those files.

apavlo:
Fixed in #931. Closing.

PR #862, opened 2017-10-30
Add Support for SSL Network Connections

haoxianghua:
This PR adds support for SSL connection based on previous PR#667.
1. This PR moves the logic of handling SSL request from NetworkConnection to ProtocolHandler. It's because different protocols have different ways to handle SSL request from first data package.
2. It corrects some logic in StateMachine() function to address SSL package. 
3. It sets the 'data' directory as the default directory for server key files and certificates. User can set the file path by setting 'private_key_file' and 'certificate_file'. The provided path is relative to the Data directory(peloton/data/).

Now, it functions in a basic way and the user can connect to Peloton using SSL. But we didn't handle some errors properly and didn't add multithread in SSL yet.


dengyuchenkit:
This PR does not run checks. Will submit a new PR later.

Issue #863, opened 2017-10-30
Successive AVG aggregations in GroupBy produce wrong output


tcm-marcel:
When using several successive AVG aggregations in a GroupBy query, some of the AVG columns produce wrong output. The wrong columns appear randomly (but deterministic between the runs). 

http://sqlfiddle.com/#!15/8fc516/1/0
```sql
CREATE TABLE test(a INT PRIMARY KEY, b INT, c INT, d VARCHAR);

INSERT INTO test VALUES (1, 1, 3, 'a');
INSERT INTO test VALUES (2, 1, 3, 'a');
INSERT INTO test VALUES (3, 2, 2, 'a');
INSERT INTO test VALUES (4, 2, 2, 'b');
INSERT INTO test VALUES (5, 3, 1, 'b');
INSERT INTO test VALUES (6, 3, 1, 'b');

SELECT AVG(c), AVG(c), AVG(c) FROM test GROUP BY d;
```
```
   avg   |   avg    |   avg   
---------+----------+---------
 1.33333 | 0.666667 | 1.33333
 2.66667 |  1.66667 | 2.66667
(2 rows)
```
All three aggregations are identical, but the second one produces a wrong output. In some cases the output is even NULL, so it might be a problem with NULL handling somewhere.

PR #847, opened 2017-10-19
LLVM support for `DISTINCT` in scans

tcm-marcel:
This PR adds support for `DISTINCT` in scans in the LLVM engine, implementing the first part of #825. 

* no batch processing/pipelining so far
* aggregations are not yet supported

PR #866, opened 2017-10-30
Fix for wrong output of successive AVG aggregations in GroupBy

tcm-marcel:
Fix for  #863. A wrong index was used to access the NullBitmap. 

The Bug only appeared when heaving an AVG aggregation, because AVGs get splitted up in three aggregations, a COUNT, a SUM and an internal AVG for the final result, that doesn't have a slot in the UpdateableStorage. Only in this case the storage index and the aggregation index differ and the bug appeared. 

Test cases have been added.


tcm-marcel:
I just realized, that the bug also should have appeared with global aggregations, so the title and the commit message is a bit misleading. However, the fix solves both problems.

test/sql/aggregate_groupby_sql_test.cpp
```diff
@@ -0,0 +1,164 @@
...
+
+    for (auto &result_row : ref_result) {
+      if (actual_result.count(result_row) == 0) {
+
+        EXPECT_EQ(ref_result, actual_result);
+        break;
+      }
+    }
+  }
+};
```
pmenon: This is a useful function that other tests can use. Can you move it to a utility class?

test/sql/aggregate_groupby_sql_test.cpp
```diff
@@ -0,0 +1,164 @@
...
+                                          "1|3|6|2",
+                                          "1|2|3|2",
+                                          "1|2|3|1",
+                                          "1|2|6|1"});
+
+
+  // free the database just created
+  txn = txn_manager.BeginTransaction();
+  catalog::Catalog::GetInstance()->DropDatabaseWithName(DEFAULT_DB_NAME, txn);
+  txn_manager.CommitTransaction(txn);
```
pmenon: The tests are good. Does it make sense to move them into the aggregate_sql_test?

test/sql/aggregate_groupby_sql_test.cpp
tcm-marcel: I just talked with @pmenon and we decided to move the function as part of this PR. After this PR is merged I will adapt the test cases in #847 and #867 to use this utility function.

Issue #825, opened 2017-10-11
Add support for LLVM DISTINCT


apavlo:
The LLVM engine does not support `DISTINCT`. The parser/planner already supports this, so you don't need to modify anything other than the LLVM engine. We need the following features for TPC-C:

1. `SELECT DISTINCT <col>`
2. `SELECT <AGGREGATE>(DISTINCT <col>)`

As always, try to follow the semantics of Postgres:

https://www.postgresql.org/docs/current/static/sql-select.html#sql-distinct

We don't need to support `DISTINCT ON (...)` right now.

@pmenon @seojungmin Can provide more guidance on where to add this new code. You should reuse the existing hash table in the LLVM engine.

tcm-marcel:
The task is basically finished. 
However, there is a strange bug that appears when executing a query with several distinct AVG aggregations, that some of them return empty results. I wasn't able to find the cause yet, because it's happens somewhere in the dynamic execution and is very hard to debug. I will continue to find the problem.

PR #867, opened 2017-10-30
LLVM support for `DISTINCT` in aggregations

tcm-marcel:
This PR adds support for `DISTINCT` inside aggregations in the LLVM engine, implementing the second part of #825.

* mostly affects `aggregations.cpp`, which is used from GlobalGroupByTranslator and HashGroupByTranslator
* small extensions to the HashTable interface
* this PR also includes changes in the codegen code for generating more meaningful labels in the IR for better debugging

This PR depends on the fix for wrong output of successive AVG aggregations in GroupBy #866 and on the PR regarding DISTINCT in scans #847.

### DIFF
--- a/src/codegen/aggregation.cpp
+++ b/src/codegen/aggregation.cpp
@@ -61,8 +61,7 @@ void Aggregation::Setup(
           // Create the hash table
           // we don't need to save any values, so value_size is zero
           auto hash_table =
-              OAHashTable{codegen, key_type, 0,
-                          "Agg" + std::to_string(source_idx) + ".HashTable"};
+              OAHashTable{codegen, key_type, 0};
 
           hash_table_infos_.emplace_back(
               std::pair<OAHashTable, RuntimeState::StateID>(
@@ -98,8 +97,7 @@ void Aggregation::Setup(
           // Create the hash table
           // we don't need to save any values, so value_size is zero
           auto hash_table =
-              OAHashTable{codegen, key_type, 0,
-                          "Agg" + std::to_string(source_idx) + ".HashTable"};
+              OAHashTable{codegen, key_type, 0};
 
           hash_table_infos_.emplace_back(
               std::pair<OAHashTable, RuntimeState::StateID>(
@@ -177,11 +175,9 @@ void Aggregation::Setup(
           // Create the hash table
           // we don't need to save any values, so value_size is zero
           auto sum_hash_table =
-              OAHashTable{codegen, key_type, 0,
-                          "Agg" + std::to_string(source_idx) + ".HashTable"};
+              OAHashTable{codegen, key_type, 0};
           auto count_hash_table =
-              OAHashTable{codegen, key_type, 0,
-                          "Agg" + std::to_string(source_idx) + ".HashTable"};
+              OAHashTable{codegen, key_type, 0};
 
           hash_table_infos_.emplace_back(
               std::pair<OAHashTable, RuntimeState::StateID>(
--- a/src/codegen/aggregation.cpp
+++ b/src/codegen/aggregation.cpp
@@ -390,8 +390,7 @@ void Aggregation::DoAdvanceValue(CodeGen &codegen, llvm::Value *space,
 void Aggregation::DoNullCheck(CodeGen &codegen, llvm::Value *space,
                               ExpressionType type, uint32_t storage_index,
                               const codegen::Value &update,
-                              UpdateableStorage::NullBitmap &null_bitmap,
-                              llvm::Value *curr_val) const {
+                              UpdateableStorage::NullBitmap &null_bitmap) const {
   // This aggregate is NULL-able, we need to check if the update value is
   // NULL, and whether the current value of the aggregate is NULL.
 
@@ -408,9 +407,9 @@ void Aggregation::DoNullCheck(CodeGen &codegen, llvm::Value *space,
   llvm::Value *update_not_null = update.IsNotNull(codegen);
   llvm::Value *agg_null = null_bitmap.IsNull(codegen, storage_index);
 
-  if (curr_val == nullptr) {
-    curr_val = null_bitmap.ByteFor(codegen, storage_index);
-  }
+  // fetch null byte value here already, as we need it to create the PHIs
+  // after the branches (MergeValues)
+  llvm::Value *curr_val = null_bitmap.ByteFor(codegen, storage_index);
 
   lang::If valid_update{codegen, update_not_null, "Agg.IfValidUpdate"};
   {
@@ -448,13 +447,12 @@ void Aggregation::DoNullCheck(CodeGen &codegen, llvm::Value *space,
   null_bitmap.MergeValues(valid_update, curr_val);
 }
 
-// Advancethe value of a specifig aggregate. Performs NULL check if necessary
+// Advance the value of a specific aggregate. Performs NULL check if necessary
 // and finally calls DoAdvanceValue()
 void Aggregation::AdvanceValue(CodeGen &codegen, llvm::Value *space,
                                const std::vector<codegen::Value> &next_vals,
                                const Aggregation::AggregateInfo &aggregate_info,
-                               UpdateableStorage::NullBitmap &null_bitmap,
-                               llvm::Value *curr_val) const {
+                               UpdateableStorage::NullBitmap &null_bitmap) const {
   const Value &update = next_vals[aggregate_info.source_index];
 
   switch (aggregate_info.aggregate_type) {
@@ -469,8 +467,7 @@ void Aggregation::AdvanceValue(CodeGen &codegen, llvm::Value *space,
         // else do NULL check, which will then call DoAdvanceValue if
         // appropriate
         DoNullCheck(codegen, space, aggregate_info.aggregate_type,
-                    aggregate_info.storage_indices[0], update, null_bitmap,
-                    curr_val);
+                    aggregate_info.storage_indices[0], update, null_bitmap);
       }
       break;
     }
@@ -490,8 +487,7 @@ void Aggregation::AdvanceValue(CodeGen &codegen, llvm::Value *space,
         // else do NULL check, which will then call DoAdvanceValue if
         // appropriate
         DoNullCheck(codegen, space, ExpressionType::AGGREGATE_SUM,
-                    aggregate_info.storage_indices[0], update, null_bitmap,
-                    curr_val);
+                    aggregate_info.storage_indices[0], update, null_bitmap);
       }
 
       // the COUNT can't be nullable, so skip the null check
@@ -555,10 +551,10 @@ void Aggregation::AdvanceValues(
 
       // Create llvm values for the bitmap bytes here already, so we can create
       // the Phis to merge the two paths that are created by the hast table
-      llvm::Value *curr_val;
-
-      // Every aggregation has at least one component
-      curr_val =
+      // so far only one snapshow needed, as the only aggregate with two
+      // components (AVG) has only one nullable component
+      llvm::Value *null_byte_snapshot;
+      null_byte_snapshot =
           null_bitmap.ByteFor(codegen, aggregate_info.storage_indices[0]);
 
       // Only process aggregation if value is distinct (key didn't exist in hash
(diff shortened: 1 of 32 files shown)


tcm-marcel:
It shows lot's of commits in this code, because it builds up on the PR regarding DISTINCT in scans #847. Maybe we should first merge in #847 and then this one should get better to understand.

pmenon:
I'll review this once #866 is merged in.

tcm-marcel:
@pmenon PR is ready for review now. 

There are 4 primary commits:
* 30d45feff3f0ced484034cfd538f9a7e04dc76ce Add names for loops and if clauses in generated hash tables
  * this commit is not actually part of the aggregation feature, but extends the naming for llvm control structures for better IR debugging (which i needed)
* 313f5e32541ffb36b0ff611108e166de71943451 Extend OAHashTable to return probing result as LLVM value
* 083783e380035ded71c864c777a0c5b8275501f3 Add DISTINCT for Aggregations
* 0ddfc5edea9e8dc1b1c8f555ab30a2bb9ee5ffe5 Add test cases for DISTICT in Aggregations 
  * test cases have been adapted to utility function for unordered rows in a later commit

tcm-marcel:
Will make changes requested by @pmenon.

src/codegen/aggregation.cpp
```diff
@@ -275,83 +534,106 @@ void Aggregation::AdvanceValues(
...
 // This function will computes the final values of all aggregates stored in the
 // provided storage space, and populates the provided vector with these values.
 void Aggregation::FinalizeValues(
     CodeGen &codegen, llvm::Value *space,
     std::vector<codegen::Value> &final_vals) const {
   // Collect all final values into this map. We need this because some
   // aggregates are derived from other component aggregates.
-  std::map<std::pair<uint32_t, ExpressionType>, codegen::Value> vals;
+  // bool shows if distinct
+  std::map<std::tuple<uint32_t, ExpressionType, bool>, codegen::Value> vals;
```
pmenon: Why does the map need to know if it's a distinct aggregate?

src/codegen/lang/loop.cpp
```diff
@@ -82,8 +82,7 @@ void Loop::CollectFinalLoopVariables(std::vector<llvm::Value *> &loop_vals) {
   PL_ASSERT(last_loop_bb_ != nullptr);
   for (auto *phi_node : phi_nodes_) {
     llvm::PHINode *end_phi =
-        cg_->CreatePHI(phi_node->getType(), 2 + break_bbs_.size(),
-                       phi_node->getName() + "End");
+        cg_->CreatePHI(phi_node->getType(), 2, phi_node->getName() + "End");
```
pmenon: I don't think this is correct.

src/include/codegen/hash_table.h
```diff
@@ -119,12 +130,23 @@ class HashTable {
...
   // Insert a new entry into the hash table with the given keys, but don't
   // perform any key matching or merging
   virtual void Insert(CodeGen &codegen, llvm::Value *ht_ptr, llvm::Value *hash,
                       const std::vector<codegen::Value> &keys,
                       InsertCallback &callback) const = 0;
 
+  // Insert a new entry into the hash table with the given keys, not caring
+  // about existing entries.
+  virtual void Insert(CodeGen &codegen, llvm::Value *ht_ptr, llvm::Value *hash,
+                      const std::vector<codegen::Value> &key) const = 0;
```
pmenon: What is the difference between the two `Insert` methods? Can you implement the one you created by using the existing `Insert` and providing a no-op callback? I ask because I don't want to complicate the interface too much.

src/include/codegen/hash_table.h
```diff
@@ -119,12 +130,23 @@ class HashTable {
                              ProbeCallback &probe_callback,
                              InsertCallback &insert_callback) const = 0;
 
+  // Probe the hash table and insert a new slot if needed, returning both the
+  // result and the data pointer
+  virtual ProbeResult ProbeOrInsert(
+      CodeGen &codegen, llvm::Value *ht_ptr, llvm::Value *hash,
+      const std::vector<codegen::Value> &key) const = 0;
```
pmenon: I briefly remember our discussion on this, but why can't you use the existing `ProbeOrInsert` method?

src/codegen/aggregation.cpp
```diff
@@ -39,15 +43,90 @@ void Aggregation::Setup(
...
+          }
+          key_type.push_back(agg_term.expression->ResultType());
+
+          // Create the hash table
+          // we don't need to save any values, so value_size is zero
+          auto hash_table = OAHashTable{codegen, key_type, 0};
+
+          hash_table_infos_.emplace_back(
+              std::pair<OAHashTable, RuntimeState::StateID>(
+                  std::move(hash_table), hash_table_id));
```
pmenon: This logic is copied in different case branches, can you extract it somehow? Also the cases are getting quite complex, consider doing multiple passes over the aggregates during setup.

src/include/codegen/aggregation.h
```diff
@@ -37,20 +39,39 @@ namespace codegen {
 class Aggregation {
  public:
   // Setup the aggregation to handle the provided aggregates
-  void Setup(CodeGen &codegen,
+  void Setup(CompilationContext &context,
+             const std::vector<planner::AggregatePlan::AggTerm> &agg_terms,
+             bool is_global, std::vector<type::Type> &grouping_ai_types);
+
+  // Setup the aggregation to handle the provided aggregates
+  void Setup(CompilationContext &context,
```
pmenon: During `Setup`, pull out `RuntimeState` and assign it as a member variable. That way, you don't need to pass around the `CompilationContext` to each method.

src/include/codegen/hash_table.h
tcm-marcel: The hash probing introduces a new branch, after which the null bitmap values have to be merged (Phi generation). When building the lang::If for the probing manually, I can pass it to the existing merge function. (https://github.com/tcm-marcel/peloton/blob/c3f7329abdbfc39ae2f941e4029431144cd5ed53/src/codegen/aggregation.cpp#L596 )

Also currently both the SUM and the COUNT part of an AVG aggregation make a hash table lookup, but I wanted to change this in the future. Therefore the lookup result has to be passed on to the next aggregation. (I can't exactly remember why I left this out, I think I will add it just now)

src/include/codegen/hash_table.h
tcm-marcel: In the subclass both call the same method. It exists, because the `InsertCallback &insert_callback` reference can't be null. 
The three possibilities I see are: 
1. Changing signature to take pointer instead of reference (can be null)
2. Overload function (what I did)
3. Create dummy InsertCallback instance. Then however the functions are actually called during the lookup, even if they are empty (if not removed by compiler).

src/include/codegen/aggregation.h
tcm-marcel: Many methods start with 
```c++
auto &codegen = context.GetCodeGen();
auto &runtime_state = context.GetRuntimeState();
```

Should I assign both references as member variable or assign codegen and call the two getters in each method?

src/codegen/aggregation.cpp
tcm-marcel: A distinct SUM and a non-distinct one have different results, and for a distinct AVG I have to pick the correct one.

src/codegen/lang/loop.cpp
tcm-marcel: I have to admit I can't remember why I changed this. I will change it back and if it had a reason I might remember when an error occurs.

src/include/codegen/aggregation.h
pmenon: Just assign `runtime_state`, and pass around the `codegen` instance through methods. We pass around `codegen` to indicate that a given method is generating code. We pass around the `CompilationContext` during class instantiation for classes to pull out the pieces they need.

src/include/codegen/hash_table.h
pmenon: If no extra work is done during insertions when a value is explicitly provided and when it is not, I'd prefer to have just one `Insert` method. Users not caring to insert values provide an `InsertCallback` that is a no-op. You can even create a `NoOnInsertCallback` implementation class in `codegen::HashTable` that is a no-op, which users of the hash table can use.

src/codegen/aggregation.cpp
```diff
@@ -234,180 +361,335 @@ void Aggregation::DoAdvanceValue(CodeGen &codegen, llvm::Value *space,
...
+  // Merge the null value
+  null_bitmap.MergeValues(valid_update, curr_val);
 }
 
 // Advance each of the aggregates stored in the provided storage space
 void Aggregation::AdvanceValues(
     CodeGen &codegen, llvm::Value *space,
-    const std::vector<codegen::Value> &next_vals) const {
+    const std::vector<codegen::Value> &next_vals,
+    const std::vector<codegen::Value> &grouping_keys) const {
```
pmenon: This method is too complex in its current form. Refactor by extracting the core aggregation logic into a separate function:
```
void Aggregation::AdvanceValue(CodeGen &codegen, llvm::Value *space, 
                               const std::vector<codegen::Value> &next_vals, 
                               const Aggregation::AggregateInfo &agg);
```

Then, `AdvanceValues` has logic to check distinct values, if necessary, and delegating to `AdvanceValue` to deal with individual aggregates. `AdvanceValue` then performs NULL-checking logic, delegating to `DoAdvanceValue` to physically perform the update.

src/include/codegen/aggregation.h
```diff
@@ -36,17 +38,39 @@ namespace codegen {
 //===----------------------------------------------------------------------===//
 class Aggregation {
  public:
+  // Constructor taking the runtime state reference
+  Aggregation(RuntimeState &runtime_state) : runtime_state_(runtime_state) {}
+
+  // Setup the aggregation to handle the provided aggregates
+  void Setup(CompilationContext &context,
```
pmenon: I don't think you need the `CompilationContext` here. Just pass the `CodeGen` instance.

src/include/codegen/aggregation.h
```diff
@@ -36,17 +38,39 @@ namespace codegen {
...
+             const std::vector<planner::AggregatePlan::AggTerm> &agg_terms,
+             bool is_global, std::vector<type::Type> &grouping_ai_types);
+
   // Setup the aggregation to handle the provided aggregates
-  void Setup(CodeGen &codegen,
+  void Setup(CompilationContext &context,
```
pmenon: Ditto.

src/include/codegen/aggregation.h
```diff
@@ -71,46 +95,59 @@ class Aggregation {
...
+  // stored at which index.
   //
   // Storing the mapping from the physical position the aggregate is stored to
   // where the caller expects them allows us to rearrange positions without
   // the caller knowing or caring.
   //===--------------------------------------------------------------------===//
-  struct AggregateInfo {
-    // The type of aggregate
-    ExpressionType aggregate_type;
+  static const unsigned int max_comp_size_ = 2;
```
pmenon: For static constants, we use the camel-casing formats. Rename this to `kMaxNumComponents`

src/include/codegen/operator/hash_group_by_translator.h
```diff
@@ -106,19 +111,24 @@ class HashGroupByTranslator : public OperatorTranslator {
   class ConsumerInsert : public HashTable::InsertCallback {
    public:
     // Constructor
-    ConsumerInsert(const Aggregation &aggregation,
-                   const std::vector<codegen::Value> &initial_vals);
+    ConsumerInsert(CompilationContext &context, const Aggregation &aggregation,
```
pmenon: I think you can remove the `CompilationContext` constructor argument.

src/codegen/aggregation.cpp
tcm-marcel: I had to extend the signature of the function to this:
```
void AdvanceValue(CodeGen &codegen, llvm::Value *space,
                                 const std::vector<codegen::Value> &next_vals,
                                 const Aggregation::AggregateInfo &agg,
                                 UpdateableStorage::NullBitmap &null_bitmap,
                                 llvm::Value *curr_val = nullptr) const;
```

When calling this function from inside the `lang::If` of the distinct check, we have to hand it the `NullBitmap`, as well as the `llvm::Value*` to be able to PHI merge this value after the `lang::If`. (same as in `DoNullCheck`)

pmenon: Just spoke with @tcm-marcel, I don't think passing along the `curr_val` is necessary.

tcm-marcel: This is true, I changed it in https://github.com/cmu-db/peloton/pull/867/commits/2fbaeacff62889daa5f6c65e1916513a9ff72b63 and named it `null_byte_snapshot` to clarify its purpose.

pmenon:
LGTM +1. Can be merged when build passes.

apavlo:
@pmenon Thanks! I will rebase and merge.

apavlo:
@tcm-marcel Can you rebase? I tried doing it quickly but I don't have time to figure out how to resolve the conflicts.

tcm-marcel:
@apavlo Rebase complete - sorry, it took me a while.

PR #868, opened 2017-10-31
Optimizer support for predicate push-down and query-derived table

hzxa21:
This PR applies bug fixes for the optimizer and extends it to support predicate push-down and query-derived table.
- Single table predicates and join predicates are bound to the operators directly rather than using predicate property.
- Binder treats regular table and query derived table differently to perform correct binding.
- New logical and physical operators are added to support scan operation on query-derived tables.
- Fix a bug when enabling join commutativity rule.

Now we can support queries like:
```sql
SELECT A.c, B.c 
FROM (SELECT a+b as a, c FROM test) as A, (SELECT a+b as a, c as c from test2) as B 
WHERE A.a=B.a ORDER BY A.a
```

WIP:
- Performance test for predicate push-down and query-derived table
- Support nested queries in WHERE clause (need support for pre-connection temp table)


apavlo:
> Now we can support queries like:
> ```sql
> SELECT A.c, B.c 
> FROM (SELECT a+b as a, c FROM test) as A, (SELECT a+b as a, c as c from test2) as B 
> WHERE A.a=B.a ORDER BY A.a
> ```

This is amazing! I love it!

apavlo:
@hzxa21 @chenboy Is this PR ready to merge?

hzxa21:
@apavlo Bowei and I are doing the framework refactoring right now. If the features supported in this PR are not critical to other teams, we think it can be merged later.

apavlo:
Replaced with #958

PR #871, opened 2017-10-31
Fix Jenkins integration

crd477:
After you merge this, I will reconfigure the project on the Jenkins server to work correctly.  It will look particularly nice in the Blue Ocean interface.  I got pull requests building to work too.
I implemented the alternative platform builds entirely in Dockerfiles so that testing can be done with them outside of Jenkins and also because the Docker/Jenkins integration sucks for anything more complicated than what is here.
Sorry for the extraordinarily long delay in submitting this - there was a lot of trial and error and several false starts.


crd477:
There's something different about the Travis environment and I'm looking into it now.

PR #874, opened 2017-10-31
Remove the uses of split() and find() in the frontend

dengyuchenkit:
This PR removes string methods in ProtocolHandler:
- ProtocolHandler calls parser to check syntax error and get query type. Add tests to check that queries with syntax error or empty query are handled correctly. 
- Store a SQLStatementList inside Statement for replanning purpose.
- Fix the response when clients attempt to operate in a transaction which has already been aborted (such as due to SQL syntax error) but is not explicitly committed/rollbacked yet.

```
postgres=# begin;
BEGIN
postgres=# create tabel foo(id int);
syntax error at or near "tabel" at 8
postgres=# select * from foo;
current transaction is aborted, commands ignored until end of transaction block
postgres=# rollback;
ROLLBACK
```
- Fix PREPARE and EXECUTE query
```
postgres=# create table foo(id decimal);
CREATE TABLE
postgres=# prepare fooplan as insert into foo values($1);
PREPARE
postgres=# execute fooplan(1);
INSERT 0 1
postgres=# execute fooplan(1+1);
INSERT 0 1
postgres=# execute fooplan(SQRT(9.0));
INSERT 0 1
postgres=# select * from foo;
 id 
----
  1
  2
  3
(3 rows)
postgres=# 
```

### DIFF
--- a/src/include/common/statement.h
+++ b/src/include/common/statement.h
@@ -19,8 +19,8 @@
 
 #include "common/printable.h"
 #include "internal_types.h"
-#include "type/types.h"
 #include "parser/sql_statement.h"
+#include "type/types.h"
 
 namespace peloton {
 namespace planner {
--- a/src/include/common/statement.h
+++ b/src/include/common/statement.h
@@ -20,7 +20,6 @@
 #include "common/printable.h"
 #include "internal_types.h"
 #include "parser/sql_statement.h"
-#include "type/types.h"
 
 namespace peloton {
 namespace planner {
--- a/src/include/common/statement.h
+++ b/src/include/common/statement.h
@@ -19,6 +19,8 @@
 
 #include "common/printable.h"
 #include "internal_types.h"
+#include "type/types.h"
+#include "parser/sql_statement.h"
 
 namespace peloton {
 namespace planner {
@@ -43,17 +45,10 @@ class Statement : public Printable {
   Statement& operator=(Statement&&) = delete;
 
   Statement(const std::string& statement_name, const std::string& query_string);
+  Statement(const std::string& statement_name, QueryType query_type,
+            std::string query_string, std::unique_ptr<parser::SQLStatementList> sql_stmt_list);
 
   ~Statement();
-
-  static void ParseQueryTypeString(const std::string& query_string,
-                             std::string& query_type_string);
-
-  static void ParseCreateTypeString(const std::string& query_string,
-                             std::string& create_type_string);
- 
-  static void MapToQueryType(const std::string& query_type_string,
-                             QueryType& query_type);
 
   std::vector<FieldInfo> GetTupleDescriptor() const;
 
@@ -83,6 +78,10 @@ class Statement : public Printable {
 
   const std::shared_ptr<planner::AbstractPlan>& GetPlanTree() const;
 
+  std::unique_ptr<parser::SQLStatementList>const& GetStmtParseTreeList() {return sql_stmt_list_;}
+
+  std::unique_ptr<parser::SQLStatementList> PassStmtParseTreeList() {return std::move(sql_stmt_list_);}
+
   inline bool GetNeedsPlan() const { return (needs_replan_); }
 
   inline void SetNeedsPlan(bool replan) { needs_replan_ = replan; }
@@ -94,16 +93,19 @@ class Statement : public Printable {
   // logical name of statement
   std::string statement_name_;
 
+  // enum value of query_type
+  QueryType query_type_;
+
   // query string
   std::string query_string_;
+
+  // query parse tree
+  std::unique_ptr<parser::SQLStatementList> sql_stmt_list_;
 
   // first token in query
   // Keep the string token of the query_type because it is returned 
   // as responses after executing commands.
   std::string query_type_string_;
-
-  // enum value of query_type
-  QueryType query_type_;
 
   // format codes of the parameters
   std::vector<int32_t> param_types_;
@@ -120,10 +122,5 @@ class Statement : public Printable {
 
   // If this flag is true, then somebody wants us to replan this query
   bool needs_replan_ = false;
-
-  // containing pairs of <query_type_string, query_type>
-  // use map to speed up searching
-  static std::unordered_map<std::string, QueryType> query_type_map_;
-
 };
 }  // namespace peloton
--- a/src/include/traffic_cop/traffic_cop.h
+++ b/src/include/traffic_cop/traffic_cop.h
@@ -75,6 +75,7 @@ class TrafficCop {
   // Prepare and bind a query from a query string
   std::shared_ptr<Statement> PrepareStatement(const std::string &statement_name,
                                               const std::string &query_string,
+                                              std::unique_ptr<parser::SQLStatementList> sql_stmt_list,
                                               std::string &error_message,
                                               size_t thread_id = 0);
 
@@ -100,6 +101,10 @@ class TrafficCop {
   }
 
   void setRowsAffected(int rows_affected) { rows_affected_ = rows_affected; }
+
+  void AbortInvalidStmt();
+
+  executor::ExecuteResult p_status_;
 
   int getRowsAffected() { return rows_affected_; }
 
@@ -169,6 +174,7 @@ class TrafficCop {
   // The current callback to be invoked after execution completes.
   void (*task_callback_)(void *);
   void *task_callback_arg_;
+  std::vector<StatementResult> result_;
 
   // pair of txn ptr and the result so-far for that txn
   // use a stack to support nested-txns
--- a/src/include/traffic_cop/traffic_cop.h
+++ b/src/include/traffic_cop/traffic_cop.h
@@ -79,6 +79,9 @@ class TrafficCop {
                                               std::string &error_message,
                                               size_t thread_id = 0);
 
+  bool BindParamsForCachePlan(const std::vector<std::unique_ptr<expression::AbstractExpression>>&,
+                              std::string &error_message, const size_t thread_id = 0);
+
   std::vector<FieldInfo> GenerateTupleDescriptor(
       parser::SQLStatement *select_stmt);
 
--- a/src/include/traffic_cop/traffic_cop.h
+++ b/src/include/traffic_cop/traffic_cop.h
@@ -104,8 +104,6 @@ class TrafficCop {
 
   void ProcessInvalidStatement();
 
-  void ProcessInvalidStatement();
-
   int getRowsAffected() { return rows_affected_; }
 
   void SetStatement(std::shared_ptr<Statement> statement) {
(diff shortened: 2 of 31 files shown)


dengyuchenkit:
@hzxa21 I change the map to unordered_map and the shared_ptr back to unique_ptr. Could you please check if it's fine?

src/type/types.cpp
```diff
@@ -549,6 +554,115 @@ std::ostream& operator<<(std::ostream& os, const StatementType& type) {
...
+    case QueryType::QUERY_ANALYZE: return "ANALYZE";
+    case QueryType::QUERY_RENAME: return "RENAME";
+    case QueryType::QUERY_PREPARE: return "PREPARE";
+    case QueryType::QUERY_EXECUTE: return "EXECUTE";
+    case QueryType::QUERY_SELECT: return "SELECT";
+    case QueryType::QUERY_OTHER: default: return "OTHER";
+  }
+}
+
+QueryType StringToQueryType(std::string str) {
```
chenboy: Since we only use the value of the string, it would be better if we use a `const std::string *` as the parameter to avoid copying. Also, seems like we will construct the hash map each time we call this function.

src/type/types.cpp
```diff
@@ -549,6 +554,115 @@ std::ostream& operator<<(std::ostream& os, const StatementType& type) {
...
+  };
+  std::unordered_map<std::string, QueryType>::iterator it  = querytype_string_map.find(str);
+  if (it != querytype_string_map.end()) {
+    return it -> second;
+  } else {
+    return QueryType::QUERY_INVALID;
+  }
+}
+
+QueryType StatementTypeToQueryType(StatementType stmt_type, parser::SQLStatement* sql_stmt) {
```
chenboy: Seems like we will not modify sql_stmt in this function, please make it a `const parser::SQLStatement*`, also the hash map seems to be created each time we call this function.

src/include/common/statement.h
```diff
@@ -81,6 +75,10 @@ class Statement : public Printable {
 
   const std::shared_ptr<planner::AbstractPlan>& GetPlanTree() const;
 
+  std::unique_ptr<parser::SQLStatementList>const& GetStmtParseTreeList() {return sql_stmt_list_;}
```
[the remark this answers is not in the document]
dengyuchenkit: This method will be called when the statement lends its parse tree to the optimizer to generate a query plan. I think it may be better to keep this method now since the optimizer takes in a unique_ptr.

src/network/postgres_protocol_handler.cpp
```diff
@@ -237,179 +235,132 @@ bool PostgresProtocolHandler::HardcodedExecuteFilter(QueryType query_type) {
...
+    traffic_cop_->AbortInvalidStmt();
+    std::string error_message = e.what();
+    SendErrorResponse(
+        {{NetworkMessageType::HUMAN_READABLE_ERROR, e.what()}});
+    SendReadyForQuery(NetworkTransactionStateType::IDLE);
+    return ProcessResult::COMPLETE;
+  }
+  // If the query contains zero sql statements, for example ";;;"
+  if (sql_stmt_list->GetNumStatements() == 0) {
+    SendEmptyQueryResponse();
```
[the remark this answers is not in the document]
dengyuchenkit: Yes, thanks!

src/network/postgres_protocol_handler.cpp
```diff
@@ -237,179 +235,132 @@ bool PostgresProtocolHandler::HardcodedExecuteFilter(QueryType query_type) {
...
-                                           results_, rows_affected_, error_message_, thread_id);
-        if (traffic_cop_->is_queuing_) {
-          return ProcessResult::PROCESSING;
-        }
-        ExecQueryMessageGetResult(status);
+      query_ = statement_->GetQueryString();
+      query_type_ = statement_->GetQueryType();
+      std::vector<int> result_format(statement_->GetTupleDescriptor().size(), 0);
+      result_format_ = result_format;
+      //TODO: We assume it's only constant value expression
```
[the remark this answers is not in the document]
dengyuchenkit: Okay, I will do that.

src/type/types.cpp
dengyuchenkit: Fixed

src/type/types.cpp
dengyuchenkit: Fixed

src/network/postgres_protocol_handler.cpp
```diff
@@ -237,179 +235,132 @@ bool PostgresProtocolHandler::HardcodedExecuteFilter(QueryType query_type) {
...
+      for (std::unique_ptr<expression::AbstractExpression>& param : exec_stmt->parameters) {
```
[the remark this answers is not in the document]
dengyuchenkit: Hi just to clarify first, the exec_stmt->parameters is a vector of unique pointers. I suppose it's more convenient to iterate by the reference of the unique pointers. Could you explain more if you still think it's better to use the raw pointer here? Thanks

src/network/postgres_protocol_handler.cpp
```diff
@@ -237,179 +235,132 @@ bool PostgresProtocolHandler::HardcodedExecuteFilter(QueryType query_type) {
...
+      // There used to be code here that would invoke this method passing
+      // in NetworkMessageType::READY_FOR_QUERY as the argument. But when
+      // I switched to strong types, this obviously doesn't work. So I
+      // switched it to be NetworkTransactionStateType::IDLE. I don't know
+      // we just don't always send back the internal txn state?
+      SendReadyForQuery(NetworkTransactionStateType::IDLE);
+      return ProcessResult::COMPLETE;
+    };
+    case QueryType::QUERY_EXECUTE: {
+      parser::ExecuteStatement* exec_stmt = (parser::ExecuteStatement*)sql_stmt.get();
```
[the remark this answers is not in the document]
dengyuchenkit: I change it to dynamic_cast. Right now this kind of casting (dynamic_cast or static_cast) exist in a large number of codes. I think maybe right now it's fine to keep it in the same way, for consistency.

test/network/exception_test.cpp
```diff
@@ -0,0 +1,204 @@
...
+      if (s) {
+        LOG_INFO("Invalid Drop Query: %s", e.base().what());
+        exception_count += 1;
+      }
+    }
+
+    // CREATE query
+    pqxx::work txn2(C);
+    try {
+      // TODO(Yuchen): If the query is 'CREATE TABLE employee(id ITN)', postgres parser will generate a valid parse tree
```
dengyuchenkit: @haojin2 I am wondering, there may be a memory leak in the parser. Could you have a look it?
It happens when issuing query _CREATE TABLE employee(id ITN)_. It is invalid since the typename _ITN_ is wrong. The postgres parser returns a valid parse tree. But Peloton will catch the error when doing node transform. The memory block allocated for 'result' object in CreateTransform() is not freed. I see that ListTransform() will free the SQLStatementList. The memory allocated in CreateTransform() may still not be freed since the CreateStatement is not added into the SQLStatementList yet. I am thinking maybe we need some changes in the exception handling. Could you check that? Thanks

src/traffic_cop/traffic_cop.cpp
```diff
@@ -281,15 +275,23 @@ void TrafficCop::ExecuteStatementPlanGetResult() {
...
+ * Prepare a statement based on parse tree. Begin a transaction if necessary.
+ * If the query is not issued in a transaction (when the query type is not 'BEGIN'
+ * and the tcop_txn_state is empty), Peloton will wrap it in a transaction and
+ * begin a transaction here. It is a single_stmt transaction.
+ * Otherwise, the query is issued inside a transaction by the client. It is a multi_stmt transaction.
+ */
 std::shared_ptr<Statement> TrafficCop::PrepareStatement(
-    const std::string &statement_name, const std::string &query_string,
-    UNUSED_ATTRIBUTE std::string &error_message,
+    const std::string &stmt_name, const std::string &query_string,
```
ChTimTsubasa: I am wondering if we still need the query string here if we already do the parsing in the protocol handler?

src/traffic_cop/traffic_cop.cpp
```diff
@@ -210,35 +212,27 @@ executor::ExecuteResult TrafficCop::ExecuteStatementPlan(
...
     threadpool::MonoQueuePool::GetInstance().SubmitTask(ExecutePlanWrapper, arg, task_callback_, task_callback_arg_);
     LOG_TRACE("Submit Task into MonoQueuePool");
-
     is_queuing_ = true;
     return p_status_;
-
   } else {
-    // otherwise, we have already aborted
-    p_status_.m_result = ResultType::ABORTED;
+    // If the transaction state is ABORTED, the transaction should be aborted
```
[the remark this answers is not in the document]
dengyuchenkit: It is fixed. It is a comment.

dengyuchenkit:
@ChTimTsubasa @chenboy Would you mind taking a second round reviewing this? It is ready to be reviewed again

src/include/traffic_cop/traffic_cop.h
```diff
@@ -129,15 +124,14 @@ class TrafficCop {
   std::unique_ptr<optimizer::AbstractOptimizer> optimizer_;
 
   // flag of single statement txn
-  bool single_statement_txn_;
```
ChTimTsubasa: Set this in the constructor.

src/include/type/types.h
```diff
@@ -12,7 +12,7 @@
 
 #pragma once
 
-#include <bitset>
```
ChTimTsubasa: Why is this blank missing?

src/traffic_cop/traffic_cop.cpp
ChTimTsubasa: I think Haoxiang has a PR that fix this problem. Just keep it in this PR.

test/network/exception_test.cpp
dengyuchenkit: Fixed in #979

dengyuchenkit:
@apavlo The problem I mentioned during the meeting is that, in ProtocolHandler when processing EXECUTE query, it does not call optimizer since it doesn't need a query plan. After talking to @chenboy , we realize that it results in the fact that we didn't call binder to check the params and evaluate the expression. The way of directly casting it to Const Expression is incorrect. Will check it

dengyuchenkit:
@tli2 @haojin2 Could you help me review this PR? I think Tianyi had a pass over the ProtocolHandler and TrafficCop part. There're some changes in parser and binder. It will be great if @haojin2 can also take a look at it. Thanks

haojin2:
LGTM so far.

dengyuchenkit:
@apavlo This PR should be ready now.

apavlo:
@dengyuchenkit I will merge this after the build passes.

PR #875, opened 2017-11-01
Design Doc template

tli2:
Moved design doc template from Google Doc to .md


script/testing/sqlite_trace/tags
```diff
@@ -0,0 +1,21 @@
+!_TAG_FILE_SORTED	2	/0=unsorted, 1=sorted, 2=foldcase/
```
apavlo: What is it this?

design_docs/README.md
```diff
@@ -0,0 +1 @@
+Placeholder for actual design doc process
```
apavlo: Should we put this into `doc` instead of `design_doc`.

tli2: I thought about that, but then doc is git ignored for some reason, so I just made a new one. I can ask Joy about this.

PR #877, opened 2017-11-02
Fix Jenkins integration (take 2)

crd477:
This reverts the changes that caused Travis to fail.
After you merge this, I will reconfigure the project on the Jenkins server to work correctly. It will look particularly nice in the Blue Ocean interface. I got pull requests building to work too.
I implemented the alternative platform builds entirely in Dockerfiles so that testing can be done with them outside of Jenkins and also because the Docker/Jenkins integration sucks for anything more complicated than what is here.


apavlo:
We still need to get you the compilation fix, right?

crd477:
On 2017-11-06 11:48, Andy Pavlo wrote:
> We still need to get you the compilation fix, right?
> 

Yes but that is unrelated to all of this noise.  I'm trying to debug a 
build problem that happens only in Travis.  In some cases, it's picking 
up a broken version of LLVM 3.9 that is never even explicitly installed 
by any of the steps.  I suspect it's been preinstalled in their VM and 
I'm trying to figure out where and why it's getting picked up by cmake 
at all.

Sorry for all the noise too...

-- 
     -Chad

apavlo:
@crd477 Actually, see #853:

https://github.com/cmu-db/peloton/pull/853/files#diff-15547c54d3d4898a882b4ab7b3cee381L36

I think that this might solve your LLVM problem. I haven't tested it.

The discussion is here:

https://github.com/cmu-db/peloton/pull/853#issuecomment-338498135

crd477:
On 2017-11-06 14:09, Andy Pavlo wrote:
> @crd477 <https://github.com/crd477> Actually, see #853 
> <https://github.com/cmu-db/peloton/pull/853>:
> 
> https://github.com/cmu-db/peloton/pull/853/files#diff-15547c54d3d4898a882b4ab7b3cee381L36
> 
> I think that this might solve your LLVM problem. I haven't tested it.
> 
> The discussion is here:
> 
> #853 (comment) 
> <https://github.com/cmu-db/peloton/pull/853#issuecomment-338498135>
> 

Thanks but I'm pretty sure that is unrelated.  The Travis problem I'm 
fighting with involves their VMs picking up a completely broken version 
of LLVM to use.  This is an environment/packaging problem and happens 
before compilation even starts.

-- 
     -Chad

crd477:
debugging this in my own fork

Issue #878, opened 2017-11-05
`Select * from Table` hangs if number of rows >  589


prashasthip:
I was trying to do some testing where I insert a large number of rows into a table and then call `Select * from Table;` on it using PSQL.

Select * works well as long as the number of rows < 590. Else, it hangs after the query's tearDown().
I suspect that there may be some bug in the network_handler that is causing some sort of a deadlock.

@ChTimTsubasa has seen this issue previously.

rohit-cmu:
I am able to reproduce this issue. I noticed another possible bug. As @prashasthip said the terminal hangs after the query's tearDown(). If you try to close this connection, you get a seg fault in the **network threads**.

Here is the stack trace -

`2017-11-05 15:02:00 [src/network/network_master_thread.cpp:126:DispatchConnection] DEBUG - Dispatching connection to worker 2
2017-11-05 15:02:00 [src/network/network_callback_util.cpp:45:WorkerHandleNewConn] DEBUG - Creating new socket fd:19
2017-11-05 15:02:00 [src/network/network_manager.cpp:50:CreateNewConnection] INFO  - create new connection: id = 19
2017-11-05 15:02:00 [src/network/protocol_handler.cpp:67:ProtocolHandler] DEBUG - Registered new ProtocolHandler [count=4]
2017-11-05 15:02:00 [src/network/protocol_handler.cpp:138:ProcessInitialPacket] INFO  - protocol version: 80877102
2017-11-05 15:02:00 [src/network/protocol_handler.cpp:166:ProcessStartupPacket] ERROR - Protocol error: Only protocol version 3 is supported.
2017-11-05 15:02:00 [src/network/protocol_handler.cpp:79:~ProtocolHandler] DEBUG - Removed ProtocolHandler [count=3]
2017-11-05 15:02:00 [src/network/protocol_handler.cpp:79:~ProtocolHandler] DEBUG - Removed ProtocolHandler [count=2]
2017-11-05 15:02:00 [src/network/protocol_handler.cpp:79:~ProtocolHandler] DEBUG - Removed ProtocolHandler [count=1]
2017-11-05 15:02:00 [src/network/protocol_handler.cpp:79:~ProtocolHandler] DEBUG - Removed ProtocolHandler [count=0]
==42675== Thread 4:
==42675== Invalid read of size 1
==42675==    at 0x1003F723C: peloton::network::NetworkThread::GetThreadIsStarted() (network_thread.h:78)
[...]

PR #882, opened 2017-11-07
Fix for 'select * ' hanging

dengyuchenkit:
Fix for #878 


dengyuchenkit:
I don't know why it is not running check. I will close it and submit it again

PR #884, opened 2017-11-09
Remove all delete calls in /test

seojungmin:
This PR removes all the `delete` calls in our `test` directory by applying smart pointers.   This work started by simplifying the `InsertPlan` code to enhance parameterizaton later in this week, so it would be nice to merge it soon.

The code changes are straightforward, so someone could spend less than quarter an hour to review them.

### DIFF
--- a/src/include/planner/insert_plan.h
+++ b/src/include/planner/insert_plan.h
@@ -6,7 +6,7 @@
 //
 // Identification: src/include/planner/insert_plan.h
 //
-// Copyright (c) 2015-16, Carnegie Mellon University Database Group
+// Copyright (c) 2015-17, Carnegie Mellon University Database Group
 //
 //===----------------------------------------------------------------------===//
 
@@ -30,12 +30,11 @@ class InsertStatement;
 namespace planner {
 class InsertPlan : public AbstractPlan {
  public:
-  // This constructor takes in neither a project info nor a tuple
-  // It must be used when the input is a logical tile
+  // Construct when input is a logical tile - WARNING Not supported by codegen
   InsertPlan(storage::DataTable *table, oid_t bulk_insert_count = 1)
     : target_table_(table), bulk_insert_count_(bulk_insert_count) {}
 
-  // This constructor takes in a project info
+  // Construct with a project info
   InsertPlan(storage::DataTable *table,
              std::unique_ptr<const planner::ProjectInfo> &&project_info,
              oid_t bulk_insert_count = 1)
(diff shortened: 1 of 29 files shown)


apavlo:
@seojungmin I think you need to rebase first.

apavlo:
@chenboy @luochenUmich Can you guys review this today? I think that we should get it in before tomorrow's hackathon. Thanks.

luochenUmich:
sure, I will review this today

src/include/planner/insert_plan.h
```diff
@@ -29,27 +30,34 @@ class InsertStatement;
...
-      const std::vector<std::vector<std::unique_ptr<expression::AbstractExpression>>> *
-          insert_values);
-
-  // Get a varlen pool (will construct the pool only if needed)
+  // Construct when input is a logical tile
+  InsertPlan(storage::DataTable *table, oid_t bulk_insert_count = 1)
+    : target_table_(table), bulk_insert_count_(bulk_insert_count) {}
+
+  // Construct with a project info
+  InsertPlan(storage::DataTable *table,
```
luochenUmich: I am not sure whether moving the implementation of some constructors to the header file is a good idea. Any specific reason for doing this?

seojungmin: These constructors that I moved are doing nothing much other than setting member variables, so I moved them to headers.  Nothing functional. Do we have a rule on this? I could apply accordingly. Thanks for the comment. I did not think much about it when I changed. 👍

test/index/testing_index_util.cpp
```diff
@@ -36,8 +36,11 @@ void TestingIndexUtil::BasicTest(const IndexType index_type) {
   std::vector<ItemPointer *> location_ptrs;
 
   // INDEX
-  std::unique_ptr<index::Index> index(
-      TestingIndexUtil::BuildIndex(index_type, false));
+  std::unique_ptr<index::Index, std::function<void(index::Index *)>> index(
+      TestingIndexUtil::BuildIndex(index_type, false),
+      [](index::Index *index) {
```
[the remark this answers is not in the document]
seojungmin: If I put this in a separate function and reuse it, the binary size(in the release mode) gets bigger by 169 bytes.  This is because the binary needs some symbol information for handling a separate function.  However, I would buy your argument because it is less error-prone.

src/include/planner/insert_plan.h
luochenUmich: I don't think there is a hard rule. Just in general, it would be nice to keep it consistent and make the header shorter.

luochenUmich:
Looks good to me. After the build passes, it should be good to go.

PR #885, opened 2017-11-09
Logging and Recovery

PauloAmora:
This PR cleans up #792. It is now ready for review. 

A first draft of the design doc can be found in https://docs.google.com/document/d/1gTW6yiRcPxf4GwX6mmWEjM4Z7dJpQ05Tx9kpYbdsMQ0/edit?usp=sharing . Suggestions are accepted.


src/include/network/network_connection.h
```diff
@@ -50,28 +51,33 @@ namespace network {
...
+
+  SSL *conn_SSL_context = nullptr;  // SSL context for the connection
+
+  NetworkThread *thread;  // reference to the network thread
+  std::unique_ptr<ProtocolHandler>
+      protocol_handler_;                      // Stores state for this socket
   ConnState state = ConnState::CONN_INVALID;  // Initial state of connection
   tcop::TrafficCop traffic_cop_;
+
+  logging::WalLogManager log_manager_;
```
ChTimTsubasa: Can we make log_manager a member of traffic_cop_? It seems that it is always passing way down to traffic_cop. We can have a setter function to bind logging event with log_manager.

src/network/network_connection.cpp
```diff
@@ -73,20 +88,28 @@ void NetworkConnection::Init(short event_flags, NetworkThread *thread,
 
   event_add(network_event, nullptr);
   event_add(workpool_event, nullptr);
+  event_add(logpool_event, nullptr);
```
ChTimTsubasa: I still doubt why we need this logpool_event. Can't we just use the workerpool_event? Since it would be no way for worker in worker pool and logger to invoke the event simultainously.

src/threadpool/logger_queue_pool.cpp
```diff
@@ -0,0 +1,32 @@
...
+
+void LoggerQueuePool::SubmitTask(void (*task_ptr)(void*), void* task_arg,
+                                 void (*task_callback_ptr)(void*),
+                                 void* task_callback_arg) {
+  task_queue_.EnqueueTask(task_ptr, task_arg, task_callback_ptr,
+                          task_callback_arg);
+}
+
+LoggerQueuePool& LoggerQueuePool::GetInstance() {
+  static LoggerQueuePool logger_queue_pool;
```
ChTimTsubasa: We should stop using singleton now.

src/network/network_connection.cpp
PauloAmora: I don't like the idea of having two different events with two different purposes sharing the same trigger. They should do the same thing (trigger the state machine) but I don't think we should mix those.

src/include/network/network_connection.h
PauloAmora: Done

src/threadpool/logger_queue_pool.cpp
PauloAmora: Fair, I've seen some refactoring on the worker pool, I'll try to do something similar

PauloAmora:
Found a bug while implementing the recovery tests, please hold on this a little.

src/logging/wal_log_manager.cpp
```diff
@@ -0,0 +1,67 @@
...
+// Static method that accepts void pointer
+void WalLogManager::WriteTransactionWrapper(void* arg_ptr) {
+  LogTransactionArg* arg = (LogTransactionArg*)arg_ptr;
+  WriteTransaction(arg->log_records_);
+  delete (arg);
+}
+
+// Actual method called by the logger thread
+void WalLogManager::WriteTransaction(std::vector<LogRecord> log_records) {
+  WalLogger* wl = new WalLogger(0, "/tmp/log");
```
seojungmin: The file path should not be hard-coded.  Use the configuration..

seojungmin: Every WriteTransaction, you are allocating an object.  Isn't this inefficient?

src/common/init.cpp
```diff
@@ -75,6 +76,9 @@ void PelotonInit::Initialize() {
   pg_catalog->CreateDatabase(DEFAULT_DB_NAME, txn);
 
   txn_manager.CommitTransaction(txn);
```
seojungmin: Is this ok to call commit, before recovery? If ok, then can you comment why?

PauloAmora: It was a design decision to let the catalog initialize and bootstrap before recovery. Therefore, since the catalog is always rebuilt on bootstrapping, we avoid logging those entries.

PauloAmora:
About the formatting issues: I ran clang-format through the python script on changed files, it should be ok.

src/logging/wal_recovery.cpp
```diff
@@ -0,0 +1,540 @@
...
+        tup->GetValue(1).ToString(), tup->GetValue(0).GetAs<oid_t>(),
+        table->GetOid(), database_oid,
+        static_cast<IndexType>(tup->GetValue(3).GetAs<oid_t>()),
+        static_cast<IndexConstraintType>(tup->GetValue(4).GetAs<oid_t>()),
+        tuple_schema, key_schema, key_attrs, tup->GetValue(4).GetAs<bool>());
+    std::shared_ptr<index::Index> index(
+        index::IndexFactory::GetIndex(index_metadata));
+    table->AddIndex(index);
+    tables_with_indexes.insert(table);
+    // Attributes must be changed once we have arraytype
```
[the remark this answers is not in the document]
PauloAmora: Currently, the indexed attributes on pg_index are stored as a string, but it is clearly supposed to be an array. The logic a little bit above this snippet deals with it as a string and will probably need to change when arraytypes are implemented.

src/logging/wal_recovery.cpp
```diff
@@ -0,0 +1,540 @@
...
+            tg->InsertTupleFromRecovery(current_cid, tg_offset, tuple.get());
+        old_tg->UpdateTupleFromRecovery(current_cid, old_tg_offset, location);
+        table->IncreaseTupleCount(1);
+        if (tuple_id == tg->GetAllocatedTupleCount() - 1) {
+          if (table->GetTileGroupById(tg->GetTileGroupId() + 1).get() ==
+              nullptr)
+            table->AddTileGroupWithOidForRecovery(tg->GetTileGroupId() + 1);
+          catalog::Manager::GetInstance().GetNextTileGroupId();
+        }
+        /*
```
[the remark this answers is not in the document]
PauloAmora: So far, `ALTER` statements are not supported in Peloton. Once they are, we will have to deal with updates in the catalog tables.

src/logging/wal_recovery.cpp
```diff
@@ -0,0 +1,540 @@
...
+              uint index = stoi(tuple->GetValue(2).ToString());
+              if (index >= columns.size()) {
+                // Made to fit index as the last element
+                columns.resize(index + 1);
+              }
+              if (column_type == type::TypeId::VARCHAR ||
+                  column_type == type::TypeId::VARBINARY) {
+                catalog::Column tmp_col(column_type,
+                                        type::Type::GetTypeSize(column_type),
+                                        tuple->GetValue(1).ToString(), false,
```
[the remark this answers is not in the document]
PauloAmora: Got it.

src/logging/wal_recovery.cpp
```diff
@@ -0,0 +1,540 @@
...
+                                        tuple->GetValue(1).GetAs<oid_t>());
+                columns[index] = tmp_col;
+              } else {
+                catalog::Column tmp_col(column_type,
+                                        type::Type::GetTypeSize(column_type),
+                                        tuple->GetValue(1).ToString(), true,
+                                        tuple->GetValue(1).GetAs<oid_t>());
+                columns[index] = tmp_col;
+                //  columns.insert(columns.begin(), catalog::Column();
```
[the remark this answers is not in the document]
PauloAmora: I missed it on my cleaning.

src/logging/wal_recovery.cpp
```diff
@@ -0,0 +1,540 @@
...
+              break;
+            }
+            case COLUMN_CATALOG_OID:  // pg_attribute
+            {
+              std::string typeId = tuple->GetValue(4).ToString();
+              type::TypeId column_type = StringToTypeId(typeId);
+              uint index = stoi(tuple->GetValue(2).ToString());
+              if (index >= columns.size()) {
+                // Made to fit index as the last element
+                columns.resize(index + 1);
```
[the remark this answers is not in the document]
PauloAmora: This was an issue that I found out about last week. So far, the log records were retrieved from the RWSet, which is unordered, and that caused some issues such as records in the `PG_ATTRIBUTE` table being out of order and it botched schema reconstruction. This line is supposed to, if the vector is smaller than the column position(index), to resize it so the column could be placed in the correct place.

I've corrected this and now the log records are flushed in proper order, making this code useless.

src/logging/wal_recovery.cpp
```diff
@@ -0,0 +1,540 @@
...
+        ItemPointer location(tg_block, tg_offset);
+        auto table = storage::StorageManager::GetInstance()->GetTableWithOid(
+            database_id, table_id);
+        auto schema = table->GetSchema();
+        auto tg = table->GetTileGroupById(tg_block);
+        // The required tile group might not have been created yet
+        if (tg.get() == nullptr) {
+          table->AddTileGroupWithOidForRecovery(tg_block);
+          tg = table->GetTileGroupById(tg_block);
+          catalog::Manager::GetInstance().GetNextTileGroupId();
```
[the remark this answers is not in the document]
PauloAmora: As explained below, if the record tilegroup didn't exist, it was created on the spot.

src/logging/wal_recovery.cpp
```diff
@@ -0,0 +1,540 @@
...
+    if (LoggingUtil::ReadNBytesFromFile(file_handle, (void *)&length_buf, 4) ==
+        false) {
+      LOG_TRACE("Reach the end of the log file");
+      break;
+    }
+    CopySerializeInput length_decode((const void *)&length_buf, 4);
+    int length = length_decode.ReadInt();
+    // Adjust the buffer
+    if ((size_t)length > buf_size) {
+      buffer.reset(new char[(int)(length * 1.2)]);
```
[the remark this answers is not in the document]
PauloAmora: Missed it.

seojungmin:
@PauloAmora Thanks for clarifications.

PauloAmora:
@apavlo Recovery tests added and PR rebased

apavlo:
@PauloAmora Can you resolve the conflicts and rebase? I don't know how I missed this over the weekend.

PauloAmora:
So @apavlo, the call was choppy but as far as I understood, someone is looking at this before merging?

PR #886, opened 2017-11-10
Remove parser hack for SELECT without FROM

hzxa21:
This PR fixes SELECT without FROM by removing the hack in the parser and conducting the check in the binder. We should get this merged in before adding support for new built-in SQL functions.

This PR also enables binder to actually bind function expression. Previously the binding is done in the optimizer since we didn't have a binder then :(


src/include/expression/expression_util.h
```diff
@@ -776,9 +761,8 @@ class ExpressionUtil {
       auto catalog = catalog::Catalog::GetInstance();
       const catalog::FunctionData &func_data =
           catalog->GetFunction(func_expr->GetFuncName(), argtypes);
-      func_expr->SetFunctionExpressionParameters(func_data.func_,
-                                                 func_data.return_type_,
-                                                 func_data.argument_types_);
+      func_expr->SetFunctionExpressionParameters(
+          func_data.func_, func_data.return_type_, func_data.argument_types_);
     }
```
luochenUmich: Can you remove deduce expression type in EvaluateExpression?

test/parser/parser_test.cpp
```diff
@@ -390,20 +390,5 @@ TEST_F(ParserTests, CopyTest) {
...
-  queries.push_back("SELECT *;");
-  // Query with wrong syntax
-  queries.push_back("SECLECT *;");
-
-  // Parsing
-  for (auto query : queries) {
-    EXPECT_THROW(parser::PostgresParser::ParseSQLString(query.c_str()),
-                 Exception);
-  }
-}
```
luochenUmich: Can you move this to binder_test instead of deleting it?

src/binder/bind_node_visitor.cpp
```diff
@@ -175,5 +178,38 @@ void BindNodeVisitor::Visit(expression::CaseExpression *expr) {
...
+
+  // Check catalog and bind function
+  std::vector<type::TypeId> argtypes;
+  for (size_t i = 0; i < expr->GetChildrenSize(); i++)
+    argtypes.push_back(expr->GetChild(i)->GetValueType());
+  // Check and set the function ptr
+  auto catalog = catalog::Catalog::GetInstance();
+  const catalog::FunctionData &func_data =
+      catalog->GetFunction(expr->GetFuncName(), argtypes);
+  LOG_INFO("Function %s found in the catalog", func_data.func_name_.c_str());
```
apavlo: Switch to `LOG_DEBUG`

hzxa21: This is not going through LLVM because the ProjectionTranslator requires the projection_plan must have a child. I will add support for that.
The test cases are already there in the optimizer_sql_test.

test/parser/parser_test.cpp
hzxa21: Sure.

malin1993ml:
I'm wondering what's the status of this? Are we going to merge this anytime soon? #901 requires some fixes in this PR to do more proper testing.

hzxa21:
@malin1993ml This PR can be merged. I will do the rebase and merge it once all the tests pass.

malin1993ml:
Cool! Thanks.

PR #887, opened 2017-11-10
Implement function Length

chenboy:
In this PR we implement `Length` SQL function for the LLVM engine. Now we can support queries like `SELECT length(a) FROM table`. We also add a test case in `string_functions_test.cpp`.

### DIFF
--- a/src/function/string_functions.cpp
+++ b/src/function/string_functions.cpp
@@ -266,7 +266,7 @@ type::Value StringFunctions::BTrim(const std::vector<type::Value> &args) {
 uint32_t StringFunctions::Length(UNUSED_ATTRIBUTE const char *str,
                                  uint32_t length) {
   PL_ASSERT(str != nullptr);
-  return length > 0 ? length - 1 : length;
+  return length;
 }
 
 // The length of the string
--- a/src/function/string_functions.cpp
+++ b/src/function/string_functions.cpp
@@ -263,8 +263,8 @@ type::Value StringFunctions::BTrim(const std::vector<type::Value> &args) {
   return (type::ValueFactory::GetVarcharValue(str));
 }
 
-}  // namespace peloton
-uint32_t StringFunctions::Length(const char *str, uint32_t length) {
+uint32_t StringFunctions::Length(UNUSED_ATTRIBUTE const char *str,
+                                 uint32_t length) {
   PL_ASSERT(str != nullptr);
   return length;
 }
--- a/src/function/string_functions.cpp
+++ b/src/function/string_functions.cpp
@@ -266,10 +266,10 @@ type::Value StringFunctions::BTrim(const std::vector<type::Value> &args) {
 uint32_t StringFunctions::Length(UNUSED_ATTRIBUTE const char *str,
                                  uint32_t length) {
   PL_ASSERT(str != nullptr);
-  return length;
+  return length > 0 ? length - 1 : length;
 }
 
-// Length of the argument.
+// The length of the string
 type::Value StringFunctions::_Length(const std::vector<type::Value> &args) {
   PL_ASSERT(args.size() == 1);
   if (args[0].IsNull()) {
--- a/src/function/string_functions.cpp
+++ b/src/function/string_functions.cpp
@@ -269,5 +269,16 @@ uint32_t StringFunctions::Length(const char *str, uint32_t length) {
   return length;
 }
 
+// Length of the argument.
+type::Value StringFunctions::_Length(const std::vector<type::Value> &args) {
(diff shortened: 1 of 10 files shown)


src/function/string_functions.cpp
```diff
@@ -186,5 +186,21 @@ type::Value StringFunctions::BTrim(const std::vector<type::Value> &args) {
   return (type::ValueFactory::GetVarcharValue(str));
 }
 
+uint32_t StringFunctions::Length(UNUSED_ATTRIBUTE const char *str, uint32_t length) {
+  PL_ASSERT(str != nullptr);
+  return length;
+}
+
+// ASCII code of the first character of the argument.
```
apavlo: Fix comment.

chenboy:
We've added a sql test for string_functions.

test/sql/stringfunctions_sql_test.cpp
```diff
@@ -0,0 +1,71 @@
+//===----------------------------------------------------------------------===//
```
seojungmin: Format the file's header comment correctly (add copyright, etc.) and set the namespace comment etc.  Run time source validator.

test/sql/stringfunctions_sql_test.cpp
```diff
@@ -0,0 +1,71 @@
...
+  catalog::Catalog::GetInstance()->Bootstrap();
+  txn_manager.CommitTransaction(txn);
+  // Create a t
+  txn = txn_manager.BeginTransaction();
+
+  TestingSQLUtil::ExecuteSQLQuery(
+      "CREATE TABLE foo(id integer, name varchar(500));");
+  int i;
+  std::string s = "'", t = "'";
+  for (i = 0; i < 500; i++) {
```
seojungmin: Is there any reason that we have 500 tuples? Otherwise, let's pick something like 32, or less. We are not testing the tuple insertion here, so we don't need to waste our time running the test cases. It is relatively small, but they all add up.

test/sql/stringfunctions_sql_test.cpp
seojungmin: While changing these, can you set the file name to be "string_functions_sql_test.cpp".  I looked at others' PR and this seems to be the mostly preferable format of the file name.

DeanChensj: Hi, what do you mean by 'Run time source validator'? I'm new to Pelton and DBMS, thanks.

seojungmin: Oops. Run 'the' source validator.  `python ./script/validators/source_validator.py`

test/sql/stringfunctions_sql_test.cpp
DeanChensj: Fix that part. Thanks.

hzxa21:
Is this PR ready to be merged?

pmenon:
I still see the file `src/include/type/value_peeker.h-0e35e24c`. Please remove it.

DeanChensj:
Just removed that file.

hzxa21:
I will do the merge once all the tests pass.

pmenon:
@allisonwang Does this implementation match up with how we're dealing with string lengths?

chenboy:
This currently does not match up with the string length fix. So I think we should merge Allison's PR first. Then we'll rebase and require a review.

hzxa21:
I see. Thanks for point out. This PR should be merged after #911.

hzxa21:
@chenboy #911 has been merged. Can you rebase and update this PR for the length issue?

chenboy:
I've rebased and update this PR. I think it's ready for review now.

chenboy:
Hi @pmenon, I found there are still potential varchar length bugs in the type system and I tried to fix them in my new commits. I'm not 100% sure if these changes are correct, could you take a quick look?

pmenon:
@chenboy I'm seeing commits from other PR's. Can you rebase?

chenboy:
I'm not quite sure if `size_.len` in `type::Value` for varchar is intended to be the length of the actual string + 1. Right now it seems like we're treating it this way. Examples are the function to initialize a varchar value https://github.com/cmu-db/peloton/blob/master/src/include/type/value_factory.h#L78 and the function to get the length https://github.com/cmu-db/peloton/blob/master/src/type/varlen_type.cpp#L52 . I think we should at least add some documentation for this design decision. I got kind of confused when I first read this piece of code.

hzxa21:
@chenboy Can you resolve the conflicts? I will do the merge after that.

pmenon:
LGTM +1

PR #889, opened 2017-11-10
Floor function addition

saatviks:
Added Floor function following the [Postgres standard](https://www.postgresql.org/docs/9.5/static/functions-math.html). Added 2 unit tests - one to check working of the Floor function internally and one with SQL.


test/sql/decimal_functions_sql_test.cpp
```diff
@@ -0,0 +1,81 @@
+//===----------------------------------------------------------------------===//
+//
+//                         Peloton
+//
+// distinct_sql_test.cpp
```
apavlo: Run the formatter script to update header:
```
$ ./script/formatting/formatter.py --add-header  test/sql/decimal_functions_sql_test.cpp
```

test/sql/decimal_functions_sql_test.cpp
```diff
@@ -0,0 +1,81 @@
...
+      TestingSQLUtil::ExecuteSQLQuery(os.str());
+    }
+    EXPECT_EQ(i, numEntries);
+
+    // Fetch values from the table
+    std::vector<StatementResult> result;
+    std::vector<FieldInfo> tuple_descriptor;
+    std::string error_message;
+    int rows_affected;
+    std::ostringstream os;
```
apavlo: Why not just use `std::string` here?

test/sql/decimal_functions_sql_test.cpp
saatviks: Addressed in latest commit

test/sql/decimal_functions_sql_test.cpp
saatviks: Addressed in latest commit

PR #894, opened 2017-11-10
Abs function addition

jiachenwang:
Added Abs function for decimal following the Postgres standard. Added 2 unit tests - one to check working of the Abs function internally and one with SQL.

### DIFF
--- a/src/codegen/query_compiler.cpp
+++ b/src/codegen/query_compiler.cpp
@@ -119,7 +119,7 @@ bool QueryCompiler::IsExpressionSupported(
     const expression::AbstractExpression &expr) {
   switch (expr.GetExpressionType()) {
     case ExpressionType::STAR:
-    case ExpressionType::FUNCTION:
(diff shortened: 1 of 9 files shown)


hzxa21:
@jiachenwang It looks like you are failing a test case. Can you fix that and request a review?

jiachenwang:
Sure, I will fix it and request again ASAP.

src/codegen/query_compiler.cpp
```diff
@@ -119,7 +119,7 @@ bool QueryCompiler::IsExpressionSupported(
     const expression::AbstractExpression &expr) {
   switch (expr.GetExpressionType()) {
     case ExpressionType::STAR:
-    case ExpressionType::FUNCTION:
+    // case ExpressionType::FUNCTION:
```
pmenon: Remove this. We'll fix when we have function parity with existing value-based functions.

apavlo:
Replaced by #1031

PR #900, opened 2017-11-11
Fix for 893 - Casting

### LINKED ISSUE #893: Codegen crashes on expression with different types
The codegen part crashes when using an arithmetic expression with two different types. Apparently the null bitmap is missing. 

```sql
CREATE TABLE a (i INT, j SMALLINT);
INSERT INTO A VALUES (2, 1), (3, 1), (4, 1);
SELECT i + j FROM a;
```
http://sqlfiddle.com/#!15/ae631/1/0

Error:
```
peloton/src/codegen/value.cpp:30: peloton::codegen::Value::Value(const peloton::codegen::type::Type&, llvm::Value*, llvm::Value*, llvm::Value*): Assertion (!type_.nullable || null_ != nullptr) failed.
```

Assertion failed here:
```c++
Value::Value(const type::Type &type, llvm::Value *val, llvm::Value *length,
             llvm::Value *null)
    : type_(type), value_(val), length_(length), null_(null) {
  // If the value is NULL-able, it better have an accompanying NULL bit
  PL_ASSERT(!type_.nullable || null_ != nullptr);
}
```

pmenon:
This PR fixed #893. The problem stems from passing along the wrong type when performing the cast. If the input is not null, we still pass a NULLable type when doing the cast. Thus, when the `Value` is created, it has a NULLable type, but no null-bit, throwing the error.

To fix, we ensure the type we're casting to is not NULLable in the branch where we know the input value is not NULL.


tcm-marcel:
(Sorry, I accidentally added a reviewer)

PR #901, opened 2017-11-12
Add date_trunc function

malin1993ml:
This PR implements the date_trunc SQL function. ~~It contains a SQL test, however I cannot do the real appropriate SQL test without adding the logic in the binder to check the DatePartType of the date_trunc function and replace the type string with its real type ID. That requires some fixes in #886.~~

### DIFF
--- a/src/binder/bind_node_visitor.cpp
+++ b/src/binder/bind_node_visitor.cpp
@@ -203,7 +203,10 @@ void BindNodeVisitor::Visit(expression::FunctionExpression *expr) {
   // we have to do the string comparison to find out the corresponding
   // DatePartType when scanning every tuple.
   auto func_name = expr->GetFuncName();
-  if (func_name == "date_trunc" || func_name == "extract") {
+  auto func_operator_id =
+      function::BuiltInFunctions::GetFuncBySQLName(func_name).op_id;
+  if (func_operator_id == OperatorId::DateTrunc ||
+      func_operator_id == OperatorId::Extract) {
     // Check the type of the first argument. Should be VARCHAR
     auto date_part = expr->GetChild(0);
     if (date_part->GetValueType() != type::TypeId::VARCHAR) {
--- a/src/binder/bind_node_visitor.cpp
+++ b/src/binder/bind_node_visitor.cpp
@@ -199,34 +199,6 @@ void BindNodeVisitor::Visit(expression::FunctionExpression *expr) {
   // Visit the subtree first
   SqlNodeVisitor::Visit(expr);
 
-  // Specialize the first argument (DatePartType) for date functions, otherwise
-  // we have to do the string comparison to find out the corresponding
-  // DatePartType when scanning every tuple.
-  auto func_name = expr->GetFuncName();
-  auto func_operator_id =
-      function::BuiltInFunctions::GetFuncBySQLName(func_name).op_id;
-  if (func_operator_id == OperatorId::DateTrunc ||
-      func_operator_id == OperatorId::Extract) {
-    // Check the type of the first argument. Should be VARCHAR
-    auto date_part = expr->GetChild(0);
-    if (date_part->GetValueType() != type::TypeId::VARCHAR) {
-      throw Exception(EXCEPTION_TYPE_EXPRESSION,
-                      "Incorrect argument type to function: " + func_name +
-                          ". Argument 0 expected type VARCHAR but found " +
-                          TypeIdToString(date_part->GetValueType()) + ".");
-    }
-
-    // Convert the first argument to DatePartType
-    auto date_part_type = StringToDatePartType(
-        date_part->Evaluate(nullptr, nullptr, nullptr).ToString());
-    auto date_part_integer = type::ValueFactory::GetIntegerValue(
-        static_cast<int32_t>(date_part_type));
-
-    // Replace the first argument with an Integer expression of the DatePartType
-    expr->SetChild(0,
-                   new expression::ConstantValueExpression(date_part_integer));
-  }
-
   // Check catalog and bind function
   std::vector<type::TypeId> argtypes;
   for (size_t i = 0; i < expr->GetChildrenSize(); i++)
@@ -239,6 +211,20 @@ void BindNodeVisitor::Visit(expression::FunctionExpression *expr) {
   LOG_DEBUG("Argument num: %ld", func_data.argument_types_.size());
   expr->SetFunctionExpressionParameters(func_data.func_, func_data.return_type_,
                                         func_data.argument_types_);
+
+  // Look into the OperatorId for built-in functions to check the first argument
+  // for timestamp functions.
+  // TODO(LM): The OperatorId might need to be changed to global ID after we
+  // rewrite the function identification logic.
+  auto func_operator_id = func_data.func_.op_id;
+  if (func_operator_id == OperatorId::DateTrunc ||
+      func_operator_id == OperatorId::Extract) {
+    auto date_part = expr->GetChild(0);
+
+    // Test whether the first argument is a correct DatePartType
+    StringToDatePartType(
+        date_part->Evaluate(nullptr, nullptr, nullptr).ToString());
+  }
 }
 
 }  // namespace binder
--- a/src/binder/bind_node_visitor.cpp
+++ b/src/binder/bind_node_visitor.cpp
@@ -203,7 +203,7 @@ void BindNodeVisitor::Visit(expression::FunctionExpression *expr) {
   // we have to do the string comparison to find out the corresponding
   // DatePartType when scanning every tuple.
   auto func_name = expr->GetFuncName();
-  if (func_name == "date_trunc" or expr->GetFuncName() == "extract") {
+  if (func_name == "date_trunc" or func_name == "extract") {
     // Check the type of the first argument. Should be VARCHAR
     auto date_part = expr->GetChild(0);
     if (date_part->GetValueType() != type::TypeId::VARCHAR) {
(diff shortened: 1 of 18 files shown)


malin1993ml:
This PR is ready for review. @pmenon Can you help do the code review for this?

pmenon:
@malin1993ml Will do.

src/binder/bind_node_visitor.cpp
```diff
@@ -197,6 +199,31 @@ void BindNodeVisitor::Visit(expression::FunctionExpression *expr) {
   // Visit the subtree first
   SqlNodeVisitor::Visit(expr);
 
+  // Specialize the first argument (DatePartType) for date functions, otherwise
+  // we have to do the string comparison to find out the corresponding
+  // DatePartType when scanning every tuple.
+  auto func_name = expr->GetFuncName();
+  if (func_name == "date_trunc" or expr->GetFuncName() == "extract") {
```
pmenon: Change to:
```
if (func_name == "date_trunc" || func_name == "extract") { 
  ...
}
```
Just to be consistent with the rest of the codebase.

malin1993ml:
@pmenon Fixed.

src/codegen/type/integer_type.cpp
```diff
@@ -374,6 +376,31 @@ struct Modulo : public TypeSystem::BinaryOperator {
   }
 };
 
+// DateTrunc
+struct DateTrunc : public TypeSystem::BinaryOperator {
```
pmenon: Add a `TODO` to move this to the `Timestamp` type once the function lookup logic is corrected.

src/type/types.cpp
```diff
@@ -285,7 +285,7 @@ type::TypeId StringToTypeId(const std::string& str) {
   return type::TypeId::INVALID;
 }
 
-std::string TypeIdArrayToString(const std::vector<type::TypeId> &types) {
+std::string TypeIdArrayToString(const std::vector<type::TypeId>& types) {
```
pmenon: Revert.

src/type/types.cpp
```diff
@@ -296,7 +296,7 @@ std::string TypeIdArrayToString(const std::vector<type::TypeId> &types) {
 
 // Get argument type vector from its string representation
 // e.g. "integer,boolean" --> vector{TypeId::INTEGER, TypeId::BOOLEAN}
-std::vector<type::TypeId> StringToTypeArray(const std::string &types) {
+std::vector<type::TypeId> StringToTypeArray(const std::string& types) {
```
pmenon: Revert.

src/type/types.cpp
```diff
@@ -1334,14 +1334,13 @@ AggregateType StringToAggregateType(const std::string &str) {
...
-        StringUtil::Format("No AggregateType conversion from string '%s'",
-                           upper_str.c_str()));
+    throw ConversionException(StringUtil::Format(
+        "No AggregateType conversion from string '%s'", upper_str.c_str()));
   }
   return AggregateType::INVALID;
 }
 
-std::ostream &operator<<(std::ostream &os, const AggregateType &type) {
+std::ostream& operator<<(std::ostream& os, const AggregateType& type) {
```
pmenon: Revert.

src/type/types.cpp
```diff
@@ -1539,14 +1536,13 @@ CopyType StringToCopyType(const std::string &str) {
...
-        StringUtil::Format("No CopyType conversion from string '%s'",
-                           upper_str.c_str()));
+    throw ConversionException(StringUtil::Format(
+        "No CopyType conversion from string '%s'", upper_str.c_str()));
   }
   return CopyType::INVALID;
 }
 
-std::ostream &operator<<(std::ostream &os, const CopyType &type) {
+std::ostream& operator<<(std::ostream& os, const CopyType& type) {
```
pmenon: Revert.

src/type/types.cpp
```diff
@@ -1865,7 +1858,7 @@ std::string ProtocolTypeToString(ProtocolType type) {
   return "INVALID";
 }
 
-ProtocolType StringToProtocolType(const std::string &str) {
+ProtocolType StringToProtocolType(const std::string& str) {
```
pmenon: Revert the remainder of these types of changes to make the pointers/references right-associative.

malin1993ml:
@pmenon Yes. I commented out the FUNCTION type in query_compiler locally and verified that it worked through cogen engine.

src/binder/bind_node_visitor.cpp
```diff
@@ -197,6 +199,31 @@ void BindNodeVisitor::Visit(expression::FunctionExpression *expr) {
   // Visit the subtree first
   SqlNodeVisitor::Visit(expr);
 
+  // Specialize the first argument (DatePartType) for date functions, otherwise
+  // we have to do the string comparison to find out the corresponding
+  // DatePartType when scanning every tuple.
+  auto func_name = expr->GetFuncName();
+  if (func_name == "date_trunc" or func_name == "extract") {
```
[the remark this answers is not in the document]
malin1993ml: Yes. They're guaranteed to be transformed into lower case before this point:
https://github.com/cmu-db/peloton/blob/master/src/parser/postgresparser.cpp#L481
But still that's a good catch:)

src/codegen/type/integer_type.cpp
malin1993ml: Do we have a plan on correcting the function lookup logic? Just curious.

src/type/types.cpp
malin1993ml: When I ran clang-format on my local machine it automatically formatted the code this way... Weird, but I'll figure it out.

malin1993ml:
@pmenon Done.

src/function/timestamp_functions.cpp
```diff
@@ -0,0 +1,157 @@
...
+#include "common/logger.h"
+#include "type/limits.h"
+#include "type/value_factory.h"
+
+namespace peloton {
+namespace function {
+
+uint64_t TimestampFunctions::DateTrunc(uint32_t date_part_type,
+                                       uint64_t value) {
+  DatePartType date_part =
```
pmenon: You should just be able to `static_cast<uint32_t>` if you change `DatePartType` to an enum class that subclasses `uint32_t`

src/binder/bind_node_visitor.cpp
```diff
@@ -197,6 +199,31 @@ void BindNodeVisitor::Visit(expression::FunctionExpression *expr) {
   // Visit the subtree first
   SqlNodeVisitor::Visit(expr);
 
+  // Specialize the first argument (DatePartType) for date functions, otherwise
+  // we have to do the string comparison to find out the corresponding
+  // DatePartType when scanning every tuple.
+  auto func_name = expr->GetFuncName();
+  if (func_name == "date_trunc" || func_name == "extract") {
```
pmenon: I didn't see the whole context here previously, but since you're working with `FunctionExpression`, you have access to the `OperatorId`. Can you use that instead of comparing strings? This way, we'll never have to worry about the case of strings, if it changes later.

malin1993ml: Yes. I think that would work!

src/function/timestamp_functions.cpp
malin1993ml: I see. In fact I can directly do `static_cast<DatePartType>` even without subclassing `uint32_t`. Is there any difference here? Also I thought `reinterpret_cast` is faster since it doesn't do type checking? (We already checked the validation of the value in `bind_node_visitor`)

pmenon: Neither `static_cast` nor `reinterpret_cast` perform any runtime checks. In general, `static_cast` is the "safe" cast, whereas `reinterpret_cast` allows casting between incompatible types. Since DateTimePart as a subclass of 32-bit ints, using `static_cast` is clear.

malin1993ml: I see. Thanks for the explanation!

src/binder/bind_node_visitor.cpp
malin1993ml: Actually I just tried and found out that it didn't work. The `OperatorId` was set after checking the name and input types of the function, which have to be set correctly for `Timestamp` functions here. (Basically we have to modify the `DatePartType` `Varchar` to `Integer` to pass the type checking to let the catalog find the `OperatorId` for the function.) So it seems to me has to be string comparison here.

malin1993ml:
@pmenon Thanks for the review and the comments. I fixed the casting but found the function name comparison has to be kept as explained.

src/binder/bind_node_visitor.cpp
apavlo: Can we put these in a hash table somewhere so that the will be easier to find instead of embedding them in the code like this? I suspect that there will be other functions that need to the same thing.

malin1993ml: Sounds reasonable. I can add a function in types.h called something like `FunctionNameToOperatorId()` to directly get the `OperatorId` through the name of a function.

apavlo: I feel like the mapping from function name to operator id should be in the catalog though.

malin1993ml: Actually while I was investigating how that should be done, I found there's already a mapping there for `BuiltInFunctions` used by the `Catalog`:
https://github.com/cmu-db/peloton/blob/master/src/catalog/catalog.cpp#L851
Assume we're only going to do this parameter specialization for built-in functions, this should be good. I can't really think of an example where we're going to manually infer the logic of a UDF and rewrite its parameters for now.

pmenon:
Spoke with @malin1993ml, we agreed that removing the optimization converting the string date component into a full `DatePartType` simplifies the code, and is functionally correct. We also verified that this is what Postgres does.

malin1993ml:
Just to note, the original optimization will have correctness issue when the user overloads the `date_trunc` function that takes in `(Integer, Timestamp)`.

pmenon:
@malin1993ml Did you get a chance to revert the optimizations we discussed?

malin1993ml:
@pmenon Doing it right now:)

malin1993ml:
@pmenon Done. I've also tested the new implementation with the cogen engine.

pmenon:
LGTM +1. Rebase and it should be good to merge.

PR #902, opened 2017-11-12
Add join cost estimations and using index-based sampling 

bechob:
- Using column_name (such as A.a) to search for column stats 
- Update join output table stats num_row 
- Update join output's column stats cardinality using [index-based sampling][1]
  - Add index map in TableStats
  - Add TupleSampler in TableStats
  - Add join sampling in TupleSampler
- Only support join type like A.a = B.b 
- Sampling only when join key has index

TODO:
- Check the correctness of cardinality estimation formula
- Fix HasPrimaryIndex method in TableStats 
- Support other join predicates (different comparisons, conjunctions etc.)

[1]: http://cidrdb.org/cidr2017/papers/p9-leis-cidr17.pdf


apavlo:
@bechob Why does the coverage go down so much?

bechob:
I think it's because the existing tests do not use ANALYZE firstly, so there's no tabla stats catalog available, which disables many functions that I wrote from being invoked. I'll add a test which with ANALYZE running in advance to test the added codes.

src/include/optimizer/stats/cost.h
```diff
@@ -137,6 +145,80 @@ class Cost {
   inline static double InnerHashJoin() { return DEFAULT_COST; }
 
   /*
+   * Join with Sampling
+   */
+  static double InnerNLJoinWithSampling(
```
allisonwang: Why not add those functions in InnerNLJoin / InnerHashJoin and use a flag to indicate if we want to use sampling? Then the cost computation can be shared and it will be much easier to switch the configuration for the system.

src/include/optimizer/stats/cost.h
```diff
@@ -137,6 +145,80 @@ class Cost {
...
+      const std::shared_ptr<TableStats>& left_input_stats,
+      const std::shared_ptr<TableStats>& right_input_stats,
+      std::shared_ptr<TableStats>& output_stats,
+      const std::shared_ptr<expression::AbstractExpression> predicate);
```
allisonwang: What if there are multiple join predicates?

src/include/optimizer/stats/cost.h
```diff
@@ -151,6 +233,10 @@ class Cost {
    */
   static size_t GetEstimatedGroupByRows(
       const std::shared_ptr<TableStats>& input_stats,
+      std::vector<std::string>& columns);
+
+  static size_t GetEstimatedGroupByRows(
```
allisonwang: This can be removed since we are using std::string to get column stats now.

src/include/optimizer/stats/table_stats.h
```diff
@@ -15,14 +15,17 @@
...
 #include "optimizer/stats.h"
+#include "optimizer/stats/tuple_sampler.h"
 #include "type/types.h"
 
 namespace peloton {
 namespace optimizer {
 
 #define DEFAULT_CARDINALITY 1
 #define DEFAULT_HAS_INDEX false
+#define DEFAULT_SAMPLE_NUM 100
```
allisonwang: DEFAULT_SAMPLE_SIZE would be more descriptive.

src/optimizer/cost_and_stats_calculator.cpp
```diff
@@ -30,10 +30,17 @@ static constexpr int RIGHT_CHILD_INDEX = 1;
...
+    const PropertyColumns *columns_prop,
+    storage::DataTable *data_table = nullptr) {
   std::vector<std::shared_ptr<ColumnStats>> output_column_stats;
   if (columns_prop->HasStarExpression()) {
     for (size_t i = 0; i < input_table_stats->GetColumnCount(); i++) {
+      auto column_stat = input_table_stats->GetColumnStats((oid_t)i);
+      if (data_table != nullptr) {
+        // add table_name for base table
+        column_stat->column_name =
+            data_table->GetName() + "." + column_stat->column_name;
```
allisonwang: This is hacky. Would be good to write a helper function for this.

src/optimizer/cost_and_stats_calculator.cpp
```diff
@@ -43,18 +50,90 @@ std::shared_ptr<TableStats> generateOutputStat(
...
       // TODO: Deal with or add column stats for complex expressions like a+b
       if (expr->GetExpressionType() == ExpressionType::VALUE_TUPLE) {
-        auto column_name =
-            (std::dynamic_pointer_cast<expression::TupleValueExpression>(expr))
-                ->GetColumnName();
+        auto tv_expr =
+            std::dynamic_pointer_cast<expression::TupleValueExpression>(expr);
+        auto column_name = tv_expr->GetColumnName();
+        if (data_table == nullptr) {
+          column_name = tv_expr->GetTableName() + "." + column_name;
```
allisonwang: Same point

src/optimizer/cost_and_stats_calculator.cpp
```diff
@@ -43,18 +50,90 @@ std::shared_ptr<TableStats> generateOutputStat(
...
+        }
         auto column_stat = input_table_stats->GetColumnStats(column_name);
         if (column_stat != nullptr) {
+          if (data_table != nullptr) {
+            // add table_name for base table
+            column_stat->column_name =
+                data_table->GetName() + "." + column_stat->column_name;
```
allisonwang: Same.

src/optimizer/cost_and_stats_calculator.cpp
```diff
@@ -43,18 +50,90 @@ std::shared_ptr<TableStats> generateOutputStat(
...
       input_table_stats->num_rows, output_column_stats);
+
+  // Set sampler
+  output_table_stats->SetTupleSampler(input_table_stats->GetSampler());
+
+  // Add indexes to index map for base table
+  if (data_table != nullptr) {
+    for (oid_t i = 0; i < data_table->GetIndexCount(); i++) {
+      auto index = data_table->GetIndex(i);
+      // If the index is for multiple columns, we don't add
```
allisonwang: Why?

src/optimizer/cost_and_stats_calculator.cpp
```diff
@@ -43,18 +50,90 @@ std::shared_ptr<TableStats> generateOutputStat(
...
+      if (index->GetMetadata()->GetKeyAttrs().size() > 1) {
+        continue;
+      }
+      auto column_id = index->GetMetadata()->GetKeyAttrs()[0];
+      // If the index column does not exist in the property column, we don't add
+      if (input_table_stats->GetColumnStats(column_id)->column_name.find(".") ==
```
allisonwang: Same. Would be good to write a separate function for this.

src/optimizer/cost_and_stats_calculator.cpp
bechob: because we only support predicate with single column now, so if the index is for multiple columns, it is not needed.

src/optimizer/cost_and_stats_calculator.cpp
bechob: Agree with you. I think another solution is to change the way we store the column name in the ColumnStats. Therefore, we don't need to to the transition every time. Which one do you prefer for now?

src/optimizer/cost_and_stats_calculator.cpp
```diff
@@ -43,18 +50,90 @@ std::shared_ptr<TableStats> generateOutputStat(
...
+}
+
+std::shared_ptr<TableStats> generateOutputStatFromTwoTable(
+    std::shared_ptr<TableStats> &left_table_stats,
+    std::shared_ptr<TableStats> &right_table_stats,
+    const PropertyColumns *columns_prop) {
+  std::vector<std::shared_ptr<ColumnStats>> output_column_stats;
+  if (columns_prop->HasStarExpression()) {
+    for (size_t i = 0; i < left_table_stats->GetColumnCount(); i++) {
+      auto column_stat = left_table_stats->GetColumnStats((oid_t)i);
```
allisonwang: If you use oid instead of string here to get columns, then it wouldn't be correct if the table is not base table.

src/include/optimizer/stats/cost.h
bechob: Multiple joins can also be represented in the format of AbstractExpression connected with conjunction expressions. So there'll be only one join predicate.

src/optimizer/cost_and_stats_calculator.cpp
```diff
@@ -43,18 +50,90 @@ std::shared_ptr<TableStats> generateOutputStat(
...
+  } else {
+    for (size_t i = 0; i < columns_prop->GetSize(); i++) {
+      auto expr = columns_prop->GetColumn(i);
+
+      // TODO: Deal with or add column stats for complex expressions like a+b
+      if (expr->GetExpressionType() == ExpressionType::VALUE_TUPLE) {
+        auto tv_expr =
+            std::dynamic_pointer_cast<expression::TupleValueExpression>(expr);
+        auto column_name =
+            tv_expr->GetTableName() + "." + tv_expr->GetColumnName();
```
allisonwang: Same point

src/optimizer/cost_and_stats_calculator.cpp
bechob: what I do here is to add all the column stats to the output stats, so the column_id does not matter. Sorry for my misleading codes here

src/include/optimizer/stats/cost.h
bechob: I think it's a great idea. I'll do that.

src/optimizer/cost_and_stats_calculator.cpp
```diff
@@ -347,20 +428,168 @@ void CostAndStatsCalculator::Visit(const PhysicalLimit *) {
   output_stats_ = output_stats;
 }
 void CostAndStatsCalculator::Visit(const PhysicalFilter *){};
-void CostAndStatsCalculator::Visit(const PhysicalInnerNLJoin *) {
+void CostAndStatsCalculator::Visit(const PhysicalInnerNLJoin *op) {
   // TODO: Replace with more accurate cost
-  output_cost_ = 1;
+  PL_ASSERT(child_stats_.size() == 2);
```
allisonwang: Magic number

src/optimizer/cost_and_stats_calculator.cpp
```diff
@@ -347,20 +428,168 @@ void CostAndStatsCalculator::Visit(const PhysicalLimit *) {
...
+  auto predicate = op->join_predicate;
+  output_cost_ += Cost::InnerNLJoinWithSampling(
+      left_table_stats, right_table_stats, output_stats, predicate);
+  output_stats_ = output_stats;
 };
-void CostAndStatsCalculator::Visit(const PhysicalLeftNLJoin *){};
-void CostAndStatsCalculator::Visit(const PhysicalRightNLJoin *){};
-void CostAndStatsCalculator::Visit(const PhysicalOuterNLJoin *){};
-void CostAndStatsCalculator::Visit(const PhysicalInnerHashJoin *) {
+void CostAndStatsCalculator::Visit(const PhysicalLeftNLJoin *op) {
```
allisonwang: We have lots of repetitive code here. Please refactor to use a helper function.

src/optimizer/cost_and_stats_calculator.cpp
```diff
@@ -392,18 +621,19 @@ void CostAndStatsCalculator::Visit(const PhysicalHashGroupBy *op) {
...
     // TODO: Deal with complex expressions like a+b
     if (column->GetExpressionType() == ExpressionType::VALUE_TUPLE) {
-      oid_t column_id = (oid_t)(
-          std::dynamic_pointer_cast<expression::TupleValueExpression>(column)
-              ->GetColumnId());
-      column_ids.push_back(column_id);
+      auto tv_expr =
+          std::dynamic_pointer_cast<expression::TupleValueExpression>(column);
+      std::string column_name =
+          tv_expr->GetTableName() + "." + tv_expr->GetColumnName();
```
allisonwang: Refactor.

src/optimizer/stats/cost.cpp
```diff
@@ -84,6 +85,44 @@ void Cost::CombineConjunctionStats(const std::shared_ptr<TableStats> &lhs,
...
 //===----------------------------------------------------------------------===//
+
+double Cost::SortGroupByCost(const std::shared_ptr<TableStats> &input_stats,
+                             std::vector<std::string> columns,
+                             std::shared_ptr<TableStats> &output_stats) {
+  PL_ASSERT(input_stats);
+  PL_ASSERT(columns.size() > 0);
+
+  //    if (output_stats != nullptr) {
+  if (false) {
```
allisonwang: if (false)?

src/optimizer/stats/cost.cpp
```diff
@@ -202,8 +241,303 @@ double Cost::OrderByCost(const std::shared_ptr<TableStats> &input_stats,
...
+// INNER HASH JOIN
+//===----------------------------------------------------------------------===//
+double Cost::InnerHashJoinWithSampling(
+    const std::shared_ptr<TableStats> &left_input_stats,
+    const std::shared_ptr<TableStats> &right_input_stats,
+    std::shared_ptr<TableStats> &output_stats,
+    const std::shared_ptr<expression::AbstractExpression> predicate) {
+  UpdateJoinOutputSize(left_input_stats, right_input_stats, output_stats,
+                       predicate, 0);
+  return (left_input_stats->num_rows + right_input_stats->num_rows) *
```
allisonwang: Hash join cost doesn't seem right. There are costs of building a hash table (left table or right table depending on the join type) and performing the actual join.

src/optimizer/stats/cost.cpp
```diff
@@ -202,8 +241,303 @@ double Cost::OrderByCost(const std::shared_ptr<TableStats> &input_stats,
...
+      left_input_stats, right_input_stats, output_stats, predicate,
+      left_input_stats->num_rows + right_input_stats->num_rows);
+  return (left_input_stats->num_rows + right_input_stats->num_rows) *
+         DEFAULT_TUPLE_COST;
+}
+
+//===----------------------------------------------------------------------===//
 // Helper functions
 //===----------------------------------------------------------------------===//
+void Cost::UpdateColumnStatsWithSampling(
```
allisonwang: Please add some high-level comments describing this function.

src/optimizer/stats/cost.cpp
```diff
@@ -202,8 +241,303 @@ double Cost::OrderByCost(const std::shared_ptr<TableStats> &input_stats,
...
+    const std::shared_ptr<TableStats> &right_input_stats,
+    std::shared_ptr<TableStats> &output_stats,
+    const std::shared_ptr<expression::AbstractExpression> predicate,
+    size_t adjustment) {
+  size_t default_join_size =
+      left_input_stats->num_rows * right_input_stats->num_rows + adjustment;
+  if (predicate == nullptr) {
+    output_stats->num_rows = default_join_size;
+  } else if (predicate->GetExpressionType() == ExpressionType::COMPARE_EQUAL) {
+    // consider only A.a = B.a case here
```
allisonwang: Add this to the spec for this function.

src/optimizer/stats/cost.cpp
```diff
@@ -202,8 +241,303 @@ double Cost::OrderByCost(const std::shared_ptr<TableStats> &input_stats,
...
+      return;
+    }
+    auto left_child =
+        reinterpret_cast<const expression::TupleValueExpression *>(
+            predicate->GetChild(0));
+    auto right_child =
+        reinterpret_cast<const expression::TupleValueExpression *>(
+            predicate->GetChild(1));
+    std::string left_column_name =
+        left_child->GetTableName() + "." + left_child->GetColumnName();
```
allisonwang: Refactor

src/optimizer/stats/cost.cpp
```diff
@@ -216,14 +550,31 @@ void Cost::UpdateConditionStats(const std::shared_ptr<TableStats> &input_stats,
 
 size_t Cost::GetEstimatedGroupByRows(
     const std::shared_ptr<TableStats> &input_stats,
+    std::vector<std::string> &columns) {
+  // Idea is to assume each column is uniformaly network and get an
```
allisonwang: This might be my original naive implement. I will change this.

apavlo:
@allisonwang Nice job on the review!

chenboy:
@apavlo This is ready for review. Basically, we'll collect stats on the fly and cache stats for each group during optimization. This PR does not utilize index-based sampling in the optimizer yet. According to the paper, we may want to implement the sampling in a separate dynamic programming phase before the optimization to derive stats. We can talk about this in the optimizer meeting this week.

apavlo:
@chenboy I forgot whether we said yesterday that this was ready to merge?

chenboy:
@apavlo Yes. We can merge this now and I'll figure out how to test the cost model later

PR #908, opened 2017-11-14
Fix for 'select * ' hanging

dengyuchenkit:
Fix for #878

### DIFF
--- a/src/network/network_connection.cpp
+++ b/src/network/network_connection.cpp
@@ -141,6 +141,10 @@ WriteState NetworkConnection::WritePackets() {
   if (protocol_handler_->force_flush == true) {
     return FlushWriteBuffer();
   }
+
+  // we have flushed, disable force flush now
+  protocol_handler_->force_flush = false;
+
   return WriteState::WRITE_COMPLETE;
 }
 
@@ -335,9 +339,6 @@ WriteState NetworkConnection::FlushWriteBuffer() {
 
   // buffer is empty
   wbuf_.Reset();
-
-  // we have flushed, disable force flush now
-  protocol_handler_->force_flush = false;
 
   // we are ok
   return WriteState::WRITE_COMPLETE;
--- a/test/network/select_all_test.cpp
+++ b/test/network/select_all_test.cpp
@@ -1,0 +1,110 @@
+//===----------------------------------------------------------------------===//
+//
+//                         Peloton
+//
+// select_all_test.cpp
+//
+// Identification: test/network/select_all_test.cpp
+//
+// Copyright (c) 2016-17, Carnegie Mellon University Database Group
+//
+//===----------------------------------------------------------------------===//
+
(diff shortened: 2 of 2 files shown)


seojungmin:
@dengyuchenkit This PR flushes the network packet whenever there is a `WritePackets()` call. Could you provide more information or comment about how buffering/flushing network packets work?

hzxa21:
@dengyuchenkit Any updates for this PR?

dengyuchenkit:
@seojungmin Oh I'm sorry. When the execution of each query is finished, the connection state will transfer to WriteState. In the write state, it calles WritePackets(https://github.com/cmu-db/peloton/blob/master/src/network/network_connection.cpp#L768). This function is mainly copying the response from the response buffer in ProtocolHandler(https://github.com/cmu-db/peloton/blob/master/src/include/network/protocol_handler.h#L51) to the buffer in NetworkConnection (packet header: https://github.com/cmu-db/peloton/blob/master/src/network/network_connection.cpp#L523. packet body: https://github.com/cmu-db/peloton/blob/master/src/network/network_connection.cpp#L552). 

But when the response size is larger than the maximum buffer size in NetworkConnection, it will flush out the content by calling FlushWriteBuffer to make room for the rest of the response. Besides, the networking layer will flush the write buffer when the force_flush flag is set to true. Usually, the force_flush flag will be set to true when the user need to get the response before sending out its next request. 

For simple query protocol (https://github.com/cmu-db/peloton/blob/master/src/network/postgres_protocol_handler.cpp#L1097), since each request needs a response, the networking layer will flush the buffer after finishing each query. For extended query protocol, there is a SYNC command by which the user requests collect the result of the previous queries (https://github.com/cmu-db/peloton/blob/master/src/network/postgres_protocol_handler.cpp#L1119). So the server will flush the result when receving a SYNC commands. User can send multiple queries between two SYNC commands. In this case, when one query is finished executing, the NetworkConnection will go to WriteState to only copy the reponse in ProtocolHandler to the local buffer. It won't flush it out until the server finally receive a SYNC or the buffer size is not enough for current response.

Currently, the result of the query execution is initially stored in the ProtocolHandler buffer as different OutputPackets. (https://github.com/cmu-db/peloton/blob/master/src/network/postgres_protocol_handler.cpp#L142). The networking layer will do an extra copy to its own byte buffer when sending it out on the network.

@hzxa21 This PR is completed but it seems it can't finish the check. I will check it.

dengyuchenkit:
@seojungmin Previously when server calls FlushWriteBuffer(), the force_flush will be reset to false inside the function. The problem was that when there is a large response, the server will call FlushWriteBuffer() and the force_flush becomes false. Then when it comes to the last part of the response, if its size is small and the force_flush is set to false, it will not trigger the flushing. This part will not be sent out.

dengyuchenkit:
Close this PR since there is a new one now.

PR #911, opened 2017-11-15
Fix varchar length difference

### LINKED ISSUE #897: Varchar length is incorrectly set
The length of the varchar type is set incorrectly to its actual length + 1 (for some reason?). The root of this problem is:
https://github.com/cmu-db/peloton/blob/02670dab2f08bd002e3958a5a122887874274148/src/type/value.cpp#L331

Once this problem is fixed, we should also fix:
https://github.com/cmu-db/peloton/blob/02670dab2f08bd002e3958a5a122887874274148/src/type/varlen_type.cpp#L154
https://github.com/cmu-db/peloton/blob/61aaf5450f0b90486cd2d2e69e60555c149c4316/src/codegen/values_runtime.cpp#L74

allisonwang:
This PR fixed #897 and add varchar sql tests. After tracing through the execution, I found the inconsistency comes from incorrectly setting the varchar length in `constant_translator`. However, it's not a best practice to +1 or -1 in the codebase. We should have a unified way of representing varchar data and length in the future.  

The equality checking should be fixed now. 

```
postgres=# CREATE TABLE foo (
postgres(# name varchar(255)
postgres(# );
CREATE TABLE
postgres=# INSERT INTO foo VALUES ('Alice');
INSERT 0 1
postgres=# SELECT * FROM foo WHERE name = 'Alice';
 name
-------
 Alice
(1 row)
```


pmenon:
Do you need a separate test file for this, or can you modify an existing test?

seojungmin:
@allisonwang can we change `ValueRuntime::OutputVarchar()` in `codegen/values_runtime.cpp` to call `type::ValueFactory::GetVarcharValue(str, len, false);` instead of `type::ValueFactory::GetVarcharValue(str, false);`?  With your change in this PR, we can now have OutputVarchar() to be corrected, which was previously a kind of temporary fix.

@hzxa21 and @luochenUmich had an issue while they were producing an output from their built-in function implementation, and I think this additional fix will solve their problem.

allisonwang:
@seojungmin Good catch. `type::ValueFactory::GetVarcharValue(str, len, false)` shoudn't be used else where other than in ValueFactory.

allisonwang:
> In the latter case, the variable 'len' does not include a null terminator at the end of the array, while it does in the former. 

I am thinking where will this function (OutputVarchar) be called so that length there won't include a null terminator?

allisonwang:
@pmenon There is no existing SQL test that covers Varchar specifically under `test/sql`. Should I add those tests in `optimizer_sql_test`?

seojungmin:
@allisonwang You are correct `type::ValueFactory::GetVarcharValue(str, len, false)` shouldn't be called if we are explicitly using a null-terminator all throughout Peloton. However, what @hzxa21, @luochenUmich were trying to do was to hand over (or return) a partial `VARCHAR` value within a pre-allocated `VARCHAR`. Since they could not allocate a new `VARCHAR` space inside the function yet, they tried to reuse the area, but have a `VARCHAR` value out of the area.   This would be a kind of (understandable) special case.

hzxa21:
@seojungmin Thanks for the clarification.
@allisonwang Just as @seojungmin said, in order to support `substr()` without allocating memory in LLVM, we set the length of the `VARCHAR` type explicitly. So we need to ensure `type::ValueFactory::GetVarcharValue(str, len, false)` is called while outputting the value.

seojungmin:
@allisonwang I think you could combine with `typelimit_sql_test.cpp`, and then name the file `type_sql_test.cpp`.  `optimizer_sql_test.cpp` is huge and it is hard to debug later if we put more into it.

Change what I requested, call `type::ValueFactory::GetVarcharValue(str, len, false)` in `ValueRuntime::OutputVarchar()` in `codegen/values_runtime.cpp`. If it passes, then it would be good to have a final review before merge.

allisonwang:
@seojungmin Gotcha. Will combine the tests into type sql test and change the value runtime.

seojungmin:
LGTM +1

PR #912, opened 2017-11-16
[WIP] Morsel-Based Parallel LLVM Checkpoint: TableScan

phisiart:
This work is in progress. I want to get the design out early so that people can check whether this is a reasonable design.

Currently I've implemented the supporting infrastructure for the parallel LLVM engine.

- `codegen::ExecutorThreadPool`: Wrapper around common/thread_pool.h.

  Each task must have the function prototype:
  ```C
  // runtime_state: per-query, i.e. shared among all tasks.
  // task_info: per-task.
  void task(char *runtime_state, TaskInfo *task_info);
  ```

- `CountDown`: After submitting tasks, the main thread waits for a counter to get to 0. Each task decreases the counter by 1 on finish. This is similar to last semester's `Barrier`, but uses a condition variable instead of polling so that we don't waste cycles in the master thread waiting for tasks to complete.

I have written 3 test cases to show how to use this infrastructure.

- Test Case 1: From C++, use `codegen::ExecutorThreadPool` to submit tasks and wait for all to finish.

- Test Case 2: Use LLVM to generate the task function. From C++, submit tasks to `codegen::ExecutorThreadPool`.

- Test Case 3: Use LLVM to generate both the task function and the code to submit tasks. The parallel LLVM engine will generate code that look exactly like this.

## Design of Parallel LLVM Engine

The main idea is to generate LLVM IR that submits tasks to a thread pool.
Each task decreases a counter on finish. The main thread waits until the counter reaches 0.

Each task is responsible for a portion of the table(s) or intermediate table(s). We only exploit in-query parallelism by splitting the table(s).

Hash join and other plans that introduce pipeline breakers are tricky to deal with. The key is that **we only generate the task submission part of the code alongside table or intermediate table traversal**. This means we only need to change the structure of the generated code in `TableScan.Produce` and the second half of `Aggregate.Produce`. The reason is that `#(pipeline breakers) = #(traversals) - 1 = #(stages) - 1`.

In this way we are not committed to a certain number of tasks in the beginning of the query. We can have different numbers of tasks for different stages.

One key observation is that every time when `Produce` gets called, we are in the context of the main plan function; every time when `Consume` gets called, we are in the context of a task function.

Let's show the design through examples.

For simplicity, suppose our DB supports these kinds of plans:
```scala
Plan ::= TableScan(table : Table)
       | Filter(child : Plan, pred : Expr)
       | HashJoin(lhs : Plan, rhs : Plan, key : Attrib)
       | Count(child : Plan, key : Attrib)
```

### Example 1

AST:
```scala
Filter(
  child = TableScan(table1),
  pred  = Less(Attrib('x'), Const(1)) // tuple.x < 1
)
```

Single Threaded:
```scala
// Filter.Produce {
  // TableScan.Produce {
    for t in table1:
      // Filter.Consume {
        if t.x < 1:
          // Consume {
            output(t)
          // } Consume
      // } Filter.Consume
  // } TableScan.Produce
// } Filter.Produce
```

Multithreaded:
```scala
// Filter.Produce {
  // TableScan.Produce {
    c = CountDown()
    for tid in range(ntasks(table1)): // We can determine how to split the table at runtime.
      pool.submit({ // Note that now we are in the task function.
        for t in table1.range(tid):
          // Filter.Consume {
            if t.x < 1:
              // Consume {
                output(t)
              // } Consume
          // } Filter.Consume
        c.Decrease()
      })
  // } TableScan.Produce
  c.Wait()
// } Filter.Produce
```

### Example 2

AST:
```scala
HashJoin(
  lhs = Filter(
    child = TableScan(table1)
    pred  = Less(Attrib('x'), Const(1))
  ),
  rhs = TableScan(table2),
  key = "key"
)
```

Single Threaded:
```scala
// HashJoin.Produce {
  h = HashTable()

  // Filter.Produce {
    // TableScan1.Produce {
      for t1 in table1:
        // Filter.Consume {
          if t1.x < 1:
          // HashJoin.ConsumeLeft {
            h.insert(t1.key, t1)
          // } HashJoin.ConsumeLeft
        // } Filter.Consume
    // } TableScan1.Produce
  // } Filter.Produce

  // TableScan2.Produce {
    for t2 in table2:
    // HashJoin.ConsumeRight {
      if (t1 = h.probe(t2.key)) != null:
        t = combine(t1, t2)
          // Consume {
            output(t)
          // } Consume
    // } HashJoinConsumeRight
  // } TableScan2.Produce

// } HashJoinProduce
```

Multithreaded:
```scala
// HashJoin.Produce {
  h = HashTable()

  // Filter.Produce {
    // TableScan1.Produce {
      c1 = CountDown()
      for tid in range(ntasks(table1)):
        pool.submit({
          for t1 in table1.range(tid):
            // Filter.Consume {
              if t1.x < 1:
                // HashJoin.ConsumeLeft {
                  h.insert(t1.key, t1)
                // } HashJoin.ConsumeLeft
            // } Filter.Consume
        c.Decrease()
      })
      c1.Wait()
    // } TableScan1.Produce
  // } Filter.Produce

  // TableScan2.Produce {
    c2 = CountDown()
    for tid in range(ntasks(table2)):
      pool.submit({
        for t2 in table2.range(tid):
          // HashJoin.ConsumeRight {
            if (t1 = h.probe(t2.key)) != null:
              t = combine(t1, t2)
              // Consume {
                output(t)
              // } Consume
          // } HashJoin.ConsumeRight
    })
    c2.Wait()
  // } TableScan2.Produce

// } HashJoin.Produce
```

### Example 3

AST:

```scala
Count(
  child = TableScan(table1)
  key = "key"
)
```

Single Threaded:

```scala
// Count.Produce {
  h = HashTable()

  // TableScan.Produce {
  for t in table1:
    // Count.Consume {
      h[t.key] += 1
    // } Count.Consume

  for entry in h:
    // Consume {
      output(entry)
    // } Consume

// } Count.Produce
```

Multithreaded:

```scala
// Count.Produce {
  h = HashTable()

  // TableScan.Produce {
    c1 = CountDown()
    for tid in range(ntasks(table1)):
      pool.submit({
        for t1 in table1.range(tid):
          // Count.Consume {
            h[t.key] += 1
          // } Count.Consume
        c1.Decrease()
      })
    c1.Wait()
  // } TableScan.Produce

  c2 = CountDown()
  for tid in range(ntasks(h)):
    pool.submit({
      for entry in h.range(tid):
        // Consume {
          output(entry)
        // } Consume
      c2.Decrease()
    })

// } Count.Produce
```

### DIFF
--- a/src/codegen/buffering_consumer.cpp
+++ b/src/codegen/buffering_consumer.cpp
@@ -65,9 +65,9 @@ BufferingConsumer::BufferingConsumer(const std::vector<oid_t> &cols,
 // Append the array of values (i.e., a tuple) into the consumer's buffer of
 // output tuples.
 void BufferingConsumer::BufferTuple(char *state, char *tuple,
-                                    uint32_t num_cols) {
+                                    int32_t task_id, uint32_t num_cols) {
   BufferingState *buffer_state = reinterpret_cast<BufferingState *>(state);
-  buffer_state->output->at(0).emplace_back(
+  buffer_state->output->at(task_id).emplace_back(
       reinterpret_cast<peloton::type::Value *>(tuple), num_cols);
 }
 
@@ -137,9 +137,13 @@ void BufferingConsumer::ConsumeResult(ConsumerContext &ctx,
     codegen.CallFunc(output_func, args);
   }
 
+  llvm::Value *task_id = ctx.GetTaskId();
+  codegen.CallPrintf("[BufferingConsumer] My task id = %d\n", {task_id});
+
   // Append the tuple to the output buffer (by calling BufferTuple(...))
   auto *consumer_state = GetStateValue(ctx, consumer_state_id_);
   std::vector<llvm::Value *> args = {consumer_state, tuple_buffer_,
+                                     task_id,
                                      codegen.Const32(output_ais_.size())};
   codegen.Call(BufferingConsumerProxy::BufferTuple, args);
 }
--- a/src/codegen/buffering_consumer.cpp
+++ b/src/codegen/buffering_consumer.cpp
@@ -17,6 +17,7 @@
 #include "codegen/proxy/value_proxy.h"
 #include "codegen/proxy/values_runtime_proxy.h"
 #include "codegen/type/sql_type.h"
+#include "common/logger.h"
 #include "planner/binding_context.h"
 
 namespace peloton {
@@ -53,11 +54,12 @@ DEFINE_METHOD(peloton::codegen, BufferingConsumer, BufferTuple);
 //===----------------------------------------------------------------------===//
 
 BufferingConsumer::BufferingConsumer(const std::vector<oid_t> &cols,
-                                     planner::BindingContext &context) {
+                                     planner::BindingContext &context)
+    : state(&output_chunks_) {
+  output_chunks_.emplace_back();
   for (oid_t col_id : cols) {
     output_ais_.push_back(context.Find(col_id));
   }
-  state.output = &tuples_;
 }
 
 // Append the array of values (i.e., a tuple) into the consumer's buffer of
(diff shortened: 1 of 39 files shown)


seojungmin:
I briefly looked at it.  If the merge overhead for the scan is high, we could just have a vector of tuple vectors as an output, and handle this outside.

PR #913, opened 2017-11-19
Fix for 'select * ' hanging

dengyuchenkit:
It's the same PR as #908 

Fix for #878

Previous PR doesn't run the check. Just to check if it can run check after changing the source repository.


dengyuchenkit:
@hzxa21 This PR should be ready to merge now.

dengyuchenkit:
@hzxa21 Sorry, I will add setter function instead of directly accessing the field.

hzxa21:
@ChTimTsubasa Can you do a quick review for this PR. I will merge it afterward.

Issue #859, opened 2017-10-28
Fix the access pattern for catalog objects


malin1993ml:
Currently catalog objects are immutable snapshots of the database catalog tables. They way to access them is to directly access the public const member variables of these objects:
https://github.com/cmu-db/peloton/blob/master/src/include/catalog/index_catalog.h#L45

However, this is not safe when we're going to support virtual catalog and "what-if" optimizer API. The correct way to do this is to make them private members, and only provide public Getters for those immutable fields. For the fields that do need to be mutable, we should provide protected/private Setters and set the classes who need those Setters friend classes.

So what need to fix now is:
(1) Change all the direct accesses to the fields in catalog objects to public Getters.
(2) Change all the public const fields in catalog objects to mutable private fields.

The refactor tool in Eclipse may be able to help with this process.

camellyx:
@malin1993ml We thought about this during the design process, since you mentioned it earlier. For "what-if" optimizer, do you need to create fake indexes or change the property of an existing index?

malin1993ml:
Yes. We need both of them.

malin1993ml:
And actually remove existing indexes as well.

malin1993ml:
Fixed in #919.

PR #919, opened 2017-11-23
Fixing access pattern for catalog objects for Issue #859

camellyx:
Issue #859: Better support for "what-if" optimizer API in catalog objects.

@malin1993ml Sorry for the delay. I was busy with something else earlier. Let me know if you need anything else.

### DIFF
--- a/src/include/catalog/column_catalog.h
+++ b/src/include/catalog/column_catalog.h
@@ -42,14 +42,14 @@ class ColumnCatalogObject {
  public:
   ColumnCatalogObject(executor::LogicalTile *tile, int tupleId = 0);
 
-  oid_t GetTableOid() { return table_oid; }
-  std::string GetColumnName() { return column_name; }
-  oid_t GetColumnId() { return column_id; }
-  oid_t GetColumnOffset() { return column_offset; }
-  type::TypeId GetColumnType() { return column_type; }
-  bool IsInlined() { return is_inlined; }
-  bool IsPrimary() { return is_primary; }
-  bool IsNotNull() { return is_not_null; }
+  inline oid_t GetTableOid() { return table_oid; }
+  inline const std::string &GetColumnName() { return column_name; }
+  inline oid_t GetColumnId() { return column_id; }
+  inline oid_t GetColumnOffset() { return column_offset; }
+  inline type::TypeId GetColumnType() { return column_type; }
+  inline bool IsInlined() { return is_inlined; }
+  inline bool IsPrimary() { return is_primary; }
+  inline bool IsNotNull() { return is_not_null; }
 
  private:
   // member variables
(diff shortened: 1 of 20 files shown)


src/include/catalog/column_catalog.h
```diff
@@ -42,6 +42,16 @@ class ColumnCatalogObject {
  public:
   ColumnCatalogObject(executor::LogicalTile *tile, int tupleId = 0);
 
+  oid_t GetTableOid() { return table_oid; }
+  std::string GetColumnName() { return column_name; }
```
malin1993ml: You can change this to const reference to avoid copying and memory allocation.

src/include/catalog/database_catalog.h
```diff
@@ -59,11 +59,14 @@ class DatabaseCatalogObject {
   std::unordered_map<oid_t, std::shared_ptr<TableCatalogObject>>
   GetTableObjects(bool cached_only = false);
 
+  oid_t GetDatabaseOid() { return database_oid; }
+  std::string GetDatabaseName() { return database_name; }
```
malin1993ml: Const reference.

src/include/catalog/index_catalog.h
```diff
@@ -45,13 +45,22 @@ class IndexCatalogObject {
  public:
   IndexCatalogObject(executor::LogicalTile *tile, int tupleId = 0);
 
+  inline oid_t GetIndexOid() { return index_oid; }
+  inline std::string GetIndexName() { return index_name; }
```
malin1993ml: Const reference.

src/include/catalog/table_catalog.h
```diff
@@ -71,12 +71,16 @@ class TableCatalogObject {
   std::shared_ptr<ColumnCatalogObject> GetColumnObject(
       const std::string &column_name, bool cached_only = false);
 
+  oid_t GetTableOid() { return table_oid; }
+  std::string GetTableName() { return table_name; }
```
malin1993ml: Const reference.

camellyx:
@malin1993ml Done. I don't usually use IDE like eclipse and clion for programming, they are too slow for me. Vim and Sublime rocks. :-)

malin1993ml:
Got it. Cool!

malin1993ml:
@camellyx The build on OSX failed after merging with the master:
https://travis-ci.org/cmu-db/peloton/builds/306943052

It passed all the Linux builds, and I know that it passed all the tests in your branch as well. Do you have a Mac machine? Can you try to reproduce the failing test case and see what's going on?

camellyx:
@malin1993ml I don't think this bug has anything to do with us... It runs ok on Leon's MacBook.
It looks like a potential data race in backend_stats_context.cpp:71. I think inserting and accessing table_metrics_ might be unsafe.

I suggest rerunning travis and the problem might go away...

malin1993ml:
Okay I see. Thanks!

pmenon:
@camellyx @malin1993ml If there is a data race or bug, please create an issue so we can track it.

malin1993ml:
I don't know what it exactly is. I first saw that as well and it only happened on Mac OS. But I'll create an issue to keep track of it.

PR #920, opened 2017-11-25
We need devel packages too in Fedora

pranavk:
And I guess same is the case with the RedHat.


pranavk:
Abandoning this for #921

PR #921, opened 2017-11-25
Build fixes for Fedora/GCC 7

pranavk:
Bunch of issues I faced when making it to compile on Fedora


pranavk:
Apart from these, I have reported a bug[1] upstream that makes it difficult for paleton or any other software that require older version of LLVM to be compiled on Fedora without manually changing the LLVMConfig.cmake file.

[1] https://bugzilla.redhat.com/show_bug.cgi?id=1517478

pranavk:
I see. So the build failures are because my 'Werror-nonnull-compare' commit uncovered undefined behavior in some of the code -- calling member function of Property class with a nullptr. I will fix them. That should fix the build failures as well.

pranavk:
Okay, so lot of fedora related changes since I submitted this PR. I will fix the conflict

apavlo:
@crd477 Can we have this PR run on the new Docker tests since that has Fedora?

crd477:
On 2017-11-30 20:35, Andy Pavlo wrote:
> @crd477 <https://github.com/crd477> Can we have this PR run on the new 
> Docker tests since that has Fedora?
> 

Yes, it will be no problem but I need to figure out that aspect of 
Jenkins configuration for pull requests that we discussed in the other 
email thread yesterday.

-- 
     -Chad

pranavk:
I updated by system to fedora 27 today (which comes with GCC 7). Found a bunch of issues that prevented peloton to compile with GCC 7. Above patches takes care of that. Peloton now runs and passes all tests on my system.

pranavk:
Other than that, I added a file util/global.h above with copyright header of CMU DB group itself. Let me know if it needs to changed, etc. since I am not a CMU member.

script/installation/packages.sh
```diff
@@ -78,7 +78,8 @@ elif [[ "$DISTRO" == *"FEDORA"* ]]; then
         llvm-devel \
         llvm-static \
         libedit-devel \
-        postgresql
+        postgresql \
+        libatomic
```
pranavk: Since this is due to GCC 7 that calls are routed to libatomic, I guess we need to install libatomic on  Ubuntu 17.10 as well which has GCC 7. I don't have any Ubuntu installation around and from a quick search couldn't find the name of the package that provides libatomic.so there.

But worth checking if it's *really* required on Ubuntu 17.10. Maybe Ubuntu packagers did some magic not requiring to link against libatomic at all :)

pranavk:
@pervazea Your comments seems code specific to different patches in this pull requests and it's not clear to me to what part of the code you are commenting against. Your review comments are not attached to any piece of code (at least how it is from here). Can you please give me some hints to what patch or piece of code you are referring to in my comments?

pervazea:
Sorry. Still learning how to use the tools. I'll clarify here for now.


Comment 1 relates to this, in CMakelists.txt. I assume you'd have done this only for third party code if it was easy... how hard is it?

-Wno-implicit-fallthrough for now
Even after we make *our* code behave properly, there is some third_party
code that emits several warnings. Ideal would be to
-Wno-implicit-fallthrough only for third_party code.


Other comment was in relation to allocator.cpp, and the change to use noexcept. What are the semantics of noexcept? Is there a behavior change?

-void *operator new(size_t size, const std::nothrow_t &) throw() {
 +void *operator new(size_t size, const std::nothrow_t &) noexcept {

pranavk:
> Comment 1 relates to this, in CMakelists.txt. I assume you'd have done this only for third party code if it was easy... how hard is it?

I didn't look much into it; so, I don't know how hard it is. I had some concerns regarding the state of third party code while I was working on it and filed #957 . I was expecting some input on that and depending on that I would have looked into this to save some extra work here. So, as a workaround for now, I have disabled the -Wimplicit-fallthrough for whole of the code so that it atleast compiles on systems with GCC7. It has no effect on systems with older compilers; they don't understand -Wimplicit-fallthrough anyways.

> Other comment was in relation to allocator.cpp, and the change to use noexcept. What are the semantics of noexcept? Is there a behavior change?

There's no behavior change. throw() just means that no exception would be thrown which is the case but it's deprecated. It has been replaced by `noexcept` since c++11 to mean the same. And since we are already post-c++11, there's no point in keeping the old syntax. You add `noexcept` specifier when you are sure that function cannot throw any exception. You can read the noexcept documentation here[1]

[1] http://en.cppreference.com/w/cpp/language/noexcept

Issue #917, opened 2017-11-21
Proposal: Change `codegen::lang::If` to hide PHI and force scoping.


phisiart:
I discussed with Tony about this. We think we should post an issue here for further discussion.

## TLDR

Provide a different `codegen::lang::If` that
- Doesn't require the user to explicit construct PHI nodes to merge values.
- Forces correct scoping.

The interface should look like this:
```cpp
llvm::Value *codegen::lang::If2(
    CodeGen &cg,
    llvm::Value *condition,
    const std::function<llvm::Value *()> &cgen_then_branch,
    const std::function<llvm::Value *()> &cgen_else_branch);
```

## Background

Since LLVM uses the SSA form, we must explicitly use a phi node to merge 2 different versions of a variable. For example, instead of writing

```python
if condition:
  printf("then branch")
  a = 1
else:
  printf("else branch")
  a = 2
```

We have to write

```python
if condition:
  printf("then branch")
  a1 = 1
else:
  printf("else branch")
  a2 = 2
a = merge(a1, a2) # This is the PHI node in LLVM
```

More specifically, to actually generate the code above, we need to use `codegen::lang::If` in this way:

```cpp
llvm::Value *condition = ...;
codegen::lang::If cond{cg, condition};
cg.CallPrintf("then branch", {});
llvm::Value *a1 = cg.Const32(1);
cond.ElseBlock();
cg.CallPrintf("else branch", {});
llvm::Value *a2 = cg.Const32(2);
cond.EndIf();
llvm::Value *a = cond.BuildPHI(a1, a2);
```

The above looks a bit messy, so we usually add scoping like this:

```cpp
llvm::Value *condition = ...;
llvm::Value *a1;
llvm::Value *a2;
codegen::lang::If cond{cg, condition};
{
  cg.CallPrintf("then branch", {});
  a1 = cg.Const32(1);
}
cond.ElseBlock();
{
  codegen::CallPrintf("else branch");
  a2 = cg.Const32(2);
}
cond.EndIf();
llvm::Value *a = cond.BuildPHI(a1, a2);
```

The problems of the current `codegen::lang::If` are:
- The user must explictly call BuildPHI, which can be confusing for beginners.
- a1 and a2 are explosed outside of the if-statement.

For example, the user might (mistakenly) directly use `a2`:

```cpp
llvm::Value *condition = ...;
llvm::Value *a1;
llvm::Value *a2;
codegen::lang::If cond{cg, condition};
{ ... }
[...]

PR #924, opened 2017-11-27
Add a new usage for `codegen::clang::If`

MalcolmUnWang:
As Issue #917 has posted,  a different usage of `If` may be needed. This PR provides a new method in `codegen::lang::If` which will make it easier for users to call `If` and reduces confusion. The new usage will be like this.
```c++
llvm::Value *condition = ...;
llvm::Value *a = codegen::lang::If::If2(cg, condition, [&] {
  cg.CallPrintf("then branch", {});
  return cg.Const32(1);
}, [&] {
  cg.CallPrintf("else branch", {});
  return cg.Const32(2);
});
```

Support for `codegen::lang::Loop` will be added later.


src/codegen/lang/if.cpp
```diff
@@ -42,6 +42,27 @@ If::If(CodeGen &cg, llvm::Value *if_condition, const std::string &name)
...
+  // Run the then-branch
+  llvm::Value *then_result = cgen_then_branch();
+  new_if->ElseBlock();
+  // Run the else-branch
+  llvm::Value *else_result = cgen_else_branch();
+  new_if->EndIf();
+  return new_if->BuildPHI(then_result,else_result);
+}
+
+llvm::Value *If::If2(CodeGen &cg, const codegen::Value &if_condition,
```
apavlo: I don't think we should make a new `If::If2`. If we want to go this way, we should just replace the old `If::If` since having two of them will confuse people.

MalcolmUnWang: The old `If::If` is a constructor which can only give a `codegen::lang::If` type variable instead of `llvm::Value *` type variable. If we want a `If` to return `llvm::Value *` directly as the issue posted, what if we make the `If::If2` less confusing by giving it a less confusing name like `If::GetIfResult` or something like that. Or, maybe we should implement this whole class completely different by makes `codegen::lang::If` be a subclass of `llvm::Value *`.

pmenon:
Before reviewing this, let's discuss back in the original thread.

pmenon:
I don't think we'll be going with this approach, so I'm closing.

PR #926, opened 2017-11-28
Add Support for SSL Network Connections

dengyuchenkit:
This PR adds support for SSL connection to postgres client in the networking layer.

Currently, peloton supports SSL connection with PSQL in some way and still needs thorough testing. Now we add a unit test which checks the correctness of building connection and mainly sending response in SSL context. 
During the test, server loads key and certificate files from peloton/test/network/ssl directory. 

TODO:
- May need to add support for OpenSSL in multi-threaded environment.
- Check correctness in SSL I/O and add tests to check the situations when unauthenticated files are given.
- Since the SSL connection is closely related with postgres client, we may need to check configurations in PSQL more.
- Decide which key files and certificates to use and how to load them.


src/network/postgres_protocol_handler.cpp
```diff
@@ -28,6 +28,9 @@
 #include "type/value_factory.h"
 #include "network/marshal.h"
 #include "settings/settings_manager.h"
+#define SSL_MESSAGE_VERNO 80877103
+#define PROTO_MAJOR_VERSION(x) x >> 16
```
tli2: Always put parens around your macro. (e.g. PROTO_MAJOR_VERSION(x) ((x) >> 16))

The reason for this is that a macro is literally a string substitution pre-compile time. If you omit parens operator precedence can get in your way and result in strange bugs. For example:

#define FOO(x) x * 16

and the expression FOO(3 + 4) will expand to 3 + 4 * 16 and result in 67 instead of 112.

or, let FOO(x) be x + 16, FOO(3) * 4 wll expand to 3 + 16 * 4 and equal to 67 instead of 76

src/network/postgres_protocol_handler.cpp
```diff
@@ -1056,6 +1059,87 @@ bool PostgresProtocolHandler::ReadPacket(Buffer &rbuf, InputPacket &rpkt) {
   return true;
 }
 
+/*
+ * process_startup_packet - Processes the startup packet
+ *  (after the size field of the header).
+ */
+bool PostgresProtocolHandler::ProcessInitialPacket(InputPacket *pkt, Client client, bool ssl_able, bool& ssl_handshake, bool& finish_startup_packet) {
+  std::string token, value;
+  std::unique_ptr<OutputPacket> response(new OutputPacket());
```
tli2: use std::make_unique<>()

https://stackoverflow.com/questions/22571202/differences-between-stdmake-unique-and-stdunique-ptr

src/network/postgres_protocol_handler.cpp
```diff
@@ -1056,6 +1059,87 @@ bool PostgresProtocolHandler::ReadPacket(Buffer &rbuf, InputPacket &rpkt) {
...
+  else {
+    LOG_TRACE("process startup packet");
+    return ProcessStartupPacket(pkt, proto_version, client, finish_startup_packet);
+  }
+}
+
+bool PostgresProtocolHandler::ProcessSSLRequestPacket(InputPacket *pkt, bool ssl_able, bool& ssl_handshake) {
+  UNUSED(pkt);
+  std::unique_ptr<OutputPacket> response(new OutputPacket());
+  // TODO: consider more about a proper response
```
tli2: nit: Can you put your name after your TODO? Also, what does it mean to be more proper?

src/network/postgres_protocol_handler.cpp
```diff
@@ -1056,6 +1059,87 @@ bool PostgresProtocolHandler::ReadPacket(Buffer &rbuf, InputPacket &rpkt) {
...
+  if (ssl_able) {
```
tli2: This is like if (true) return true; else return false;

Instead:
ssl_handshake = ssl_able;
response->msg_type = ssl_able ? NetworkMessageType::SSL_YES : NetworkMessageType::SSL_NO

src/network/postgres_protocol_handler.cpp
```diff
@@ -1056,6 +1059,87 @@ bool PostgresProtocolHandler::ReadPacket(Buffer &rbuf, InputPacket &rpkt) {
...
+    ssl_handshake = true;
+    LOG_TRACE("SSL support");
+  } else {
+    response->msg_type = NetworkMessageType::SSL_NO;
+    ssl_handshake = false;
+    LOG_TRACE("SSL not support");
+  }
+  responses.push_back(std::move(response));
+  SetFlushFlag(true);
+  return true;
```
tli2: Will we ever return false?

src/network/protocol_handler.cpp
```diff
@@ -27,6 +27,14 @@ namespace network {
   //  bool ManageStartupPacket();
   void ProtocolHandler::SendInitialResponse() {}
 
+  bool ProtocolHandler::ProcessInitialPacket(UNUSED_ATTRIBUTE InputPacket* pkt,
```
tli2: Why is this concrete and not pure virtual?

CMakeLists.txt
```diff
@@ -155,3 +158,4 @@ peloton_print_configuration_summary()
 
 # ---[ Export configs generation
 peloton_generate_export_configs()
+
```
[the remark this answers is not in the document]
dengyuchenkit: @tli2 I will remove it.

src/CMakeLists.txt
```diff
@@ -97,4 +97,3 @@ endforeach()
 add_custom_target(benchmark)
 add_dependencies(benchmark tpcc ycsb sdbench)
 
-
```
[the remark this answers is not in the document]
dengyuchenkit: @tli2  Will remove it.

src/network/postgres_protocol_handler.cpp
dengyuchenkit: @tli2 Fixed. Thanks!

src/network/postgres_protocol_handler.cpp
```diff
@@ -1056,6 +1059,87 @@ bool PostgresProtocolHandler::ReadPacket(Buffer &rbuf, InputPacket &rpkt) {
...
+    return ProcessSSLRequestPacket(pkt, ssl_able, ssl_handshake);
+  }
+  else {
+    LOG_TRACE("process startup packet");
+    return ProcessStartupPacket(pkt, proto_version, client, finish_startup_packet);
+  }
+}
+
+bool PostgresProtocolHandler::ProcessSSLRequestPacket(InputPacket *pkt, bool ssl_able, bool& ssl_handshake) {
+  UNUSED(pkt);
```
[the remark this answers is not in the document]
dengyuchenkit: @tli2 It is unnecessary since the SSL startup packet only contains the version number for SSL, which is already checked outside the call. I will just remove the parameter from the function prototype.

src/network/postgres_protocol_handler.cpp
dengyuchenkit: @tli2 It is outdated. I will remove it.

src/network/postgres_protocol_handler.cpp
dengyuchenkit: @tli2 It won't return false, like the ProcessStartupPacket. Will refactor them.

src/network/postgres_protocol_handler.cpp
```diff
@@ -1056,6 +1059,87 @@ bool PostgresProtocolHandler::ReadPacket(Buffer &rbuf, InputPacket &rpkt) {
...
+  std::string token, value;
+
+
+  // Only protocol version 3 is supported
+  if (PROTO_MAJOR_VERSION(proto_version) != 3) {
+    LOG_ERROR("Protocol error: Only protocol version 3 is supported.");
+    exit(EXIT_FAILURE);
+  }
+
+  // TODO: check for more malformed cases
```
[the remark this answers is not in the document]
dengyuchenkit: @tli2 It is an old comment. I will check what this comment refers to.

src/network/protocol_handler.cpp
dengyuchenkit: @tli2 This may be because we are not quite clear about how to handle the initial packet of different protocols except wire protocol. I think we can make it virtual. Will double check it with @ChTimTsubasa

src/network/postgres_protocol_handler.cpp
```diff
@@ -1056,6 +1059,87 @@ bool PostgresProtocolHandler::ReadPacket(Buffer &rbuf, InputPacket &rpkt) {
...
+    else {
+      if (pkt->ptr >= pkt->len) break;
+      GetStringToken(pkt, value);
+      client.cmdline_options[token] = value;
+    }
+
+  }
+  finished_startup_packet = true;
+
+  // Send AuthRequestOK to client (skip client authentication)
```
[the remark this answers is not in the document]
dengyuchenkit: @tli2 Will add that.

src/network/postgres_protocol_handler.cpp
dengyuchenkit: @tli2 Fixed. Thanks

script/testing/jdbc/src/SSLTest.java
```diff
@@ -0,0 +1,69 @@
...
+
+    public void Basic() throws SQLException {
+        Statement stmt = conn.createStatement();
+        stmt.execute(INSERT_1);
+        stmt.execute(INSERT_2);
+        stmt.execute(SELECT_1);
+        ResultSet rs = stmt.getResultSet();
+        while (rs.next()) {
+            String city_name = rs.getString("city");
+            System.out.println(city_name);
```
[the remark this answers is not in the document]
dengyuchenkit: @tli2 Thanks. We will add that. Right now we didn't enable this test since we need to do something with certificate loading in Java. Will add that later!

src/network/network_manager.cpp
```diff
@@ -116,44 +207,19 @@ void NetworkManager::StartServer() {
...
+      if (GetSSLLevel() != SSLLevel::SSL_DISABLE) {
+        SSL_CTX_free(ssl_context);
+      }
       throw ConnectionException("Failed binding socket.");
     }
 
     if (listen(listen_fd, conn_backlog) < 0)
     {
-      SSL_CTX_free(ssl_context);
+      if (GetSSLLevel() != SSLLevel::SSL_DISABLE) {
```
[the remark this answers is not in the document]
dengyuchenkit: @tli2 Okay, fixed

src/network/postgres_protocol_handler.cpp
dengyuchenkit: @tli2 It won't return false. Have refactored it. Thanks

apavlo:

pmenon:
`ssl_test`, introduced in this PR, is failing on dev4, dev5, and my machine due to a memory leak. Here is the output from the test on my machine: [test output](https://gist.github.com/pmenon/c7f5929198d9bdf26753101afc040cb5). Can you take a look at it?

dengyuchenkit:
@pmenon The document says that memory leak exists in old version 
of openssl. It is fixed in 1.1.0+ (https://www.openssl.org/docs/faq.html#PROG13). My laptop is ubuntu 16.04 and apt brings openssl of v1.0.2. I also see the memory leak now. Let me try and check the new version.

PR #930, opened 2017-12-01
Support 'sqrt' function

prasoontelang:
Implementation was done during the Hackathon. Supports and tests sqrt operation on datatypes such as decimal, tinyint, smallint, integer, bigint.


pmenon:
@prasoontelang It looks like you have some bugs in the code. Can you try to fix them?

apavlo:
@pmenon Fixed this separately. Wu tang

PR #932, opened 2017-12-04
Fix INSERT SELECT by filtering out the tuples inserted in the same execution

seojungmin:
Edited: One of the problems that the original PR solved is not in #934.

Without this PR, inserting tuples with SELECT may insert more tuples than required, e.g. `INSERT INTO foo1 SELECT a, b FROM foo1;` will insert more than the number of tuples already there in `foo1`.

This problem was there because the tuples inserted were scanned again by the table scanner depending on the tile group selection.  This is fixed by keeping  a separate filter in INSERT to filter out those.  It could also implemented in other ways, e.g. the modules underneath the INSERT logic may take a sort of filter. It may be more efficient, but more complex, so for the moment, we do this at INSERT.

### DIFF
--- a/src/codegen/inserter.cpp
+++ b/src/codegen/inserter.cpp
@@ -13,6 +13,7 @@
 #include "codegen/inserter.h"
 #include "codegen/transaction_runtime.h"
 #include "common/container_tuple.h"
+#include "common/item_pointer.h"
 #include "concurrency/transaction_manager_factory.h"
 #include "executor/executor_context.h"
 #include "executor/logical_tile.h"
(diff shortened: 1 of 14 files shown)


pmenon:
@huangminchn This is the fixed I referred to that should fix your index scan into insert problem.

@seojungmin Can you provide some more details on the second bug this PR fixes? I can't reproduce it locally on master. From the look of it, you're duplicating information stored in the transaction's read/write set. At best, it's as simple as a missed check somewhere in TXN management; at worst it's a more insidious transaction bug. Either way, I'd prefer pulling these apart

seojungmin:
The second one is reproducible, when the insertion of tuples are done to the tile groups that have not been selected to be read.  In those cases, when the table scan is executed, it does not distinguish whether it has been inserted by the current query or by the ones before this query. 

The TXN manager may check this, but as far as I know, it also requires an additional read-write set to check whether the tuple is inserted by _this query_, since the concept of a read-write set at the TXN manager should be built around a transaction.  Do we have such facility inside or outside the read-write set?

pmenon:
@seojungmin Can you provide a SQL session that reproduces the bug?

seojungmin:
@pmenon Sure. I am interested that it has not been reproducible in your session. I checked this in dev4 and 5. I did not check it at my laptop(OSX), though.
```
postgres=# create table foo(a int);
CREATE TABLE
postgres=# insert into foo values(1),(2);
INSERT 0 2
postgres=# select * from foo;
 a 
---
 1
 2
(2 rows)

postgres=# insert into foo select * from foo;
INSERT 0 10
postgres=# select * from foo;
 a 
---
 1
 1
 2
 2
 1
 2
 1
 2
 1
 2
 1
 2
(12 rows)

postgres=#
```

seojungmin:
I close this PR, since a similar case may occur in other queries. 

This should be re-written by introducing a generalized filter for filtering out the tuples inserted by the current execution "context".  The execution engine translator, e.g. INSERT in this case, builds a generalized filter, and the table scanner should apply this filter only when it is required, e.g. scanning while inserting to the same table.

PR #934, opened 2017-12-04
Fix column attribute binding at INSERT-SELECT

seojungmin:
This PR fixes a INSERT SELECT bug that is mentioned at #932.  I separate this out for a better management of PRs. 

Inserting tuples with SELECT on some specific order of columns may not be working correctly. e.g. `INSERT INTO foo1 SELECT b, a FROM foo2;` may behave like `INSERT INTO foo1 SELECT a, b FROM foo2;` when `foo1` has columns order of `a` and `b`.


pmenon:
LGTM +1

seojungmin:
This PR is ready to be merged.

PR #936, opened 2017-12-05
Simplify NULL handling in type system

pmenon:
**Problem**: Right now, all the operators/functions in codegen assume non-null inputs, and rely on the `codegen::Value` class to wrap operators in "null-aware" wrappers to handle NULL inputs. This was okay as long as all calls go through `codegen::Value`. But, with the introduction of function expressions, this is no longer the case as these expressions directly call into the type system. Exposing the null logic externally is a bad idea; moreover, the functions themselves know how to handle NULLs.

**Solution**: Provide base classes that perform the same NULL-handling logic, deferring to derived classes to perform non-null-aware logic. This encapsulates NULL logic to the operator, which makes a lot more sense.

**Details**: In this PR, we introduce four base classes for casting, comparison, unary, and binary operators. The base performs common NULL handling in the evaluation method, and call into derived implementations. Base classes can also elide the NULL-check if it knows the input is not null-able; but, even this can be overridden.


seojungmin:
Looks good! It should be merged ASAP.

apavlo:

PR #939, opened 2017-12-05
Check constraints while updating a tuple

### LINKED ISSUE #938: Constraints violated on update
It looks like column constraints are not being checked on update in neither the codegen or interpreted engine.  Example session:

```
postgres=# create table test (a int, b int not null);
CREATE TABLE
postgres=# insert into test values (NULL, 1);
INSERT 0 1
postgres=# select * from test;
 a | b 
---+---
   | 1
(1 row)

postgres=# update test set b=a;
UPDATE 1
postgres=# select * from test;
 a | b 
---+---
   |  
(1 row)
```

seojungmin:
This PR fixes #938. It used to check the constraints only when a tuple is newly inserted. This PR also checks them at `InsertVersion()`, which is essentially called at UPDATE.

Note: After updating a tuple which violates the constraints, it produces an exception as in INSERT, however, the exception is correctly caught and the peloton instance gets **aborted**.  This is a another bug which is not related to this, but to be solved.

[SQL session]
```
postgres=# create table test (a int, b int not null);
CREATE TABLE
postgres=# insert into test values(null, 1);
INSERT 0 1
postgres=# update test set b=a;
```

[Peloton output]
```
Exception Type :: Constraint
Message :: NOTNULL constraint violated on column 'b' : FIXME
terminate called after throwing an instance of 'peloton::ConstraintException'
  what():  NOTNULL constraint violated on column 'b' : FIXME
Aborted (core dumped)
```

### DIFF
--- a/src/storage/data_table.cpp
+++ b/src/storage/data_table.cpp
@@ -222,7 +222,6 @@ bool DataTable::CheckConstraints(const AbstractTuple *tuple) const {
 // however, when performing insert, we have to copy data immediately,
 // and the argument cannot be set to nullptr.
 ItemPointer DataTable::GetEmptyTupleSlot(const storage::Tuple *tuple) {
-
   //=============== garbage collection==================
   // check if there are recycled tuple slots
   auto &gc_manager = gc::GCManagerFactory::GetInstance();
@@ -308,6 +307,11 @@ bool DataTable::InstallVersion(const AbstractTuple *tuple,
                                const TargetList *targets_ptr,
                                concurrency::Transaction *transaction,
                                ItemPointer *index_entry_ptr) {
+  if (CheckConstraints(tuple) == false) {
+    LOG_TRACE("InsertVersion(): Constraint violated");
+    return false;
+  }
+
   // Index checks and updates
   if (InsertInSecondaryIndexes(tuple, targets_ptr, transaction,
                                index_entry_ptr) == false) {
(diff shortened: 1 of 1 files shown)


apavlo:
@vittvolt Can you review this?

seojungmin:
@vittvolt The API signature changes in this PR has been already done from recent other PRs. The only change is that `CheckConstraints()` is called at `InsertVersion()`.  AFAIK, `CheckConstaints()` was used to be called at `GetEmptyTuple()` initially when checking constraints is first added, and is called now in `InsertTuple()` and `InsertVersion()`. This is because, with the recent execution engine modification, `GetEmptyTuple()` is not always called with a `storage::Tuple`, but a lot of times with `nullptr`.

seojungmin:
@vittvolt Do you have time to review this? If not, I will assign someone else. Thanks!

vittvolt:
@seojungmin Sure no problem.

vittvolt:
@seojungmin The abort problem after exception is probably because it's never handled. I'll open another PR for this, it was once done but never merged. I might need to ask about this in later meetings.

seojungmin:
@vittvolt For your information, the exception handling is related to the thread pool.

seojungmin:
@vittvolt Thanks for the review!!

apavlo:
@seojungmin Can you rebase?

seojungmin:
@apavlo Rebased. This should work.

PR #942, opened 2017-12-06
Reorganize existing protobuf files

tli2:
We used to have different versions of the same protobuf files scattered across the database. I moved both of them into the same location and eliminated duplicates.

Should we remove some of these at some point though? They don't seem to be used anywhere.


apavlo:

PR #945, opened 2017-12-07
Reorganize and clean up settings

tli2:
Added comments explaining the macro logic.
Removed redundant -h flag for help and only use -help built in from gflag
Moved flag definitions from settings.cpp to settings_manager.cpp for clarity of reference. (It looks like settings.cpp is mostly out-dated constants before settings manager)


apavlo:
I will merge this next when the build passes.

PR #950, opened 2017-12-11
Support for basic UDFs in Peloton

prashasthip:
This PR adds PL/pgSQL UDF support to Peloton and some other features/ changes which are needed by the same.

**PL/pgSQL UDFs**

There are 2 parts to adding support for UDFs:
1. UDF Registration : This involves parsing the UDF, compiling it into native code and saving it in an in-memory map for quick access in case of future invocations.
2. UDF Invocation : This identifies that the SELECT query is invoking a UDF and extracts the previously compiled function and executes it.

This PR can be divided into 2 Parts:

1. **UDF Registration support**
a. Adding support for handling `CREATE OR REPLACE FUNCTION` statement which dissects the `CREATE FUNCTION` statement into UDF function_name, function parameter names, function parameter types, function return type and function body.
b. Handling the UDF function body which includes, parsing into Abstract Syntax Tree form and Just-in-Time compiling it during UDF registration.

2. **UDF Invocation support**
a. Adding support to help identify a function_expression to be a UDF.
b. Registering the previously compiled function into the current code_context to allow for invocation.

This PR currently supports UDFs with:
1. Expression statements
2. If-Else constructs
3. Recursive Function Calls


pmenon:
@prashasthip Can you rebase and fix the tests?

src/catalog/catalog.cpp
```diff
@@ -819,6 +819,42 @@ void Catalog::AddBuiltinFunction(
...
+    const std::string &name, const std::vector<type::TypeId> &argument_types,
+    const type::TypeId return_type, oid_t prolang, const std::string &func_src,
+    std::shared_ptr<peloton::codegen::CodeContext> code_context,
+    concurrency::TransactionContext *txn) {
+  // Check if UDF already exists
+  auto proc_catalog_obj =
+      ProcCatalog::GetInstance().GetProcByName(name, argument_types, txn);
+
+  if (proc_catalog_obj == nullptr) {
+    if (!ProcCatalog::GetInstance().InsertProc(name, return_type,
```
pmenon: Why is this check needed? Can't you add the 'plpgsql' language on startup?

src/catalog/catalog.cpp
```diff
@@ -832,28 +868,47 @@ const FunctionData Catalog::GetFunction(
...
+          "Function " + name +
+          " is internal, but doesn't have a function address");
+    }
+  } else if (lang_catalog_obj->GetName() == "plpgsql") {
+    // If the function is a "plpgsql" udf, perform the lookup in the plpgsql
+    // functions map (i.e., function::PlpgsqlFunctions) to get the function
+    // code_context
+    result.func_context_ = function::PlpgsqlFunctions::GetFuncContextByOid(
+        proc_catalog_obj->GetOid());
+    result.isUDF_ = true;
```
pmenon: Rename `isUDF_` to `is_udf_`

src/catalog/catalog.cpp
```diff
@@ -832,28 +868,47 @@ const FunctionData Catalog::GetFunction(
...
+
+    if (result.func_context_->GetUDF() == nullptr) {
```
pmenon: Is this ever possible?

src/codegen/code_context.cpp
```diff
@@ -158,18 +159,15 @@ void CodeContext::RegisterFunction(llvm::Function *func) {
...
 }
 
-void CodeContext::RegisterExternalFunction(
-    llvm::Function *func_decl, UNUSED_ATTRIBUTE llvm::Function *external,
-    CodeContext::FuncPtr func_impl) {
+void CodeContext::RegisterExternalFunction(llvm::Function *func_decl,
+                                           CodeContext::FuncPtr func_impl) {
   PL_ASSERT(func_decl->isDeclaration() &&
             "The first argument must be a function declaration");
-  PL_ASSERT(!external->isDeclaration() &&
```
pmenon: Why remove these checks?

src/include/expression/expression_util.h
```diff
@@ -514,6 +514,26 @@ class ExpressionUtil {
...
+    } else if (expr->GetExpressionType() == ExpressionType::FUNCTION) {
+      auto func_expr = (expression::FunctionExpression *)expr;
+      std::vector<type::TypeId> argtypes;
+      for (size_t i = 0; i < children_size; i++)
+        argtypes.push_back(expr->GetChild(i)->GetValueType());
+      // Check and set the function ptr
+      auto catalog = catalog::Catalog::GetInstance();
+      const catalog::FunctionData &func_data =
+          catalog->GetFunction(func_expr->GetFuncName(), argtypes);
+      LOG_INFO("Function %s found in the catalog",
```
pmenon: Change to `LOG_DEBUG`

src/include/expression/function_expression.h
```diff
@@ -30,7 +30,8 @@ class FunctionExpression : public AbstractExpression {
                      const std::vector<AbstractExpression *> &children)
       : AbstractExpression(ExpressionType::FUNCTION),
         func_name_(func_name),
-        func_(OperatorId::Invalid, nullptr) {
```
pmenon: Does it make sense to separate this into `BuiltinFunctionExpression` and `UDFExpression` ?

src/catalog/catalog.cpp
prashasthip: This is not for adding the plpgsql language. This is the ProcCatalogObject returned for the queried UDF name and parameter types. 

This check basically achieves the functionality of checking whether the UDF has already been registered in the catalog or not, and if it has, doesn't not try to register the UDF again.

src/catalog/catalog.cpp
prashasthip: Technically, if there are any exceptions raised while parsing the UDF definition, this code will be entered.

src/codegen/code_context.cpp
prashasthip: I only require a `llvm::Function *func_decl` to get the _function_name_ and the `void *` pointing to the _function_.

src/include/expression/function_expression.h
prashasthip: It would require a bit of re-work on the pg_proc_catalog side to separate the contents of FunctionData https://github.com/cmu-db/peloton/blob/f25ff8ddb452302f178d9a3b62305f95ef4545f7/src/include/catalog/catalog.h#L57 and return the corresponding catalog object. (We would need one for Built-in functions and one for UDF types).

prashasthip: We could, but it seems like a lot of work just to separate `function::BuiltInFuncType` and `codegen::CodeContext * `.

pervazea:
Where are we on this PR. More changes required?

apavlo:
@prashasthip @pmenon @sivaprasadsudhir @pervazea I just talked with Siva. This is a lot of code. It works but there are some parts that we want to refactor. I will merge it now when the build passes and then we can continue working.

apavlo:
@sivaprasadsudhir I merged this by hand after a minor fix.

@prashasthip Thanks!!!

PR #951, opened 2017-12-13
Statement Cache

ChTimTsubasa:
This is the plan cache that is used by Postgres protocol handler to support Prepare or extended protocol.

It would be updated when schema or index changes. The assumption is that the table schema or index rarely changes and the plan cache is not going to last for a very long time (drop when the connection tears down).

### DIFF
--- a/src/common/statement_cache_manager.cpp
+++ b/src/common/statement_cache_manager.cpp
@@ -26,6 +26,7 @@ void StatementCacheManager::InvalidateTableOid(oid_t table_id) {
   // Lock the table by grabing the iterator
   auto iterator = statement_caches_.GetIterator();
 
+  // Iterate each plan cache
   for (auto &iter : iterator) {
     iter.first->NotifyInvalidTable(table_id);
   }
--- a/src/common/statement_cache_manager.cpp
+++ b/src/common/statement_cache_manager.cpp
@@ -14,11 +14,11 @@
 
 namespace peloton {
 
-void StatementCacheManager::RegisterStatementCache(StmtCachePtr stmt_cache) {
+void StatementCacheManager::RegisterStatementCache(StatementCache *stmt_cache) {
   statement_caches_.Insert(stmt_cache, stmt_cache);
 }
 
-void StatementCacheManager::UnRegisterStatementCache(StmtCachePtr stmt_cache) {
+void StatementCacheManager::UnRegisterStatementCache(StatementCache *stmt_cache) {
   statement_caches_.Erase(stmt_cache);
 }
 
--- a/src/common/statement_cache_manager.cpp
+++ b/src/common/statement_cache_manager.cpp
@@ -1,0 +1,35 @@
+//===----------------------------------------------------------------------===//
+//
+//                         Peloton
+//
+// statement.h
+//
+// Identification: src/include/common/statement_cache_manager.cpp
+//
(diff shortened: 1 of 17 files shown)


src/common/statement_cache_manager.cpp
```diff
@@ -0,0 +1,56 @@
...
+  LOG_TRACE("Locking the table to get the iterator");
+  auto iterator = statement_caches_.GetIterator();
+  // Iterate each plan cache
+  for (auto &iter : iterator) {
+    iter.first->NotifyInvalidTable(table_id);
+  }
+  // Automatically release the table;
+}
+
+void StatementCacheManager::InvalidateTableOids(std::set<oid_t> &table_ids) {
```
pervazea: Is there a simple/clean way for InvalidateTableOids to call InvalidateTableOid and avoid duplication?

src/common/statement_cache_manager.cpp
```diff
@@ -0,0 +1,56 @@
...
+void StatementCacheManager::UnRegisterStatementCache(StatementCache *stmt_cache) {
+  statement_caches_.Erase(stmt_cache);
+}
+
+void StatementCacheManager::InvalidateTableOid(oid_t table_id) {
+
+  if (statement_caches_.IsEmpty()) 
+    return;
+
+  // Lock the table by grabing the iterator
```
pervazea: grabbing

src/common/statement_cache_manager.cpp
```diff
@@ -0,0 +1,56 @@
...
+    iter.first->NotifyInvalidTable(table_id);
+  }
+  // Automatically release the table;
+}
+
+void StatementCacheManager::InvalidateTableOids(std::set<oid_t> &table_ids) {
+  if (table_ids.empty())
+    return;
+
+  // Lock the table by grabing the iterator
```
pervazea: As above, grabbing.

test/common/statement_cache_manager_test.cpp
```diff
@@ -0,0 +1,88 @@
...
+
+  // Invalidate table oid 0, 1
+  statement_cache_manager->InvalidateTableOids(ref_table);
+
+  // Both plans need to replan now
+  for (auto oid : ref_table) {
+    std::string string_name = "foo" + oid;
+    auto statement = statement_cache.GetStatement(string_name);
+
+    EXPECT_TRUE(statement->GetNeedsPlan());
```
pervazea: GetNeedsPlan and the corresponding SetNeedsPlan, should in my opinion, be GetNeedsReplan and SetNeedsReplan, because the flag (boolean) being set is replan.

src/common/statement_cache_manager.cpp
ChTimTsubasa: InvalidateTableOids calling InvalidateTableOid would cause multiple times of mutex grabbing every time the iterator of CuckooMap is generated. I was wondering this would affect the performance.

Further, neither of these function is long or complex.

pervazea: It can stay as is. 
We don't want to add mutex acquisitions.

apavlo:
@pervazea Is this ready to be merged?

@ChTimTsubasa Can you fix the conflicts + rebase?

apavlo:
@pervazea This fails to build with `LOG_LEVEL_TRACE` for me, but I don't think the problems are in @ChTimTsubasa's PR. We should fix that separately.

PR #952, opened 2017-12-13
Fix path

schedutron:
I think under Peloton's root folder, the path should be `./script/git-hooks/pre-commit` instead of `../../script/git-hooks/pre-commit`.


schedutron:
Just realized that the original path is correct!

PR #962, opened 2017-12-19
[Cleanup] Factor Network State Machine out of NetworkConnection

tli2:
Make Network State Machine into its own class and pull some Macro trick to make transition declaration more readable.

I collapsed some states together but there are definitely more opportunities. I marked them with TO-DOs and with any luck we can collapse them as we progress along in the refactor.

Next step would be to clear up the mess with NetworkThread, NetworkMasterThread and NetworkWorkerThread. Will rename those respectively into NotifiableTask, ConnectionDispatcherTask and ConnectionHandlerTask to avoid confusion, and shuffle code around for better encapsulation


tli2:
There seems to be some flakiness in the build. This passed on my fork's Travis under (I think) the same environment

tcm-marcel:
I just restarted the build and it worked. This seems to happen quite often after Travis switched to Ubuntu Trusty on Dec 12th.

src/network/connection_handle.cpp
```diff
@@ -0,0 +1,89 @@
...
+// Copyright (c) 2015-17, Carnegie Mellon University Database Group
+//
+//===----------------------------------------------------------------------===//
+
+#include "network/connection_handle.h"
+#include "network/network_connection.h"
+
+
+namespace peloton {
+  namespace network {
```
apavlo: Formatting.

src/network/connection_handle.cpp
```diff
@@ -0,0 +1,89 @@
...
+              switch (c) {
+      #define DEFINE_STATE(s) case ConnState::s: { switch (t) {
+      #define END_DEF default: throw std::runtime_error("undefined transition"); }}
+      #define ON(t) case Transition::t: return
+      #define SET_STATE_TO(s) {ConnState::s,
+      #define AND_INVOKE(m) ([] (NetworkConnection &w) { return w.m(); })};
+      #define AND_WAIT ([] (NetworkConnection &){ return Transition::NONE; })};
+    }
+
+    DEF_TRANSITION_GRAPH
```
apavlo: This is all very clever you need to add some comments about what's going on.

tli2:
Done

src/include/network/network_state.h
```diff
@@ -40,5 +37,17 @@ enum class WriteState {
   WRITE_ERROR,      // Some error happened
 };
 
+enum class Transition {
+  NONE,
+  WAKEUP,
+  PROCEED,
+  NEED_DATA,
+  // TODO(tianyu) generalize this symbol, this is currently only used in process
+  GET_RESULT,
```
dengyuchenkit: Since we have a `CONN_GET_RESULT` state, it may be better to change the `GET_RESULT` transition to some other name.

tli2: The intention is for this to have the same name since this is the only state that makes use of this transition symbol.

src/include/network/connection_handle.h
```diff
@@ -0,0 +1,68 @@
...
+  /**
+   * delta is the transition function that defines, for each state, its behavior and the
+   * next state it should go to.
+   */
+  static transition_result Delta_(ConnState state, Transition transition);
+  ConnState current_state_;
+};
+
+}
+}
```
ChTimTsubasa: Add one more blank line here

src/include/network/network_state.h
```diff
@@ -22,10 +22,7 @@ enum class ConnState {
   CONN_WAIT,       // State for waiting for some event to happen
   CONN_PROCESS,    // State that runs the network protocol on received data
   CONN_CLOSING,    // State for closing the client connection
-  CONN_CLOSED,     // State for closed connection
-  CONN_INVALID,    // Invalid STate
```
ChTimTsubasa: If I was right, CONN_INVALID is used to disable the system?

src/network/connection_handle.cpp
```diff
@@ -0,0 +1,103 @@
...
+void ConnectionHandleStateMachine::Accept(Transition action, NetworkConnection &connection)   {
+  Transition next = action;
+  while (next != Transition::NONE) {
+    transition_result result = Delta_(current_state_, next);
+    current_state_ = result.first;
+    next = result.second(connection);
+  }
+}
+}
+}
```
ChTimTsubasa: Add the new line here

src/include/network/network_state.h
```diff
@@ -22,10 +22,7 @@ enum class ConnState {
...
   CONN_GET_RESULT, // State when triggered by worker thread that completes the task.
```
ChTimTsubasa: I am wondering if we may merge CONN_GET_RESULT and CONN_WAIT?

src/network/network_connection.cpp
```diff
@@ -615,202 +609,132 @@ void NetworkConnection::Reset() {
...
-        PL_ASSERT(false);
-        break;
-      }
-    }
+  // Should not be here
+  return Transition::ERROR;
+}
+
+Transition NetworkConnection::GetResult() {
+  // TODO(tianyu) We probably can collapse this state with some other state.
```
ChTimTsubasa: WAIT state I suppose.

tli2: Will do that in a separate PR to clean up all the states

src/include/network/network_state.h
tli2: Will do that in a separate PR to clean up all the states

src/include/network/network_state.h
tli2: It is used as a dummy value for an invalid state. It also doubles as a sensible default value since a state machine can start either on CONN_LISTENING or CONN_READ.

This does not make much sense because:
1. If an instance is not valid, it should not be there, available for interaction with the other components (object lifecycle management, an invalid object should also be out of scope)
2. It does not make much sense to have a CONN_LISTENTING state since it is disconnected with the rest of the graph. I will put in a separate PR to make that into its own callback function and take it out of the state machine.

PR #965, opened 2017-12-19
Changed StatementResult to ResultValue

poojanilangekar:
The second element of the StatementResult type was used to store the value of a column of a tuple in a result set. The first element of the pair was **never** used. So I have changed `std::pair<std::vector<unsigned char>, std::vector<unsigned char>>` to `std::vector<unsigned char>` and everything else seems to be working as expected. I have renamed the type to `ResultValue`.

We are copying it from `std::string` to `std::vector<unsigned char>` because it is later used by `SendDataRows` function of the `PostgresProtocolHandler`. I am not sure, but I am guessing there must be a way we can avoid this conversion and hence save the copy.

This PR changes 37 files, however only 7 are a part of the engine,  the rest are tests.

### DIFF
--- a/src/executor/plan_executor.cpp
+++ b/src/executor/plan_executor.cpp
@@ -44,7 +44,7 @@ void CleanExecutorTree(executor::AbstractExecutor *root);
 void PlanExecutor::ExecutePlan(
     std::shared_ptr<planner::AbstractPlan> plan,
     concurrency::Transaction *txn, const std::vector<type::Value> &params,
-    std::vector<StatementResult> &result,
+    std::vector<ResultValue> &result,
     const std::vector<int> &result_format, executor::ExecuteResult &p_status) {
   PL_ASSERT(plan != nullptr && txn != nullptr);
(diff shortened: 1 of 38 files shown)


phisiart:
Wait... since its a vector<unsigned char>, how about just use string?

poojanilangekar:
I thought the same. However, @malin1993ml has commented on [Line 451 in logical_tile.cpp](https://github.com/cmu-db/peloton/blob/5ca11d3adc039c9e63aded78968bac9229d65e6c/src/executor/logical_tile.cpp#L451) that we need to use a vector of unsigned char so we don't have to care about the endianness while sending the data. 
@malin1993ml or @ChTimTsubasa can you confirm if this can/can't be changed?

malin1993ml:
I don’t know about other parts of the networking layer, but for the
function that I modified(as you referred), it does return the result as
vector of strings. And we do care about the edianness. We need to reverse
the order of the binary result for non-varchar(varbinary) type. Having said
that, there might be ways to achieve that without copying. And I don’t see
how using a vector of unsigned char can solve the endianness problem...

PR #969, opened 2017-12-19
[Cleanup] change all direct print to LOG_* macro

ksaito7:
This PR changes all direct printf/cout calls and original logging schema to LOG_* macro for code cleanup. In addition, direct assert calls change to PL_ASSERT macro.


src/include/index/bwtree.h
```diff
@@ -108,25 +101,7 @@ namespace bwtree {
...
-
-#else
-
-#define bwt_printf(fmt, ...)   \
-  do {                         \
-    dummy(fmt, ##__VA_ARGS__); \
-  } while (0);
-
-#endif
+static void dummy(const char *, ...) {}
```
apavlo: Should we just get rid of this now?

src/include/index/bwtree.h
```diff
@@ -346,127 +303,123 @@ class BwTreeBase {
    * This function must be called when the garbage pool is empty
    */
   void DestroyThreadLocal() {
-    bwt_printf("Destroy %lu thread local slots\n", thread_num);
-    
+    LOG_TRACE("Destroy %lu thread local slots\n", thread_num);
```
apavlo: Remove the newline character after all of these new LOG_TRACE

src/include/index/bwtree.h
ksaito7: I think these should be deleted because these are used nowhere.

src/include/index/bwtree.h
ksaito7: Sorry I will fix it soon.

src/include/index/bwtree.h
ksaito7: I've deleted its dummy function.

apavlo:
I resolved the conflicts by hand on Github but it doesn't look like that triggered the build.

apavlo:
Rebased + merged manually.

PR #971, opened 2017-12-19
[Cleanup] Cleaned up settings in Peloton

rohit-cmu:
No setting in settings.cpp were being used apart from peloton_layout_mode. Cleaned up that and layout mode is now set as a parameter with a default value same as the earlier setting. Appropriate Getter and Setter functionality is also added. The only affected test was tile_group_layout_test which has also been modified to use the new functionality.


rohit-cmu:
cc @tli2

rohit-cmu:
I dont think we want this in settings manager. Its real use is only when you want to load tables in a row manner/ column manner in benchmarks. I provide that functionality now as well. The sdbench benchmark uses this now. So I think this PR can be merged now. :)

chenboy:
LGTM

PR #977, opened 2017-12-20
[Cleanup] Change ResultValue to std::string. One result copy gone.

phisiart:


src/include/network/marshal.h
```diff
@@ -182,7 +182,7 @@ extern void PacketPutInt(OutputPacket *pkt, int n, int base);
 extern void PacketPutCbytes(OutputPacket *pkt, const uchar *b, int len);
 
 /* packet_put_bytes - used to write a uchar vector into a packet */
-extern void PacketPutBytes(OutputPacket *pkt, const std::vector<uchar> &data);
+extern void PacketPutBytes(OutputPacket *pkt, const std::string &data);
```
seojungmin: @phisiart Can you revert this back, and could you change the code in the network layer to use `PacketPutString`?

phisiart: Actually `PacketPutBytes` and `PacketPutString` are different.
`PacketPutString` pushes an extra null terminator.
Here in `PacketPutBytes` I'm using `std::string` as a byte array. I guess to make things more clear I could change `PacketPutBytes` to take `ResultValue`, though its just a typedef?

seojungmin: I think you got my point.  The signature is not really intuitive here. Any name that suits `PacketPutStringWithoutNullTerminator` will do.  I cannot think of the best name, though.

phisiart: Will change names, wait for an update ...

seojungmin:
This is ready to be merged.

PR #979, opened 2017-12-20
Fix for memory leak in failed CREATE transform

haojin2:
This PR includes the fix for memory leak issues when transformation of a CREATE statement throws an exception.

### DIFF
--- a/src/parser/postgresparser.cpp
+++ b/src/parser/postgresparser.cpp
@@ -886,6 +886,7 @@ parser::ColumnDefinition* PostgresParser::ColumnDefTransform(ColumnDef* root) {
         try {
           result->default_value.reset(ExprTransform(constraint->raw_expr));
         } catch (NotImplementedException e) {
+          delete result;
           throw NotImplementedException(StringUtil::Format(
               "Exception thrown in default expr:\n%s", e.what()));
         }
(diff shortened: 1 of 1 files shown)


seojungmin:
@haojin2 First, was there an GITHUB issue related to this, or was this something that you found that there was a leak? If it is the former case, please link the issue.

Second, is there any reason why we did not use smart pointers?

haojin2:
One of the networking guys found this issue but did not create one on Github. I talked to one of them yesterday afternoon and started to work on this.

For the smart pointer issue, it was because the expression system does not use that. Smart pointers are good, but the migration cost is high. If I remember correctly, at the time when PostgresParser was created, we prioritized the speed of development as we wanted it to work in a short period of time thus did not have the time to migrate the whole expression system to smart pointers.

seojungmin:
@haojin2 Thanks! I got it. 
@dengyuchenkit The code looks all right.  Can you confirm that this fix solves your problem? If it does, I think this is ready to go.

haojin2:
It should be ready to go.

dengyuchenkit:
Sorry, but can you hold on for a moment? I'm checking and there is still some leaks. I am checking where is that from

dengyuchenkit:
Sorry, it's good to go! @haojin2 @seojungmin

seojungmin:
This is ready to be merged.

PR #980, opened 2017-12-21
[Clean-Up] Add the ClearGarbage back to StopGC.

malin1993ml:
This adds the call to ClearGarbage function back to the Garbage Collector. It turns out that the Garbage Collector already has a StopGC function that will be called when Peloton shuts down. So I just added the call there.


malin1993ml:
@tcm-marcel Can you help review this when it passes the build? I only touched two functions so it should be very easy.

PR #985, opened 2017-12-21
[Cleanup] Clean up network threads

tli2:
In this PR I have renamed network threads to avoid confusion with the other thread classes in our codebase, and encapsulated all libevent related stuff into the class NotifiableTask.

Some other minor simplification of logic was also performed.

Based on this PR I want to next rewrite NetworkConnection and see if I can breakup that class into smaller pieces. Finally, I will redo the error handling logic to use exceptions and save all the error propagation code.

I've left some other logical changes I am not sure about in the TODOs, @ChTimTsubasa  can you take a look and tell me what you think?


tli2:
Unfortunately a bunch of other PRs got mixed in because I pulled from master this morning...

Issue #978, opened 2017-12-20
Change Travis to use gcc 4.8


apavlo:
* Change travis.yml to enforce gcc version 4.8. 
* Add new travis target for clang builds on Ubuntu

This may fix @crd477's docker build problem.

crd477:
On 2017-12-20 13:18, Andy Pavlo wrote:
>   * Change travis.yml to enforce gcc version 4.8.
>   * Add new travis target for clang builds on Ubuntu
> 
> This may fix @crd477 <https://github.com/crd477>'s docker build problem.
> 

That problem was already fixed by 
4399ebf99c7553741f7bbbdf3170fa583ff8a7ee.  I'm not aware of any 
outstanding Docker build problems.

The Jenkins gcc builds for Ubuntu Trusty use 4.8 and are currently working.


-- 
     -Chad

crd477:
Just to clarify, I'm not aware of any outstanding problems with the Docker build **mechanism**.  There are still some build failures in Jenkins but they are genuine compilation or linking problems not to do with Docker.

tcm-marcel:
The error is Jenkins is following:
```
/usr/bin/../lib/gcc/x86_64-redhat-linux/4.8.5/../../../../include/c++/4.8.5/cstdio:120:11: error: no member named 'gets' in the global namespace
  using ::gets;
        ~~^
```
It may happen because of clang using the wrong `libc++`. 

Comment from https://stackoverflow.com/questions/24342312/clang-seems-to-use-the-gcc-libraries:
>I had similar issue: GCC (g++) already was installed on my LinuxMint (Ubuntu base) so when compile with clang, was getting an " error: no member named 'gets' in the global namespace using ::gets ".
>
>resolved by installing libc++-dev (sudo apt-get install libc++-dev) and compiling with -stdlib++ (clang++ -g -std=c++1y -stdlib=libc++ helloworld.cpp -o helloworld)

crd477:
Nice find.  libc++-dev is already installed on the Jenkins Trusty image but -stdlib=libc++ is not being passed on the command line.  I can test this fix if you're not already down that path.

crd477:
I take that back - libc++-dev is not installed, only libstdc++-4.8-dev.  I'll add it.

tcm-marcel:
1. done in #989 
2. WIP in #1052

pervazea:
So... is this OK to close now? #1052 remains to track clang work.

Issue #538, opened 2017-02-05
Test Travis fast fail for 


apavlo:
Can someone modify `.travis.yml` to have it fail right away when `source_validator.py` fails?
We still want to run all the other builds and tests though to the very end even if they fail.

https://blog.travis-ci.com/2013-11-27-fast-finishing-builds/

danielcompton:
I've started working on this, but am having trouble making fast_finish actually finish any faster. https://travis-ci.org/danielcompton/peloton/builds/201264982. Just leaving a note here so work isn't duplicated.

tcm-marcel:
Implemented in #989

PR #989, opened 2017-12-21
 [Cleanup] Change travis coniguration to run gcc 4.8 and gcc 5.0

tcm-marcel:
This changes the travis configuration to build with gcc 4.8 and clang as requested in #978. Also the tests now only run twice (Debug+Release) instead of every time, and the travis jobs fast fail if the source validator fails and do not run the tests if the build failed. 

  * Debug _and_ Release builds are ran on gcc4.8 and gcc5.0 ~~~clang3.7 each (Linux)~~~
  * one Debug build under Mac OS X with clang
  * the gcc4.8-builds also include tests
  * coveralls is ran once
  * build is skipped if source validator fails (#538)
  * do not runs tests if build fails

**This PR got split up into this one introducing some general changes and #1052 adding clang support.**


tcm-marcel:
The clang builds on Linux still fail because of things clang handles differently to gcc. I will take care of this.

tcm-marcel:
I added gcc5 as requested by @apavlo and changed clang to version 3.9, hoping that the clang errors would disappear. They didn't... (https://travis-ci.org/tcm-marcel/peloton/builds/320329682)

Fixing the clang errors will take some time, I will continue there on Jan 5th.

tcm-marcel:
I still work on making peloton build with clang under Linux. 

Following files have classes with unused private members. I thought adding `UNUSED_ATTRIBUTE` could temporary help to get rid of the clang warnings, but apparently gcc doesn't support this for class members. So we should take a look at all of these and remove them if possible. 
```
/src/include/codegen/operator/update_translator.h
/src/include/codegen/type/type_system.h
/src/include/common/exception.h
/src/include/concurrency/local_epoch.h
/src/include/executor/hash_join_executor.h
/src/include/index/bloom_filter.h
/src/include/logging/logical_checkpoint_manager.h
/src/include/logging/logical_log_manager.h
/src/include/network/postgres_protocol_handler.h
/src/include/optimizer/child_property_generator.h
/src/include/optimizer/cost_and_stats_calculator.h
/src/include/optimizer/property_enforcer.h
/src/include/optimizer/query_property_extractor.h
/src/include/optimizer/query_to_operator_transformer.h
/src/include/optimizer/stats.h
/src/include/optimizer/stats/hyperloglog.h
/src/include/optimizer/tuple_sample.h
/src/include/storage/tile_group_header.h
```

In addition the linker fails because of missing symbols in protobuf, I haven't yet figured out why. The linker doesn't get any paths for protobuf, but with gcc it works somehow. 
```
../lib/libpeloton-d.so.0.0.5: undefined reference to `google::protobuf::Message::InitializationErrorString() const'
../lib/libpeloton-d.so.0.0.5: undefined reference to `google::protobuf::Message::GetTypeName() const'
../lib/libpeloton-d.so.0.0.5: undefined reference to `google::protobuf::internal::empty_string_'
../lib/libpeloton-d.so.0.0.5: undefined reference to `__atomic_compare_exchange'
../lib/libpeloton-d.so.0.0.5: undefined reference to `__atomic_load'
../lib/libpeloton-d.so.0.0.5: undefined reference to `google::protobuf::internal::StringTypeHandlerBase::New()'
```

Cmake can be configured to use clang as follows:
```
cmake -DCMAKE_BUILD_TYPE=Debug -DCMAKE_C_COMPILER=clang-3.7 -DCMAKE_CXX_COMPILER=clang++-3.7 ..
```

tcm-marcel:
**This PR got split up into this one introducing some general changes and #1052 adding clang support.**

tcm-marcel:
@apavlo I tested the early fail once: https://travis-ci.org/tcm-marcel/peloton/builds/319756560 (activated namespace check in source validator)

tcm-marcel:
@pavlo Do you want someone to take another look at this? Otherwise this PR is ready to merge.

Issue #770, opened 2017-07-31
Additional Static Source Code Checks


apavlo:
We should add the following checks to `scripts/formatting/formatter.py`

1. Make sure that the namespace is properly commented (#768) 
2. Make sure that testing utility files follow the proper naming convention
3. Make sure that we don't have the prefix "include/" in our own `#include`'s

tcm-marcel:
1. and 3. were added in #968 and #991, 2. is still missing.

tcm-marcel:
Remaining part moved to #1192.

PR #991, opened 2017-12-22
[Cleanup] Add more checks in source validator

tcm-marcel:
This PR adds more checks to the source validator as requested in #770. 

1. **Make sure that the namespace is properly commented (#768)**
  The check works, but all namespace checks are still disabled, until refactoring is finished (see #968)
2. **Make sure that we don't have the prefix "include/" in our own #include's**
  PR also includes commit that fixes violating files. 

The second request in #770 "Make sure that testing utility files follow the proper naming convention" is still missing. Also the checks are added to the source validator and **not** in the formatter.

### DIFF
--- a/script/validators/source_validator.py
+++ b/script/validators/source_validator.py
@@ -67,7 +67,8 @@ VALIDATOR_PATTERNS = [re.compile(patt) for patt in [
     " \:\:memcpy\(",
     " std\:\:memset\(",
     " std\:\:memcpy\(",
-    "\_\_attribute\_\_\(\(unused\)\)"
+    "\_\_attribute\_\_\(\(unused\)\)",
+    r'^#include "include/'
     ]
 ]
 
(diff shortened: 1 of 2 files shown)


tcm-marcel:
@apavlo "Make sure that testing utility files follow the proper naming convention" is still missing. Can you tell me what to check here? Is it about file name/location or class naming?
This is what I found about naming conventions: https://github.com/cmu-db/peloton/blob/master/test/README.md

apavlo:
@tcm-marcel For starters, we could check that the class name and the individual tests follow the proper naming convention. Enforcing it for the filenames is a bit tougher.

pmenon:
@tcm-marcel Can you add a check to ensure the namespaces are also closed properly? Brace, two spaces and double-slash comment in reverse namespace nesting order?

Another check would be sorted + partitioned includes, but that's probably a different PR.

tcm-marcel:
@pmenon At the moment, I get a list of namespaces that should exist, and then check if a proper opening and closing exists for every namspace. I can add more logic to also check the order of the closings. 
I don't check for "wrong" namespace tags so far, as there may exist more due to forward declarations.

apavlo:
@tcm-marcel I am moving this back to "In Progress". You can finish it when you get back from the mountains.

tcm-marcel:
Build succeeds in Travis, but is somehow not reported: https://travis-ci.org/tcm-marcel/peloton/builds/326898829

Test case checks will be done in a separate PR, this PR is ready to merge now.

pmenon:
@tcm-marcel @apavlo Did commit a613847e8086dbf90729f3f546ba9073b98d406b remove a bunch of new optimizer tests?

apavlo:
Ugh yes. Not sure how I missed that.

@tcm-marcel Please fix.

PR #992, opened 2017-12-22
Fix return type for floor

pmenon:
Looks like the return type of the floor function as set incorrectly. This PR fixes it.

@saatviks Since this was your function, you can review the fix :)


apavlo:
I will merge this next.

PR #941, opened 2017-12-06
Extend formatter script to run on all staged files, merge add/strip action

tcm-marcel:
This is especially useful when using the new pre-commit hook that prevents me from committing. 

* add -f option to formatter that applies the selected action to all staged files (git)
* adapt pre-commit hook to only check stages files

(since -a, -s and -c were already taken, I just picked any other letter that somehow looks like "staged files" when looking from far away)

*Testing:* I run Ubuntu, can someone test this on a Mac?

PR #993, opened 2017-12-22
[Cleanup] Fix formatter script

tcm-marcel:
Small fixes that came up during a conversation with @dengyuchenkit  and @haoxianghua. 

  * use any clang-format version
  * catch error if clang-format is not installed
    (similar to source validator changed in #941)
  * fix wrong path in pre-commit hook fail message


tcm-marcel:
@apavlo Sorry for pushing another commit, but I saw that this PR again would change the script not to use a specified version of clang-format. For now I decided to always call a specified version (see https://github.com/cmu-db/peloton/issues/1015#issuecomment-356962660 )

4c1c90e just reverts this change.

apavlo:
@tcm-marcel Sounds good. I will merge when this passes.

PR #995, opened 2017-12-22
Print out stack trace when throws an exception

mengranwo:
Using libunwind to print out stack trace whenever throw an exception. 
The following is a basic example.

**Client Terminal:** 
```
postgres=# select * from foo1;
Table foo1 is not found
```

**Server side**
```
Exception Type :: Catalog
Message :: Table foo1 is not found
#1  pc = 0x7fb2c866cdaf peloton::CatalogException::CatalogException(std::string) + 0x14f
#2  pc = 0x7fb2c86986f6 peloton::catalog::Catalog::GetTableObject(std::string const&, std::string const&, peloton::concurrency::Transaction*) + 0x176
#3  pc = 0x7fb2c85cc7ed peloton::binder::BinderContext::AddTable(peloton::parser::TableRef*, std::string, peloton::concurrency::Transaction*) + 0x9d
#4  pc = 0x7fb2c85c790c peloton::binder::BindNodeVisitor::Visit(peloton::parser::TableRef*) + 0x10c
#5  pc = 0x7fb2c85c8f3c peloton::binder::BindNodeVisitor::Visit(peloton::parser::SelectStatement*) + 0x13c
.....
```

### DIFF
--- a/src/include/common/exception.h
+++ b/src/include/common/exception.h
@@ -14,7 +14,7 @@
 
 #include <cxxabi.h>
 #include <errno.h>
-#include <execinfo.h>
+#include <libunwind.h>
 #include <signal.h>
 #include <cstdio>
 #include <cstdlib>
@@ -22,8 +22,8 @@
 #include <memory>
 #include <stdexcept>
 
-#include "type/type.h"
 #include "internal_types.h"
+#include "type/type.h"
 
 namespace peloton {
 
@@ -68,9 +68,9 @@ class Exception : public std::runtime_error {
 
   Exception(ExceptionType exception_type, std::string message)
       : std::runtime_error(message), type(exception_type) {
-    exception_message_ = "Exception Type :: " +
-                                    ExpectionTypeToString(exception_type) +
-                                    "\nMessage :: " + message;
+    exception_message_ =
+        "Exception Type :: " + ExpectionTypeToString(exception_type) +
+        "\nMessage :: " + message;
   }
 
   std::string ExpectionTypeToString(ExceptionType type) {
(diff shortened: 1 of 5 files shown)


mengranwo:
I can pass all the test cases on my local machine(with debug mode and thread sanitizer on), but on travis I always failed on one test: **value_integrity_test**
Here is the log info I got:
```
72: Test command: /home/travis/build/cmu-db/peloton/build/test/value_integrity_test "--gtest_color=yes" "--gtest_output=xml:/home/travis/build/cmu-db/peloton/build/test/value_integrity_test.xml"
72: Test timeout computed to be: 9.99988e+06
72: Running main() from gmock_main.cc
72: [0;32m[==========] [mRunning 2 tests from 1 test case.
72: [0;32m[----------] [mGlobal test environment set-up.
72: [0;32m[----------] [m2 tests from ValueIntegrityTest
72: [0;32m[ RUN      ] [mValueIntegrityTest.IntegerOverflow
72: unknown file: Failure
72: C++ exception with description "ERROR: overflow" thrown in the test body.
72: [0;31m[  FAILED  ] [mValueIntegrityTest.IntegerOverflow (263 ms)
72: [0;32m[ RUN      ] [mValueIntegrityTest.IntegerDivideByZero
72: 
72: Exception Type :: Divide by Zero
72: Message :: ERROR: division by zero
72: #1  pc = 0x544ca9 peloton::Exception::Exception(peloton::ExceptionType, std::string) + 0x329
72: #2  pc = 0x2b1c5e028993 peloton::DivideByZeroException::DivideByZeroException(std::string) + 0x5d
72: #3  pc = 0x2b1c5e0283c2 peloton::codegen::RuntimeFunctions::ThrowDivideByZeroException() + 0x92
72:  -- error: unable to obtain symbol name for this frame
72: #4  pc = 0x481e7d peloton::test::ValueIntegrityTest_IntegerDivideByZero_Test::TestBody() + 0x1d9
72: #5  pc = 0x5119a8 void testing::internal::HandleSehExceptionsInMethodIfSupported<testing::Test, void>(testing::Test*, void (testing::Test::*)(), char const*) + 0x158
72: #6  pc = 0x500fc0 void testing::internal::HandleExceptionsInMethodIfSupported<testing::Test, void>(testing::Test*, void (testing::Test::*)(), char const*) + 0x84
72: #7  pc = 0x4a40c2 testing::Test::Run() + 0x226
72: #8  pc = 0x4a56ca testing::TestInfo::Run() + 0x312
72: unknown file: Failure
72: C++ exception with description "ERROR: division by zero" thrown in the test body.
72: [0;31m[  FAILED  ] [mValueIntegrityTest.IntegerDivideByZero (91 ms)
```
There suppose to be eight Exception Messages print out in total (according to my local machine and the test content), but as you can see there is only one in the log and I got the **" unknown file: Failure"** error. Anyone has any idea?

seojungmin:
@tcm-marcel Would you give a shot on this? I think `libunwind` that is linked with `peloton` does not seem to work with a codegen'ed object throwing an exception.  If you run this at a debug mode and run `value_integrity_test`, you will get there, quickly.

tcm-marcel:
While looking at this, I realized that we have another `PrintStackTrace` function in `/common/stack_trace.cpp`, which is also used by the signal handler. Maybe we should move the changes there. 

(btw I know the guy from which the old `PrintStackTrace` code was taken from, he is from my university: https://github.com/tcm-marcel/peloton/blob/f844c864ffe75de89bc6c509c03896a1e6cb8e1d/src/common/stack_trace.cpp#L70)

tcm-marcel:
I was not able to find the actual problem yet. What I got so far: 

The two error lines are created by gmock. It catches the exception and usually would print file and line number, but here it cannot retrieve the file (maybe because the stack trace is incomplete). 
  ```
  unknown file: Failure
  C++ exception with description "ERROR: division by zero" thrown in the test body.
  ```
The exceptions are not thrown by the codegen code itself, but the codegen code calls `RuntimeFunctions::ThrowDivideByZeroException()` [1], which is a one-liner. So the location where it is thrown exists as actual c++ code. Maybe the problem is in gmock and the way it catches the exceptions. 

And as we don't call any function in libunwind, linking it to peloton should not affect the execution at all - but somehow it does. 

[1] https://github.com/tcm-marcel/peloton/blob/8d9e716228a3fb757df3ea271bbeefa5e7e42a53/src/codegen/runtime_functions.cpp#L124

test/codegen/value_integrity_test.cpp
```diff
@@ -68,6 +68,7 @@ void DivideByZeroTest(const codegen::type::Type &data_type, ExpressionType op) {
 
   typedef void (*func)(CType);
   func f = (func)code_context.GetRawFunctionPointer(function.GetFunction());
+
   EXPECT_THROW(f(0), DivideByZeroException);
```
ChTimTsubasa: You may also want to check the error message in this exception. This is what this PR is trying to do.

src/include/common/exception.h
```diff
@@ -128,76 +128,43 @@ class Exception : public std::runtime_error {
     }
   }
 
-  // Based on :: http://panthema.net/2008/0901-stacktrace-demangled/
-  static void PrintStackTrace(FILE *out = ::stderr,
-                              unsigned int max_frames = 63) {
-    ::fprintf(out, "Stack Trace:\n");
+  static void PrintStackTrace(FILE *out = ::stderr) {
```
ChTimTsubasa: Can you provide a comment on what is this one is trying to do as you have deleted the previous reference link?

Besides, I know it is not you did this but it is actually bad to have the implementations in header files. Can you move them to the exception.cpp?

tcm-marcel:
I just took a look at the Mac build. I forked the branch to try some changes with travis: https://github.com/tcm-marcel/peloton/commit/a19cff29428f4a1e1119827f4e8bbc1793f118ea

Installing *libunwind* on Mac only installs the missing official headers, so the library itself is already available somehow. [1] But CMake then fails when trying to find the actual executable of the library to link. [2] I tried not adding libunwind as a linking dependency on Mac, only the headers are included. If the linking works I can only see if the other parts run through. 

For *liblzma* I simply added `brew install xz` to the install script. 

But now the Mac build throws different compiler errors:  [3]
```
In file included from /Users/travis/build/tcm-marcel/peloton/src/binder/bind_node_visitor.cpp:13:
In file included from /Users/travis/build/tcm-marcel/peloton/src/include/expression/expression_util.h:19:
In file included from /Users/travis/build/tcm-marcel/peloton/src/include/catalog/catalog.h:18:
In file included from /Users/travis/build/tcm-marcel/peloton/src/include/function/functions.h:18:
/Users/travis/build/tcm-marcel/peloton/src/include/type/value.h:34:29: error: expected unqualified-id
  return boolean ? CmpBool::TRUE : CmpBool::FALSE;
                            ^
/usr/include/mach/boolean.h:81:14: note: expanded from macro 'TRUE'
#define TRUE    1
```

On first sight it looks like if a Mac-specific header file (mach/boolean.h) is messing up our identifiers with pre-processor definitions. A simple hack (ifdef/undef) in our header files may help, but I haven't looked at this yet. 

[1] https://github.com/cmu-db/peloton/pull/995/files#diff-46454d00238f98e83096b2d44703df7eR193
[2] https://github.com/cmu-db/peloton/blame/master/cmake/Modules/FindLibunwind.cmake#L15
[3] https://travis-ci.org/tcm-marcel/peloton/jobs/342621704#L2445

PR #1001, opened 2017-12-24
Improved network error handling

tli2:
This is the last network layer refactor PR. I mainly worked on improving error handling using wrappers and exceptions. There is much unfinished work but I should have gotten to most things above protocol handler.

Protocol Handler and traffic cop are both so complicated we probably need a summer rewriting them (if we ever want to)

### DIFF
--- a/src/include/network/error_util.h
+++ b/src/include/network/error_util.h
@@ -12,7 +12,6 @@
 
 #pragma once
 
-#include <exception>
 #include <string>
 #include <utility>
 
--- a/src/include/network/error_util.h
+++ b/src/include/network/error_util.h
@@ -1,0 +1,89 @@
+//===----------------------------------------------------------------------===//
+//
+//                         Peloton
+//
+// error_util.h
+//
+// Identification: src/include/network/error_util.h
+//
+// Copyright (c) 2015-17, Carnegie Mellon University Database Group
+//
+//===----------------------------------------------------------------------===//
+
+#pragma once
+
+#include <exception>
+#include <string>
+#include <utility>
+
+namespace peloton {
+namespace network {
+/**
+ * Used to signal that something has gone wrong when processing network connections.
+ *
+ * More details should be included in the error message and accessed with what().
+ */
+struct NetworkProcessException : public std::exception {
+public:
+  explicit NetworkProcessException(std::string error_msg): error_msg_(std::move(error_msg)) {}
+
+  const char *what() const _NOEXCEPT override {
+    return exception::what();
+  }
+private:
+  std::string error_msg_;
+};
+
+class EventUtil {
+private:
+  template <typename T>
+  static bool NotNull(T *ptr) {
+    return ptr != nullptr;
+  }
+
+  static bool IsZero(int arg) {
+    return arg == 0;
+  }
+
+  static bool NonNegative(int arg) {
+    return arg >= 0;
+  }
+
+  template <typename T>
+  static T Wrap(T value, bool(*check)(T), const char *error_msg) {
+    if (!check(value)) throw NetworkProcessException(error_msg);
+    return value;
+  }
+
+public:
+  EventUtil() = delete;
+
+  static struct event_base *EventBaseNew() {
+    return Wrap(event_base_new(), NotNull<struct event_base>, "Can't allocate event base");
+  }
+
+  static int EventBaseLoopExit(struct event_base *base, const struct timeval *timeout) {
+    return Wrap(event_base_loopexit(base, timeout), IsZero, "Error when exiting loop");
+  }
+
+  static int EventDel(struct event *event) {
+    return Wrap(event_del(event), IsZero, "Error when deleting event");
+  }
+
+  static int EventAdd(struct event *event, const struct timeval *timeout) {
+    return Wrap(event_add(event, timeout), IsZero, "Error when adding event");
+  }
+
+  static int EventAssign(struct event *event, struct event_base *base, int fd,
+                         short flags, event_callback_fn callback, void *arg) {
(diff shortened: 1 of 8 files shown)


schedutron:
This might be lame, but in [error_util.h](src/include/network/error_util.h), wouldn't it be better to write `LogErrno()` like the following?

```
for (int i=0; i < ERRNOS.size(); i++){  //ERRNOS contains EINTR, EAGAIN, EBADF, ...
    if (errno = ERRNOS[i]){
        LOG_TRACE("Error Writing: %s", #ERRNOS[i]);
        break;
    }
}

if (i < ERRNOS.size())
    LOG_TRACE("Error Writing: UNKNOWN");
```

tli2:
This matters little because:

- this is hidden in a function anyways
- this is inherited code from before

I agree that if I were to do it I would do something differently, like keeping a static std::map<int, std::string> from errno to their names. However, a loop does not seem like a good idea since it is first of all slower and second of all non-idiomatic.

src/include/network/notifiable_task.h
```diff
@@ -32,6 +26,7 @@
 #include "common/logger.h"
 #include "container/lock_free_queue.h"
 #include "network_state.h"
+#include "network/error_util.h"
```
dengyuchenkit: Can we remove `network/`?

src/include/network/notifiable_task.h
```diff
@@ -59,20 +54,14 @@ class NotifiableTask {
...
-  explicit NotifiableTask(int task_id)
-      : task_id_(task_id) {
-    base_ = event_base_new();
-    // TODO(tianyu) Error handling here can be better perhaps?
-    if (base_ == nullptr) {
-      LOG_ERROR("Can't allocate event base\n");
-      exit(1);
-    }
+  explicit NotifiableTask(int task_id) : task_id_(task_id) {
+    base_ = EventUtil::EventBaseNew();
```
dengyuchenkit: Are we handling exceptions when we create `NotifiableTask` like `ConnectionDispatcherTask` or `ConnectionHandlerTask` now?

vittvolt:
@dengyuchenkit One question, does the postgres parser handle exceptions well and send through the network thread?

dengyuchenkit:
@vittvolt The network layer will catch the exception thrown by the parser and send back the error message. The parser detects different exceptions now but I am not sure what do you mean by `parser handle exception well`? Are you talking about particular exceptions that should be detected by parser? If it's only about network thread catching exception, then it's yes.

src/include/network/notifiable_task.h
tli2: They should both only interact with libevent through the NotifiableTask interface. (the base is private and not accessible to child classes)

vittvolt:
@dengyuchenkit Thanks for the clarification. The problem I found was that exceptions thrown in postgresparser cause peloton the terminate, though the exception info is printed.

Not sure if it is related to this PR.

dengyuchenkit:
@vittvolt >< I think this PR is more about refactoring. Can you please share the queries you run? Will check that...

src/include/network/notifiable_task.h
tli2: Will do.

dengyuchenkit:
@vittvolt Sorry, some corrections. This PR is removing `EXIT`. And I'm not sure about the terminate you mention. Could you share the queries?

apavlo:
@vittvolt We should catch those and send them back to the client.

@dengyuchenkit This PR is separate from your work on passing exceptions from the worker threads to the networking thread, right?

src/include/network/error_util.h
```diff
@@ -0,0 +1,135 @@
...
+namespace peloton {
+namespace network {
+/**
+ * Static utility class for some common error handling code
+ */
+class ErrorUtil {
+public:
+  ErrorUtil() = delete;
+
+  static void LogErrno() {
```
apavlo: What is this function? And why are we using `LOG_TRACE` instead of `LOG_ERROR`.

dengyuchenkit:
@apavlo Yes, it's separate.

src/include/network/notifiable_task.h
```diff
@@ -171,29 +160,21 @@ class NotifiableTask {
...
-//      LOG_ERROR("Failed to update event");
-//      PL_ASSERT(false);
-//    }
-//    event_add(event, timeout);
-//  }
-//  void UpdateManualEvent(struct event *event, event_callback_fn callback, void *arg) {
-//    UpdateEvent(event, -1, EV_PERSIST, callback, arg);
-//  }
+// TODO(tianyu): The original network code seems to do this as an optimization. I am leaving this out until we get numbers
+  void UpdateEvent(struct event *event,
```
apavlo: Can we add documentation for all of these parameters?

src/network/connection_handle.cpp
```diff
@@ -448,36 +374,33 @@ bool ConnectionHandle::ProcessInitialPacket(InputPacket *pkt) {
...
-    LOG_ERROR("Protocol error: Only protocol version 3 is supported.");
-    exit(EXIT_FAILURE);
-  }
+  if (PROTO_MAJOR_VERSION(proto_version) != 3)
+    throw NetworkProcessException("Protocol error: Only protocol version 3 is supported.");
 
   // TODO: check for more malformed cases
   // iterate till the end
+  // TODO(tianyu): WTF is this?
   for (;;) {
```
apavlo: Yeah seriously?? What is this???

src/include/network/error_util.h
tli2: This is migrated from clearly copy-and-pasted errno handling in the protocol handler. I suppose I can make it better as well.

src/include/network/notifiable_task.h
tli2: Yes

PR #1004, opened 2017-12-24
Add support for AExpr and FuncCall as parameter in ParamListTransform

haojin2:
This PR provides support for using AExpr and FuncCall as parameters in ParamListTransform together with some fixes for possible memory leaks under exceptions.
- The parser will be able to support queries like "EXECUTE plan(1+1)" and "EXECUTE plan(sqrt(1))"
- Adds corresponding tests for the newly-supported queries.
- Adds extensive fixes for handling memory leak issue in transform functions for various statments.


apavlo:
@sivaprasadsudhir @aaron-tian This adds support for `EXECUTE` SQL in the parser. This will allow us to implement real stored procedures.

Issue #1003, opened 2017-12-24
Refactor CreateExecutor::DExecute


apavlo:
The CreateExecutor is the one of the few executors that we plan on keeping after we switch over the LLVM engine. Right now it's main implementation method is a giant switch statement. We need to break this up into separate functions (one for each thing that we need to create).

apavlo:
Merged in #1037

PR #1005, opened 2017-12-25
[Cleanup] Refactor CreateExecutor

apavlo:
Eh, so you this does what I want for #1003, but we still need to clean up the logic for how we report errors in the different methods.


apavlo:
I am going to abandon this for now and come back to it later.

PR #1011, opened 2017-12-28
Drop executor fix

ChTimTsubasa:
Fix the drop database executor that previously only drop database from default databases;
Add drop database function.

### DIFF
--- a/src/executor/drop_executor.cpp
+++ b/src/executor/drop_executor.cpp
@@ -15,9 +15,7 @@
 #include "catalog/catalog.h"
 #include "catalog/trigger_catalog.h"
 #include "common/logger.h"
-#include "concurrency/transaction_context.h"
 #include "executor/executor_context.h"
-#include "planner/drop_plan.h"
 
 namespace peloton {
 namespace executor {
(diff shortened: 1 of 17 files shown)


src/executor/drop_executor.cpp
```diff
@@ -43,22 +43,48 @@ bool DropExecutor::DExecute() {
   const planner::DropPlan &node = GetPlanNode<planner::DropPlan>();
   DropType dropType = node.GetDropType();
   switch (dropType) {
+    case DropType::DB: {
```
apavlo: I know that you didn't write all of this code, but can you break this up the logic for the different drop targets into separate functions? See #1003 and my attempt to do this in `CreateExecutor` here:

https://github.com/cmu-db/peloton/pull/1005/commits/958681019845e35598b7cc7c4d06e63c881aaf1d

src/executor/drop_executor.cpp
```diff
@@ -43,22 +43,48 @@ bool DropExecutor::DExecute() {
...
+      auto database_name = node.GetDatabaseName();
+      auto current_txn = context_->GetTransaction();
+      if (node.IsMissing()) {
+        try {
+          auto database_object = catalog::Catalog::GetInstance()->GetDatabaseObject(
+              database_name, current_txn);
+        } catch (CatalogException &e) {
+          LOG_TRACE("Database %s does not exist.", database_name.c_str());
```
apavlo: Do we want to make this `LOG_ERROR` or `LOG_WARN`? Actually, we need to document what the correct error handling logic is here (and `CreateExecutor`). Should we just return false or should let the exception go up the stack.

src/executor/drop_executor.cpp
```diff
@@ -43,22 +43,48 @@ bool DropExecutor::DExecute() {
...
+    }
     case DropType::TABLE: {
+      auto database_name = node.GetDatabaseName();
       auto table_name = node.GetTableName();
       auto current_txn = context_->GetTransaction();
 
       if (node.IsMissing()) {
         try {
           auto table_object = catalog::Catalog::GetInstance()->GetTableObject(
-              DEFAULT_DB_NAME, table_name, current_txn);
```
apavlo: Yes!!!

src/include/parser/parsenodes.h
```diff
@@ -571,6 +571,12 @@ typedef struct DropStmt {
   bool concurrent;       /* drop index concurrently? */
 } DropStmt;
 
+typedef struct DropdbStmt {
```
apavlo: Maybe rename to `DropDatabaseStmt`?

src/include/parser/postgresparser.h
```diff
@@ -214,6 +214,9 @@ class PostgresParser {
   // transform helper for drop statement
   static parser::DropStatement *DropTransform(DropStmt *root);
 
+  // transform helper for drop database statement
+  static parser::DropStatement* DropDbTransform(DropdbStmt* root);
```
apavlo: Rename to `DropDatabaseTransform` to be consistent with the struct typedef.

src/include/planner/drop_plan.h
```diff
@@ -60,8 +62,11 @@ class DropPlan : public AbstractPlan {
   DropType drop_type = DropType::TABLE;
```
apavlo: Public data members :-1: 
I will write an issue for @haojin2 to fix these.

src/parser/postgresparser.cpp
```diff
@@ -1071,7 +1071,11 @@ parser::SQLStatement *PostgresParser::CreateTriggerTransform(
 parser::SQLStatement *PostgresParser::CreateDbTransform(CreatedbStmt *root) {
   parser::CreateStatement *result =
       new parser::CreateStatement(CreateStatement::kDatabase);
-  result->database_name = root->dbname;
+  result->table_info_.reset(new parser::TableInfo());
+  result->table_info_->database_name = root->dbname;
+
+  // TODO(Tianyi) More options need to be converted
```
apavlo: Such as?

src/parser/postgresparser.cpp
```diff
@@ -1088,7 +1092,21 @@ parser::DropStatement *PostgresParser::DropTransform(DropStmt *root) {
   }
 }
 
-parser::DropStatement *PostgresParser::DropTableTransform(DropStmt *root) {
+// This function takes in a Postgres Dropdbstmt parsenode
```
apavlo: Doxygen format. Move to header declaration too.

src/planner/drop_plan.cpp
```diff
@@ -25,14 +25,31 @@ DropPlan::DropPlan(const std::string &name) {
...
-  if (parse_tree->type == parser::DropStatement::EntityType::kTable) {
-    table_name = parse_tree->GetTableName();
-    missing = parse_tree->missing;
-  } else if (parse_tree->type == parser::DropStatement::EntityType::kTrigger) {
-    // note parse_tree->table_name is different from parse_tree->GetTableName()
-    table_name = std::string(parse_tree->table_name_of_trigger);
-    trigger_name = std::string(parse_tree->trigger_name);
-    drop_type = DropType::TRIGGER;
+  switch (parse_tree->type) {
+    case parser::DropStatement::EntityType::kDatabase :{
```
apavlo: extra space before ":"

src/planner/drop_plan.cpp
```diff
@@ -25,14 +25,31 @@ DropPlan::DropPlan(const std::string &name) {
...
+    case parser::DropStatement::EntityType::kTrigger: {
+      // note parse_tree->table_name is different from parse_tree->GetTableName()
+      database_name = parse_tree->GetDatabaseName();
+      table_name = std::string(parse_tree->table_name_of_trigger);
+      trigger_name = std::string(parse_tree->trigger_name);
+      drop_type = DropType::TRIGGER;
+      break;
+    }
+    default: {
+      LOG_ERROR("Not supported Drop type");
```
apavlo: @haojin2 Should we throw an exception here?

test/executor/drop_test.cpp
```diff
@@ -51,34 +90,32 @@ TEST_F(DropTests, DroppingTable) {
   std::unique_ptr<catalog::Schema> table_schema2(
       new catalog::Schema({id_column, name_column}));
 
-  catalog->CreateDatabase(DEFAULT_DB_NAME, txn);
+  catalog->CreateDatabase("test_db", txn);
```
apavlo: Maybe add `#define TESTING_DEFAULT_DB_NAME` for this?

src/include/parser/parsenodes.h
ChTimTsubasa: The reason I name this to DropdbStmt is because Create is named "CreatedbStmt". Do you want me to rename both?

src/parser/postgresparser.cpp
ChTimTsubasa: There is a list of options in the CreateDbNode and they are discarded in the node transform.
https://github.com/cmu-db/peloton/blob/master/src/include/parser/parsenodes.h#L625

I think these options support more complex create database queries mentioned in https://www.postgresql.org/docs/10/static/sql-createdatabase.html

I can add the above reference to the comments here in case others can take a chance and implement it.

src/executor/drop_executor.cpp
ChTimTsubasa: Sure, I would change these.

src/include/parser/parsenodes.h
apavlo: Yes, please rename both.

PR #946, opened 2017-12-07
Codegen Index Scan

huangminchn:
This PR adds support for index scan code generation. (require fixes from #936 and #937 )

PR #1013, opened 2017-12-28
Codegen (Block) Nested Loop Joins

pmenon:
This PR adds block nest-loop joins to the codegen engine. This brings us one step closer to achieving feature parity with the interpreted engine (and one step closer to phasing it out).

Initially this PR had sort-merge joins and union/intersect/extract too, but with all the required infrastructure changes the PR was too complex. I'll submit those once this goes through.

In this PR:
1. The main BNLJ logic.
2. Ability to define auxiliary producer functions aside from the "main" plan function. This will be needed for outer joins later on.
3. Separate a function's declaration from definition so that the function can be used before it is defined (i.e., implemented).
4. Ability to reset parameter cache so that it can be reused across multiple functions.
5. Optimize parameter cache to only generate null checks if parameter is nullable.
6. Optimize zone map scanning logic.
7. Removed all local (i.e., on-stack) from runtime state. Now local variables can be safely defined **anywhere** during code generation. This makes it much easier for beginners since they don't have to worry about whether allocation happens deeply in a loop.
8. Enabled NLJ plans to participate in parameterization and caching.
9. Random formatting.

Tests for all of the above are included.

### DIFF
--- a/src/codegen/vector.cpp
+++ b/src/codegen/vector.cpp
@@ -16,7 +16,7 @@ namespace peloton {
 namespace codegen {
 
 // The default vector size is 2048 elements
-std::atomic<uint32_t> Vector::kDefaultVectorSize{2048};
+std::atomic<uint32_t> Vector::kDefaultVectorSize{1024};
 
 // The default byte-alignment of all vectors is 32 bytes
 uint32_t Vector::kDefaultVectorAlignment = 32;
--- a/src/codegen/vector.cpp
+++ b/src/codegen/vector.cpp
@@ -15,7 +15,7 @@
 namespace peloton {
 namespace codegen {
 
-// The default vector size is 2048 elements
+// The default vector size is 1024 elements
 std::atomic<uint32_t> Vector::kDefaultVectorSize{1024};
 
 // The default byte-alignment of all vectors is 32 bytes
--- a/src/codegen/buffering_consumer.cpp
+++ b/src/codegen/buffering_consumer.cpp
@@ -76,11 +76,6 @@ void BufferingConsumer::Prepare(CompilationContext &ctx) {
   auto &runtime_state = ctx.GetRuntimeState();
   consumer_state_id_ =
       runtime_state.RegisterState("consumerState", codegen.CharPtrType());
-
-  // Introduce our output tuple buffer as local (on stack)
-  auto *value_type = ValueProxy::GetType(codegen);
-  tuple_output_state_id_ = runtime_state.RegisterState(
-      "output", codegen.ArrayType(value_type, output_ais_.size()), true);
 }
 
 // For each output attribute, we write out the attribute's value into the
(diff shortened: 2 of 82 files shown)


apavlo:
@pmenon How easy will it be to modify this to use indexes once #946 is merged?

seojungmin:
@apavlo I think the index scan, if implemented correctly, should be orthogonal to this.

pmenon:
@apavlo The short answer is that modifying BNLJ to support indexes is easy, but we probably shouldn't do it this way. This BNLJ materializes tuples that we can avoid when doing index nested-loop joins.

phisiart:
I encountered the same error when doing #1016. This is a failure that happens randomly.

The reason is that you allocate a buffer `type::Value[N]` on the stack without initializing it. Therefore, when calling functions in https://github.com/cmu-db/peloton/blob/master/src/codegen/values_runtime.cpp, you are effectively destructing objects that don't exist.

For example, if you do `vals[idx] = type::ValueFactory::GetTinyIntValue(val);`, then you are calling `operator=` on `vals[idx]`, which will first free the resources in `vals[idx]`. But `vals[idx]` is garbage memory before assignment.

The solution is to directly call the constructor on the buffer, like this:
```
new (&vals[idx]) type::Value(type::ValueFactory::GetTinyIntValue(val));
```

pmenon:
@phisiart Thanks for the catch.  

@apavlo I think there is another more subtle bug that this PR exposes. On our dev machines, master fails a few tests with LLVM 3.9, but passes with 3.7. I spent a little time and tracked it down to how we're doing vectorized scans. @seojungmin Can you take a look at this?

seojungmin:
@pmenon I just read this. I was busy clearing the snow on my door way haha :-) I hope you don't have to do this there.  I think this is *not* related to this PR.  The current branch also crashes at  `ScanWithSubtractColumnsPredicate`.  I will keep it posted when I find more.

pmenon:
@seojungmin It isn't related to this PR. Master seems to be broken on dev4/5 for the last few weeks. If I remember correctly, it fails predictably only on `ScanWithSubtractColumnsPredicate` and `ScanWithAddColumnsPredicate` and only only LLVM 3.9 (4.0 works).

pmenon:
@seojungmin All comments addressed. Take a second look.

src/codegen/vector.cpp
```diff
@@ -16,7 +16,7 @@ namespace peloton {
 namespace codegen {
 
 // The default vector size is 2048 elements
-std::atomic<uint32_t> Vector::kDefaultVectorSize{2048};
+std::atomic<uint32_t> Vector::kDefaultVectorSize{1024};
```
seojungmin: Can you comment this change?

apavlo:
@pmenon I will merge this one next when the build passes.

apavlo:
@pmenon This got messed up when I tried to fix the conflict. I will try to rebase/merge by hand but I may need you to do it.

apavlo:
@pmenon I got it to rebase cleanly. I will merge manually.

apavlo:
Merged in a0426df19e9057d0d7d16ec9796cb556f5ca8d98

PR #1014, opened 2017-12-28
Parser support for a range of types of queries

haojin2:
This PR adds support for a few types of queries in parser with corresponding tests
- With this PR the parser will be able to support queries like "CREATE INDEX ind ON t USING SKIPLIST (col);"
- With this PR the parser will be able to support queries like "CREATE SCHEMA schema_name;"
- With this PR the parser will be able to support queries like "CREATE VIEW view_name AS query;"
- Corresponding tests
Note: Postgres' default index type is "BTREE", while Peloton's is "BWTREE", as we do not currently support BTREE index type, the default index type is now set to BWTREE if no type is specified in the query.

### DIFF
--- a/src/common/internal_types.cpp
+++ b/src/common/internal_types.cpp
@@ -2370,9 +2370,9 @@ std::string LayoutTypeToString(LayoutType type) {
       return "HYBRID";
     }
     default: {
-      throw ConversionException(StringUtil::Format(
-          "No string conversion for LayoutType value '%d'",
-          static_cast<int>(type)));
+      throw ConversionException(
+          StringUtil::Format("No string conversion for LayoutType value '%d'",
+                             static_cast<int>(type)));
     }
   }
   return "INVALID";
@@ -2382,9 +2382,6 @@ std::ostream &operator<<(std::ostream &os, const LayoutType &type) {
   os << LayoutTypeToString(type);
   return os;
 }
-
-
-
 
 type::TypeId PostgresValueTypeToPelotonValueType(PostgresValueType type) {
   switch (type) {
--- a/src/common/internal_types.cpp
+++ b/src/common/internal_types.cpp
@@ -841,6 +841,8 @@ IndexType StringToIndexType(const std::string &str) {
   std::string upper_str = StringUtil::Upper(str);
   if (upper_str == "INVALID") {
     return IndexType::INVALID;
+  } else if (upper_str == "BTREE") {
+    return IndexType::BWTREE;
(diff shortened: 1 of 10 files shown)


test/parser/postgresparser_test.cpp
```diff
@@ -1100,5 +1105,36 @@ TEST_F(PostgresParserTests, TypeCastTest) {
...
+
+    if (result->is_valid == false) {
+      LOG_ERROR("Message: %s, line: %d, col: %d", result->parser_msg,
+                result->error_line, result->error_col);
+    }
+    EXPECT_EQ(result->is_valid, true);
+
+    LOG_TRACE("%d : %s", ++ii, result->GetInfo().c_str());
+  }
+  std::vector<std::string> invalid_queries;
```
apavlo: Add a comment here that says that these are index types that we do **not** currently support.

src/parser/parser_utils.cpp
```diff
@@ -24,7 +24,7 @@
 namespace peloton {
 namespace parser {
 
-std::string ParserUtils::GetTableRefInfo(const TableRef* table,
+std::string ParserUtils::GetTableRefInfo(const TableRef *table,
```
apavlo: Stupid question: Is there a reason why this code shouldn't be in `TableRef::GetInfo()` instead of a separate util class?

haojin2: I guess we should ask Joy, this file was some of the older ones in this folder and carried over from the previous version of parser.

test/parser/postgresparser_test.cpp
haojin2: Done.

apavlo:
@haojin2 Can you rebase and resolve conflicts?

haojin2:
Sure, will do after I finish lunch (on my way to lunch right now)

haojin2:
Just rebased and resolved conflicts, should be ready for merge once tests pass.

src/common/internal_types.cpp
```diff
@@ -2368,9 +2370,9 @@ std::string LayoutTypeToString(LayoutType type) {
       return "HYBRID";
     }
     default: {
-      throw ConversionException(StringUtil::Format(
-          "No string conversion for LayoutType value '%d'",
-          static_cast<int>(type)));
+      throw ConversionException(
+          StringUtil::Format("No string conversion for LayoutType value '%d'",
```
apavlo: This should be "IndexType" not "LayoutType"

PR #1016, opened 2017-12-29
[Multithread] Step 1: Introduce lang::LocalVariable

phisiart:
When we change into multithreaded execution, we will have various task functions.

Each function has its own local variables.

It will be no longer feasible store track local variables in RuntimeState, because we
cannot initialize them at once.

### DIFF
--- a/src/codegen/buffering_consumer.cpp
+++ b/src/codegen/buffering_consumer.cpp
@@ -13,6 +13,7 @@
 #include "codegen/buffering_consumer.h"
 
 #include "codegen/lang/if.h"
+#include "codegen/lang/local_variable.h"
 #include "codegen/proxy/proxy.h"
 #include "codegen/proxy/value_proxy.h"
 #include "codegen/proxy/values_runtime_proxy.h"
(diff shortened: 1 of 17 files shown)


seojungmin:
@phisiart I briefly looked at it, and I like the abstraction over the local variable.  Apparently, @pmenon has changed this part recently, and I think you can improve the implementation of `lang::LocalVariable` with `AllocBuffer`, which @pmenon has used in #1013.

pmenon:
@phisiart The changes I made in #1013 allow definition of local variables without runtime state. I think this achieves what you want in this PR. You may want to still wrap `codegen.Allocate..` calls in your `lang::LocalVariable` class, but at that point the class isn't doing much other than clearly marking some special values as local variables without special behavior.

phisiart:
I feel like with stuff in #1013 we don't need this PR anymore. I don't really like introducing new classes. Let me reconsider.

phisiart:
I'm killing this PR. Let's start with Step 2 #1017 .

PR #1017, opened 2017-12-29
[Multithread] Step 2: Task Chain

phisiart:
Due to pipeline breakers, each query is naturally split up into "stages".

One stage must be completed before the next one starts.

In this PR we change the compiled engine so that each stage is generated as a separate function.

More specifically, the `Produce` function now returns `std::vector<CodeGenStage>` which contains a list of `llvm::Function *`'s.

In execution, each stage function is submitted to MonoQueuePool, and when that stage finishes, it is responsible for submitting the next stage to the pool.


phisiart:
I'm working on this.

src/codegen/query.cpp
```diff
@@ -70,44 +92,42 @@ void Query::Execute(std::unique_ptr<executor::ExecutorContext> executor_context,
...
+    result.m_result = ResultType::SUCCESS;
+    result.m_processed = context->num_processed;
+    on_complete(result);
 
-  // No need to cleanup if we get an exception while cleaning up...
-  if (stats != nullptr) {
-    timer.Stop();
-    stats->tear_down_ms = timer.GetDuration();
-  }
+    delete[] param;
```
seojungmin: Reason going back to non-smart pointers?

src/include/codegen/stage.h
```diff
@@ -0,0 +1,35 @@
...
+enum class StageKind {
+  SINGLE_THREADED,
+};
+
+struct CodeGenStage {
+  StageKind kind_;
+  llvm::Function *llvm_func_;
+};
+
+struct Stage {
```
seojungmin: What are we expecting these for?

src/codegen/query.cpp
phisiart: These objects need to have a lifetime throughout the execution of the query.

Now that the query is split into separate stage functions, all these functions need to have access to the objects. This is done by lambda capture.

However in C++11 we don't have move-capture. A broader problem is that C++ doesn't have non-copy-able `std::function`s (`folly::Function` does that).

Therefore, we just go back to manually destroying the objects in the last function.

src/include/codegen/stage.h
phisiart: In the next PR, we will have a different kind of stage - multi-threaded scan. Going forward, we could also have multi-thread aggregate stage.

For the multi-threaded scan stage, we have a single task function, and rely on `ExecuteStages` to submit tasks to the queue. Note that this means:
- Each "stage function" invocation happens in a single thread, so we don't need to do synchronization stuff within LLVM;
- Submitting tasks is all done in C++, which makes code more clear (hopefully?) and exception handling easier (because we can just wrap the call to task functions with try-catch).

PR #1022, opened 2017-12-29
[Cleanup] Cleaning up types.h

apavlo:
Three things in this one. The goal of this PR is to remove confusion between 'type/types.h' and 'type/type.h'.

* `CmpBool` is now in 'types.h' and is now in the `peloton` namespace (it used to be `peloton::type`).
* Removed some unused variables from 'types.h'
* Renamed 'type/types.h' to ' common/internal_types.h' 

https://www.youtube.com/watch?v=BHrkdzvSs_Q


malin1993ml:
I think this is good to go. It's a subtle video, though.

PR #898, opened 2017-11-11
Adding ceiling function

selwyni:
Added the ceiling function code, including the alias for CEIL and CEILING, in addition to a SQL test and a unit test for the utility function. Ceiling function has support for type Decimal, Integer, Big/Small/Tiny/Integer.

PR #1023, opened 2017-12-30
Minor type related fixes/additions to Floor function

### LINKED ISSUE #927: Changes needed in floor SQL function
Some changes are needed in the `floor` function implementation/tests added via PR #889  as are being suggested in the `ceiling` function PR #898 .

saatviks:
Just a few minor additions needed in the Floor function to make it correctly compatible across different numeric types. More or less following the additional suggestions provided in #898.

There was also some duplicate code related to the `ceiling` function - just removed the duplicate part.

Resolves #927

### DIFF
--- a/src/catalog/catalog.cpp
+++ b/src/catalog/catalog.cpp
@@ -1064,41 +1064,6 @@ void Catalog::InitializeFunctions() {
                                     function::DecimalFunctions::_Ceil},
           txn);
 
-      AddBuiltinFunction(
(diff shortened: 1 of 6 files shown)


apavlo:
@saatviks Do we have test cases for these different types?

saatviks:
Yea, this had been tested in `decimal_functions_test.cpp` earlier itself.
The SQL test however is only for decimal types, since `floor` is basically a no-op for other integer types.

PR #1028, opened 2018-01-03
Fix packages.sh execution on Ubuntu 15.x and 16.x

### LINKED ISSUE #1027: packages.sh execution fails on Ubuntu 16.04+
Running `sudo bash packages.sh` fails on Ubuntu 16.04 due to a command rename and deprecated command line option of apt-get.

Ubuntu 16.04+ renamed `cmake3` command to just `cmake`. Also, the apt-get's  `--force-yes` was deprecated and only runs if we remove it (or change it to one of the `--allow` prefixes). After doing those changes it will install the dependencies and successfully proceed the building of Peloton. Didn't check Ubuntu 15+ to see if it builds though. Finally, `cmake` version installed was `3.5.1`, not the latest it seems.

eribeiro:
Fix #1027

### DIFF
--- a/script/installation/packages.sh
+++ b/script/installation/packages.sh
@@ -31,11 +31,25 @@ if [ "$DISTRO" = "UBUNTU" ]; then
         sudo apt-get update -qq
     fi
 
-    sudo apt-get -qq --force-yes --ignore-missing -y install \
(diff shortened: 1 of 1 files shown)


eribeiro:
@apavlo I am sorry, I had a last minute typo that passed over my head. More details on PR https://github.com/cmu-db/peloton/pull/1030

PR #1031, opened 2018-01-03
Support Abs SQL Function

pervazea:
The tests provide similar coverage to other functions. They need
to be enhanced to cover all possible types supported.

Tested (so far) with debug build on Ubuntu 14.04, using pull on 3 Jan 2018.


apavlo:
This replaces #894

apavlo:
@pervazea I think you're missing the additional functions for the integer types. See @pmenon's comment here:

https://github.com/cmu-db/peloton/pull/894#pullrequestreview-79600329

You should follow what we do for the `floor` function. It supports both decimals and integers.

pervazea:
You are correct, the integer functions are missing. So are the tests to exercise the integer functions.  Some of this is now clearer as I work on the sqrt support.

The sql tests are going to become repetitive and should be re-factored to minimize code. Probably will do the initial verbose version for expediency, then follow up with cleaned up tests, unless there is objection.

pervazea:
Please review again, to see if I'm still missing anything.

src/catalog/catalog.cpp
```diff
@@ -995,6 +1000,34 @@ void Catalog::InitializeFunctions() {
...
+
+      AddBuiltinFunction(
+          "abs", {type::TypeId::INTEGER}, type::TypeId::DECIMAL, internal_lang,
+          "Abs",
+          function::BuiltInFuncType{OperatorId::Abs,
+                                    function::DecimalFunctions::_Abs},
+          txn);
+
+      AddBuiltinFunction(
+          "abs", {type::TypeId::BIGINT}, type::TypeId::DECIMAL, internal_lang,
```
pmenon: The return types of these built-ins need to match the input types for `abs`

src/codegen/type/integer_type.cpp
```diff
@@ -189,6 +189,23 @@ struct Floor : public TypeSystem::UnaryOperatorHandleNull {
...
+    return type.GetSqlType() == Integer::Instance();
+  }
+
+  Type ResultType(UNUSED_ATTRIBUTE const Type &val_type) const override {
+    return Type{Integer::Instance()};
+  }
+
+  Value Impl(UNUSED_ATTRIBUTE CodeGen &codegen,
+               const Value &val) const override {
+    PL_ASSERT(SupportsType(val.GetType()));
```
pmenon: You need to implement the `abs` function here. I'd do something like:

```c++
Value Impl(CodeGen &codegen, const Value &val) const override {
  // The integer subtraction implementation
  Sub sub;
  // Zero place-holder
  auto zero = codegen::Value{type::Integer::Instance(), codegen.Const32(0)};

  // We want: raw_ret = (val < 0 ? 0 - val : val)
  auto sub_result = sub.Impl(codegen, zero, left, OnError::Exception);
  auto *lt_zero = codegen->CreateICmpSLT(left.GetValue(), zero.GetValue());
  auto *raw_ret = codegen->CreateSelect(lt_zero, sub_result.GetValue(), left.GetValue());
  return Value{Integer::Instance(), raw_ret};
}
```

pervazea:
Please review updated version. Thanks.

pmenon:
@pervazea You'll need similar implementations for `tinyint`, `smallint`, and `bigint` too.

pervazea:
WIP changes. At a minimum, need to move abs so it is in the unary functions block of code.

src/codegen/type/bigint_type.cpp
```diff
@@ -464,6 +370,127 @@ struct Modulo : public TypeSystem::BinaryOperatorHandleNull {
   }
 };
 
+////////////////////////////////////////////////////////////////////////////////
+///
+/// Unary operations
+///
+////////////////////////////////////////////////////////////////////////////////
+
+// Abs
```
pmenon: Any reason for moving all these functions? Convention for all other types is to put all unary functions above the binary ones.

test/sql/decimal_functions_sql_test.cpp
```diff
@@ -76,6 +114,221 @@ class DecimalFunctionsSQLTest : public PelotonTest {};
...
+                                    rows_affected, error_message);
+    for (i = 0; i < numEntries; i++) {
+      std::string result_id(
+              TestingSQLUtil::GetResultValueAsString(result, (2 * i)));
+      std::string result_income(
+              TestingSQLUtil::GetResultValueAsString(result, (2 * i) + 1));
+      int id = std::stoi(result_id);
+      int income = std::stoi(result_income);
+      EXPECT_EQ(id, i);
+      EXPECT_EQ(income, abs(inputs[i]));
```
pmenon: Can you use [`TestingSQLUtil::ExecuteSQLQueryAndCheckResult`](https://github.com/cmu-db/peloton/blob/master/test/include/sql/testing_sql_util.h#L81)? I find it to be much clearer to see what the test is doing.

pervazea:
@pmenon - changes made as requested.

pmenon:
@pervazea Looks good. Just add tests that insert NULL and it should be good to go.

PR #1034, opened 2018-01-06
Remove C++14 usage 

dengyuchenkit:
This PR removes the `make_unique` introduced in #785 
@aaron-tian Could you check if it's good now? Thanks!

### DIFF
--- a/test/network/exception_test.cpp
+++ b/test/network/exception_test.cpp
@@ -72,7 +72,7 @@ void *ParserExceptionTest(int port) {
     } catch (const pqxx::pqxx_exception &e) {
       const pqxx::sql_error *s = dynamic_cast<const pqxx::sql_error*>(&e.base());
       if (s) {
-        LOG_INFO("Invalid Drop Query: %s", e.base().what());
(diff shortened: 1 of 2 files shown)


test/network/exception_test.cpp
```diff
@@ -72,7 +72,7 @@ void *ParserExceptionTest(int port) {
     } catch (const pqxx::pqxx_exception &e) {
       const pqxx::sql_error *s = dynamic_cast<const pqxx::sql_error*>(&e.base());
       if (s) {
-        LOG_INFO("Invalid Drop Query: %s", e.base().what());
+        LOG_TRACE("Invalid Drop Query: %s", e.base().what());
```
apavlo: Maybe switch to `LOG_WARN`?

dengyuchenkit: @apavlo We want this test to catch the exception, to check the Peloton reports error when user inputs an invalid query (at parsing stage). Then shall we put `log_warn` here?

aaron-tian:
LGTM. Sorry for the late reply. Thanks for removing and merging it.

Issue #1043, opened 2018-01-10
oltpbench: tatp issues


pervazea:
Running oltpbench tatp, peloton complains about an unsupported data type and terminates.

```
pp:49:CreateNewConnection] INFO  - Create new connection: id = 66
2018-01-10 10:15:54 [/home/proj/peloton-pa/peloton/src/network/network_connection.cpp:425:ProcessInitialPacket] INFO  - protocol version: 196608
2018-01-10 10:15:55 [/home/proj/peloton-pa/peloton/src/network/postgres_protocol_handler.cpp:777:ReadParamValue] ERROR - Do not support data type: 21
peloton: /home/pakhtar/proj/peloton-pa/peloton/src/network/postgres_protocol_handler.cpp:781: static size_t peloton::network::PostgresProtocolHandler::ReadParamValue(peloton::network::InputPacket*, int, std::vector<int>&, std::vector<std::pair<peloton::type::TypeId, std::basic_string<char> > >&, std::vector<peloton::type::Value>&, std::vector<short int>&): Assertion `(param_values[param_idx].GetTypeId() != type::TypeId::INVALID)' failed.
Aborted (core dumped)
```

[tatp_config.xml.txt](https://github.com/cmu-db/peloton/files/1619356/tatp_config.xml.txt)

Configuration file attached, as generated by benchmark.py. Ran manually, to capture output from peloton. Config:
- Ubuntu 14.04 LTS
- debug build of Peloton

apavlo:
`PostgresValueType` "21" is `SMALLINT`. How do we not support that???

pervazea:
The fixes in #1045 do not resolve the problem. Ran a debug build, built from branch apavlo/jan2018-cleanup

Last bit of Peloton output:
```
CreateNewConnection] INFO  - Create new connection: id = 66
2018-01-11 15:07:02 [/home/proj/apavlo/peloton/src/network/network_connection.cpp:425:ProcessInitialPacket] INFO  - protocol version: 196608
terminate called after throwing an instance of 'peloton::Exception'
  what():  Numeric value out of range.
```
Appended is the stack trace from gdb, for what seems to be the offending thread. Someone familiar with the network code should take a look.

```
#0  0x00007fd29bf81858 in __gnu_cxx::operator!=<unsigned char const*, std::vector<unsigned char, std::allocator<unsigned char> > > (__lhs=56 '8', 
[...]

PR #1045, opened 2018-01-10
Postgres Binary Protocol Fixes for SMALLINT + TINYINT

apavlo:
This is to fix the problem in #1043. I have to admit that I have not tested this with a JDBC connection.

https://www.youtube.com/watch?v=qPr1UTU2UO0


src/network/postgres_protocol_handler.cpp
```diff
@@ -730,7 +730,30 @@ size_t PostgresProtocolHandler::ReadParamValue(
         PL_ASSERT(param_values[param_idx].GetTypeId() != type::TypeId::INVALID);
       } else {
         // BINARY mode
-        switch (static_cast<PostgresValueType>(param_types[param_idx])) {
+        PostgresValueType pg_value_type = static_cast<PostgresValueType>(param_types[param_idx]);
+        switch (pg_value_type) {
+          case PostgresValueType::TINYINT: {
+            int int_val = 0;
```
apavlo: This should probably be int8_t

src/network/postgres_protocol_handler.cpp
```diff
@@ -730,7 +730,30 @@ size_t PostgresProtocolHandler::ReadParamValue(
...
+            for (size_t i = 0; i < sizeof(int); ++i) {
+              int_val = (int_val << 8) | param[i];
```
apavlo: Can somebody explain what this bit-shifting is doing?

src/network/postgres_protocol_handler.cpp
```diff
@@ -730,7 +730,30 @@ size_t PostgresProtocolHandler::ReadParamValue(
...
+            }
+            bind_parameters[param_idx] =
+                std::make_pair(type::TypeId::TINYINT, std::to_string(int_val));
+            param_values[param_idx] =
+                type::ValueFactory::GetTinyIntValue(int_val).Copy();
+            break;
+          }
+          case PostgresValueType::SMALLINT: {
+            int int_val = 0;
+            for (size_t i = 0; i < sizeof(int); ++i) {
```
pervazea: This one I'd expect to be sizeof(int6_t)

src/network/postgres_protocol_handler.cpp
```diff
@@ -730,7 +730,30 @@ size_t PostgresProtocolHandler::ReadParamValue(
         PL_ASSERT(param_values[param_idx].GetTypeId() != type::TypeId::INVALID);
       } else {
         // BINARY mode
-        switch (static_cast<PostgresValueType>(param_types[param_idx])) {
+        PostgresValueType pg_value_type = static_cast<PostgresValueType>(param_types[param_idx]);
+        switch (pg_value_type) {
+          case PostgresValueType::TINYINT: {
+            int int_val = 0;
+            for (size_t i = 0; i < sizeof(int); ++i) {
```
pervazea: Why sizeof(int)? Why not something like sizeof(int8_t)?

src/network/postgres_protocol_handler.cpp
pervazea: (I followed some of the code to lower layers, but didn't get the clarity desired). However, if we need to pull data off the wire, are there not functions to do this, rather than this approach?

apavlo:
@pervazea This adds a new test in the JDBC validator. From what I can tell, this is working now.

pervazea:
I had hacked up a variant of the JDBC test(s) to exercise tinyint, and fail to see (extra) logging that shows it going through the binary transfer path. The test passes though. So I'll probably run some further tests with your new JDBC test.

PR #1047, opened 2018-01-11
Add support for extract and date_part function

### LINKED ISSUE #637: Invoke EXTRACT function correctly
The Postgres parser parses the EXTRACT function into date_part function so now Peloton cannot invoke the EXTRACT function correctly.
``` sql
postgres=# create table a(t timestamp);
create table a(t timestamp) 0
postgres=# insert into A values('2017-01-02 12:13:14.999999+00');
insert into A values('2017-01-02 12:13:14.999999+00') 1
postgres=# select extract(century from t) from a;
function date_part not found.
```

date_part should be made as an alias to EXTRACT function and the parser should be modified to feed the right DatePartType into the arguments of the EXTRACT function.
https://github.com/cmu-db/peloton/blob/master/src/expression/date_functions.cpp#L33
https://github.com/cmu-db/peloton/blob/master/src/parser/postgresparser.cpp#L411
https://github.com/cmu-db/peloton/blob/master/src/catalog/catalog.cpp#L904

sivaprasadsudhir:
This PR adds support for date_part sql function. The parser maps calls to extract function to appropriate calls to the date_part function. 

This fixes #637


test/function/timestamp_functions_test.cpp
```diff
@@ -168,5 +168,141 @@ TEST_F(TimestampFunctionsTests, DateTruncYearTest) {
...
+  // Invoke the DatePart method and get back the result
+  auto result =
+      function::TimestampFunctions::DatePart(char_date_part, int_date);
+
+  LOG_INFO("COMPARED: %s, %f = %f \n",
+        expected.c_str(), double_expected,
+        result);
+
+  // Check that result is *not* null
+  EXPECT_NE(result, type::PELOTON_DECIMAL_NULL);
```
pmenon: For these expectations, the expected value is the first argument and the actual value is the second argument. Can you change these `EXPECT_...` assertions appropriately?

test/sql/timestamp_functions_sql_test.cpp
```diff
@@ -86,5 +86,97 @@ TEST_F(TimestampFunctionsSQLTest, DateTruncTest) {
...
+  std::vector<ResultValue> result;
+  std::vector<FieldInfo> tuple_descriptor;
+  std::string error_message;
+  int rows_affected;
+
+  // wrong argument type
+  std::string test_query = "select date_part('123', value) from foo;";
+  TestingSQLUtil::ExecuteSQLQuery(test_query.c_str(), result, tuple_descriptor,
+                                  rows_affected, error_message);
+  EXPECT_EQ(result.size(), 0);
```
pmenon: You're not testing the return value here. In fact, for all tests here, can you use `TestingSQLUtil::ExecuteSQLQueryAndCheckResult` to verify the return value?

sivaprasadsudhir:
@apavlo , Yes.

test/sql/timestamp_functions_sql_test.cpp
sivaprasadsudhir: The result of a few queries is empty and that is where the return value is not checked. Instead `EXPECT_EQ(result.size(), 0);` check is done. 

I think this was done in tests as
`TestingSQLUtil::ExecuteSQLQueryAndCheckResult` has a line `unsigned int rows = result.size() / tuple_descriptor.size();` which gives a floating point exception when the result is empty.

I can add a check whether the expected result is empty if `result.size() == 0`  in `TestingSQLUtil::ExecuteSQLQueryAndCheckResult`. Should I add it or is the current format fine?

I have changed other tests to use `TestingSQLUtil::ExecuteSQLQueryAndCheckResult`

pervazea:
Looks complete now, yes? Needs Prashanth to confirm.

sivaprasadsudhir:
@pervazea I have resolved the conflicts and rebased the branch. It's mostly complete. But I am waiting for Prashanth's reply to a query on `TestingSQLUtil::ExecuteSQLQueryAndCheckResult`.

test/sql/timestamp_functions_sql_test.cpp
sivaprasadsudhir: @pervazea What do you think?

pmenon:
@sivaprasadsudhir Can you run the formatter on the files again? Then it should be good to go.

sivaprasadsudhir:
Done. This can be merged after the build passes.

apavlo:
@sivaprasadsudhir Ack. I will merge.

PR #1051, opened 2018-01-11
Enable memory allocations from codegen

pmenon:
This PR enables calling functions that allocate memory from codegen. We simply pass along the executor context (which has a memory pool) to function implementations. I've also included some cleanup.


seojungmin:
I am happy to review, and I will do it tonight.

src/include/planner/plan_util.h
```diff
@@ -4,7 +4,7 @@
 //
 // plan_util.h
 //
-// Identification: /peloton/src/include/planner/plan_util.h
+// Identification: src/include/planner/plan_util.h
 //
 // Copyright (c) 2015, Carnegie Mellon University Database Group
```
seojungmin: @pmenon has missed this! I don't believe this!

PR #1052, opened 2018-01-11
 [Cleanup] Add clang builds in travis

tcm-marcel:
This PR 
  * adds clang builds in travis and 
  * adapts the sources not to rise any warnings/errors in clang *(in progress)*

---

Following files have classes with unused private members. I thought adding `UNUSED_ATTRIBUTE` could temporary help to get rid of the clang warnings, but apparently gcc doesn't support this for class members. So we should take a look at all of these and remove them if possible. 
```
/src/include/codegen/operator/update_translator.h
/src/include/codegen/type/type_system.h
/src/include/common/exception.h
/src/include/concurrency/local_epoch.h
/src/include/executor/hash_join_executor.h
/src/include/index/bloom_filter.h
/src/include/logging/logical_checkpoint_manager.h
/src/include/logging/logical_log_manager.h
/src/include/network/postgres_protocol_handler.h
/src/include/optimizer/child_property_generator.h
/src/include/optimizer/cost_and_stats_calculator.h
/src/include/optimizer/property_enforcer.h
/src/include/optimizer/query_property_extractor.h
/src/include/optimizer/query_to_operator_transformer.h
/src/include/optimizer/stats.h
/src/include/optimizer/stats/hyperloglog.h
/src/include/optimizer/tuple_sample.h
/src/include/storage/tile_group_header.h
```

In addition the linker fails because of missing symbols in protobuf, I haven't yet figured out why. The linker doesn't get any paths for protobuf, but with gcc it works somehow. 
```
../lib/libpeloton-d.so.0.0.5: undefined reference to `google::protobuf::Message::InitializationErrorString() const'
../lib/libpeloton-d.so.0.0.5: undefined reference to `google::protobuf::Message::GetTypeName() const'
../lib/libpeloton-d.so.0.0.5: undefined reference to `google::protobuf::internal::empty_string_'
../lib/libpeloton-d.so.0.0.5: undefined reference to `__atomic_compare_exchange'
../lib/libpeloton-d.so.0.0.5: undefined reference to `__atomic_load'
../lib/libpeloton-d.so.0.0.5: undefined reference to `google::protobuf::internal::StringTypeHandlerBase::New()'
```

Cmake can be configured to use clang as follows:
```
cmake -DCMAKE_BUILD_TYPE=Debug -DCMAKE_C_COMPILER=clang-3.7 -DCMAKE_CXX_COMPILER=clang++-3.7 ..
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,7 +27,7 @@ matrix:
                     - g++-4.8
           env:
             # NAME has no actual use, just to make the travis jobs overview more clear
-            - NAME="gcc4.8 Debug (Coveralls)"
+            - NAME="gcc4.8 Debug+Tests+Coveralls"
             - CXX=g++-4.8
             - CC=gcc-4.8
             - PELOTON_BUILD_TYPE=Debug
@@ -62,7 +62,7 @@ matrix:
                     - g++-4.8
           env:
             # NAME has no actual use, just to make the travis jobs overview more clear
-            - NAME="gcc4.8 Release"
+            - NAME="gcc4.8 Release+Tests"
             - CXX=g++-4.8
             - CC=gcc-4.8
             - PELOTON_BUILD_TYPE=Release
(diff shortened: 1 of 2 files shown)


pervazea:
I discussed this PR with Marcel last week. He stated that he would likely not complete this PR, but create a new PR for clang builds in Travis. Therefore closing this PR.

Issue #1038, opened 2018-01-08
Make sure that we don't use brackets for #include


apavlo:
I fixed some `#include`'s where we were using the brackets (`<...>`) instead of quotation marks (`"..."`):

https://github.com/cmu-db/peloton/pull/1037/commits/9e72651dcf1ff89c2a9fb117094843fa629a7994

It would nice if the validator script could check this automatically and prevent this from happening. For each `#include` that uses brackets, it can simply check to see whether it's one of our own files and throw an error if it is.

tli2:
Just as an aside that CLion-generated includes are always `<>` for some reason (it is stupid).  This could be one of the reasons people use `<>`.

Also, is there a  style guide on whether to use always `#include "dir/file"` or is `#include file` fine for a file in dir?

apavlo:
I thought that the convention was `<>` for something you expect to be installed in the system and `""` for something that is local to the project?

malin1993ml:
Yes. At least that's the convention that I'm following.

schedutron:
Can I work on adding this to the validator script?

tcm-marcel:
@schedutron Sure, give it a try. Make sure to rebase your repo, as #991 got merged in today. You can then create a PR and I will review it.

schedutron:
Do we need to check only the header files in the `src/include` folder?

tcm-marcel:
All files (header and implementation) must be checked for includes that use brackets `<...>`. If `...` is an existing file in `src/include/...`, an error should be thrown. Maybe check for relatives path both with and without `include/`.

schedutron:
No, I mean the file mentioned in `#include <...>` would be present only in `src/include`, so I was wondering that, for a given `#include <...>` statement, should I check for the file's presence in only `src/include` or other directories as well?

tcm-marcel:
Only in `src/include`, but make sure that the check also works if the include path starts with `include/...` (which is wrong, but otherwise me miss those).

tcm-marcel:
Added in #1054.

PR #1054, opened 2018-01-13
Add include check in validator script

schedutron:


schedutron:
Fixes #1038.

script/validators/source_validator.py
```diff
@@ -219,8 +219,35 @@ def check_namespaces(file_path):
...
+                    status = False
+                    LOG.info("Line %s: %s" % (linenum, line))
+        if status:
+            LOG.info('Use "..." for including local headers.')
+    return status
+
+
 VALIDATORS = [
     check_common_patterns,
+    check_format,
```
tcm-marcel: Don't add this here. `check_format` is already in the list a few lines further down.

script/validators/source_validator.py
```diff
@@ -219,8 +219,35 @@ def check_namespaces(file_path):
...
+def check_includes(file_path):
+    """Checks whether local includes are done via #include<...>"""
+    with open(file_path, "r") as f:
+        path_pattern = re.compile(r'<(.*?)>')
+        includes = os.listdir('src/include')
+        linenum = 0
+        status = True
+        for line in f:
+            linenum += 1
+            if line.startswith('#include <'):
```
tcm-marcel: Did you know, you can actually test this also in the regex: https://regexr.com/3j66j (with multiline flag)

script/validators/source_validator.py
```diff
@@ -219,8 +219,35 @@ def check_namespaces(file_path):
...
+                word = path[0]
+                if word == 'include':
+                    word = path[1]
+                if word in includes:
+                    if status:
+                        LOG.info("Invalid include in %s" % file_path)
+                    status = False
+                    LOG.info("Line %s: %s" % (linenum, line))
+        if status:
+            LOG.info('Use "..." for including local headers.')
```
tcm-marcel: I think it should say `if not status:`. And please make the hint more useful, write something like "includes for peloton header files have must not have brackets".

script/validators/source_validator.py
```diff
@@ -219,8 +219,35 @@ def check_namespaces(file_path):
...
+        status = True
+        for line in f:
+            linenum += 1
+            if line.startswith('#include <'):
+                path = path_pattern.search(line).groups()[0]
+                path = path.split('/')
+                word = path[0]
+                if word == 'include':
+                    word = path[1]
+                if word in includes:
```
tcm-marcel: Be aware that if you only check the first directory, we might get false positives. For now it works, but if you want to make it more fancy, create a set with *all* files of /src/include at the beginning of the script and check if the path exists in this list (path without `include/`).

schedutron: _All_ means all the possible paths of `src/include`?

script/validators/source_validator.py
schedutron: That's quite cool! But then I'd have to dump all the lines in the source at once (using `f.readlines()`), right?

schedutron:
Pardon, but I don't understand what are "PRs for changes that are in progress". Can you please elaborate further?

tcm-marcel:
Oh, sorry. I wanted to tell you, that we not only create pull requests (PRs) when we are finished with our changes, but often when we still work on them (PR in progress). So if you want someone to review your pull request, just write a short comment saying that it is ready for review - so we know you are done with your changes. (Otherwise we might oversee it)

script/validators/source_validator.py
tcm-marcel: You can test every line with the regex and check if a match was found or not. You can also regex the whole file (cmp. [1]), but then you need to output the whole line as you are missing the line numbers. 

[1] https://github.com/schedutron/peloton/blob/cbd505df0b3b3539bc2cf9fa8f4ee15f17c6b299/script/validators/source_validator.py#L172

script/validators/source_validator.py
tcm-marcel: It's a lot, I know. (474, see below) But if we put them in a set (not a list) I hope that python does a fast lookup. And once at the beginning of the script, not for every file. I think you can use `glob` for that. 

```
$ find ./src/include/ -type f | wc -l
474
```

schedutron:
@tcm-marcel Can you add the "ready for review" badge for #1056? It is ready.

script/validators/source_validator.py
schedutron: I used `glob` and I'm getting 470. Perhaps because I traversed only the paths ending with `.h`?

schedutron:
@tcm-marcel Thanks for the advice you mentioned! I've modified the validator accordingly, have a look.

script/validators/source_validator.py
```diff
@@ -219,8 +223,31 @@ def check_namespaces(file_path):
     return status
 
 
+def check_includes(file_path):
+    """Checks whether local includes are done via #include<...>"""
+    with open(file_path, "r") as f:
+        path_pattern = re.compile(r'^#include <(.*?)>')
+        includes = os.listdir('src/include')
```
tcm-marcel: variable `includes` not needed anymore

script/validators/source_validator.py
```diff
@@ -219,8 +223,31 @@ def check_namespaces(file_path):
...
+        linenum = 0
+        status = True
+        for line in f:
+            linenum += 1
+            res = path_pattern.match(line)
+            if res:
+                path = res.groups()[0].lstrip("include/")
```
tcm-marcel: `lstrip` also doesn't work here. But I just realized you can deal with this also in the regex itself: `^#include <(include\/)?(.*?)>` (and then take the second match group)

script/validators/source_validator.py
```diff
@@ -51,6 +52,9 @@
     os.path.join(PELOTON_DIR, "test")
 ]
 
+# To be used in check_includes.
+PATHS = set([path.lstrip('src/include') for path in glob.glob('src/include/**/*.h')])
```
tcm-marcel: `lstrip` doesn't work as intended here, as it takes a set of characters, and not a string. Use string replace or just cut off the first x characters array access.

script/validators/source_validator.py
```diff
@@ -219,8 +223,31 @@ def check_namespaces(file_path):
...
+        for line in f:
+            linenum += 1
+            res = path_pattern.match(line)
+            if res:
+                path = res.groups()[0].lstrip("include/")
+                if path in PATHS:
+                    if status:
+                        LOG.info("Invalid include in %s" % file_path)
+                    status = False
+                    LOG.info("Line %s: %s" % (linenum, line))
```
tcm-marcel: Please strip the line before printing it to avoid additional new lines. (cmp. https://github.com/schedutron/peloton/blob/84b8947b388b7686a3b4e19aef2be1c5fc03bba9/script/validators/source_validator.py#L152)

schedutron:
> Can you add a new commit that fixes the includes in those files
So I should run the validator on all the source files, and fix the includes where needed, right?

schedutron:
Hey, is the source validator script written for Python2 or Python3?

tcm-marcel:
The script is written in Python 2. Btw the build failed because of a package problem we currently have, I restarted the job and it should work then.

PR #1056, opened 2018-01-14
Fix segfault for query on nonexistent tables

### LINKED ISSUE #1055: Segfault when inserting into a nonexistent table
When I write a query like `insert into foo values(33);`, where `foo` is a table not yet created, the client hangs and Peloton server logs show `Segmentation fault: 11`.

schedutron:
Fixes #1055. I had to dig deep into `lldb` for this! I hope I've added the `throw` at the appropriate place. Using throw here also prevents the segfault from happening on, say `update` queries operating on a table that hasn't been created yet.

### DIFF
--- a/src/include/common/exception.h
+++ b/src/include/common/exception.h
@@ -129,8 +129,6 @@ class Exception : public std::runtime_error {
     }
   }
 
-  ExceptionType GetType() { return type; }
-
   // Based on :: http://panthema.net/2008/0901-stacktrace-demangled/
   static void PrintStackTrace(FILE *out = ::stderr,
                               unsigned int max_frames = 63) {
--- a/src/include/common/exception.h
+++ b/src/include/common/exception.h
@@ -56,7 +56,7 @@ enum class ExceptionType {
   CONNECTION = 21,        // connection related
   SYNTAX = 22,            // syntax related
   SETTINGS = 23,          // settings related
-  BINDER = 24,             // settings related
+  BINDER = 24,            // settings related
   NETWORK = 25
 };
 
@@ -70,8 +70,8 @@ class Exception : public std::runtime_error {
   Exception(ExceptionType exception_type, std::string message)
       : std::runtime_error(message), type(exception_type) {
     exception_message_ = "Exception Type :: " +
-                                    ExceptionTypeToString(exception_type) +
-                                    "\nMessage :: " + message;
+                         ExceptionTypeToString(exception_type) +
+                         "\nMessage :: " + message;
   }
 
   std::string ExceptionTypeToString(ExceptionType type) {
@@ -200,7 +200,7 @@ class Exception : public std::runtime_error {
     }
   }
 
-  friend std::ostream& operator<<(std::ostream& os, const Exception& e);
+  friend std::ostream &operator<<(std::ostream &os, const Exception &e);
 
  private:
   // type
@@ -433,8 +433,9 @@ class ConnectionException : public Exception {
 class NetworkProcessException : public Exception {
   NetworkProcessException() = delete;
 
-  public:
-  NetworkProcessException(std::string msg) : Exception(ExceptionType::NETWORK, msg) {}
+ public:
+  NetworkProcessException(std::string msg)
+      : Exception(ExceptionType::NETWORK, msg) {}
 };
 
 class SettingsException : public Exception {
--- a/src/include/common/exception.h
+++ b/src/include/common/exception.h
@@ -129,6 +129,8 @@ class Exception : public std::runtime_error {
     }
   }
 
+  ExceptionType GetType() { return type; }
+
   // Based on :: http://panthema.net/2008/0901-stacktrace-demangled/
   static void PrintStackTrace(FILE *out = ::stderr,
                               unsigned int max_frames = 63) {
--- a/src/optimizer/query_to_operator_transformer.cpp
+++ b/src/optimizer/query_to_operator_transformer.cpp
@@ -26,7 +26,6 @@
 #include "parser/statements.h"
 
 #include "catalog/manager.h"
-#include "common/exception.h"
 
 using std::vector;
 using std::shared_ptr;
--- a/src/optimizer/query_to_operator_transformer.cpp
+++ b/src/optimizer/query_to_operator_transformer.cpp
@@ -26,6 +26,7 @@
 #include "parser/statements.h"
 
 #include "catalog/manager.h"
+#include "common/exception.h"
 
 using std::vector;
 using std::shared_ptr;
@@ -231,6 +232,10 @@ void QueryToOperatorTransformer::Visit(parser::InsertStatement *op) {
       catalog::Catalog::GetInstance()
           ->GetDatabaseObject(op->GetDatabaseName(), txn_)
           ->GetTableObject(op->GetTableName());
+  
+  if (target_table == nullptr)
+      throw CatalogException("Table " + op->GetTableName() + " is not found");
+
   if (op->type == InsertType::SELECT) {
     auto insert_expr = std::make_shared<OperatorExpression>(
         LogicalInsertSelect::make(target_table));
--- a/src/optimizer/query_to_operator_transformer.cpp
+++ b/src/optimizer/query_to_operator_transformer.cpp
@@ -232,9 +232,6 @@ void QueryToOperatorTransformer::Visit(parser::InsertStatement *op) {
       catalog::Catalog::GetInstance()
           ->GetDatabaseObject(op->GetDatabaseName(), txn_)
           ->GetTableObject(op->GetTableName());
-  
-  if (target_table == nullptr)
-      throw CatalogException("Table " + op->GetTableName() + " is not found");
 
   if (op->type == InsertType::SELECT) {
     auto insert_expr = std::make_shared<OperatorExpression>(
--- a/test/sql/insert_sql_test.cpp
+++ b/test/sql/insert_sql_test.cpp
@@ -63,10 +63,12 @@ void CreateAndLoadTable4() {
       "i CHAR, j VARCHAR, k VARBINARY, l BOOLEAN);");
 
   // Insert tuples into table
-  TestingSQLUtil::ExecuteSQLQuery("INSERT INTO test4 VALUES "
+  TestingSQLUtil::ExecuteSQLQuery(
+      "INSERT INTO test4 VALUES "
       "(1, 2, 3, 4, 5.1, 6.1, '2017-10-10 00:00:00-00', "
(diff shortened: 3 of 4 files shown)


tcm-marcel:
Your build fails, fix this first!
https://travis-ci.org/cmu-db/peloton/jobs/328775088#L1405

schedutron:
Oh, sorry! I was missing a semicolon!

seojungmin:
Wouldn't this be a partial solution to the whole problem?  What happens if somebody _delete_ something from a table?

schedutron:
So what should be the appropriate place for checking this? The `select` statements are already validated in `catalog.cpp`, but there's no equivalent for `insert` ones. I think this is the appropriate place to check. I mentioned earlier that this also fixes the problem for `update` queries. And it _does_ work on `delete` queries as well. Quite universal! :)

pmenon:
You need to move this check earlier in processing, ideally during binding. Add the check [here](https://github.com/cmu-db/peloton/blob/master/src/binder/bind_node_visitor.cpp#L177) for `InsertStatement`, similarly to how it's done for `DeleteStatement` [here](https://github.com/cmu-db/peloton/blob/master/src/binder/bind_node_visitor.cpp#L160). The call to `BinderContext:: AddRegularTable()` will throw a CatalogException for you if the table is not in the catalog.

schedutron:
@pmenon Done, and it works as expected!

schedutron:
@pervazea Can you please add back the "ready-for-review" tag?

src/optimizer/query_to_operator_transformer.cpp
```diff
@@ -231,6 +232,10 @@ void QueryToOperatorTransformer::Visit(parser::InsertStatement *op) {
       catalog::Catalog::GetInstance()
           ->GetDatabaseObject(op->GetDatabaseName(), txn_)
           ->GetTableObject(op->GetTableName());
+  
+  if (target_table == nullptr)
```
pmenon: At this point, I don't think `target_table` can be null. Have you found a scenario where the expectation is false?

schedutron: I don't think so. Should I remove this check? Is this code reachable from some other part, apart from a query coming from the postgres front-end?

schedutron: @pmenon  Should I remove this check?

pmenon: @schedutron Yes, it should be safe to remove.

pmenon:
@schedutron You also need to add a unit test.

schedutron:
@pmenon What should be an ideal place for the test? What about [`test/sql/insert_sql_test.cpp`](https://github.com/schedutron/peloton/blob/d69704797aceb3bd014caa9b7b3bc973ebe9276c/test/sql/insert_sql_test.cpp)?

pmenon:
@schedutron Sure, that looks okay.

schedutron:
@pmenon I've added a test.

src/optimizer/query_to_operator_transformer.cpp
pervazea: Wasn't this going to be removed?

schedutron: @pervazea Yes, I'll do that in another commit.

schedutron:
@pervazea Done :)

test/sql/insert_sql_test.cpp
```diff
@@ -570,5 +567,34 @@ TEST_F(InsertSQLTests, UniqueColumn) {
...
+          TestingSQLUtil::GeneratePlanWithOptimizer(optimizer, query, txn);
+      EXPECT_EQ(plan->GetPlanNodeType(), PlanNodeType::INSERT);
+      txn_manager.CommitTransaction(txn);
+    } catch (peloton::Exception &ex) {
+      EXPECT_EQ(ExceptionType::CATALOG, ex.GetType());
+      EXPECT_STREQ("Table NotExistTestTable is not found", ex.what());
+      throw peloton::CatalogException(ex.what());
+    }
+  }, peloton::CatalogException);
+  EXPECT_EQ(0, rows_changed);
```
pmenon: It would make the test clearer if you reduced the scope of the check to where you expect the exception to be thrown. Something like:

```c++
// Setup transaction, optimizer etc. here
std::string query = "...";
EXPECT_THROW(TestingSQLUtil::GeneratePlanWithOptimizer(optimizer, query, txn), 
             peloton::CatalogException);
```

src/optimizer/query_to_operator_transformer.cpp
```diff
@@ -26,6 +26,7 @@
 #include "parser/statements.h"
 
 #include "catalog/manager.h"
+#include "common/exception.h"
```
pmenon: Is this needed?

src/include/common/exception.h
```diff
@@ -129,6 +129,8 @@ class Exception : public std::runtime_error {
     }
   }
 
+  ExceptionType GetType() { return type; }
```
pmenon: With the changes to your test, do you still need this?

src/optimizer/query_to_operator_transformer.cpp
schedutron: We can still catch `CatalogException` without it, so I think it isn't needed. I included it while adding the test just to be safe. I'll remove it.

src/include/common/exception.h
schedutron: Yes, because we need to ensure that the exception of the appropriate type was thrown.

schedutron:
I don't know why the build is failing. The tests fail even for the master branch when I run them locally.

tcm-marcel:
@schedutron we had problems with the build system before, maybe your branch is based on a faulty version. Try to rebase and see if this solves the problem.

schedutron:
Should I rebase with master? The problem is that the build in the master branch is successful, but it makes the Peloton server break when a client connects (see #1141).

tcm-marcel:
I see... I do not know what's the cause of this. Try finding the last working commit in master or simply rebase with the current version, as you currently just need to know, if the build succeeds.

pmenon:
@schedutron It's failing the build because your test is failing. Please make the changes are suggested in my review.

schedutron:
@pmenon I think it's failing because the table name implementation gives lower case table names by default. I Googled this up to check whether SQL table names should be case insensitive and found that DBMSs provide the functionality to enable/disable it. Is there such a provision in Peloton?

The test will pass if I change the expected error message to have the table name in lowercase.

schedutron:
Also, I'd tried earlier to catch `peloton::CatalogException` directly as mentioned in your review, but that hadn't worked. That's why I have to catch the general `peloton::Exception` first and then check its type.

schedutron:
@pmenon The build passed this time.

pmenon:
@schedutron Can you explain why the suggestion of using `EXPECT_THROW` doesn't work? If you're throwing an exception, but you aren't able to catch it is your solution correct?

The test, as it is, is a little brittle - it won't work if we change how we print exceptions.

schedutron:
@pmenon When I was originally writing tests for this, my code was almost identical to what you suggested - `EXPECT_THROW(TestingSQLUtil::GeneratePlanWithOptimizer(optimizer, query, txn), 
             peloton::CatalogException)`, but rather than getting `peloton::CatalogException`, it got `peloton::Exception`. That's why I had to employ the style in my current test.

Why won't the test work if the exception print format changes? Yes, we'll have to change the expected string, but the type of the exception would stay the same, right?

schedutron:
@pmenon But I think I didn't make the test entirely atomic, I'll do that soon, then the test code will be much closer to what you suggested.

pmenon:
@schedutron That probably means we're catching your exception and throwing a different exception rather than re-throwing the original `exception::CatalogException`. Can you dig into where this is happening? I suspect somewhere during optimization.

schedutron:
@pmenon I'll do a debugger session.

schedutron:
@pmenon I'll do a debugger session, but if that's the case, why does `EXPECT_EQ(ExceptionType::CATALOG, ex.GetType())` executes without error?

schedutron:
@pmenon Catching `peloton::CatalogException` works fine, the problem with exception types mentioned above actually occurs in #1063. Sorry for the confusion.

schedutron:
Why did Jenkins fail? The Travis build passed though. @pmenon I think the test is fine now, right?

schedutron:
@pmenon is this PR mergeable now?

src/include/common/exception.h
pmenon: Is this being used anymore? If not remove it.

test/sql/insert_sql_test.cpp
```diff
@@ -570,5 +567,26 @@ TEST_F(InsertSQLTests, UniqueColumn) {
...
+      new optimizer::Optimizer());
+  // Insert an int into a non-existent table.
+  std::string query("INSERT INTO NonExistentTable VALUES(3);");
+  txn = txn_manager.BeginTransaction();
+  EXPECT_THROW({
+    try {
+      TestingSQLUtil::GeneratePlanWithOptimizer(optimizer, query, txn);
+    } catch (peloton::CatalogException &ex) {
+      EXPECT_STREQ("Table nonexistenttable is not found", ex.what());
+      throw peloton::CatalogException(ex.what());
```
pmenon: Just rethrow here with `throw;`

I'm still not a fan of string equality checks in tests. It creates a tight coupling and because in the long run they'll cause problems if/when we change the string format.

Can you simplify the check here to just fail if you get past the call to `TestingSQLUtil::GeneratePlanWithOptimizer`? You can remove the outer `EXPECT_THROW` .

schedutron: But if I'll rethrow, the test suite will report that as an error, right? If that's fine, I'll remove the outer `EXCEPT_THROW`. Otherwise, I can just do something like `EXCEPT_THROW(TestingSQLUtil::GeneratePlanWithOptimizer(optimizer, query, txn), peloton::CatalogException);`, just like what you suggested sometime earlier.

schedutron:
@pmenon I've removed the `GetType()` method, now I'll remove the `EXPECT_STREQ`s. Should I also remove you want, the outer `EXPECT_THROW`?

schedutron:
@pmenon Any update on this?

test/sql/insert_sql_test.cpp
pmenon: Remove the outer throw, just fail if you get past the call to `GeneratePlanWithOptimizer()`.

schedutron:
This is perhaps the best thing that happened to me today!

PR #1059, opened 2018-01-16
Move old string functions

pmenon:
This PR moves the string functions used by the interpreted engine into `OldEngineStringFunctions`. This is in preparation for 15-721. I also added the `Repeat` function to codegen to serve as an example.

### DIFF
--- a/src/catalog/catalog.cpp
+++ b/src/catalog/catalog.cpp
@@ -27,7 +27,7 @@
 #include "concurrency/transaction_manager_factory.h"
 #include "function/date_functions.h"
 #include "function/decimal_functions.h"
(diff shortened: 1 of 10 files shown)


seojungmin:
A small, can-be-ignored suggestion:  Would it be better to put the old ones in `src/deprecated/include/...` and to have `DeprecatedStringFunctions` as a namespace, so that we could keep the deprecated code in one place and let nobody add something on it unless it is _approved_?

apavlo:
@seojungmin We need this for tomorrow, so let's keep it simple.

apavlo:
Travis is busted. I am merging this in tomorrow morning.

Issue #1015, opened 2017-12-29
Formatter script not seeming to work


tli2:
```
MacBook-Pro:formatting Tianyu$ python formatter.py -c ../../src/include/network
12-28-2017 20:31:20 [<module>:170] INFO : Scanning directory ../../src/include/network
12-28-2017 20:31:20 [format_file:115] INFO : clang-format-3.6 -style=file  -i ../../src/include/network/connection_dispatcher_task.h
Traceback (most recent call last):
  File "formatter.py", line 171, in <module>
    format_dir(x, args.update_header, args.clang_format_code)
  File "formatter.py", line 132, in format_dir
    format_file(file_path, update_header, clang_format_code)
  File "formatter.py", line 116, in format_file
    subprocess.call([CLANG_FORMAT, "-style=file", "-i", file_path])
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/subprocess.py", line 522, in call
    return Popen(*popenargs, **kwargs).wait()
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/subprocess.py", line 710, in __init__
    errread, errwrite)
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/subprocess.py", line 1335, in _execute_child
    raise child_exception
OSError: [Errno 2] No such file or directory
```

In addition, the command-line flag -d mentioned in the wiki does not seem to be supported any more.

schedutron:
Is `clang-format-3.6` executing separately as a terminal command on your machine?

apavlo:
There are three outstanding PRs for the formatting script (#994, #993, #991).
I also thought that we agreed to not require an exact version of `clang-format`?

@tcm-marcel Can you figure out what's going on with all the PRs and this problem when you get back?

schedutron:
Even if an exact version is not used, it might happen that even `clang-format` command doesn't work on a person's machine. If one does a fresh install of `clang-format` (latest version), the `clang-format` command *still* won't work on their machine, but `clang-format-3.7` will.

So the formatter script needs to find the correct version/command for `clang-format` installed, but that may not be a trivial task to do.

tli2:
I can confirm that changing the script to use clang-format-3.7 fixes the issue on my machine.

tcm-marcel:
[...]

PR #1061, opened 2018-01-17
Formatter/Validator: find installed version of clang-format

tcm-marcel:
Adds some code that checks for different versions of `clang-format` to avoid hard-coding a single version. 
(as proposed by @pmenon in https://github.com/cmu-db/peloton/issues/1015#issuecomment-358187024)


script/formatting/formatter.py
```diff
@@ -140,6 +142,17 @@ def format_dir(dir_path, update_header, clang_format_code):
     #END FOR [os.walk]
 #END ADD_HEADERS_DIR(DIR_PATH)
 
+#find clang-format executable
+def find_clangformat():
```
pervazea: This is duplicated in source_validator.py. Would it be reasonable to move this into a library or common file and have a single implementation?

script/formatting/formatter.py
```diff
@@ -140,6 +142,17 @@ def format_dir(dir_path, update_header, clang_format_code):
...
+    global CLANG_FORMAT
+    #check for possible clang-format versions
+    for exe in ["clang-format", "clang-format-3.6", "clang-format-3.7", "clang-format-3.8"]:
+        path = distutils.spawn.find_executable(exe)
+        if not path is None:
```
pervazea: I find this harder to read than:
if path is not None:
or just
if path:

Issue #829, opened 2017-10-13
Peloton crashes on invalid INSERT statements instead of throwing an error


tcm-marcel:
The function  `planner::InsertPlan::InsertPlan` checks the validity of INSERT statements only with assertions instead of throwing an exception that can be handled. 

## Example 1
```sql
CREATE TABLE A (val1 INT, val2 INT,  val3 INT,  val4 INT);
INSERT INTO A (val1, val2, val3, val4, val5) VALUES (1, 1, 1, 1, 1)
```
```
peloton: peloton/src/planner/insert_plan.cpp:109: peloton::planner::InsertPlan::InsertPlan(peloton::storage::DataTable*, std::vector<char*>*, std::vector<std::vector<peloton::expression::AbstractExpression*>*>*): Assertion `(columns->size() <= table_schema->GetColumnCount())' failed.
```
## Example 2
```sql
CREATE TABLE A (val1 INT, val2 INT,  val3 INT,  val4 INT);
INSERT INTO A (id, val) VALUES (1, 1, 1, 1);
```
```
peloton: peloton/src/planner/insert_plan.cpp:122: peloton::planner::InsertPlan::InsertPlan(peloton::storage::DataTable*, std::vector<char*>*, std::vector<std::vector<peloton::expression::AbstractExpression*>*>*): Assertion `(col_cntr != INVALID_OID)' failed.
```

Ubuntu 16.04 x86_64 gcc

apavlo:
@hzxa21 @chenboy @luochenUmich -- Can you help @tcm-marcel with this problem?

apavlo:
This is likely. @haojin2 is working on a related fix in #808

haojin2:
I think if the query is having invalid syntax then it's the parser's responsibility to handle it. If it has correct syntax but the underlying operation is illegal, then other parts should be responsible for handling that. I'm now working on throwing exceptions for bad syntax, not all invalid queries.

tcm-marcel:
I may was inaccurate, it's not a syntax problem. So your PR hopefully will solve these ones.

apavlo:
Actually, I think that @tcm-marcel's example should be caught in the planner. The parser/binder doesn't "know" the number of columns that the table has.

Please confirm with @hzxa21 @chenboy @luochenUmich

hzxa21:
[...]

Issue #1062, opened 2018-01-17
Peloton crashes (segfault) during incorrect insert


schedutron:
If a table has just 2 columns and one tries to insert a tuple with 3 values, Peloton crashes. The server logs show: 
```
Assertion failed: ((values.size() <= schema->GetColumnCount())), function InsertPlan, file /Users/binaryboy/Desktop/peloton/src/planner/insert_plan.cpp, line 40.
Abort trap: 6
```

On the `psql` client, I get the following message:
```
server closed the connection unexpectedly
	This probably means the server terminated abnormally
	before or while processing the request.
The connection to the server was lost. Attempting reset: Failed.
```

PR #1063, opened 2018-01-17
Refactor insert plan for better error handling and more efficiency

schedutron:
For #1062. PL_ASSERT logs the exception on the server but doesn't care about the client, which suffers from unexpected disconnection. Instead, why not report the exception to the client directly?

### DIFF
--- a/src/planner/insert_plan.cpp
+++ b/src/planner/insert_plan.cpp
@@ -13,6 +13,7 @@
 #include "planner/insert_plan.h"
 
 #include "catalog/catalog.h"
+#include "common/exception.h"
 #include "expression/constant_value_expression.h"
 #include "storage/data_table.h"
 #include "type/ephemeral_pool.h"
@@ -21,8 +22,8 @@
 namespace peloton {
 namespace planner {
 
-InsertPlan::InsertPlan(storage::DataTable *table,
-    const std::vector<std::string> *columns,
+InsertPlan::InsertPlan(
+    storage::DataTable *table, const std::vector<std::string> *columns,
     const std::vector<std::vector<
         std::unique_ptr<expression::AbstractExpression>>> *insert_values)
     : target_table_(table), bulk_insert_count_(insert_values->size()) {
@@ -37,7 +38,10 @@ InsertPlan::InsertPlan(storage::DataTable *table,
     for (uint32_t tuple_idx = 0; tuple_idx < insert_values->size();
          tuple_idx++) {
       auto &values = (*insert_values)[tuple_idx];
-      PL_ASSERT(values.size() <= schema->GetColumnCount());
+      if (values.size() > schema->GetColumnCount())
+        throw CatalogException(
+            "ERROR: INSERT has more expressions than target columns");
+
       uint32_t param_idx = 0;
       for (uint32_t column_id = 0; column_id < values.size(); column_id++) {
         auto &exp = values[column_id];
@@ -66,28 +70,46 @@ InsertPlan::InsertPlan(storage::DataTable *table,
   }
   // INSERT INTO table_name (col1, col2, ...) VALUES (val1, val2, ...);
   else {
-    PL_ASSERT(columns->size() <= schema->GetColumnCount());
+    uint32_t tup_size = ((*insert_values)[0]).size();  // size of each tuple
+    auto &table_columns = schema->GetColumns();
+    auto table_columns_num = schema->GetColumnCount();
+    // check for equality of columns size and insert values size
+    if (columns->size() > tup_size)
+      throw CatalogException(
+          "ERROR:  INSERT has more target columns than expressions");
+    else if (columns->size() < tup_size)
+      throw CatalogException(
+          "ERROR:  INSERT has more expressions than target columns");
+
+    std::vector<size_t> specified(
+        columns->size());  // containing the indices of columns specified.
+    // Would it be more appropriate for 'specified' to be a map?
+    for (size_t idx = 0; idx != columns->size(); ++idx) {
+      auto col = (*columns)[idx];
+      // for(size_t idx = 0; idx < columns->size(); idx++){
+      // auto col = (*columns)[idx];
+      auto found = std::find_if(table_columns.begin(), table_columns.end(),
+                                [&col](const peloton::catalog::Column &x) {
+                                  return col == x.GetName();
+                                });
+      if (found == table_columns.end())
+        throw CatalogException("ERROR:  column \"" + col + "\" of relation \"" +
+                               target_table_->GetName() + "\" does not exist");
+      size_t column_id = std::distance(table_columns.begin(), found);
+      specified[idx] = column_id;
+    }
+
+    values_.resize(table_columns_num);
+
     for (uint32_t tuple_idx = 0; tuple_idx < insert_values->size();
          tuple_idx++) {
       auto &values = (*insert_values)[tuple_idx];
 
-      auto &table_columns = schema->GetColumns();
-      auto table_columns_num = schema->GetColumnCount();
       uint32_t param_idx = 0;
-      for (size_t column_id = 0; column_id < table_columns_num; column_id++) {
-        auto col = table_columns[column_id];
+
+      for (size_t idx = 0; idx != specified.size(); ++idx) {
+        auto column_id = specified[idx];
         const type::TypeId type = schema->GetType(column_id);
-        auto found = std::find_if(columns->begin(), columns->end(),
-            [&col](const std::string &x) { return col.GetName() == x; });
-        if (found == columns->end()) {
-          type::Value *v = schema->GetDefaultValue(column_id);
-          if (v == nullptr)
-            values_.push_back(type::ValueFactory::GetNullValueByType(type));
-          else
-            values_.push_back(*v);
-          continue;
-        }
-        auto idx = std::distance(columns->begin(), found);
         auto &exp = values[idx];
         if (exp->GetExpressionType() == ExpressionType::VALUE_PARAMETER) {
           std::tuple<oid_t, oid_t, oid_t> pair =
@@ -99,16 +121,32 @@ InsertPlan::InsertPlan(storage::DataTable *table,
           auto *const_exp =
               dynamic_cast<expression::ConstantValueExpression *>(exp.get());
           type::Value value = const_exp->GetValue().CastAs(type);
-          values_.push_back(value);
+          values_[column_id] = value;
         }
+      }
+
+      // Write defaults for unspecified columns.
+      size_t specified_idx = 0;
+      std::sort(specified.begin(),
+                specified.end());  // for efficient comparision
+      for (size_t column_id = 0; column_id < table_columns_num; column_id++) {
+        if (column_id == specified[specified_idx]) {
+          specified_idx++;
+          continue;  // column with index idx already specified
+        }
+        const type::TypeId type = schema->GetType(column_id);
+        type::Value *v = schema->GetDefaultValue(column_id);
+        if (v == nullptr)
+          values_[column_id] = type::ValueFactory::GetNullValueByType(type);
+        else
+          values_[column_id] = *v;
       }
     }
   }
 }
 
 type::AbstractPool *InsertPlan::GetPlanPool() {
-  if (pool_.get() == nullptr)
-    pool_.reset(new type::EphemeralPool());
+  if (pool_.get() == nullptr) pool_.reset(new type::EphemeralPool());
   return pool_.get();
 }
 
@@ -157,23 +195,19 @@ hash_t InsertPlan::Hash() const {
 }
 
 bool InsertPlan::operator==(const AbstractPlan &rhs) const {
-  if (GetPlanNodeType() != rhs.GetPlanNodeType())
-    return false;
+  if (GetPlanNodeType() != rhs.GetPlanNodeType()) return false;
 
   auto &other = static_cast<const planner::InsertPlan &>(rhs);
 
   auto *table = GetTable();
   auto *other_table = other.GetTable();
   PL_ASSERT(table && other_table);
-  if (*table != *other_table)
-    return false;
+  if (*table != *other_table) return false;
 
   if (GetChildren().size() == 0) {
-    if (other.GetChildren().size() != 0)
-      return false;
+    if (other.GetChildren().size() != 0) return false;
 
-    if (GetBulkInsertCount() != other.GetBulkInsertCount())
-      return false;
+    if (GetBulkInsertCount() != other.GetBulkInsertCount()) return false;
   }
 
   return AbstractPlan::operator==(rhs);
@@ -189,8 +223,9 @@ void InsertPlan::VisitParameters(
     for (uint32_t i = 0; i < values_.size(); i++) {
       auto value = values_[i];
       auto column_id = i % columns_num;
-      map.Insert(expression::Parameter::CreateConstParameter(value.GetTypeId(),
-          schema->AllowNull(column_id)), nullptr);
+      map.Insert(expression::Parameter::CreateConstParameter(
+                     value.GetTypeId(), schema->AllowNull(column_id)),
+                 nullptr);
       values.push_back(value);
     }
   } else {


schedutron:
@tcm-marcel Doesn't this also fix #829?

schedutron:
@tcm-marcel My latest commit solves the sort of problems mentioned in #829. I've "swapped" the order of nesting involved in creating the insert plan for better error checking. This also leads to some optimization as the earlier version was computing `table_columns` and `table_columns_num` repeatedly, but they really need to be computed just once. Please have a look and add a `ready for review` badge.

schedutron:
This also fixes #1069.

schedutron:
Also, should I delete the `specified` vector after the final loop where it is used? Or does C++ already do that, implicitly?

src/planner/insert_plan.cpp
```diff
@@ -66,28 +70,46 @@ InsertPlan::InsertPlan(storage::DataTable *table,
...
+    auto table_columns_num = schema->GetColumnCount();
+    // check for equality of columns size and insert values size
+    if (columns->size() > tup_size)
+      throw CatalogException(
+          "ERROR:  INSERT has more target columns than expressions");
+    else if (columns->size() < tup_size)
+      throw CatalogException(
+          "ERROR:  INSERT has more expressions than target columns");
+
+    std::vector<size_t> specified(
```
schedutron: Should I delete the `specified` vector defined here after the final loop where it is used? Or does C++ already do that, implicitly? Is there a potential memory leak here?

src/planner/insert_plan.cpp
```diff
@@ -66,28 +70,46 @@ InsertPlan::InsertPlan(storage::DataTable *table,
...
+        columns->size());  // containing the indices of columns specified.
+    // Would it be more appropriate for 'specified' to be a map?
```
schedutron: Should I delete the `specified` vector defined here after the final loop where it is used? Or does C++ already do that, implicitly? Is there a potential memory leak here?

Also, would it be more appropriate for `specified` to be a `map`, with `idx : column_id` pairs?

schedutron:
@pervazea I'm sorry. I'll add the appropriate tests ASAP! Considering point 2, this PR refactors `InsertPlan` to support better error handling and perhaps is more optimized, as I mentioned in an earlier comment. 

I'll keep functional changes separate from formatting changes from now on...

schedutron:
@pervazea I've added some tests, have a look.

schedutron:
@pervazea Have I added tests at the appropriate place?

schedutron:
@tcm-marcel Can you please add the `ready for review` label here?

pmenon:
@schedutron I think the tests are valid, but they shouldn't be done inside the planner object. Ideally, it should be done much earlier, during initial plan generation when you have access to the catalog and all input expressions (perhaps here: https://github.com/cmu-db/peloton/blob/master/src/optimizer/query_to_operator_transformer.cpp#L227).

schedutron:
@pmenon cool, I'll try to add the test to the location you mentioned. Can you please have a look at #1056 as well?

schedutron:
@pmenon Should I add the checks in `void QueryToOperatorTransformer::Visit(
    UNUSED_ATTRIBUTE parser::CreateStatement)` or `void QueryToOperatorTransformer::Visit(
    UNUSED_ATTRIBUTE parser::InsertStatement)`?

schedutron:
@pmenon I've moved the checks from the planner to the transformer. Please have a look and provide feedback :)

src/optimizer/query_to_operator_transformer.cpp
```diff
@@ -246,6 +249,52 @@ void QueryToOperatorTransformer::Visit(parser::InsertStatement *op) {
     insert_expr->PushChild(output_expr_);
     output_expr_ = insert_expr;
   } else {
+    auto storage_manager = storage::StorageManager::GetInstance();
+    auto database =
+        storage_manager->GetDatabaseWithOid(target_table->GetDatabaseOid());
+    auto table = database->GetTableWithOid(target_table->GetTableOid());
+
+    auto *schema = table->GetSchema();
```
pmenon: You shouldn't be accessing the physical table, but rather access the catalog to retrieve table schema information. Look at how `parser::TableRef` is handled.

src/optimizer/query_to_operator_transformer.cpp
```diff
@@ -246,6 +249,52 @@ void QueryToOperatorTransformer::Visit(parser::InsertStatement *op) {
...
+    else {
+      auto num_columns = op->columns.size();
+      // range based for-loop shows "call to implicitly-deleted copy
+      // constructor" error
+      for (uint32_t i = 0; i != op->insert_values.size();
+           ++i) {  // check size of each tuple
+        if (op->insert_values[i].size() > num_columns)
+          throw CatalogException(
+              "ERROR:  INSERT has more expressions than target columns");
+        else if (op->insert_values[i].size() < num_columns)
```
pmenon: Braces each clause in `if`'s.

src/optimizer/query_to_operator_transformer.cpp
```diff
@@ -246,6 +249,52 @@ void QueryToOperatorTransformer::Visit(parser::InsertStatement *op) {
...
+      // auto table_columns_num = schema->GetColumnCount();
+
+      for (size_t idx = 0; idx != num_columns; ++idx) {
+        auto col = (op->columns)[idx];
+        auto found = std::find_if(table_columns.begin(), table_columns.end(),
+                                  [&col](const peloton::catalog::Column &x) {
+                                    return col == x.GetName();
+                                  });
+        if (found == table_columns.end())
+          throw CatalogException("ERROR:  column \"" + col +
```
pmenon: 1. Braces.
2. Is this formatted correctly?

test/sql/insert_sql_test.cpp
```diff
@@ -140,6 +151,24 @@ TEST_F(InsertSQLTests, InsertOneValue) {
...
+      plan = TestingSQLUtil::GeneratePlanWithOptimizer(optimizer, query, txn);
+      EXPECT_EQ(plan->GetPlanNodeType(), PlanNodeType::INSERT);
+      txn_manager.CommitTransaction(txn);
+    } catch (peloton::Exception &ex) {
+      EXPECT_EQ(ExceptionType::CATALOG, ex.GetType());
+      EXPECT_STREQ("ERROR:  INSERT has more expressions than target columns",
+                   ex.what());
+      throw peloton::CatalogException(ex.what());
+    }
+  }, peloton::CatalogException);
```
pmenon: Use the same exception handling test logic as your previous PR.

src/optimizer/query_to_operator_transformer.cpp
schedutron: You mean `TableRef` in the `Visit()` method?

schedutron: I went through `parser::TableRef`'s source but it doesn't have any getter method for table schema.

schedutron:
@pmenon I've tried not to use the physical table this time. Once you think the procedure is good this time, I'll remove the commented-out obsolete code.

schedutron:
I'm now editing the tests.

schedutron:
@pmenon I've edited the tests, but there is more to be done.

src/optimizer/query_to_operator_transformer.cpp
```diff
@@ -246,6 +250,63 @@ void QueryToOperatorTransformer::Visit(parser::InsertStatement *op) {
...
+    // INSERT INTO table_name VALUES (val1, val2, ...), (val_a, val_b, ...), ...
+    if (op->columns.empty()) {
+      for (uint32_t tuple_idx = 0; tuple_idx < op->insert_values.size();
+           tuple_idx++) {
+        auto &values = (op->insert_values)[tuple_idx];
+        if (values.size() > column_objects.size()) {
+          throw CatalogException(
+              "ERROR:  INSERT has more expressions than target columns");
+        } else if (values.size() < column_objects.size()) {
+          throw CatalogException(
```
schedutron: Here we need to check if the remaining columns allow for null values.

src/optimizer/query_to_operator_transformer.cpp
```diff
@@ -246,6 +250,63 @@ void QueryToOperatorTransformer::Visit(parser::InsertStatement *op) {
...
+      auto num_columns = op->columns.size();
+      // range based for-loop shows "call to implicitly-deleted copy
+      // constructor" error
+      for (uint32_t i = 0; i != op->insert_values.size();
+           ++i) {  // check size of each tuple
+        if (op->insert_values[i].size() > num_columns) {
+          throw CatalogException(
+              "ERROR:  INSERT has more expressions than target columns");
+        } else if (op->insert_values[i].size() < num_columns) {
+          throw CatalogException(
```
schedutron: Same here.

src/optimizer/query_to_operator_transformer.cpp
pmenon: @schedutron Correct. You also need to check if the inserted type is "compatible" with the column type. Also, try to ensure you have the same error message as Postgres for these types of errors.

schedutron: @pmenon  I'm working on the null part, but I've added some compatibility checks in #1088. Do I need to add them here as well?

src/optimizer/query_to_operator_transformer.cpp
```diff
@@ -246,6 +250,67 @@ void QueryToOperatorTransformer::Visit(parser::InsertStatement *op) {
...
+    if (op->columns.empty()) {
+      for (uint32_t tuple_idx = 0; tuple_idx < op->insert_values.size();
+           tuple_idx++) {
+        auto &values = (op->insert_values)[tuple_idx];
+        if (values.size() > column_objects.size()) {
+          throw CatalogException(
+              "ERROR:  INSERT has more expressions than target columns");
+        } else if (values.size() < column_objects.size()) {
+          for (oid_t i = values.size(); i != column_objects.size(); ++i) {  // check whether null values are allowed in the rest of the columns
+            if (column_objects[i]->IsNotNull()) {
```
schedutron: Here we also need to check whether a default value is available for the current column. How do I do that? The `ColumnCatalogObject` instances don't have any method for that.

schedutron: @pmenon Here we also need to check whether a default value is available for the current column. How do I do that? The `ColumnCatalogObject` instances don't have any method for that.

schedutron:
@pmenon Should I add a new attribute to `ColumnCatalogObject` representing the default value for a given column?

pmenon:
@schedutron We don't support default values at the moment and doing so will probably be a bit more involved, so let's punt on it for now. Just finish up this PR by adding the basic checks, i.e., column count, type compatibility, and nullability checks.

schedutron:
@pmenon Alright, I'll add the remaining checks, but I think I did use default values in the tests and they worked fine.

schedutron:
@pmenon Also, as I mentioned earlier, I'm working on type compatibility in #1088, should I add those checks here? I'm unsure about what to do if we want to implicitly convert the types, i.e., when I insert `'3'` in an `int` column instead of `3`. My work in #1088 checks for values which cannot be converted, say `'h'`, but I'm not sure whether adding type checks here, in `query_to_operator_transformer.cpp` would prevent implicit conversions from happening.

src/optimizer/query_to_operator_transformer.cpp
pmenon: @schedutron Correct, but we don't currently store default values in the catalog tables. Punt on this for now, but leave a `TODO`.

In the mean time, simplify this function's logic; there are currently about five levels of nesting. One easy thing is to use early exits, i.e., return immediately in the insert-from-select case, thus removing the first `else` clause. Also use range-for loops with const reference loop variables. As far as I can tell, you don't directly use loop indexes. Finally, run the code formatter. Put `else` on the closing brace of the `if`.

test/sql/insert_sql_test.cpp
```diff
@@ -580,6 +620,76 @@ TEST_F(InsertSQLTests, NonExistentTable) {
...
+  EXPECT_THROW(TestingSQLUtil::GeneratePlanWithOptimizer(optimizer, query, txn),
+               peloton::CatalogException);
+  EXPECT_EQ(0, rows_changed);
+
+  // Insert a tuple with more target columns than values
+  query = "INSERT INTO test8(num1, num4) VALUES(3, 4);";
+  txn = txn_manager.BeginTransaction();
+  EXPECT_THROW(TestingSQLUtil::GeneratePlanWithOptimizer(optimizer, query, txn),
+               peloton::CatalogException);
+  EXPECT_EQ(0, rows_changed);
```
pmenon: Be sure to add tests cases to check each of the four code paths you've added. I think you're missing the case where target list is implicit and number of expressions is less than the number of columns.

schedutron:
@pmenon The not null check also happens in lower in the hierarchy, in [`data_table.cpp`](https://github.com/cmu-db/peloton/blob/c21f80aa841b02f6ddcdf062bc49e5cca5b77a8a/src/storage/data_table.cpp#L168). Since we're able to add the same test upper int he hierarchy, should I remove that test from`data_table.cpp`?

test/sql/insert_sql_test.cpp
schedutron: @pmenon There's an underlying bug in that test case. I built the master branch and then tried to insert a single value in a table with two attributes - and that resulted in segmentation fault. I'll try to fix it in this PR, or should I open another PR for that?

schedutron:
@pmenon I've added the code to handle the segfault which previously occurred for inserts like the following:

```
test=# create table foo(id int, id2 int);
CREATE TABLE
test=# insert into foo values(3);
server closed the connection unexpectedly
	This probably means the server terminated abnormally
	before or while processing the request.
The connection to the server was lost. Attempting reset: Failed.
!>
```

Please review the insert plan and suggest whether the added code can be made more elegant.

src/planner/insert_plan.cpp
```diff
@@ -39,25 +39,29 @@ InsertPlan::InsertPlan(storage::DataTable *table,
       auto &values = (*insert_values)[tuple_idx];
       PL_ASSERT(values.size() <= schema->GetColumnCount());
       uint32_t param_idx = 0;
-      for (uint32_t column_id = 0; column_id < values.size(); column_id++) {
-        auto &exp = values[column_id];
+      for (uint32_t column_id = 0; column_id < schema->GetColumnCount(); column_id++) {
+        auto index = column_id;
```
schedutron: I had to use an `index` variable which becomes `0` for values which are not specified in the `INSERT` (i.e., `id2` for a table `foo(id int, id2 int)` for, say, `INSERT INTO foo VALUES(3)`). Following would've been a more elegant approach:

```
for (uint32_t column_id = 0; column_id < schema->GetColumnCount(); column_id++) {
        if (column_id < values.size()){
          auto &expr = values[column_id];
        }
        const type::TypeId type = schema->GetType(column_id);
        if (column_id >= values.size() || expr == nullptr) {  // expr isn't checked when 1st condition is true
        ...
        ...
```

However, that resulted in a compilation error because of the declaration within an `if` lead to `unidentified error` for `expr` in the code that followed.

Furthermore, I wasn't able to get around this by declaring `expr` outside the `if` and assigning it within the `if` because `expr` requires initialization while declaration.

So, I think this is the only way to prevent the error mentioned and it's working as per my tests. Please suggest a more elegant approach if possible.

test/sql/insert_sql_test.cpp
schedutron: @pmenon I've added that test case and the additional code to handle it.

schedutron:
@pmenon Finally, all that's remaining now is

1. Removal of now-unnecessary headers
2. Linting

I'll do these once you approve the hitherto changes :)

src/planner/insert_plan.cpp
schedutron: @pmenon Is there a way to do this thing in `query_to_operator_transformer.cpp`?

src/optimizer/query_to_operator_transformer.cpp
```diff
@@ -14,6 +14,8 @@
 #include "settings/settings_manager.h"
 
 #include "catalog/database_catalog.h"
+#include "catalog/table_catalog.h"
+#include "catalog/column_catalog.h"
```
pmenon: Sort includes.

src/optimizer/query_to_operator_transformer.cpp
```diff
@@ -245,11 +249,62 @@ void QueryToOperatorTransformer::Visit(parser::InsertStatement *op) {
...
+            // TODO: Add check for default value's existence for the current column
+            throw CatalogException(
+            "ERROR:  null value in column \"" + column_objects[i]->GetColumnName() + "\" violates not-null constraint");
+          }
+        }
+      }
+    }
+  }
+  // INSERT INTO table_name (col1, col2, ...) VALUES (val1, val2, ...), ...
+  else {
```
pmenon: Put `else` on same line as closing `if` brace.

src/optimizer/query_to_operator_transformer.cpp
```diff
@@ -245,11 +249,62 @@ void QueryToOperatorTransformer::Visit(parser::InsertStatement *op) {
...
+    auto num_columns = op->columns.size();
+    // range based for-loop shows "call to implicitly-deleted copy
+    // constructor" error
+    for (uint32_t i = 0; i != op->insert_values.size();
```
pmenon: Can't you use a range-for loop with a `const` reference loop variable?

src/optimizer/query_to_operator_transformer.cpp
```diff
@@ -245,11 +249,62 @@ void QueryToOperatorTransformer::Visit(parser::InsertStatement *op) {
...
-        LogicalInsert::make(target_table, &op->columns, &op->insert_values));
-    output_expr_ = insert_expr;
+    return;
+  }
+  // column_objects represents the columns for the current table as defined in
+  // its schema
+  auto column_objects = target_table->GetColumnObjects();
+  // INSERT INTO table_name VALUES (val1, val2, ...), (val_a, val_b, ...), ...
+  if (op->columns.empty()) {
+    for (uint32_t tuple_idx = 0; tuple_idx != op->insert_values.size(); ++tuple_idx) {
```
pmenon: Can you use a range-for loop with a `const` reference variable? That'll clean up the visual noise.

src/optimizer/query_to_operator_transformer.cpp
```diff
@@ -245,11 +249,62 @@ void QueryToOperatorTransformer::Visit(parser::InsertStatement *op) {
...
+      auto &values = op->insert_values[tuple_idx];
+      if (values.size() > column_objects.size()) {
+        throw CatalogException(
+            "ERROR:  INSERT has more expressions than target columns");
+      } else if (values.size() < column_objects.size()) {
+        for (oid_t i = values.size(); i != column_objects.size(); ++i) {  // check whether null values are allowed in the rest of the columns
```
pmenon: Put the comment on a new line.

src/optimizer/query_to_operator_transformer.cpp
```diff
@@ -245,11 +249,62 @@ void QueryToOperatorTransformer::Visit(parser::InsertStatement *op) {
...
+          if (column_objects[i]->IsNotNull()) {
+            // TODO: Add check for default value's existence for the current column
+            throw CatalogException(
+            "ERROR:  null value in column \"" + column_objects[i]->GetColumnName() + "\" violates not-null constraint");
```
pmenon: I think this needs to be formatted.

src/optimizer/query_to_operator_transformer.cpp
```diff
@@ -245,11 +249,62 @@ void QueryToOperatorTransformer::Visit(parser::InsertStatement *op) {
...
+      if (op->insert_values[i].size() > num_columns) {
+        throw CatalogException(
+            "ERROR:  INSERT has more expressions than target columns");
+      } else if (op->insert_values[i].size() < num_columns) {
+        throw CatalogException(
+            "ERROR:  INSERT has more target columns than expressions");
+      }
+    }
+
+    for (size_t idx = 0; idx != num_columns; ++idx) {
```
pmenon: Again, since you don't need `idx` aside from accessing elements, can you change this to a range-for loop with a `const` reference loop variable?

src/optimizer/query_to_operator_transformer.cpp
```diff
@@ -245,11 +249,62 @@ void QueryToOperatorTransformer::Visit(parser::InsertStatement *op) {
...
+      auto col = (op->columns)[idx];
+      auto found =
+          std::find_if(column_objects.begin(), column_objects.end(),
+                        [&col](const decltype(column_objects)::value_type &x) {
+                          return col == x.second->GetColumnName();
+                        });
+      if (found == column_objects.end()) {
+        throw CatalogException(
+            "ERROR:  column \"" + col + "\" of relation \"" +
```
pmenon: Can you use `StringUtil::Format()` to make the formatting of this message clearer?

src/optimizer/query_to_operator_transformer.cpp
schedutron: I will retry by adding `const`.

schedutron:
@pmenon Done :)

src/optimizer/query_to_operator_transformer.cpp
```diff
@@ -245,11 +243,72 @@ void QueryToOperatorTransformer::Visit(parser::InsertStatement *op) {
...
+  auto column_objects = target_table->GetColumnObjects();
+  // INSERT INTO table_name VALUES (val1, val2, ...), (val_a, val_b, ...), ...
+  if (op->columns.empty()) {
+    for (const auto &values : op->insert_values) {
+      if (values.size() > column_objects.size()) {
+        throw CatalogException(
+            "ERROR:  INSERT has more expressions than target columns");
+      } else if (values.size() < column_objects.size()) {
+        for (oid_t i = values.size(); i != column_objects.size(); ++i) {
+          // check whether null values are allowed in the rest of the columns
```
schedutron: @pmenon I think I can't replace this loop with a range-based equivalent because `column_objects` is a `HashMap` instead of a vector, and I need to start the index `i` in the middle.

Otherwise, I've replaced all the loops you mentioned with range-based ones.

src/optimizer/query_to_operator_transformer.cpp
```diff
@@ -245,11 +243,72 @@ void QueryToOperatorTransformer::Visit(parser::InsertStatement *op) {
...
+        throw CatalogException(
+            "ERROR:  INSERT has more expressions than target columns");
+      } else if (tuple.size() < num_columns) {
+        throw CatalogException(
+            "ERROR:  INSERT has more target columns than expressions");
+      }
+    }
+
+    // vector below contains column ids of the columns mentioned in the insert statement
+    std::vector<oid_t> specified;
```
schedutron: I require this vector to store the ids of the specified columns. This is needed because we also need to check the `not null` constraint for the unspecified ones, just like we did in the implicit case.

schedutron:
@pmenon Any update here :)

schedutron:
@pmenon This is almost done! I understand that you're busy, but please spare some time and review this so that PR gets done.

pervazea:
I haven't been following these changes... but note that insert_plan.cpp is going to substantially change. Not a lot in lines of code, but the algorithm is much different so the changes you have, will have to be re-worked into the new code. (The reason for the change is #1197). 
I'll take a closer look at this PR as a starting point and then we can discuss.

schedutron:
@pervazea My instinct says that won't be a problem because the net change I've added in the insert plan amounts to just about 3-4 lines (although lines of code is not the main problem as you mentioned). However, I've also mentioned in the comments that it's probable that the task accomplished by adding those lines can also be done by modifying the query to operator transformer a bit and leaving the insert plan unchanged.

src/optimizer/query_to_operator_transformer.cpp
```diff
@@ -245,11 +243,72 @@ void QueryToOperatorTransformer::Visit(parser::InsertStatement *op) {
...
+    for (const auto col : op->columns) {
+      auto found =
+          std::find_if(column_objects.begin(), column_objects.end(),
+                        [&col](const decltype(column_objects)::value_type &x) {
+                          return col == x.second->GetColumnName();
+                        });
+      if (found == column_objects.end()) {
+        throw CatalogException(StringUtil::Format("ERROR:  column \"%s\" of relation \"%s\" does not exist", col.c_str(), target_table->GetTableName().c_str()));
+      } else {
+        specified.push_back(found->first);
```
pervazea: I believe the way you are checking columns is not correct.
1. The columns specified do not have to be in schema column order. The user can specify an arbitrary order for the columns to be inserted.

2. Because of 1. the assumption that the specified vector can be initialized using push_back, is not valid

src/planner/insert_plan.cpp
pervazea: What is the problem preventing these checks being done in the query_to_operator_transformer.cpp?

src/optimizer/query_to_operator_transformer.cpp
schedutron: @pervazea I think it's correct. I tried specifying the columns in a different order and the insert ran correctly.

src/optimizer/query_to_operator_transformer.cpp
```diff
@@ -245,11 +243,72 @@ void QueryToOperatorTransformer::Visit(parser::InsertStatement *op) {
...
+      } else if (tuple.size() < num_columns) {
+        throw CatalogException(
+            "ERROR:  INSERT has more target columns than expressions");
+      }
+    }
+
+    // vector below contains column ids of the columns mentioned in the insert statement
+    std::vector<oid_t> specified;
+    
+    for (const auto col : op->columns) {
```
pmenon: It's an improvement. Can you simplify this using a map? I think the catalog even has a way to get a map of column names to column objects (if not, add one). That should remove this loop, and simplify the next loop.

src/optimizer/query_to_operator_transformer.cpp
```diff
@@ -245,11 +243,72 @@ void QueryToOperatorTransformer::Visit(parser::InsertStatement *op) {
...
+    int32_t index = 0;  // index for 'specified' vector
+    for (oid_t id = 0; id != column_objects.size(); ++id) {
+      // this loop checks not null constraint for unspecified columns
+      if (id == specified[index]) {
+        index++;
+        continue;
+      }
+      if (column_objects[id]->IsNotNull()) {
+        // TODO: Add check for default value's existence for the current column
+        throw CatalogException(StringUtil::Format("ERROR:  null value in column \"%s\" violates not-null constraint", column_objects[id]->GetColumnName().c_str()));
```
pmenon: Format.

pervazea:
Maybe it does execute :-) I'm not yet convinced that makes it correct. :-)


src/optimizer/query_to_operator_transformer.cpp
pervazea: Look more carefully at the code, and consider two scenarios:
1. If the columns are out of order, is the check on line 303 still correct?
2. What if you have fewer columns specified than in the table columns? 

Maybe it will be easier to see if you add test cases to cover such cases. Be sure to run with the address sanitizer enabled if (when) you exercise these cases.

schedutron: @pervazea Thanks for pointing that out. The check at line 303 is still correct, however, the `index` increment isn't. I think I did keep this in mind while adding my code in the insert plan. However, when moving my code from insert plan to the transformer, the sort for `specified` vector got removed. Sorting the `specified` vector at line 295 will solve this problem.

src/optimizer/query_to_operator_transformer.cpp
schedutron: @pmenon Thanks, I'll try doing that. I did mention earlier in source comments whether `specified` will be more suitable as a map. Now that you've asked for it, I'll make it a map :)

schedutron:
@pmenon Done :) @pervazea The problem you mentioned is solved now, with `specified` implemented as an `unordered_map`.

schedutron:
@pmenon Is this PR mergeable now?

pervazea:
Not yet, just a little bit more.
See the Jenkins log. The changes to the insert_sql_test have memory leaks. Because the transactions are not terminated, they result in memory leaks. Take a look at what other tests that expect exceptions do. I think it may be as simple as calling transaction commit.

schedutron:
@pervazea For my tests, should I commit them or perform a rollback?

src/catalog/table_catalog.cpp
```diff
@@ -257,6 +257,22 @@ TableCatalogObject::GetColumnObjects(bool cached_only) {
   return column_objects;
 }
 
+/* @brief   get all column objects of this table into cache
+ * @return  map from column name to cached column object
+ */
+std::unordered_map<std::string, std::shared_ptr<ColumnCatalogObject>>
+TableCatalogObject::GetColumnNames(bool cached_only) {
+  if (!valid_column_objects && !cached_only) {
+    auto column_objects = this->GetColumnObjects();
```
pmenon: Remove `this`

src/catalog/table_catalog.cpp
```diff
@@ -257,6 +257,22 @@ TableCatalogObject::GetColumnObjects(bool cached_only) {
...
+    std::unordered_map<std::string, std::shared_ptr<ColumnCatalogObject> > column_names;
+    for (auto pair : column_objects) {
```
pmenon: Probably should make this a reference to avoid copy.

src/optimizer/query_to_operator_transformer.cpp
```diff
@@ -11,21 +11,19 @@
...
 #include "optimizer/operators.h"
 #include "optimizer/query_node_visitor.h"
 #include "optimizer/query_to_operator_transformer.h"
-
-#include "planner/seq_scan_plan.h"
-
 #include "parser/statements.h"
-
-#include "catalog/manager.h"
+#include "planner/seq_scan_plan.h"
```
pmenon: Why do you need this header?

test/sql/insert_sql_test.cpp
```diff
@@ -619,6 +627,72 @@ TEST_F(InsertSQLTests, NonExistentTable) {
...
+      new optimizer::Optimizer());
+  // Insert a tuple with more values than target columns
+  std::string query("INSERT INTO test8 VALUES(1, 2, 3, 4);");
+  txn = txn_manager.BeginTransaction();
+  EXPECT_THROW(TestingSQLUtil::GeneratePlanWithOptimizer(optimizer, query, txn),
+               peloton::CatalogException);
+
+  // Insert a tuple with more target columns (implicit) than values, violating
+  // not null constraint for num3
+  query = "INSERT INTO test VALUES(3, 4);";
```
pmenon: This test case is the same as the preceding one? If so, we can remove it.

test/sql/insert_sql_test.cpp
```diff
@@ -619,6 +627,72 @@ TEST_F(InsertSQLTests, NonExistentTable) {
...
+               peloton::CatalogException);
+
+  // Insert a tuple with a nonexistent target column
+  query = "INSERT INTO test8(numx) VALUES(3);";
+  txn = txn_manager.BeginTransaction();
+  EXPECT_THROW(TestingSQLUtil::GeneratePlanWithOptimizer(optimizer, query, txn),
+               peloton::CatalogException);
+
+  // Insert a tuple with a nonexistent target column (non-singleton tuple)
+  query = "INSERT INTO test8(num1, num4) VALUES(3, 4);";
```
pmenon: You need to add another test case where a non-nullable column isn't specified in the insert clause.

pervazea:
@schedutron re: commit or rollback. No preference on my part. Commit for transactions that are correct and rollback for those that aren't would be reasonable. Run make check on a address sanitizer build to make sure there are no leaks if you can. It will be more efficient than waiting for Jenkins to report the results.

src/catalog/table_catalog.cpp
schedutron: You mean the loop variable `pair`?

test/sql/insert_sql_test.cpp
schedutron: It's just a bit different - columns aren't explicitly specified unlike in the previous case.

src/planner/insert_plan.cpp
schedutron: @pervazea They were simply easy here. I'll try migrating them to `query_to_operator_transformer.cpp` and let you know the results :)

schedutron:
@pervazea I've resolved the merge conflict in insert plan now.
@pmenon I've addressed the changes you requested.

src/optimizer/query_to_operator_transformer.cpp
```diff
@@ -245,11 +242,74 @@ void QueryToOperatorTransformer::Visit(parser::InsertStatement *op) {
...
+            "ERROR:  INSERT has more expressions than target columns");
+      } else if (tuple.size() < num_columns) {
+        throw CatalogException(
+            "ERROR:  INSERT has more target columns than expressions");
+      }
+    }
+
+    // map below contains column names of the columns mapped to column objects,
+    // as mentioned in the insert statement
+    std::unordered_map<std::string,
```
pmenon: Does this need to be a map, or is a set sufficient? And how about the case of the column names? At this point, have they all been lower or upper-cased?

src/planner/insert_plan.cpp
```diff
@@ -66,6 +66,11 @@ InsertPlan::InsertPlan(
           is_prepared_stmt = true;
         }
       }
+      // for remaining columns, insert defaults
+      for (uint32_t column_id = values.size(); column_id != schema_col_count;
+           ++column_id) {
+        SetDefaultValue(column_id);
+      }
```
pervazea: Good. Thank you. I missed this.
No test added for this though? 
I will add one in the JUnit tests, since I'm working on those now.

pervazea:
You'll need to look at the insert_sql_test. There are memory leaks being generated by the test. Check the Jenkins test results.

src/planner/insert_plan.cpp
schedutron: Cool. So should I add a test in insert_sql_test or not?

src/optimizer/query_to_operator_transformer.cpp
```diff
@@ -245,11 +242,74 @@ void QueryToOperatorTransformer::Visit(parser::InsertStatement *op) {
...
+            "ERROR:  column \"%s\" of relation \"%s\" does not exist",
+            col.c_str(), target_table->GetTableName().c_str()));
+      } else {
+        specified[col] = column_names[col];
+      }
+    }
+
+    for (auto column : column_names) {
+      // this loop checks not null constraint for unspecified columns
+      if (specified.find(column.first) == specified.end() &&
```
schedutron: @pmenon I think if we have `specified` as a set instead of a map, the lookup here would look like `specified.find(column)` instead of `specified.find(column.first)` and that would be slower since we'd be comparing whole `ColumnCatalogObject`s instead of more elementary `string`s for search.

schedutron:
@pervazea Hey, I don't think I dismissed your review (I rather addressed it) even though GitHub says so!

src/planner/insert_plan.cpp
pervazea: No. I would not worry. I've got it covered. Will be in a PR as soon as I have it working as desired on Travis and Jenkins.

src/optimizer/query_to_operator_transformer.cpp
pmenon: Make `specified` a set of strings for the column names.

schedutron: @pmenon Done.

test/sql/insert_sql_test.cpp
```diff
@@ -619,6 +627,86 @@ TEST_F(InsertSQLTests, NonExistentTable) {
   txn = txn_manager.BeginTransaction();
```
pervazea: You still have a memory leak. This transaction is not terminated.

schedutron: Fixed.

schedutron:
@pervazea @pmenon Thanks for the constructive feedback along the way!

PR #1076, opened 2018-01-19
Query History Logging

sivaprasadsudhir:
This PR implements #833 - query templatization by logging query string, its hash, and other metrics into an internal catalog table for later use in the self-driving component.


malin1993ml:
@sivaprasadsudhir @pbollimp There are other two things that I forgot to add last time:
1. You need to add a bool flag in [settings.h ](https://github.com/cmu-db/peloton/blob/master/src/include/settings/settings.h) called "brain". And you should check that flag before all our features (create the catalog, start the thread pool, compute the fingerprint, etc.) using [`SettingsManager`](https://github.com/cmu-db/peloton/blob/master/src/include/settings/settings_manager.h). And the default value of that flag should be false.
2. You need to add a SQL test to this to test whether you insert the query into the history table correctly:
create a table, execute a query, and check whether that query exists in your history table. You can reference how they wrote the [optimizer_sql_test](https://github.com/cmu-db/peloton/blob/master/test/sql/optimizer_sql_test.cpp).

pbollimp:
@malin1993ml I have made all the changes needed. Can you please review the PR again?

pbollimp:
The removal of pg_*.h files in the third party folders can be done as part of another PR.

pbollimp:
All the changes are done.

malin1993ml:
Sounds good! I'll merge this after the tests pass.

pmenon:
I was just perusing and noticed `BrainThreadPool` which looks like a copy of `MonoQueuePool`. Why is this needed?

malin1993ml:
They're different singletons for different purposes. `MonoQueuePool` is the worker queue pool to execute queries, whereas `BrainThreadPool` is responsible for executing all the brain related stuff. Right now it's only recording queries, but there will be other tasks in the future. We think it's better to separate those two queues apart to avoid brain contending resources with normal workers. Brain may also deploy a different scheduling strategy than normal workers in future as well.

Actually as I'm looking at this, it's probably better to name it `BrainQueuePool`. Both the two of them are implemented using `WorkerPool` and that's the "real" thread pool. The `QueuePool` just wraps a task queue around it.

pmenon:
I agree they should be separate pool/queue instances, but they can use the same class. Just add a second static instance `MonoQueuePool` for the brain. Call it `MonoQueuePool::GetBrainInstance()` or something like that that holds a singleton instance of `MonoQueuePool` for the brain. Otherwise, we have duplicated code.

malin1993ml:
@pmenon I think that's a good idea! We need to add another set of size parameters and change the construction logic a bit to make them take those parameters. But I think this would indeed make the code cleaner. Thanks!

If fact, just as what the TODO in this file said, we should add all these four parameters to the settings manager. But I think that could be done in another PR. I'll write an issue after this PR is in.

@sivaprasadsudhir @pbollimp Can you guys first merge the two queue pool together as Prashanth suggested, and change the construction logic to take another two size parameters? Thanks!

pbollimp:
@malin1993ml Please review the latest changes

src/include/threadpool/mono_queue_pool.h
```diff
@@ -4,32 +4,29 @@
...
 /**
  * @brief Wrapper class for single queue and single pool.
  * One should use this if possible.
  */
 class MonoQueuePool {
  public:
-  MonoQueuePool()
-      : task_queue_(kDefaultTaskQueueSize),
-        worker_pool_(kDefaultWorkerPoolSize, &task_queue_),
+  MonoQueuePool(int task_queue_size = 32, int worker_pool_size = 4)
```
malin1993ml: I don't think we want the default numbers here. I think we should explicitly remove the default constructor (`MonoQueuePool() = delete;`), and remove the default values here. People should know what they're doing and give explicit values to the QueuePool.

PR #1083, opened 2018-01-22
Fix superfluous and incorrect capnp cmake exports

pranavk:
Some of these variables are already exported by cmake files in capnproto
in third_party. And correctly.

The problem I see in my machine (Fedora 27, gcc7) is that capnproto writes the static libs
to prefix/lib64/ but hardcoded CAPNP_LIBRARIES says that they are in the
prefix/lib/ directory. This makes peloton fail to link with capnp static
libs.

Let capnp cmake files in third_party handle it correctly.


pranavk:
Hmm, the build failure is unrelated to this pull request.

crd477:
Confirmed working on Fedora 26.  @pervazea I'd say feel free to enable the Fedora 26 gcc builds in Jenkins if you'd like.  It looks like Fedora 27 updated the default LLVM package to 5.0 in the past few days which causes some separate breakage there.

malin1993ml:
Hi @pranavk,  thanks for helping out with this issue! So in the master branch we haven't started using capnproto right now, but in my own [local branch](https://github.com/malin1993ml/peloton/tree/capnp-example) I have a fork which compiles a capnproto example. Also feeling strange, I noticed some variables will not be set by capnproto. So when I manually set those variables I could get it compiled. After I merge this PR into that branch, I got the following error with cmake:
```
CMake Error at cmake/CapnProtoMacros.cmake:37 (message):
  Could not locate capnp executable (CAPNP_EXECUTABLE).
Call Stack (most recent call first):
  src/CMakeLists.txt:11 (capnp_generate_cpp)


CMake Error at cmake/CapnProtoMacros.cmake:40 (message):
  Could not locate capnpc-c++ executable (CAPNPC_CXX_EXECUTABLE).
Call Stack (most recent call first):
  src/CMakeLists.txt:11 (capnp_generate_cpp)


CMake Error at cmake/CapnProtoMacros.cmake:43 (message):
  Could not locate capnp header files (CAPNP_INCLUDE_DIRECTORY).
Call Stack (most recent call first):
  src/CMakeLists.txt:11 (capnp_generate_cpp)
```
I don't know what is exactly happening, but an easy fix would be only set those variables...

pranavk:
@malin1993ml I see. I wasn't aware that capnproto is not being used right now in master. I removed whatever looked unused to me. However, the only problematic part that was preventing it to compile master branch on Fedora (as #1106 claims) is CAPNP_LIBRARIES and CAPNP_LIBRARY_DIRS variables. And these variables doesn't seem to be problematic for you as our output above suggests. You can resurrect all variables (modulo CAPNP_LIBRARIES and CAPNP_LIBRARY_DIRS) and be done with it. We both will be happy then, I guess :)

malin1993ml:
@pranavk Sounds reasonable!

Issue #1085, opened 2018-01-23
LENGTH() is incorrect


schedutron:
The inbuilt function `length()` doesn't compute lengths correctly:

```
binaryBoy=# select  name, length(name) from foo;
 name | length(name) 
------+--------------
 Sam  |            4
 Andy |            5
(2 rows)
```

pervazea:
The required fix is more involved than what you have in #1086. (I'm not familiar with all the code, so I'm relaying information).

-  You need to look at the representation of strings. The bug arises from the inconsistency between the representation, the stored data and how the code handles those. You'll probably want to start at the lowest level and make sure there is a consistent approach all the way up.
- There are tests for string length. They need to be examined and possibly fixed.
- If you undertake this, you should also add or update comments as needed, so the implementation is clear and to help any future maintainers.

amanmehta-maniac:
I would like to take up this issue. Any heads up about where/how to start looking for fixes?

apavlo:
@amanmehta-maniac This is a bit advanced. It requires a major rewrite of how we represent VARCHAR everywhere.

Issue #1084, opened 2018-01-23
Unable to Start Peloton Server


schedutron:
Here's what I get in the logs:

```
peloton $ peloton 
2018-01-23 12:56:50 [src/catalog/catalog_cache.cpp:37:InsertDatabaseObject] DEBUG - Database 16777216 already exists in cache!
Assertion failed: ((success == true)), function GetDatabaseObject, file /Users/binaryboy/Desktop/peloton/src/catalog/database_catalog.cpp, line 345.
Abort trap: 6
```

pervazea:
Since the fix for #1086 is incomplete (and therefore not completely correct), this issue is probably invalid. I'm going to close it for now.

PR #1086, opened 2018-01-23
Fix length

schedutron:
Using `lldb`, I figured out the origin of #1085, and tried to fix it. But after removing the additional boolean factor and recompiling, I get #1084.


schedutron:
This is ready for review.

schedutron:
Perhaps the checks are failing because of the problem mentioned in #1084.

apavlo:
@schedutron I am closing this for now. We need to think about it and plan out a real solution later this semester.

PR #1095, opened 2018-01-26
Jenkins (for real this time)

crd477:
Correct the Jenkins build pipelines to run all builds in Docker, including the test generation stage.  The results of this pipeline should be interpreted as true results.  Platforms known to be not working at this time are commented out.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -133,7 +133,7 @@ script:
     # create build directory
     - mkdir build
     - cd build
-    # run cmake
+    # run cmake. NOTE: the PATH is made explicit to avoid automatic selection of the preinstalled llvm version in the Travis trusty image
     - PATH=/usr/lib/llvm-3.7/bin:/usr/bin:$PATH cmake -DCOVERALLS=$COVERALLS -DCMAKE_PREFIX_PATH=`llvm-config-3.7 --prefix` -DCMAKE_BUILD_TYPE=$PELOTON_BUILD_TYPE -DUSE_SANITIZER=Address ..
     # build
     - make -j4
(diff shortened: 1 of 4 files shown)


apavlo:
@crd477 Thanks! This is a big deal for us. I really appreciate it.

Is it currently configured to only build on commits to the master branch? How do we setup pull requests from trusted accounts?

crd477:
@apavlo The underlying Jenkins configuration hasn't changed yet.  It will build commit to master and any other branches included in cmu-db/peloton.

After this is merged, I would like to recreate the main peloton pipeline in Jenkins one more time and then I **think** it will work on PRs from trusted collaborators.  I documented what I did last time in [this Wiki page](https://github.com/cmu-db/peloton/wiki/Jenkins). and you can see the PRs that Jenkins discovered and started to build [here](http://jenkins.db.cs.cmu.edu:8080/job/peloton/view/change-requests/).
I went through and manually stopped those builds at that time though because they sent the build queue through the roof; it probably would've still been working on all those jobs.

I talked with @pervazea a little about it yesterday and I suggested that perhaps a good continued use for Travis going forward is to let it build the PRs from random outsiders, and if they succeed in Travis, someone from the team could scrutinize the changes and merge them into master (or separate branch) in cmu-db/peloton where they'd then get tested by Jenkins.  The risky thing is PRs that include changes to the Jenkinsfile - if one can change that, they can do whatever you want on the build machine.  I looked through options and I did not seen an easy way to do what we want with manual Jenkins configuration.  From everything I've seen so far, it looks like maintaining the Jenkinsfile in SCM is the way of the future.

Does that make sense?

crd477:
@pervazea, @apavlo: BTW, I've tested the Debian gcc builds and they work just fine.  Feel free to enable them in Jenkins if you like.  That platform uses gcc6 which is isn't in the mix yet and is probably a useful addition.

PR #1098, opened 2018-01-26
Fix trace build failures

pranavk:


pranavk:
@crd477 I noticed you removed it from Jenkins file in a70c7d4cd295243bf7b326b2ef17aa0bcf15e9e5 

I don't quite understand the commit message. What's stopping us to guard the build with lowest log level so that Jenkins keep protecting it from incoming changes?

I think, it's best if there is atleast one trace level build in CI.

crd477:
4de6856dbea5a18ebcdf89bea6b2d432e48745d3 supersedes the commit you refer to and should be merged soon.  There will be at least one trace build in Jenkins.

pranavk:
Wow, what a coincidence! :)

THanks.

src/binder/binder_context.cpp
```diff
@@ -94,7 +94,7 @@ bool BinderContext::GetColumnPosTuple(
     value_type = column_object->GetColumnType();
     return true;
   } catch (CatalogException &e) {
-    LOG_TRACE("Can't find table %d! Return false", std::get<1>(table_id_tuple));
+    LOG_TRACE("Can't find table %s! Return false", table_obj->GetTableName().c_str());
```
pervazea: Remove the trace line. 
As you can see the old line refers to finding a table, in the current code a table_obj is passed in (which you use). As far as this particular trace line is concerned, it isn't useful any more.

Jenkinsfile
```diff
@@ -5,7 +5,7 @@ pipeline {
             steps {
                 sh 'rm -rf build && mkdir build'
                 dir('build') {
-                    sh 'cmake -DCMAKE_BUILD_TYPE=Debug -DUSE_SANITIZER=Address ..'
+                    sh 'cmake -DCMAKE_BUILD_TYPE=Debug -DUSE_SANITIZER=Address -DCMAKE_CXX_FLAGS="-DLOG_LEVEL=LOG_LEVEL_TRACE" ..'
```
pervazea: I'd not touch the Jenkinsfile for now, we have a bunch of changes to be incorporated shortly. They do include at least one trace build. After those are in.. if further changes are desired, you can suggest via PR

src/binder/binder_context.cpp
pranavk: Ah, good point. I revisited it and not even the trace this try, catch block is superfluous too since there's nothing in that code anymore that throws CatalogException.

Jenkinsfile
pranavk: Fine with me. I saw that in the other PR by crd, he has already done the same thing. I will remove this change.

pranavk:
@pervazea I have changed the PR as requested. I have force-pushed the changes to the branch instead of creating a new commit. Not sure if that's the desired workflow.

pervazea:
@pranavk, you have the correct work flow, i.e. pushing additional changes to the PR. 

You say "force-pushed" though. I'm not a Git expert, but usually no force is required. Only when a branch update via the GUI failed, I had to force the push.

pranavk:
I didn't push the "additional" changes this time to the PR; instead, I amended the previous commit and force-pushed it so that this PR has only 1 commit against it.

PR #1102, opened 2018-01-27
Allow setting number of GC threads from command line

### LINKED ISSUE #1096: Move Garbage Collector Thread Setting into SettingsManager
Right now the number of garbage collector threads is hardcoded to one:

https://github.com/cmu-db/peloton/blob/master/src/gc/gc_manager_factory.cpp#L21

We need to move this into the `SettingsManager` so that we can change this more easily:

https://github.com/cmu-db/peloton/blob/master/src/include/settings/settings_manager.h#L29

Eventually we will want to add the ability to dynamically change this parameter and modify the system's configuration through triggers. For now we just need to move it into the `SettingsManager`. We can definite a new parameter called `gc_num_threads` in 'settings.h':

https://github.com/cmu-db/peloton/blob/master/src/include/settings/settings.h

pranavk:
Fixes #1096


src/settings/settings_manager.cpp
```diff
@@ -102,7 +102,7 @@ const std::string SettingsManager::GetInfo() const {
   info.append(StringUtil::Format("%28s:   %-28s\n", "Index Tuner", GetBool(SettingId::index_tuner) ? "enabled" : "disabled"));
   info.append(StringUtil::Format("%28s:   %-28s\n", "Layout Tuner", GetBool(SettingId::layout_tuner) ? "enabled" : "disabled"));
   info.append(StringUtil::Format("%28s:   %-28s\n", "Code-generation", GetBool(SettingId::codegen) ? "enabled" : "disabled"));
-
+  info.append(StringUtil::Format("%28s:   %-28s\n", "Number of GC threads", GetInt(SettingId::gc_num_threads)));
```
apavlo: I don't think we want to have it be set from the command-line. Let's remove this.

pranavk: This is a normal GetInfo debug function, as far as I see; it has nothing to do with the option being available in command line.

Nevertheless, I don't think there's any way one can inhibit flags from appearing in ./peloton --help output and also setting them up from the command line. GFlags doesn't support that, I think (?)

src/common/init.cpp
```diff
@@ -50,6 +50,7 @@ void PelotonInit::Initialize() {
   concurrency::EpochManagerFactory::GetInstance().StartEpoch();
 
   // start GC.
+  gc::GCManagerFactory::Configure(settings::SettingsManager::GetInt(settings::SettingId::gc_num_threads));
```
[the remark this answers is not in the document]
pranavk: Sure. I think making it unsigned would be better then. I will do so.

Thanks for reviewing :)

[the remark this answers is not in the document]
pranavk: Actually, no. I take that back. Type::Value doesn't have an unsigned type. So I will keep it `int` but add the bounds checking.

malin1993ml:
@apavlo Do you think this PR is good except the conflict to fix?

pranavk:
I fixed the conflicts :)

src/common/init.cpp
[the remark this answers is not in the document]
pmenon: I agree with @apavlo. You should change the constructor for `GCManagerFactory` to take a `uint32_t`. You should also do some error checking here to ensure the command-line arguments are reasonable. For example, if someone enters a negative number of GC threads, the integer wrap-around will cause spinning up millions of threads.

pervazea:
I have not reviewed all the code, but I see a few problems that you'll need to address.
1. You'll notice the a couple of the Travis builds failed. If you look at the raw logs, you should be able to see that the attempt to run psql_test.sh fails. Same error in both failures.
1. Now, it  looks like you are setting incorrect upper bound for port number.

pranavk:
@pervazea Oh! I didn't notice there's a test that gives such a high port number. I've changed the bounds now. Let's wait for the CI.

Good thing that test was able to test the min-max bounds feature, though :)

src/include/settings/settings.h
```diff
@@ -22,12 +22,14 @@
 SETTING_int(port,
            "Peloton port (default: 15721)",
            15721,
+           512, 64000,
```
pervazea: See [](url)https://en.wikipedia.org/wiki/Registered_port
Lower bound on acceptable port value should be 1024
Upper bound should be 65535

Strictly speaking, we should not be allowing selection of the 1024 to 49151 range, except for registered or unassigned ports. 15721 is, AFAIK, not registered, but is an unassigned value. So, lets go with the values I listed above.

src/include/settings/settings.h
```diff
@@ -22,12 +22,14 @@
...
            false, false)
 
 // Maximum number of connections
 SETTING_int(max_connections,
            "Maximum number of connections (default: 64)",
            64,
+           0, 512,
```
pervazea: I'm not seeing where this is used... so maybe doesn't matter..
but min # of connections 0? Seems wrong.

src/settings/settings_manager.cpp
```diff
@@ -112,10 +113,22 @@ void SettingsManager::DefineSetting(SettingId id, const std::string &name,
...
     throw SettingsException("settings " + name + " already exists");
   }
+
+  // Only below types support min-max bound checking
+  if (value.GetTypeId() == type::TypeId::INTEGER ||
+      value.GetTypeId() == type::TypeId::SMALLINT ||
+      value.GetTypeId() == type::TypeId::TINYINT ||
+      value.GetTypeId() == type::TypeId::DECIMAL) {
+    if (!value.CompareBetweenInclusive(min_value, max_value))
+      throw SettingsException("Value given for \"" + name + "\" is not in its min-max bounds.");
```
pervazea: Would be helpful for the exception message to include the min-max bounds.

pranavk:
@pervazea I have changed the min-max bounds there as you suggested. My main focus was on introducing the min-max functionality. Even though I have changed the values there, I am still unsure what are the sensible min-max bounds that each property should have. My hope is that once this feature is merged, authors of those features would subsequently modify the values as they seem suitable to them.

pranavk:
@pervazea @apavlo  Is there anything pending according to you in this PR? This is not 'in-progress' from my side; I think it's ready to merge unless you have something in mind?

pervazea:
I'll take a look, later today.

src/settings/settings_manager.cpp
```diff
@@ -112,10 +113,25 @@ void SettingsManager::DefineSetting(SettingId id, const std::string &name,
...
+
+  // Only below types support min-max bound checking
+  if (value.GetTypeId() == type::TypeId::INTEGER ||
+      value.GetTypeId() == type::TypeId::SMALLINT ||
+      value.GetTypeId() == type::TypeId::TINYINT ||
+      value.GetTypeId() == type::TypeId::DECIMAL) {
+    if (!value.CompareBetweenInclusive(min_value, max_value))
+      throw SettingsException("Value given for \"" + name +
+                              "\" is not in its min-max bounds (" +
+                              std::to_string(min_value.GetAs<int32_t>()) + "-" +
```
pervazea: Is this part correct? 
The range checking looks ok, but if it fails, the min and max values are converted to int32_t in order to report the error. Can't decimal type can be larger than int32_t?

pranavk: Aha! Yup, this is incorrect indeed. Just updated in latest patch.

pranavk:
@pervazea Your last comment is fixed in latest commit.

src/main/peloton/peloton.cpp
```diff
@@ -32,11 +32,16 @@ int main(int argc, char *argv[]) {
...
-    settings.ShowInfo();
+  try {
+    // Print settings
+    if (peloton::settings::SettingsManager::GetBool(
+            peloton::settings::SettingId::display_settings)) {
+      auto &settings = peloton::settings::SettingsManager::GetInstance();
+      settings.ShowInfo();
+    }
+  } catch (peloton::SettingsException &exception) {
+    std::cerr << "Cannot load settings. Failed with " << exception.GetMessage() << std::endl;
```
pervazea: I believe we don't use stdout or stderr. This should be a log statement, probably a LOG_ERROR, if it is being treated as a fatal error.

pranavk: Hmm, I wanted to write this error to 'real' stderr, not stdout which is what LOG_ERROR is currently doing. I can change it to use LOG_ERROR for consistency sake but ...

pervazea: You don't want to write to stdout or stderr. Consider Postgres or MySQL or any "service" that is running. Output to stdout or stderr is either not generated or sent to /dev/null. All the information that these processes wish a user to see, is done via logging.

pranavk: Okay, I was thinking from the point of view of peloton running as a systemd service or something like that. Logging errors to stderr would then be easier for filtering. Similar to how apache logs goes to error_logs and other things go to access_logs. But you are probably correct. I am updating the PR with a new commit.

malin1993ml:
@pervazea I'm wondering what's the status of this PR?

pervazea:
@pranavk  If you could resolve the merge conflict please. 
Keep both the optimization timeout and your change, no. of GC threads.
Then we can merge and close this out.

pervazea:
Ping @pranavk
Please resolve the merge conflict. Then we can close this out.

malin1993ml:
All of the Jenkins tests are failing, though:
http://jenkins.db.cs.cmu.edu:8080/blue/organizations/jenkins/peloton/detail/PR-1102/55/pipeline/
They all failed on the same error, which I've never seen before:
```
wrapper script does not seem to be touching the log file in /tmp/workspace/peloton_PR-1102-5AVMRIJGZKAERNXSBY6RLKHN3DM52BDZ3RZVIBHXTKMYQVSLXTEQ@tmp/durable-7a3cdbb7
(JENKINS-48300: if on a laggy filesystem, consider -Dorg.jenkinsci.plugins.durabletask.BourneShellScript.HEARTBEAT_CHECK_INTERVAL=300)
script returned exit code -1
```
I restarted the test to see whether there's any difference.

pranavk:
@malin1993ml maybe it's related to this error:

> 
> /tmp/workspace/peloton_PR-1102-5AVMRIJGZKAERNXSBY6RLKHN3DM52BDZ3RZVIBHXTKMYQVSLXTEQ@3/src/udf/udf_handler.cpp:110:1: fatal error: error writing to /tmp/cccXCRVR.s: No space left on device
>  }  // namespace peloton
>  ^

You can see this if you scroll up in the Jenkins logs.

malin1993ml:
@pranavk Interesting. Thanks for pointing that out. I guess that answer has to be answered by @pervazea.

pervazea:
The most recent Jenkins builds were fine (except for the known memory leak failures). Now just waiting for branch update results from Travis. If those are good... we'll get this merged.

PR #1104, opened 2018-01-28
Move 'util' directory under 'common'.

apavlo:
* Moved 'util' directory under common.
* Moved 'util/portable_endian.h' to be just under 'common'.
* Updated the header on all of these files.

I decided to not change the namespace for now. 

https://www.youtube.com/watch?v=sHwbgzvlVJc


apavlo:
Too many conflicts. I will recreate

Issue #1097, opened 2018-01-26
New Utility Class for Extracting Information about Query Plan


apavlo:
@saatviks needs to be able to find out what indexes a query will modify without having to actually run it. If it is an INSERT/DELETE, then it will update all tuples. If it is an UPDATE, then it will only affect the indexes that are based on the attributes being modified by the query. The key way this will work is that the function has to be given a `CatalogCache` object, since we want the catalog look-up to be consistent.

We should start building up some utility classes in 'src/include/planner/util' for extracting this information. We can call it `QueryPlanUtil`.

pranavk:
@schedutron  Hmm, I'm also looking into it. Sorry for not updating the status beforehand. Have you already started working on it? If not, do you mind me taking care of this till the weekend ends? :)

Otherwise, no problems at all; you can ignore this message :)

PR #1105, opened 2018-01-28
Information about indexes affected by a given query

pranavk:
This is the first cut for #1097 

I can move the common functions like isDisjoint to some other more common utility class but first I want to get an idea if this is going in the right direction. :)

Also, I didn't create a new class as was described in the issue since there was already this PlanUtil in planner::

### DIFF
--- a/src/include/catalog/catalog_cache.h
+++ b/src/include/catalog/catalog_cache.h
@@ -18,6 +18,11 @@
 #include "common/internal_types.h"
 
 namespace peloton {
+
+namespace planner {
+    class PlanUtil;
+}
+
 namespace catalog {
 
 class DatabaseCatalogObject;
@@ -32,6 +37,7 @@ class CatalogCache {
   friend class DatabaseCatalogObject;
   friend class TableCatalogObject;
   friend class IndexCatalogObject;
+  friend class planner::PlanUtil;
 
  public:
   CatalogCache() {}
--- a/src/include/catalog/catalog_cache.h
+++ b/src/include/catalog/catalog_cache.h
@@ -20,8 +20,8 @@
 namespace peloton {
 
 namespace planner {
-    class PlanUtil;
-}
+class PlanUtil;
+}  // namespace planner
 
 namespace catalog {
 
--- a/src/include/parser/update_statement.h
+++ b/src/include/parser/update_statement.h
@@ -56,10 +56,6 @@ class UpdateStatement : public SQLStatement {
 
   virtual void Accept(SqlNodeVisitor *v) override { v->Visit(this); }
 
-  inline void TryBindDatabaseName(std::string default_database_name) {
-    table->TryBindDatabaseName(default_database_name);
-  }
-
   const std::string GetInfo(int num_indent) const override;
 
   const std::string GetInfo() const override;
--- a/src/include/parser/update_statement.h
+++ b/src/include/parser/update_statement.h
@@ -14,8 +14,8 @@
 
 #include <cstring>
 
-#include "expression/abstract_expression.h"
 #include "common/sql_node_visitor.h"
+#include "expression/abstract_expression.h"
 #include "parser/sql_statement.h"
 #include "parser/table_ref.h"
 
@@ -55,6 +55,10 @@ class UpdateStatement : public SQLStatement {
   virtual ~UpdateStatement() {}
 
   virtual void Accept(SqlNodeVisitor *v) override { v->Visit(this); }
+
+  inline void TryBindDatabaseName(std::string default_database_name) {
+    table->TryBindDatabaseName(default_database_name);
+  }
 
   const std::string GetInfo(int num_indent) const override;
 
--- a/src/include/planner/plan_util.h
+++ b/src/include/planner/plan_util.h
@@ -15,17 +15,6 @@
 #include <set>
 #include <string>
 
-#include "catalog/catalog_cache.h"
-#include "catalog/database_catalog.h"
-#include "catalog/table_catalog.h"
-#include "catalog/column_catalog.h"
-#include "catalog/index_catalog.h"
-
-#include "parser/delete_statement.h"
-#include "parser/insert_statement.h"
-#include "parser/sql_statement.h"
-#include "parser/update_statement.h"
-
 #include "planner/abstract_plan.h"
 #include "planner/abstract_scan_plan.h"
 #include "planner/delete_plan.h"
@@ -36,6 +25,15 @@
 #include "util/string_util.h"
 
 namespace peloton {
+
+namespace catalog {
+class CatalogCache;
+}
+
+namespace parser {
+class SQLStatement;
+}
+
 namespace planner {
 
 class PlanUtil {
@@ -62,7 +60,7 @@ class PlanUtil {
   * @return set of affected index object ids
   */
   static const std::set<oid_t> GetAffectedIndexes(
-      catalog::CatalogCache& catalog_cache, parser::SQLStatement* sql_stmt);
+      catalog::CatalogCache &catalog_cache, parser::SQLStatement *sql_stmt);
 
  private:
   ///
@@ -167,93 +165,6 @@ inline void PlanUtil::GetTablesReferenced(const planner::AbstractPlan *plan,
       GetTablesReferenced(child.get(), table_ids);
     }
   }
-}
-
-namespace {
-
-template <typename T>
-bool isDisjoint(const std::set<T> &setA, const std::set<T> &setB) {
-  bool disjoint = true;
-  if (setA.empty() || setB.empty())
-    return disjoint;
-
-  typename std::set<T>::const_iterator setA_it = setA.begin();
-  typename std::set<T>::const_iterator setB_it = setB.begin();
-  while (setA_it != setA.end() && setB_it != setB.end() && disjoint) {
-    if (*setA_it == *setB_it)
-      disjoint = false;
-    else if (*setA_it < *setB_it)
-      setA_it++;
-    else
-      setB_it++;
-  }
-
-  return disjoint;
-}
-
-}  // namespace
-
-inline const std::set<oid_t> PlanUtil::GetAffectedIndexes(
-    catalog::CatalogCache &catalog_cache, parser::SQLStatement *sql_stmt) {
-  std::set<oid_t> index_oids;
-  std::string db_name, table_name;
-  switch (sql_stmt->GetType()) {
-    // For INSERT, DELETE, all indexes are affected
-    case StatementType::INSERT: {
-      auto insert_stmt = static_cast<parser::InsertStatement *>(sql_stmt);
-      db_name = insert_stmt->GetDatabaseName();
-      table_name = insert_stmt->GetTableName();
-    }
-    PELOTON_FALLTHROUGH;
-    case StatementType::DELETE: {
-      if (table_name.empty() || db_name.empty()) {
-        auto delete_stmt = static_cast<parser::DeleteStatement *>(sql_stmt);
-        db_name = delete_stmt->GetDatabaseName();
-        table_name = delete_stmt->GetTableName();
-      }
-      auto indexes_map = catalog_cache.GetDatabaseObject(db_name)
-                             ->GetTableObject(table_name)
-                             ->GetIndexObjects();
-      for (auto &index : indexes_map) {
-        index_oids.insert(index.first);
-      }
-    }
-    break;
-    case StatementType::UPDATE: {
-      auto update_stmt = static_cast<parser::UpdateStatement *>(sql_stmt);
-      db_name = update_stmt->table->GetDatabaseName();
-      table_name = update_stmt->table->GetTableName();
-      auto db_object = catalog_cache.GetDatabaseObject(db_name);
-      auto table_object = db_object->GetTableObject(table_name);
-
-      auto &update_clauses = update_stmt->updates;
-      std::set<oid_t> update_oids;
-      for (const auto &update_clause : update_clauses) {
-        LOG_TRACE("Affected column name for table(%s) in UPDATE query: %s",
-                  table_name.c_str(), update_clause->column.c_str());
-        auto col_object = table_object->GetColumnObject(update_clause->column);
-        update_oids.insert(col_object->GetColumnId());
-      }
-
-      auto indexes_map = table_object->GetIndexObjects();
-      for (auto &index : indexes_map) {
-        LOG_TRACE("Checking if UPDATE query affects index: %s",
-                  index.second->GetIndexName().c_str());
-        const std::vector<oid_t> &key_attrs =
-            index.second->GetKeyAttrs();  // why it's a vector, and not set?
-        const std::set<oid_t> key_attrs_set(key_attrs.begin(), key_attrs.end());
-        if (!isDisjoint(key_attrs_set, update_oids)) {
-          LOG_TRACE("Index (%s) is affected",
-                    index.second->GetIndexName().c_str());
-          index_oids.insert(index.first);
-        }
-      }
-    } break;
-    default:
-      LOG_TRACE("Does not support finding affected indexes for query type: %d",
-                static_cast<int>(sql_stmt->GetType()));
-  }
-  return (index_oids);
 }
 
 }  // namespace planner
--- a/src/include/planner/plan_util.h
+++ b/src/include/planner/plan_util.h
@@ -19,8 +19,8 @@
 #include "planner/abstract_scan_plan.h"
 #include "planner/delete_plan.h"
 #include "planner/insert_plan.h"
-#include "planner/update_plan.h"
 #include "planner/populate_index_plan.h"
+#include "planner/update_plan.h"
 #include "storage/data_table.h"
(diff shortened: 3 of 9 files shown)


src/include/planner/plan_util.h
```diff
@@ -149,5 +169,92 @@ inline void PlanUtil::GetTablesReferenced(const planner::AbstractPlan *plan,
...
+        if (!isDisjoint(key_attrs_set, update_oids)) {
+          LOG_TRACE("Index (%s) is affected",
+                    index.second->GetIndexName().c_str());
+          index_oids.insert(index.first);
+        }
+      }
+    } break;
+    default:
+      LOG_TRACE("Does not support finding affected indexes for query type: %d",
+                static_cast<int>(sql_stmt->GetType()));
```
malin1993ml: I think you should add a switch case for `SELECT` queries, which is just going to return an empty set.

src/include/planner/plan_util.h
```diff
@@ -149,5 +169,92 @@ inline void PlanUtil::GetTablesReferenced(const planner::AbstractPlan *plan,
...
+  typename std::set<T>::const_iterator setA_it = setA.begin();
+  typename std::set<T>::const_iterator setB_it = setB.begin();
+  while (setA_it != setA.end() && setB_it != setB.end() && disjoint) {
+    if (*setA_it == *setB_it)
+      disjoint = false;
+    else if (*setA_it < *setB_it)
+      setA_it++;
+    else
+      setB_it++;
+  }
```
malin1993ml: Put brackets for all if statements for style consistency.

src/include/planner/plan_util.h
```diff
@@ -149,5 +169,92 @@ inline void PlanUtil::GetTablesReferenced(const planner::AbstractPlan *plan,
...
+  bool disjoint = true;
+  if (setA.empty() || setB.empty())
+    return disjoint;
+
+  typename std::set<T>::const_iterator setA_it = setA.begin();
+  typename std::set<T>::const_iterator setB_it = setB.begin();
+  while (setA_it != setA.end() && setB_it != setB.end() && disjoint) {
+    if (*setA_it == *setB_it)
+      disjoint = false;
+    else if (*setA_it < *setB_it)
```
malin1993ml: It seems that you're assuming an ascending order for both sets, which may not be true. I think you need to sort the two sets before doing this.

malin1993ml:
@saatviks You should also look at this PR if you have time, because this is related to the stuff you're going to do later.

src/include/planner/plan_util.h
pranavk: Hi,

The std::set stores the element inside in an ascending order already by default; so, there's no need to sort explicitly here.

pranavk: But actually, yeah. Let's sort them to be safe. Though the current implementation provide only sorted sets to this utility function, being a utility function, our users can give us all sorts of crazy input sets; so, better safe than sorry :)

malin1993ml:
@pranavk These all look good! However, you need to add test cases to test whether your `GetAffectedIndexes` function can extract the affected indexes for the four types of queries correctly. To do that, you need to create a table, create several indexes, and use your function to analyze some queries. Actually as I'm thinking of this, [this test case ](https://github.com/cmu-db/peloton/blob/master/test/optimizer/old_optimizer_test.cpp) is probably a better reference for you. You can look at how it creates the database, table, and gets the statement from the SQL string.

src/include/planner/plan_util.h
```diff
@@ -44,6 +53,15 @@ class PlanUtil {
...
       const planner::AbstractPlan *plan);
 
+  /**
+  * @brief Get the indexes affected by a given query
+  * @param CatalogCache
+  * @param SQLStatement
+  * @return set of affected index object ids
+  */
+  static const std::set<oid_t> GetAffectedIndexes(
+      catalog::CatalogCache &catalog_cache, parser::SQLStatement *sql_stmt);
```
pmenon: Can `sql_stmt` be null? If not, it's probably a good idea to make it a reference. If the function doesn't modify it, even better to make it `const`.

src/include/planner/plan_util.h
```diff
@@ -25,6 +25,15 @@
...
 
 namespace peloton {
+
+namespace catalog {
+class CatalogCache;
+}
+
+namespace parser {
+class SQLStatement;
+}
```
pmenon: Namespace closures for all these.

src/include/catalog/catalog_cache.h
```diff
@@ -18,6 +18,11 @@
 #include "common/internal_types.h"
 
 namespace peloton {
+
+namespace planner {
+    class PlanUtil;
+}
```
pmenon: 1. Namespace closures.
2. Indentation.

src/include/util/set_util.h
```diff
@@ -0,0 +1,63 @@
...
+
+    while (setA_it != setA_end && setB_it != setB_end && disjoint) {
+      if (*setA_it == *setB_it) {
+        disjoint = false;
+      } else if (*setA_it < *setB_it) {
+        setA_it++;
+      } else {
+        setB_it++;
+      }
+    }
```
pmenon: 1. Rename to `SetUtil::IsDisjoint()`
2. Move implementation outside class definition as per [style guide](https://github.com/cmu-db/peloton/blob/master/script/coding_style.md).
3. Is the sortedness check required? `std::set` provides ordered iteration (ascending by default).

pranavk: > Is the sortedness check required? std::set provides ordered iteration (ascending by default).

Yeah, true. I did the same thing before, i.e., didn't sort it. But changed the behavior after @malin1993ml 's comment above thinking that the input set could have been initialized by the user with a different comparator. But now when I think about it again, it's impossible since the input params to this function are only set<T> and not set<T, CmpFunc>. I'll remove the superfluous sorting operation again.

malin1993ml: Sounds reasonable! I didn't know that `std::set` comes with default ascending order.

src/include/planner/plan_util.h
pranavk: Good point. I have made it a const reference and made appropriate changes required for it.

src/include/util/set_util.h
```diff
@@ -0,0 +1,46 @@
...
+};
+
+template <typename T>
+inline bool SetUtil::IsDisjoint(const std::set<T> &setA,
+                                const std::set<T> &setB) {
+  bool disjoint = true;
+  if (setA.empty() || setB.empty()) return disjoint;
+
+  typename std::set<T>::const_iterator setA_it = setA.begin();
+  typename std::set<T>::const_iterator setB_it = setB.begin();
```
pmenon: You can use `auto` here to reduce visual noise.

src/planner/plan_util.cpp
```diff
@@ -0,0 +1,100 @@
+//===----------------------------------------------------------------------===//
+//
+//                         Peloton
+//
+// plan_util.cpp
+//
+// Identification: src/planner/plan_util.cpp
+//
+// Copyright (c) 2015-18, Carnegie Mellon University Database Group
```
pmenon: Fix copyright.

src/include/util/set_util.h
pranavk: I thought it doesn't work in C++11 but it does!! great! :) I will fix it.

src/planner/plan_util.cpp
pranavk: Change what? Write it as 2018 instead of 18? We should also fix it in the formatter.py script so that it complains for mistakes like this.

pranavk: And FWIW, the same pattern, eg: 2015-17 is being followed in many other files as well. See create_plan.cpp , eg:.

pmenon: Correct, years should be full. In general, if you run the formatter script [here](https://github.com/cmu-db/peloton/blob/master/script/formatting/formatter.py) to add the header, it should do it correctly.

You're correct that other files don't adhere to the rules. This is something we're fixing incrementally rather than submitting a giant PR that breaks other ongoing PRs.

pranavk: cool; pushed the new changes. Yeah, turns out I missed the `-u` flag to the script.

pranavk:
@pbollimp There's nothing in-progress about this PR. It's ready for further review or merge, if it looks good to someone.

malin1993ml:
@pranavk Thanks for updating the PR. However, we still need a test case like I commented before:
https://github.com/cmu-db/peloton/pull/1105#issuecomment-363864343

After that I think we're good to go.

pranavk:
@malin1993ml Ah, I commented on a wrong PR. 

And yes, thanks for reminding.

pranavk:
So, I tried to make use of helper functions in `TestingExecutorUtil` as mentioned in the test/README file; I tried using `CreateAndPopulateTable` (this is what is recommended in the README file). But this utility function only creates the table on the storage layer and doesn't insert table, column, or the index into their respective catalog objects. Since the new function `GetAffectedIndexes` relies on data being available in the `CatalogCache`, things doesn't work as expected inside `GetAffectedIndexes`.

The next thing  to try in my list is to use the `Catalog` object's methods such as AddTable, CreateTable, CreateIndex, etc. so that things really are available in the CatalogCache before I call `GetAffectedIndex`. But there are some more tests already that does this thing (cf. DropTests). My question is should I factor this db, table, index creation under a new set of `TestingExecutorUtil` or modify the existing `CreateAndPopulateTable` to add things to Catalog after creating these?

pranavk:
@malin1993ml I've added the test as requested :)

src/include/parser/update_statement.h
```diff
@@ -56,6 +56,10 @@ class UpdateStatement : public SQLStatement {
 
   virtual void Accept(SqlNodeVisitor *v) override { v->Visit(this); }
 
+  inline void TryBindDatabaseName(std::string default_database_name) {
+    table->TryBindDatabaseName(default_database_name);
+  }
```
malin1993ml: We can't add a function that is only used in the test to the source code. We have to use something else.

test/planner/plan_util_test.cpp
```diff
@@ -0,0 +1,160 @@
...
+  sql_stmt = sql_stmt_list->GetStatement(0);
+  static_cast<parser::InsertStatement *>(sql_stmt)->TryBindDatabaseName(
+      TEST_DB_NAME);
+  affected_indexes =
+      planner::PlanUtil::GetAffectedIndexes(txn->catalog_cache, *sql_stmt);
+
+  // all indexes are affected
+  EXPECT_EQ(static_cast<int>(affected_indexes.size()), 2);
+  expected_oids = std::set<oid_t>({id_idx_oid, fname_idx_oid});
+  EXPECT_EQ(expected_oids, affected_indexes);
```
malin1993ml: You also want to add a select case which will affect no index.

test/planner/plan_util_test.cpp
```diff
@@ -0,0 +1,160 @@
...
+  auto catalog = catalog::Catalog::GetInstance();
+  catalog->Bootstrap();
+
+  auto &txn_manager = concurrency::TransactionManagerFactory::GetInstance();
+  auto txn = txn_manager.BeginTransaction();
+
+  catalog->CreateDatabase(TEST_DB_NAME, txn);
+  auto db = catalog->GetDatabaseWithName(TEST_DB_NAME, txn);
+  // Insert a table first
+  auto id_column = catalog::Column(
```
malin1993ml: We should change how we do this and creating indexes.

test/planner/plan_util_test.cpp
```diff
@@ -0,0 +1,160 @@
...
+  // create index on 'id' and 'first_name'
+  col_id = source_table->GetSchema()->GetColumnID(fname_column.column_name);
+  LOG_INFO("fname: %d", static_cast<int>(col_id));
+  source_col_ids.push_back(col_id);
+
+  catalog->CreateIndex(TEST_DB_NAME, "test_table", source_col_ids,
+                       "test_fname_idx", false, IndexType::BWTREE, txn);
+  txn_manager.CommitTransaction(txn);
+
+  // dummy txn to get the catalog_cache object
```
malin1993ml: The rest of the test looks good.

malin1993ml:
@pranavk oh also it seems that you need to update your PR with the master and pass the Travis test.

test/planner/plan_util_test.cpp
pranavk: You mean using the approach in old_optimizer_test? I looked at both the approaches and this looked better to me (using methods instead of manually typing queries to create tables, indexes in). And I couldn't find any downsides of this approach.

pranavk:
@malin1993ml I didn't update the PR since master, for me on F27, doesn't compile.

test/planner/plan_util_test.cpp
malin1993ml: Oh if you can get it work without adding the extra TryBindDatabaseName function then I think it's fine. Then like you said, it would would be better to factor this db, table, index creation under a new set of `TestingExecutorUtil`, and ideally update the DropTests to use this as well.

pranavk: I have already done so. :)

malin1993ml:
@pranavk I see. I'll check the compilation issue with others.

test/planner/plan_util_test.cpp
malin1993ml: Cool! Can you factor this out to TestingExecutorUtil as a new utility function as well? Then other tests can also use this. Then I think it’s only the compilation issue left.

malin1993ml:
Hi @pranavk, so we fixed a compilation issue with Fedora 27. Please try to see whether you can compile it now. If so, there're three minors things left:
1. Update your PR with master
2. Fix the Travis build
3. Factor the db, table, index creation in the test under a new set of TestingExecutorUtil

pranavk:
@malin1993ml I updated the PR with master and fixed the Travis build.

Regarding factoring out the code, I was thinking it might be better to handle this in a separate PR since the  code copy & pasta was already there before the PR. It's not just the drop tests with which my code matches. There are several tests like constraint_tests, etc. See `git grep CreateDatabase test/` for more.

malin1993ml:
@pranavk Make sense. Then can you write an issue about what should we factor out later and where are they? In terms of the code, I think we're done here:)

Can you update your PR again? Github tells me that it's still out-of-date.

test/planner/plan_util_test.cpp
```diff
@@ -0,0 +1,170 @@
...
+  auto &peloton_parser = parser::PostgresParser::GetInstance();
+  auto sql_stmt_list = peloton_parser.BuildParseTree(query_string);
+  auto sql_stmt = sql_stmt_list->GetStatement(0);
+  static_cast<parser::UpdateStatement *>(sql_stmt)->table->TryBindDatabaseName(
+      TEST_DB_NAME);
+  std::set<oid_t> affected_indexes =
+      planner::PlanUtil::GetAffectedIndexes(txn->catalog_cache, *sql_stmt);
+
+  // id and first_name are affected
+  EXPECT_EQ(static_cast<int>(affected_indexes.size()), 2);
```
pmenon: For these assertions/expectations, the expected value should be the first argument.

test/planner/plan_util_test.cpp
```diff
@@ -0,0 +1,170 @@
...
+  stmt.reset(new Statement("SELECT", query_string));
+  sql_stmt_list = peloton_parser.BuildParseTree(query_string);
+  sql_stmt = sql_stmt_list->GetStatement(0);
+  affected_indexes =
+      planner::PlanUtil::GetAffectedIndexes(txn->catalog_cache, *sql_stmt);
+
+  // no indexes are affected
+  EXPECT_EQ(static_cast<int>(affected_indexes.size()), 0);
+}
+}
```
pmenon: Properly close the namespaces.

test/util/set_util_test.cpp
```diff
@@ -0,0 +1,46 @@
+//===----------------------------------------------------------------------===//
+//
+//                         Peloton
+//
+// set_util_test.cpp
+//
+// Identification: test/set_util_test.cpp
```
pmenon: Path is wrong.

test/planner/plan_util_test.cpp
```diff
@@ -0,0 +1,170 @@
...
+
+  txn = txn_manager.BeginTransaction();
+  catalog->CreateTable(TEST_DB_NAME, "test_table", std::move(table_schema),
+                       txn);
+  txn_manager.CommitTransaction(txn);
+
+  txn = txn_manager.BeginTransaction();
+  auto source_table = db->GetTableWithName("test_table");
+  oid_t col_id = source_table->GetSchema()->GetColumnID(id_column.column_name);
+  LOG_INFO("id: %d", static_cast<int>(col_id));
```
pmenon: Just curious, are these logs needed in the test? What information do they convey?

pranavk: Ah, these are the debug logs leftover from my testing. Not needed.

pranavk:
@pmenon Cleaned up

PR #1108, opened 2018-01-28
Support for EXPLAIN statement

### LINKED ISSUE #569: Add support for EXPLAIN
Once we get the new parser (#521) and the new optimizer, it should be trivial for us to add support for `EXPLAIN`. We currently do not support it:

```
postgres=# CREATE TABLE foo (id INT PRIMARY KEY, val INT);
CREATE TABLE foo (id INT PRIMARY KEY, val INT) 0
postgres=# EXPLAIN SELECT * FROM foo WHERE id = 1;
Error parsing SQL statement
```

It would also be nice to support JSON output like in postgres:

```
pavlo=# EXPLAIN (FORMAT JSON) SELECT * FROM foo WHERE id = 1;
             QUERY PLAN             
------------------------------------
 [                                 +
   {                               +
     "Plan": {                     +
       "Node Type": "Index Scan",  +
       "Scan Direction": "Forward",+
       "Index Name": "foo_pkey",   +
       "Relation Name": "foo",     +
       "Alias": "foo",             +
       "Startup Cost": 0.15,       +
       "Total Cost": 8.17,         +
       "Plan Rows": 1,             +
       "Plan Width": 8,            +
       "Index Cond": "(id = 1)"    +
     }                             +
   }                               +
 ]
(1 row)
```

https://www.postgresql.org/docs/current/static/sql-explain.html

pranavk:
Fixes #569 

As a first iteration, there's no JSON support. Pasting some sample output here:
```

postgres=# explain select * from test where a = 1;
                          Query plan                           
---------------------------------------------------------------
 -------------------------------------------------------------
 -> Plan Type: INDEXSCAN
    Info: IndexScan
    NumChildren: 0
 -------------------------------------------------------------
(5 rows)

```


src/network/postgres_protocol_handler.cpp
```diff
@@ -326,6 +328,33 @@ ProcessResult PostgresProtocolHandler::ExecQueryMessage(
...
+                                                     };
+        stmt->SetTupleDescriptor(tuple_descriptor);
+        traffic_cop_->SetResult(plan_info);
+        status = ResultType::SUCCESS;
+      } else {
+        traffic_cop_->SetErrorMessage(error_message);
+        status = ResultType::FAILURE;
+      }
+      ExecQueryMessageGetResult(status);
+      return ProcessResult::COMPLETE;
```
pmenon: Please extract this logic into a separate member function. In general, avoid large/complex switch cases.

src/network/postgres_protocol_handler.cpp
```diff
@@ -326,6 +328,33 @@ ProcessResult PostgresProtocolHandler::ExecQueryMessage(
       ExecQueryMessageGetResult(status);
       return ProcessResult::COMPLETE;
     };
+    case QueryType::QUERY_EXPLAIN: {
+      parser::ExplainStatement *explain_stmt =
```
pmenon: Use `auto` here.

src/network/postgres_protocol_handler.cpp
```diff
@@ -326,6 +328,33 @@ ProcessResult PostgresProtocolHandler::ExecQueryMessage(
...
+          static_cast<parser::ExplainStatement *>(sql_stmt.get());
+      std::unique_ptr<parser::SQLStatementList> unnamed_sql_stmt_list(new parser::SQLStatementList());
+      unnamed_sql_stmt_list->PassInStatement(
+          std::move(explain_stmt->real_sql_stmt));
+      auto stmt = traffic_cop_->PrepareStatement(
+          "explain", query, std::move(unnamed_sql_stmt_list), error_message);
+      ResultType status = ResultType::UNKNOWN;
+      if (stmt) {
```
pmenon: `stmt != nullptr`

src/include/parser/explain_statement.h
```diff
@@ -0,0 +1,40 @@
...
+ * @class ExplainStatement
+ * @brief Represents "EXPLAIN <query>"
+ */
+class ExplainStatement : public SQLStatement {
+ public:
+  ExplainStatement()
+      : SQLStatement(StatementType::EXPLAIN) {}
+  virtual ~ExplainStatement() {}
+
+  virtual void Accept(SqlNodeVisitor* /*v*/) override { }
```
pmenon: Remove `virtual` keyword.

src/include/parser/explain_statement.h
```diff
@@ -0,0 +1,40 @@
+//===----------------------------------------------------------------------===//
+//
+//                         Peloton
+//
+// explain_statement.h
+//
+// Identification: src/include/parser/explain_statement.h
+//
+// Copyright (c) 2018, Pranav Kant
```
pmenon: Fix copyright.

src/include/parser/explain_statement.h
pranavk: This is consistent with other *_statement.h files. So, I didn't want to deviate unless you have a strong reason to do so?

src/network/postgres_protocol_handler.cpp
pranavk: No problems in changing :) that but just curious, is that some sort of style we want to follow here? or?

src/network/postgres_protocol_handler.cpp
pranavk: Can you elaborate here a bit more? This switch case for `EXPLAIN` is a bit different than other switch cases around, except only 2-3 lines matching with other switch cases, so possibility to factor out anything meaningful doesn't look obvious to me.

src/include/parser/explain_statement.h
pmenon: The `virtual` keyword is unnecessary because we have `override`. Ideally, all other statement classes should change, but we'll update them incrementally.

src/network/postgres_protocol_handler.cpp
pmenon: Try to be explicit, especially because `auto` is hiding the fact that it's a pointer.

src/network/postgres_protocol_handler.cpp
pmenon: The function is fairly complex as it is, the brunt being a giant complex switch statement. Ideally, each case should be a separate function, but we'll update it incrementally. For now, extract your case into a separate function.

src/include/parser/explain_statement.h
pranavk: Okay. I will change it to "Carnegie Mellon University Database Group" but note that I am not part of the group. So writing that seemed wrong to me.

malin1993ml:
@pmenon What's the take on this PR? Do you think the review comments are addressed except fixing the conflicts?

src/include/network/postgres_protocol_handler.h
```diff
@@ -23,6 +23,7 @@
 #include "common/statement.h"
 #include "common/statement_cache.h"
 #include "traffic_cop/traffic_cop.h"
+#include "parser/postgresparser.h"
```
pmenon: Do you need this header? I think you can just forward declare `parser::ExplainStatement` to satisfy usage in `ExecQueryExplain`

In general, forward declare classes from namespaces different than the one you're in if you don't need the full definition. This reduces coupling between components, and reduces compilation times when files change.

src/include/parser/explain_statement.h
```diff
@@ -0,0 +1,40 @@
+//===----------------------------------------------------------------------===//
+//
+//                         Peloton
+//
+// explain_statement.h
+//
+// Identification: src/include/parser/explain_statement.h
+//
+// Copyright (c) 2018, Carnegie Mellon University Database Group
```
pmenon: Fix the copyright using the `formatter.py` script.

src/include/parser/explain_statement.h
```diff
@@ -0,0 +1,40 @@
...
+//
+//===----------------------------------------------------------------------===//
+
+#pragma once
+
+#include "common/logger.h"
+#include "common/sql_node_visitor.h"
+#include "expression/abstract_expression.h"
+#include "parser/sql_statement.h"
+#include "parser/table_ref.h"
```
pmenon: Do you need `expression/abstract_expression.h`, `common/logger.h`, or `parser/table_ref.h` here? If not, remove it.

src/include/parser/explain_statement.h
```diff
@@ -0,0 +1,40 @@
...
+class ExplainStatement : public SQLStatement {
+ public:
+  ExplainStatement()
+      : SQLStatement(StatementType::EXPLAIN) {}
+  virtual ~ExplainStatement() {}
+
+  void Accept(SqlNodeVisitor* /*v*/) override {}
+
+  std::unique_ptr<parser::SQLStatement> real_sql_stmt;
+};
```
pmenon: 1. Has this file been formatter using the `formatter.py` script?
2. Either add in the variable name in `Accept()` with an `UNUSED_ATTRIBUTE` or remove it altogether.

src/include/parser/explain_statement.h
```diff
@@ -0,0 +1,38 @@
...
+/**
+ * @class ExplainStatement
+ * @brief Represents "EXPLAIN <query>"
+ */
+class ExplainStatement : public SQLStatement {
+ public:
+  ExplainStatement() : SQLStatement(StatementType::EXPLAIN) {}
+  virtual ~ExplainStatement() {}
+
+  virtual void Accept(UNUSED_ATTRIBUTE SqlNodeVisitor *v) override {}
```
pmenon: 1. Any reason why this isn't implemented?
2. You can remove the `virtual` keyword. It's redundant since you specify `override`.

src/network/postgres_protocol_handler.cpp
```diff
@@ -34,14 +34,18 @@
...
 #include "planner/abstract_plan.h"
 #include "planner/delete_plan.h"
 #include "planner/insert_plan.h"
+#include "planner/plan_util.h"
 #include "planner/update_plan.h"
 #include "settings/settings_manager.h"
 #include "traffic_cop/traffic_cop.h"
 #include "type/value.h"
 #include "type/value_factory.h"
+#include "network/marshal.h"
```
pmenon: Do you need this header?

src/include/network/postgres_protocol_handler.h
```diff
@@ -23,6 +23,7 @@
 #include "common/portal.h"
 #include "common/statement.h"
 #include "common/statement_cache.h"
+#include "traffic_cop/traffic_cop.h"
```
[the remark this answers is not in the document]
pranavk: Yeah, this header is already added in this file somewhere up. No idea from where it came into my work tree. Smells like something wrong with my setup adding includes left & right. Or maybe some bug in formatting script (?). Anyhow, I will be more attentive looking out for these things.

src/include/parser/explain_statement.h
pranavk: 1) I could implement it like other *Statement classes by inserting a `v->Visit(this)` but all the statement classes currently doesn't implement `Visit` function. So, implementing it here would be a no-op. Anyhow, I am implementing it like other statements just for consistency sake.

pranavk: 2) This use of `virtual` keyword is inspired by other such *Statement classes in the code. Sure, I can remove it but I left it there for consistency sake, again. I am removing it in my latest patchset.

src/network/postgres_protocol_handler.cpp
pranavk: Again, no idea from where it came in. Removing this.

src/include/parser/explain_statement.h
```diff
@@ -0,0 +1,36 @@
...
+// Copyright (c) 2015-2018, Carnegie Mellon University Database Group
+//
+//===----------------------------------------------------------------------===//
+
+#pragma once
+
+#include "parser/sql_statement.h"
+
+namespace peloton {
+
```
pmenon: Remove empty space.

src/include/parser/explain_statement.h
```diff
@@ -0,0 +1,36 @@
...
+namespace parser {
+
+/**
+ * @class ExplainStatement
```
pmenon: Remove class name from comment.

src/network/postgres_protocol_handler.cpp
```diff
@@ -351,6 +360,30 @@ ProcessResult PostgresProtocolHandler::ExecQueryMessage(
...
+      new parser::SQLStatementList());
+  unnamed_sql_stmt_list->PassInStatement(std::move(explain_stmt.real_sql_stmt));
+  auto stmt = traffic_cop_->PrepareStatement(
+      "explain", query, std::move(unnamed_sql_stmt_list));
+  ResultType status = ResultType::UNKNOWN;
+  if (stmt != nullptr) {
+    traffic_cop_->SetStatement(stmt);
+    std::vector<std::string> plan_info = StringUtil::Split(
+        planner::PlanUtil::GetInfo(stmt->GetPlanTree().get()), '\n');
+    const std::vector<FieldInfo> tuple_descriptor = {
```
pmenon: Curious, why is the brace needed here?

src/include/parser/explain_statement.h
pranavk: There are many files where this style is being followed, i.e., include the class name in the comment. See update_statement.h or insert_statement.h, etc. I just followed the same pattern here. I think, if we want to change these kind of things, it makes more sense to do it separately for all of the classes in a separate PR.

src/network/postgres_protocol_handler.cpp
pranavk: I am using brace-initialization here to initialize the first element of the vector to be what's inside the brace. Do you prefer some other way of doing the same?

pranavk:
@pmenon Regarding tests, I don't quite understand how these psql tests are even running. There are `.sql` files everywhere in the test/ directory and I found a `psql_test.sh` which runs `basic1.sql` using `TEST_NAME` environment variable. How are other `.sql` files are being executed. I don't see anything which changes `TEST_NAME` variable and iterate over all the `.sql` files. Do you have it somewhere documented how to add a new `psql` test?

pervazea:
@pranavk regarding your comment about psql tests.
First, you are correct in your observation(s). Only the basic1.sql test is currently run.
In my opinion, the psql tests are not very user friendly in many different ways. At some point, I'd like to re-write them. So, if we separate out the issue of the test script and invocation from the test sql, what would be most valuable for you, would be to write a .sql file along with expected output, so we have a well written test case.

src/include/parser/explain_statement.h
pmenon: Huh, didn't realize this has proliferated in parser code. Sure, leave it the way it is.

pranavk:
The project doesn't compile for me any longer on Fedora. And I doubt I'll have time to look and fix those issues anytime soon. I'd appreciate if someone, who still has a buildable peloton, can adopt this PR, rebase, and work on adding the test as commented above.

Thanks.

pervazea:
This is useful work. If you have no objection, I will try adopting your PR and carry to completion.


I'm not sure how this is done, but I'd pull the changes into a branch I create and go from there, unless educated otherewise.



pranavk:
@pervazea Thank you so much for agreeing to adopt the leftover work. Solving merge conflicts was easy, so for now I've fixed the merged conflicts (without compiling). I think it's best if we merge this PR without the tests if CI passes and then you can have another PR adding the test?

pervazea:
Sounds alright to me. We'll want to get the PR through the Travis and Jenkins tests before merging.


pranavk:
@pervazea CI has passed :) 

Same with the other PR of mine. Can you please approve them while all is green.

pranavk:
Regarding the Jenkins failure, I just noticed this in the jenkins logs:

> 
> /tmp/workspace/peloton_PR-1108-RVUVICOKWADGXVKWZQGL3EHB63B7ZH3WX3E4Q3IPN2V5S7KJEK7A/test/executor/create_index_test.cpp:207:1: fatal error: error writing to /tmp/ccpfQc46.s: No space left on device
> 
>  }  // namespace peloton

PR #1109, opened 2018-01-29
ART Index

pmenon:
This PR brings in the Adaptive Radix Tree (ART) index with Optimistic Lock Coupling (OLC). Details:

1. Added third-party ART library.
2. New index interface for ART indexes.
3. Added single- and multi-threaded tests.
4. Support creation of ART indexes through PSQL using Postgres syntax (i.e., `CREATE INDEX index_name ON table USING art (...columns...)` ).

The default index is still BwTree.


pervazea:
@apavlo, suggest other reviewers, if anyone else is familiar or interested in this area.

PR #1113, opened 2018-01-31
Add back some variable settings for capnproto.

malin1993ml:
This PR adds back some manual variable settings that are removed by #1083. Those are not set by capnproto automatically, and we need them to compile the proto files.


malin1993ml:
@pervazea Could you check this after the build passes? This should be simple. Thanks!

Issue #1106, opened 2018-01-28
build and install does not work on Fedora


klimagor:
make -j4

[  1%] Built target capnp
[  3%] Built target peloton-proto
[ 16%] Built target pg_query
make[2]: *** No rule to make target 'external/capnp-install/lib/libcapnp-rpc.a', needed by 'lib/libpeloton.so.0.0.5'.  Stop.
make[1]: *** [CMakeFiles/Makefile2:1194: src/CMakeFiles/peloton.dir/all] Error 2
make: *** [Makefile:141: all] Error 2

pranavk:
#1083 fixes this issue.

klimagor:
Thanks a lot. 
I removed the relevant lines manually from the `capnproto.cmake` file.

Now I have errors related to the LLVM:
```
make[2]: *** No rule to make target 'LLVMBinaryFormat-NOTFOUND', needed by 'lib/libpeloton.so.0.0.5'.  Stop.
make[2]: *** Waiting for unfinished jobs....
[ 98%] Building CXX object src/CMakeFiles/peloton.dir/__/third_party/libcount/utility.cc.o
make[1]: *** [CMakeFiles/Makefile2:1194: src/CMakeFiles/peloton.dir/all] Error 2
make: *** [Makefile:141: all] Error 2

```

klimagor:
These errors were with LLVM version 4.0.1.
After upgrading to LLVM version 5.0.1, there are different errors:

```
../lib/libpeloton.so.0.0.5: undefined reference to `vtable for llvm::GetElementPtrInst'
../lib/libpeloton.so.0.0.5: undefined reference to `llvm::AllocaInst::AllocaInst(llvm::Type*, llvm::Twine const&, llvm::BasicBlock*)'
../lib/libpeloton.so.0.0.5: undefined reference to `llvm::User::anchor()'
../lib/libpeloton.so.0.0.5: undefined reference to `typeinfo for llvm::Instruction'
../lib/libpeloton.so.0.0.5: undefined reference to `llvm::Value::assertModuleIsMaterialized() const'
../lib/libpeloton.so.0.0.5: undefined reference to `vtable for llvm::ICmpInst'
../lib/libpeloton.so.0.0.5: undefined reference to `llvm::CmpInst::anchor()'
../lib/libpeloton.so.0.0.5: undefined reference to `vtable for llvm::UnaryInstruction'
[...]

PR #1115, opened 2018-01-31
added llvm3.9 packages to installation in case of fedora 27

speedskater:
I had problems building peloton with fedora 27 due to fedora's default LLVM Version (could be related to issue #1106 ). Therefore I propose this changes to the package.sh install script to automatically install LLVM3.9 on fedora 27.

Additionally I would like you to ask, if it would be possible to add the following Line to the wiki for building peloton with Fedora (27):

cmake -DLLVM_DIR=/usr/lib64/cmake/llvm3.9/ -DCMAKE_BUILD_TYPE=Release ../

With this two changes building peloton works on my fedora 27 system out of the box.


crd477:
The Fedora 27 Docker image in our Jenkins setup was already configured with LLVM 4.0 before they updated it to LLVM 5.0 so I suggest [a slightly different patch](https://github.com/cmu-db/peloton/files/1690756/fedora-llvm.diff.txt).

speedskater:
@crd477 I tried to incorporate your suggestion, but unfortuntely there seems to be a problem with the cmake configuration provided with the llvm4.0 package in fedora 27 resulting in the following build error:
```/home/rbinna/VCS/peloton/src/include/codegen/code_context.h:18:10: fatal error: llvm/IR/IRBuilder.h: No such file or directory
 #include "llvm/IR/IRBuilder.h"
```
Therefore, I changed the script to use LLVM 3.9 for all fedora versions >= 27. Furthermore I updated the fedora 27 docker file to the use the LLVM 3.9 version to build peloton. 

Do you have any other idea to build peloton with llvm >= 4.0 ?

pervazea:
We have decided that we are not going to support Fedora at this time.

Issue #1117, opened 2018-01-31
UNIQUE causes transaction abort


pervazea:
Observed when initially working on issue #1043, i.e. when running OLTPBench, TATP benchmark.

Postgres ddl has:
`...
CREATE TABLE subscriber (
   s_id INTEGER NOT NULL PRIMARY KEY,
   sub_nbr VARCHAR(15) NOT NULL UNIQUE,
...

Peloton schema modified to omit UNIQUE.

With the original DDL, when the loader starts inserting records into the subscriber table, we get:
- first record is successfully inserted
- fails with "conflicting transaction" on the second (or third insert). 

So, I'd speculate that the insert transaction is not being terminated correctly when UNIQUE is specified.

pervazea:
Using 
CREATE TABLE t ( 
  id INTEGER NOT NULL PRIMARY KEY,
  st VARCHAR(15) NOT NULL UNIQUE);

I get an assertion failure at schema.cpp line 156  
PL_ASSERT(column_index < schema->columns.size());

Incorrect assertion?

pervazea:
Using 
CREATE TABLE t ( 
  id INTEGER NOT NULL PRIMARY KEY,
  st VARCHAR(15) NOT NULL UNIQUE);

I get an assertion failure at schema.cpp line 156  
PL_ASSERT(column_index < schema->columns.size());

Incorrect assertion?

pervazea:
There is an interaction between UNIQUE and batch updates. Doing inserts via psql, works fine, as does setting the batch size to 1 for the TATP loader. 

So, there is an interaction between UNIQUE, batch updates and transactions. The TATP loader calls commit after executing the batch, so I assume autocommit is turned off.

pervazea:
With UNIQUE, errors are reported in the trace logs, but none without. 
 
[/home/proj/peloton-pa/peloton/src/network/postgres_protocol_handler.cpp:887:ExecExecuteMessageGetResult] ERROR - Failed to execute: No string conversion for ResultType value '7

so... update std::string ResultTypeToString to include "TO_ABORT"

pervazea:
Minimal SQL for reproducing was:
CREATE TABLE t ( 
  id INTEGER NOT NULL PRIMARY KEY,
  st VARCHAR(15) NOT NULL UNIQUE);

INSERT into t values (1, 'abc');
INSERT into t values (2, 'def');

PR #1120, opened 2018-02-02
Fix for issue #1117, UNIQUE 

pervazea:
- Fix for UNIQUE not working
- Remove inappropriately noisy LOG instance


apavlo:
This great! I will look at it tomorrow

apavlo:
@pervazea This was such a simple mistake. Good find.

Actually, is there a test case that we can add to make sure that this works correctly?

Issue #1101, opened 2018-01-27
ORDER BY is broken


pervazea:
If you run psql_test.sh
(run from the build directory i.e. ../script/testing/psql/psql_test.sh)

the test fails. If you modify the script to leave the psql output file, or run the commands by hand, you can see that the output is not sorted correctly.

crd477:
Is this the cause of the failure in:
http://jenkins.db.cs.cmu.edu:8080/blue/organizations/jenkins/crd%2Fcrdp/detail/jenkins/40/pipeline/18/
for example?

apavlo:
Ugh, we never testing this?

pmenon:
Strange, this is working for me on master.

pervazea:
Prashanth did additional diagnosis using my machine. I'll try to record his findings as accurately as possible below.

The index constructed, is out of order. The plan executed when the test executes "order by", is an index scan. 

Some variations that make this more complicated:

1.  If the table is created with a PRIMARY  KEY, the bug occurs. If you do not create column "id" as a PRIMARY KEY, it works.
1. Prashanth observed that on his machine, a single thread does the work and data is stored in a single tile group. Everything works fine. On my machine, we observed 2 tile groups and the index order is wrong.
1. I tried reducing the number of threads in the thread pool in init.cpp, as Prashanth suggested, for debugging. Peloton segfaults when I do this. I tried 1, 2, 3, 4 threads, with segfaults each time. I would assert that this is not the desired failure mode in response to lack of resources or a configuration error. We'll want to address this later.
1. My machine has 8 cores (4 real, 4 hyperthreads).   Cores can be disabled by setting
/sys/devices/system/cpu/cpuxxx/online e.g. echo 0 |sudo tee /sys/devices/system/cpu/cpu4/online
If I disable the 4 hyperthread cores, leaving my machine as a 4 core machine, everything works. The test passes.

Now... we need to debug why the index has the wrong order.

crd477:
Just for another data point, hyperthreading is enabled on the Jenkins system:
```
$ lscpu
Architecture:          x86_64
CPU op-mode(s):        32-bit, 64-bit
Byte Order:            Little Endian
CPU(s):                12
On-line CPU(s) list:   0-11
Thread(s) per core:    2
Core(s) per socket:    6
Socket(s):             1
NUMA node(s):          1
...
```
Let me know if you want me to disable it there.

pervazea:
[...]

PR #1122, opened 2018-02-03
Fix for ORDER BY (#1101)

poojanilangekar:
This pull request fixes the ORDER BY bug which caused the Jenkins builds to fail. The bug was in the `index_scan_executor.cpp`. The `BWTree` returned the tuples in the correct order, however we were trying to optimize the executor by consolidating all the tuples from a single physical tile into a single logical tile (thereby violating the order by clause). I have fixed the logic and it passes all the order by tests. However, I am running the code on a vagrant machine and I am not able to recreate the parallelism required to create two workers. To test if this works, @apavlo can you trigger the Jenkins build?


apavlo:
@poojanilangekar Thanks!

@pervazea Were you able to recreate this bug in the LLVM engine?

pervazea:
Yes. The bug is independent of  LLVM / codegen.


apavlo:
@pervazea I was stupid. I forgot that we don't support index scan in the LLVM engine yet.

@poojanilangekar Good find!

pervazea:
1. I ran psql_test.sh on this PR on a release build, with 8 cores enabled. It passed, with the following fix:
1. Need to modify psql_test.sh according to the following diff:

-PELOTON_USER="postgres"
+PELOTON_USER="default_database"

PR #1123, opened 2018-02-05
Fix memory leak in query history logging

pmenon:
Tonight, I rebased against master and ran into two tests failing because of memory leaks:
1. `query_logger_test`
2. `ssl_test`

As to how these changes made it into master without failing Travis is a separate issue. But, this PR fixes the first issue. 

`pg_query_fingerprint()` mallocs memory that must be freed by the caller through a companion call to `pg_query_free_fingerprint_result()`. This wasn't being done. I added a wrapper class that leverages RAII to make the appropriate call to safely free up allocated memory.

### DIFF
--- a/src/catalog/query_history_catalog.cpp
+++ b/src/catalog/query_history_catalog.cpp
@@ -6,25 +6,23 @@
 //
 // Identification: src/catalog/query_history_catalog.cpp
 //
-// Copyright (c) 2015-18, Carnegie Mellon University Database Group
+// Copyright (c) 2015-2018, Carnegie Mellon University Database Group
 //
 //===----------------------------------------------------------------------===//
 
 #include "catalog/query_history_catalog.h"
 
 #include "catalog/catalog.h"
-#include "executor/logical_tile.h"
-#include "parser/pg_query.h"
 #include "storage/data_table.h"
 #include "type/value_factory.h"
 
 namespace peloton {
 namespace catalog {
 
-QueryHistoryCatalog *QueryHistoryCatalog::GetInstance(
+QueryHistoryCatalog &QueryHistoryCatalog::GetInstance(
     concurrency::TransactionContext *txn) {
   static QueryHistoryCatalog query_history_catalog{txn};
-  return &query_history_catalog;
+  return query_history_catalog;
 }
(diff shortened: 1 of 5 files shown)


src/catalog/query_history_catalog.cpp
```diff
@@ -36,22 +34,23 @@ QueryHistoryCatalog::QueryHistoryCatalog(concurrency::TransactionContext *txn)
...
+  auto val1 = type::ValueFactory::GetVarcharValue(fingerprint);
   auto val2 = type::ValueFactory::GetTimestampValue(timestamp);
 
-  tuple->SetValue(ColumnId::QUERY_STRING, val0, pool);
-  tuple->SetValue(ColumnId::FINGERPRINT, val1, pool);
-  tuple->SetValue(ColumnId::TIMESTAMP, val2, pool);
+  tuple->SetValue(ColumnId::QUERY_STRING, val0,
+                  pool != nullptr ? pool : &pool_);
+  tuple->SetValue(ColumnId::FINGERPRINT, val1, pool != nullptr ? pool : &pool_);
+  tuple->SetValue(ColumnId::TIMESTAMP, val2, pool != nullptr ? pool : &pool_);
```
malin1993ml: Actually could you remind me what's the difference between using the `pool` pointer from the argument, using `pool_` from the object itself, and just passing in a `nullptr`?

pmenon: A pool has to be used, either the one from the function argument or the member pool, otherwise memory is `new`-ed inside `SetValue` that is never freed.

malin1993ml: I see! And when to use the one from the function argument, and when the member pool?

pmenon: There is a much deeper rooted problem w.r.t. string handling. All these pools never claim memory until they're destroyed, meaning they don't reuse memory. Someone needs to fix this ...

malin1993ml: Okay...

malin1993ml: Thanks for the clarification!

PR #1125, opened 2018-02-05
Read the parameters of thread pool correctly from settings

sivaprasadsudhir:
Minor fix to read the correct parameters for the worker and brain thread pool. This fixes the broken release build issue.

### DIFF
--- a/src/include/threadpool/mono_queue_pool.h
+++ b/src/include/threadpool/mono_queue_pool.h
@@ -24,9 +24,7 @@ namespace threadpool {
  */
 class MonoQueuePool {
  public:
-  MonoQueuePool() = delete;
-
-  MonoQueuePool(int task_queue_size, int worker_pool_size)
+  MonoQueuePool(uint32_t task_queue_size, uint32_t worker_pool_size)
       : task_queue_(task_queue_size),
         worker_pool_(worker_pool_size, &task_queue_),
         is_running_(false) {}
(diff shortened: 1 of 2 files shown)


src/include/threadpool/mono_queue_pool.h
```diff
@@ -24,7 +24,9 @@ namespace threadpool {
  */
 class MonoQueuePool {
  public:
-  MonoQueuePool(int task_queue_size = 32, int worker_pool_size = 4)
+  MonoQueuePool() = delete;
```
pmenon: I don't think you need this because you've defined a constructor with arguments. This is just noise.

test/optimizer/cost_test.cpp
```diff
@@ -42,7 +42,7 @@ namespace test {
 
 using namespace optimizer;
 
-const int N_ROW = 100;
+// const int N_ROW = 100;
```
pmenon: If this isn't being used, remove it entirely.

src/include/threadpool/mono_queue_pool.h
```diff
@@ -24,7 +24,9 @@ namespace threadpool {
...
+
+  MonoQueuePool(int task_queue_size, int worker_pool_size)
```
pmenon: These arguments probably shouldn't be signed, right? Can you change them to `uint32_t`?

test/optimizer/cost_test.cpp
sivaprasadsudhir: The entire test is commented out except for this definition which was giving me a compilation error. So I just commented this out as well.

pervazea: Is the test obsolete now? If so, why not delete it too. We have a lot of commented out code in this file.

pmenon: I think this was commented out in #902. @chenboy I believe said he would make changes to uncomment it?

pmenon: If this is all that's blocking the PR, I'd leave it. Getting the fix in is more important.

PR #1126, opened 2018-02-05
Fix travis build to check status of tests and benchmarks

tcm-marcel:
I made a terrible mistake in #989, that allows travis builds to succeed, even if the tests and benchmarks fail. 
The section `after_success` ignores the exit code of the executed scripts. [1] I changed those scripts to be executed in the section `script`. 

Thanks to @pervazea for finding this and sorry to everyone for making this dump mistake :(

[1] https://docs.travis-ci.com/user/customizing-the-build/#Breaking-the-Build

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,7 +34,14 @@ matrix:
               # COVERALLS: we run coveralls only for one compiler
               - COVERALLS=On
           # only run tests if the build itself succeeded
-          after_success:
+          script:
+              # create build directory
+              - mkdir build
+              - cd build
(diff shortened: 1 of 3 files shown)


tcm-marcel:
@pervazea I also added https://github.com/cmu-db/peloton/pull/1126/commits/56957bf1266387ad762030a960abd9a2e9b96599 which fixes the failing JDBC tests. Before the network refactor peloton ignored if a non-existing database was requested, but now the it fails. I changed it to use `default_database`. 

Together with #1120, #1122 and #1125 it should fix all problems.

pervazea:
Nice. Thank you.

Issue #1094, opened 2018-01-25
Refactor python scripts


tcm-marcel:
As suggested from @pervazea in #1061:

* convert source validator and source formatter to python3
* run Pylint
* remove globals where possible
* put duplicated functions in library
  * Call to clang-format together with finding its correct veesion

schedutron:
Can I work on this one?

tcm-marcel:
Sure. I added two more things to the description.

schedutron:
What happens sometimes is, I edit a source file, and only the edited part isn't formatted well. The validator complains, and then I run the formatter and it formats the entire file. When I push my commits, it's a bit difficult for the reviewers to separate my work from the formatter's cleaning in the diff. Can we somehow add functionality to make the formatter run only on the edited part of a source file?

tcm-marcel:
You can create a first commit with `commit --no-verify` and then format the file in a second commit. This way the reviewer can only use the first commit for the review.

schedutron:
Since the formatter and validator are in different folders, the common library functions have to be put in a package so that both can import them. So should I create a package for the common functions?

schedutron:
Should other Python scripts, like `warning-fixer.py` also be formatted?

pervazea:
FYI, there are accepted coding standards (best practices) for Python, i.e. PEP-8. If you are reformatting the Python code, you should look at and follow PEP-8.

________________________________

schedutron:
@pervazea I know, I have contributed to Python's core documentation :)

I'm using `pylint` here, which uses `PEP-8` as a default configuration. What I want to know here is whether scripts other than the formatter and the validator should also be formatted and ported.

pervazea:
That's good. Thanks.


________________________________

apavlo:
Maybe we should put the formatter and validator in the same folder?

schedutron:
@apavlo Indeed, we can do that, but it may not be a good choice in the long run. I'll try to create a package soon, it simply involves creating a bunch of `__init__.py` files.

schedutron:
@apavlo I've packaged the `script` folder in #1127. Should I also refactor the Python scripts you recently added?

tcm-marcel:
Done in #1127

PR #1127, opened 2018-02-06
Lint Python files and port them to Python3

schedutron:
For #1094.


schedutron:
@tcm-marcel I have linted the validator file and ported it to Python 3. Tested it with some bad formatting cases and it seemed to work. Please test it yourself and let me know if something works incorrectly.

script/validators/stat_validator.py
```diff
@@ -112,20 +124,23 @@ def parse(file_path):
...
+        ['exec ' + PELOTON_BIN + ' -port 54321 -stats_mode 1'], shell=True
+    )
+
+    # start jdbc test
+    os.chdir(PELOTON_JDBC_SCRIPT_DIR)
+    JDBC = subprocess.Popen(
+        ['/bin/bash test_jdbc.sh copy ' + PELOTON_STAT_CSV],
+        shell=True, stdout=subprocess.PIPE, stderr=subprocess.STDOUT
+    )
+    JDBC.stdout.readlines()
```
schedutron: What's the purpose of this line?

script/formatting/formatter.py
```diff
@@ -34,15 +35,14 @@
...
 DEFAULT_DIRS.append(PELOTON_TESTS_DIR)
 
-CLANG_FORMAT = None
 CLANG_FORMAT_FILE = os.path.join(PELOTON_DIR, ".clang-format")
 
 ## ==============================================
 ##             HEADER CONFIGURATION
 ## ==============================================
 
 #header framework, dynamic information will be added inside function
```
schedutron: What should we do to format this section correctly? I was thinking of merging these individual strings to a complete whole. This will eliminate all these unwanted global variables as well.

script/formatting/formatter.py
```diff
@@ -56,81 +56,82 @@
 
 header_comment_1 = header_comment_line_1 + header_comment_line_2
```
schedutron: What should we do to format this section correctly? I was thinking of merging these individual strings to a complete whole. This will eliminate all these unwanted global variables as well.

script/validators/stat_validator.py
tcm-marcel: I don't see any purpose of this line. Remove it and see if the behavior changes.

schedutron: But how do I test this script?

script/formatting/formatter.py
tcm-marcel: Feel free to find find a better way to put this into a string. But make sure one is still able to see the output somehow, so that it can be changed easily.

schedutron:
Why is the build failing for the latest commit?

crd477:
Probably because I am in the middle of invasive Jenkins reconfiguration and aborted it.

tcm-marcel:
@schedutron You can click on the red cross and then check which CI failed and why. 

The formatter and the validator seem fine. I am not familiar with `stat_validator.py`, I will try to find out what it does.

schedutron:
I've packaged the script folder. Now Python scripts can have inter-directory access to each other.

tcm-marcel:
The build fails because of the source validator [1], I think you have to check your relative paths in the scripts. I will take a closer look at the end of the week. Let's finish this PR first and then target the recently added scripts if necessary. 
Btw, I haven't figured out the roots of `stat_validator.py` yet. 

[1] https://travis-ci.org/cmu-db/peloton/jobs/339461799#L2034

schedutron:
@tcm-marcel Cool. I think I'll have to edit the `travis` file as well. That should fix this build issue.

tcm-marcel:
Why? The script should work when called from any working directory.

schedutron:
I've just changed the `/`s to `.`s, the `-m` flag expects dots instead of slashes.

schedutron:
Also, it's a Python standard to run package scripts from the top level directory, one can do workarounds, but they're eccentric.

Jenkinsfile
```diff
@@ -14,7 +14,7 @@ pipeline {
                     }
                     steps {
                         sh 'sudo /bin/bash -c "source ./script/installation/packages.sh"'
-                        sh 'python ./script/validators/source_validator.py'
+                        sh 'python3 -m script.validators.source_validator'
```
apavlo: Why do we want to do this?

schedutron: Since `script` is now a Python package, this is the correct way to access the internal scripts.

schedutron:
The build failed and the log says `Couldn't find any package by regex 'g++-5'`. Why?

schedutron:
The build fails only for Linux with `gcc 5`.

schedutron:
@crd477 Can you restart Jenkins?

schedutron:
@tcm-marcel The build is passing on Travis now, Jenkins failed because @crd477 did an abort.

schedutron:
@tcm-marcel Any updated feedback here?

tcm-marcel:
Sry, forgot about this. I don't think it was aborted. The log shows, that the validator is called the old way (before your changes). 
```
[workspace@4] Running shell script
+ python ./script/validators/source_validator.py
Traceback (most recent call last):

  File "./script/validators/source_validator.py", line 15, in <module>
    from script.helpers import find_clangformat
ImportError: No module named script.helpers
script returned exit code 1
```

I am not familiar with Jenkins, but it might not use the configuration from this commit, but the one from master. In this case it is ok to fail right now. @crd477 can you confirm this?

schedutron:
@tcm-marcel @crd477 Sorry for the confusion, the abort was perhaps done on the previous commit. So leaving Jenkins, is this PR mergeable?

tcm-marcel:
@schedutron One last thing: I think it makes sense to move the complete clang-format call to the helpers module. Create two functions there, one that formats a given file, and one that checks if the given file is already formatted (using the diff as before). You can hand over the logger instance to produce the output. 
Alternatively we can merge the PR first and do this later. 

After this PR the python scripts cannot be called with `python whatever_script.py` anymore, which at least for me is strange, as the error message is not very informative. 
@apavlo Should we simply update the Wiki on how to call the scripts, or create one-liner bash scripts that call the python scripts the right way?

schedutron:
@tcm-marcel I don't think that calling formatter script from any non-root directory is crucial. Ultimately, the usage mentioned in the git hook runs the formatter on all the staged files (and I've tested that it works fine with the new call), so I don't think it's an issue.

schedutron:
@tcm-marcel Also, I don't understand the purpose of creating the functions you mentioned in the helper module. Won't that be like creating a single script for both formatting and validating? Please correct me if I'm wrong.

tcm-marcel:
@schedutron The purpose is to encapsulate the calls to clang-format. We still have the problem, that different version of clang-format produce different output, but not all version are available on all systems. If we need to change anything there, it is all encapsulated in the helpers module, the validator and the formatter do not have to be touched.

schedutron:
@tcm-marcel Thanks, that makes sense. I'll add a commit soon.

crd477:
On 2018-02-17 14:56, Marcel Kost wrote:
> I am not familiar with Jenkins, but it might not use the configuration 
> from this commit, but the one from master. In this case it is ok to fail 
> right now. @crd477 <https://github.com/crd477> can you confirm this?
> 

Correct.  As a security measure, Jenkins only honors changes to the 
Jenkinsfile in PRs from users that already have admin or write 
privileges in the base branch.  Otherwise, random github users could run 
whatever they wanted on the Jenkins server just by sending a PR.

FWIW, I'm not a fan of this PR.  I'd much rather be able to run the 
Python programs simply from the command line or other scripts rather 
than jump through hoops.  In large Python projects like OpenStack that 
have lots of modular interaction, sure this type of change makes sense, 
but these are just utility programs.  I suspect this is what Professor 
Pavlo meant when he asked "why do we want to do this"?

Anyway, @schedutron: if you intend this to work, you should probably not 
assume that Python3 is already installed.  The Docker and Vagrant 
images, for example, start out pretty minimal are unlikely to 
automatically have it.

schedutron:
@crd477 You're right, but there isn't another way to have modular interaction (which is needed here) without involving some hacks like `sys.path.append()` calls. If we're okay with that, I'll unpackage the `script/` folder and add that.

schedutron:
@tcm-marcel Can you review the latest commit? Also, if you guys want to access the scripts in the traditional way, I can try adding `sys.path.append()` calls to both the scripts. Should I do that?

schedutron:
@tcm-marcel I've tried to implement what you said with the validator script, but I'm confused regarding what you mentioned with formatter. The subprocess call for formatter is `[CLANG_FORMAT, "-style=file", "-i", file_path]`, but for validator, it is `[CLANG_FORMAT, "-style=file", file_path]`. Are both (almost) equivalent?

tcm-marcel:
`-i` just performs the changes inplace in the file, while the other call returns the formatted file on stdout, because we want to check it against the original file. So they are almost identical. But the main reason is to hide the subprocess/piping stuff in the helper.

schedutron:
So I can use either one in the helper?

tcm-marcel:
You can still make one call with `-i` and one without. I just want all the subprocess/calling stuff to be in the helper instead of distributed over several files. 

You can create some function `call_clang_format(file, bool inline)` and either make the changes inline (needed for formatter) oder return the output in the function (needed for validator, from the function that does the diff stuff).

tcm-marcel:
@schedutron  I have a question about calling the scripts: I think it would really be a lot easier for people to use them, if the scripts can still be called with `python script.py` (without `-m`) and from any directory. 
Do we still need a hack to achieve this, if the formatter and the validator would be in the same directory together with the helper file/module? It think it may worth it.

schedutron:
@tcm-marcel If all three scripts are in the same folder, no hack is needed, but if they're in different directories and we don't want to use `-m` flag, then we need to add `sys.path.append()` calls to the formatter and validator scripts.

schedutron:
@tcm-marcel I've tried implementing what you suggested, both with the formatter and validator. Please let me know if the scripts need further change. If not, then we can move to addressing the path access problem. Please let me know if we want to move the 3 (or more?) scripts in the same folder. Or should I add the calls I mentioned before?

src/main/peloton/peloton.cpp
```diff
@@ -34,8 +34,7 @@ int main(int argc, char *argv[]) {
 
   // Print settings
   if (peloton::settings::SettingsManager::GetBool(
-          peloton::settings::SettingId::display_settings)) {
-    auto &settings = peloton::settings::SettingsManager::GetInstance();
+          peloton::settings::SettingId::display_settings)) {auto &settings = peloton::settings::SettingsManager::GetInstance();
```
tcm-marcel: This change accidentally got moved into this commit, please remove it.

script/helpers.py
```diff
@@ -84,3 +84,13 @@ def clang_check(file_path):
                 file_status = False
 
         return file_status
+
+
+def clang_format(file_path):
+    """Formats the file at file_path"""
+    if CLANG_FORMAT is None:
```
tcm-marcel: Directly print the error (like in https://github.com/schedutron/peloton/blob/2f4825a022ce86f20de2b181411f879657711225/script/helpers.py#L57) and make the function return void.

schedutron:
@tcm-marcel Done :) And now we can run the scripts as usual!

tcm-marcel:
@schedutron Thanks for your patience!

schedutron:
A conflict had arisen in the pre-commit hook, which has been resolved in the previous commit.

malin1993ml:
@tcm-marcel @pervazea What's the status of this PR? It's said accepted so are we ready to merge this?

@schedutron Can you resolve the conflicts?

tcm-marcel:
@malin1993ml yes, it is ready to merge.

schedutron:
@malin1993ml I've fixed the merge conflicts by accepting incoming changes on the `Jenkinsfile`. Now I think this PR is ready for merging.

PR #1133, opened 2018-02-08
Move parse tree binding from optimizer to catalog

nappelson:
Goal was to refactor the optimizer to pull out the parse tree binding logic as it doesn't belong in the optimizer code. 

In future changes, I'd like to add some utilities to make it easier to modify our tests so that we don't need to touch all of the tests below when making a change like this.


test/executor/copy_test.cpp
```diff
@@ -120,8 +120,7 @@ TEST_F(CopyTests, Copying) {
   auto copy_stmt = peloton_parser.BuildParseTree(copy_sql);
 
   LOG_TRACE("Building plan tree...");
-  auto copy_plan =
-      optimizer->BuildPelotonPlanTree(copy_stmt, DEFAULT_DB_NAME, txn);
+  auto copy_plan = optimizer->BuildPelotonPlanTree(copy_stmt, txn);
```
chenboy: Do we want to bind the parse tree first in this test?

nappelson: Yes done thanks. I am actually curious why this test passes without binding.

apavlo:
@nappelson Can you rebase correctly? It looks like you have duplicate commits.

nappelson:
Yep just fixed the issue thanks - sorry about that.

src/catalog/abstract_catalog.cpp
```diff
@@ -10,11 +10,12 @@
 //
 //===----------------------------------------------------------------------===//
 
+#include "binder/bind_node_visitor.h"
```
pmenon: Follow the style guide for the order of includes [here](https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes) (i.e., the first include should be `catalog/abstract_catalog.h` in `catalog/abstract_catalog.cpp`).

src/catalog/abstract_catalog.cpp
```diff
@@ -37,6 +38,8 @@
 #include "storage/storage_manager.h"
 #include "storage/table_factory.h"
 
+using std::make_shared;
```
pmenon: Don't use these `using` directives.

nappelson: @pmenon thanks! just pushed these changes. I have avoided the use of `using` directives in other projects but noticed that the codebase has a fair amount of these. I will do my best to refactor them out in the future for other diffs.

nappelson:
@apavlo @pmenon sorry about that - now there is no more `using`.

nappelson:
@pmenon just pinging again as this should be ready now.

pmenon:
It looks good overall. General questions:
1. It seems like `BindNodeVisitor` is a single-use class that doesn't outlive its scope of use. If so, does it make sense to allocate on stack rather than heap-allocate with `std::make_shared`?
2. `BindNodeVisitor` exposes an implementation detail (i.e., visitor pattern). Does it make sense to have a generic `Binder` class that implements binding with `BindNodeVisitor`?

mengranwo:
@nappelson 
Hi, can you take a look at this PR and @pmenon's comments? If you have addressed his requests, please change the tag into "accept" 

Thanks!

nappelson:
@pmenon just pushed change addressing your comment regarding stack allocation for BindNodeVisitor. 

For your second comment, that makes sense however the BindNodeVisitor currently implements SqlNodeVisitor (which also exposes an implementation detail). I think a larger refactor like this should be another explicit task to help keep the PRs smaller. 

What are your thoughts?

nappelson:
@pmenon just wanted to check in again

apavlo:
@nappelson We need to come back around and get this in. Can you resolve the conflicts and make sure you address all of the comments?

@malin1993ml Can you work with Nate to get this finished? I think Prashanth is tied up with other things. Thanks

nappelson:
Yep just resolved the conflicts. I responded to one of Prashanth's comments
a while back and checked in but haven't heard back from him. It says the
merge is blocked until I address the requested changes (I am assuming
either Prashanth has to review it again or he needs to be removed as a
reviewer).

Best,
Nate

malin1993ml:
Hi @crd477 @pervazea, it looks like some of our PRs including this one cannot download the repo with the Jenkins build again. For example:
http://jenkins.db.cs.cmu.edu:8080/blue/organizations/jenkins/peloton/detail/PR-1322/1/pipeline/
http://jenkins.db.cs.cmu.edu:8080/blue/organizations/jenkins/peloton/detail/PR-1321/1/pipeline/
http://jenkins.db.cs.cmu.edu:8080/blue/organizations/jenkins/peloton/detail/PR-1133/86/pipeline/
http://jenkins.db.cs.cmu.edu:8080/blue/organizations/jenkins/peloton/detail/PR-1310/7/pipeline/

I'm wondering is this because we have another machine that has networking problem? Would really appreciate your help.

crd477:
Ugh...

Yes, this was the default route problem again that originally cropped up in #1295.  I thought I'd fixed it the other day but obviously it happened again.  I've implemented a more permanent fix now and it should not happen again.

Sorry (again) for the inconvenience...

PR #1138, opened 2018-02-09
Small code cleanups

pervazea:
- remove unnecessary UNUSED_ATTRIBUTE designations
- LOG_ERROR -> LOG_WARNING for SSL startup


src/storage/data_table.cpp
```diff
@@ -821,6 +821,8 @@ bool DataTable::CheckForeignKeyConstraints(
   return true;
 }
 
+
```
pmenon: Remove added lines.

src/network/peloton_server.cpp
```diff
@@ -125,7 +125,7 @@ void PelotonServer::SSLInit() {
   //TODO(Yuchen): change this.
   //load trusted CA certificates (peer authentication)
   if (SSL_CTX_load_verify_locations(ssl_context, certificate_file_.c_str(), nullptr) != 1) {
-    LOG_ERROR("Exception when loading root_crt!");
+    LOG_WARN("Exception when loading root_crt!");
```
pmenon: @ChTimTsubasa Are these actually errors? Can we safely downgrade them to warnings?

ChTimTsubasa: I consulted @haoxianghua and @dengyuchenkit today and this line can be downgraded to warning with further information. I would address the richer information as well as a fix for the log throwing here in an upcoming PR.

src/network/peloton_server.cpp
```diff
@@ -120,7 +120,7 @@ void PelotonServer::SSLInit() {
   // load trusted CA certificates (peer authentication)
   if (SSL_CTX_load_verify_locations(ssl_context, certificate_file_.c_str(),
                                     nullptr) != 1) {
-    LOG_ERROR("Exception when loading root_crt!");
+    LOG_WARN("Exception when loading root_crt!");
```
pmenon: Do we want to add a more descriptive message here to help people diagnose the problem?

pervazea: Yes, the message could be more helpful. Let me do that in my next PR... more clean up to do.

PR #1140, opened 2018-02-09
fix building under Fedora 27

crd477:
fix building under Fedora 27 after https://src.fedoraproject.org/rpms/llvm/c/3d29fda1ea7b5957812adb24bcc478d68f58091a?branch=f27


script/docker/fedora27/Dockerfile-jenkins
```diff
@@ -7,13 +7,13 @@ RUN dnf -q -y install sudo clang
 RUN /bin/bash -c "source ./peloton/script/installation/packages.sh"
 
 RUN echo -n "Peloton Debug build with "; g++ --version | head -1
-RUN mkdir /peloton/build && cd /peloton/build && cmake -DCMAKE_BUILD_TYPE=Debug -DCOVERALLS=False .. && make -j4 && make install
+RUN mkdir /peloton/build && cd /peloton/build && PATH=/usr/lib64/llvm4.0/bin:$PATH cmake -DCMAKE_CXX_FLAGS="-isystem /usr/include/llvm4.0" -DCMAKE_BUILD_TYPE=Debug -DCOVERALLS=False .. && make -j4 && make install
```
tcm-marcel: I don't think the commands for the gcc builds have to be modified.

script/docker/fedora27/Dockerfile-jenkins
```diff
@@ -7,13 +7,13 @@ RUN dnf -q -y install sudo clang
...
 
 RUN echo -n "Peloton Release build with "; g++ --version | head -1
-RUN rm -rf /peloton/build && mkdir /peloton/build && cd /peloton/build && cmake -DCMAKE_BUILD_TYPE=Release -DCOVERALLS=False .. && make -j4 && make install
+RUN rm -rf /peloton/build && mkdir /peloton/build && cd /peloton/build && PATH=/usr/lib64/llvm4.0/bin:$PATH cmake -DCMAKE_CXX_FLAGS="-isystem /usr/include/llvm4.0" -DCMAKE_BUILD_TYPE=Release -DCOVERALLS=False .. && make -j4 && make install
```
tcm-marcel: same here

crd477: Unfortunately, I think they're necessary.  I tested it without specifying those options and I get:
```CMake Error at cmake/Dependencies.cmake:65 (find_package):
  Could not find a package configuration file provided by "LLVM" with any of
  the following names:

    LLVMConfig.cmake
    llvm-config.cmake

  Add the installation prefix of "LLVM" to CMAKE_PREFIX_PATH or set
  "LLVM_DIR" to a directory containing one of the above files.  If "LLVM"
  provides a separate development package or SDK, be sure it has been
  installed.
Call Stack (most recent call first):
  CMakeLists.txt:41 (include)
```

Ultimately, all of the ```Dockerfile-jenkins``` files can be removed from the tree at this point anyway since they're no longer used.

PR #1142, opened 2018-02-11
Support Ubuntu 17.10

apavlo:
This PR has the necessary fixes to get the system to build + run on Ubuntu 17.10. There was a problem with getting C++11 to work correctly where you have to add compiler directives. From what I can tell online this seems to be a problem specific to g++5.4, so I limited the fix to just that version.

Note that I added the header to 'value_factory.h' and for some reason it changed every line.

https://www.youtube.com/watch?v=b0MFNM9oOoU

### DIFF
--- a/script/installation/packages.sh
+++ b/script/installation/packages.sh
@@ -88,20 +88,21 @@ if [ "$DISTRO" = "UBUNTU" ]; then
         $PKG_CLANG \
         git \
         g++ \
-        libgflags-dev \
-        libprotobuf-dev \
-        protobuf-compiler \
         bison \
         flex \
+        valgrind \
+        lcov \
+        protobuf-compiler \
+        libgflags-dev \
+        libprotobuf-dev \
         libevent-dev \
         libboost-dev \
         libboost-thread-dev \
         libboost-filesystem-dev \
         libjemalloc-dev \
-        valgrind \
-        lcov \
         libpqxx-dev \
         libedit-dev \
+        libssl-dev \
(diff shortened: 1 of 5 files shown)


script/installation/packages.sh
```diff
@@ -55,26 +66,43 @@ if [ "$DISTRO" = "UBUNTU" ]; then
...
+
+    # Fix for cmake name change on Ubuntu 14.x and 16.x plus --force-yes deprecation
+    if [ "$MAJOR_VER" == "14" ]; then
+        PKG_CMAKE="cmake3"
+        FORCE_Y="--force-yes"
+    fi
+    # Fix for llvm on Ubuntu 17.x
+    if [ "$MAJOR_VER" == "17" ]; then
+        PKG_LLVM="llvm-3.9"
+        PKG_CLANG="clang-3.8"
```
tcm-marcel: Can you explain why the package is *llvm-3.9*, but we use *clang-3.8*? Or is this a typo?

I also want to remind that I still have this weird behavior on *Ubuntu 16.04* when using *llvm-3.8*, that the build succeeds, but LLVM creates unsupported processor instructions during runtime. We maybe should check if this is only a problem for me.

apavlo: llvm-3.9 was the only version that is available on their repository. I looked for a PPA but I couldn't find it.

This is running correctly on my new workstation at home. I am going to merge this now and then we can revisit when Ubuntu 18.04 LTS comes out.

PR #1145, opened 2018-02-12
Match setting's ssl file name and script's name to compress SSL failing LOG error

ChTimTsubasa:
@haoxianghua and @dengyuchenkit helped me to identify why we were having LOG ERROR printed when server startup.

There are two reasons:
1. We did not explicitly call the certificate generating script at compile time. [https://github.com/cmu-db/peloton/blob/master/script/installation/create_certicates.sh]
2. The certificate name and the and default setting's name did not match. One is "server.crt" and another is "peloton_insecure_server.crt". 

This PR address 2 by changing the default setting. So if the certificate is generated, there would be no longer ERROR LOG printed. @apavlo Should we add a line in Wiki to remind people to use our script if they need?


ChTimTsubasa:
Instructions on how to enable SSL by creating certicicates are added in WiKi (https://github.com/cmu-db/peloton/wiki/Installation)

ChTimTsubasa:
@pbollimp I noticed that you added an in progress flag. Anything wrong in this PR?

pbollimp:
@ChTimTsubasa I assumed that since Pervaze did not approve the review and request some changes, there is some work to be done. If that's not the case please put the label to ready for review and I will email Pervaze to review and approve the changes.

apavlo:
@ChTimTsubasa I will merge this next when the build passes.

PR #1148, opened 2018-02-13
Fix order-by in codegen

pmenon:
This PR fixes a few issues related to sorting in codegen.

1. Fix sorting logic in the presence of NULLs.
2. Fix sorting of BigInt types, which was broken before due issuing a wrong sign-extension instruction.
3. Fix sorting of Decimal types, which was broken due to issuing wrong `sub` instruction.
4. Simplify the generated sort function.
5. Add sorting tests that include NULL data for both integers and strings.

### DIFF
--- a/src/codegen/type/decimal_type.cpp
+++ b/src/codegen/type/decimal_type.cpp
@@ -141,7 +141,7 @@ struct CompareDecimal : public TypeSystem::SimpleComparisonHandleNull {
                            const Value &right) const override {
     // For integer comparisons, just subtract left from right and cast the
     // result to a 32-bit value
-    llvm::Value *diff = codegen->CreateSub(left.GetValue(), right.GetValue());
+    llvm::Value *diff = codegen->CreateFSub(left.GetValue(), right.GetValue());
     return Value{Integer::Instance(),
                  codegen->CreateFPToSI(diff, codegen.Int32Type()), nullptr,
                  nullptr};
--- a/src/codegen/type/type_system.cpp
+++ b/src/codegen/type/type_system.cpp
@@ -141,12 +141,53 @@ Value TypeSystem::SimpleComparisonHandleNull::EvalCompareGte(
   GEN_COMPARE(CompareGteImpl(codegen, left, right));
 }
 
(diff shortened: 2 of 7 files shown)


src/codegen/type/decimal_type.cpp
```diff
@@ -141,7 +141,7 @@ struct CompareDecimal : public TypeSystem::SimpleComparisonHandleNull {
                            const Value &right) const override {
     // For integer comparisons, just subtract left from right and cast the
     // result to a 32-bit value
-    llvm::Value *diff = codegen->CreateSub(left.GetValue(), right.GetValue());
+    llvm::Value *diff = codegen->CreateFSub(left.GetValue(), right.GetValue());
```
tcm-marcel: I wonder why the IR verification doesn't catch instructions which value types don't match.

src/codegen/type/type_system.cpp
```diff
@@ -210,11 +251,34 @@ Value TypeSystem::ExpensiveComparisonHandleNull::EvalCompareGte(
...
+  if (!left.IsNullable() && !right.IsNullable()) {
     return CompareForSortImpl(codegen, left, right);
-  };
-  const auto &result_type = Integer::Instance();
-  return GenerateBinaryHandleNull(codegen, result_type, left, right, impl);
+  }
+
+  /// The logic here is very similar to
+  /// SimpleComparisonHandleNull::EvalCompareForSort() except we **generate**
+  /// an if-clause becuase the non-null check is unsafe.
```
tcm-marcel: Just for clarification: The difference to *SimpleComparisonHandleNull* is, that calling `CompareForSortImpl` in this case is unsafe if one of the values is null?

src/codegen/type/decimal_type.cpp
pmenon: It will. The problem was that this code path was not being exercised in the tests.

src/codegen/type/type_system.cpp
pmenon: Correct.

PR #1150, opened 2018-02-14
Support for Join reordering

GustavoAngulo:
This PR adds support for join reordering, or join associativity, to the optimizer. Supports the reordering of joins of the form (A x B) x C  ---->  A x (B x C). 

Note: Small optimizer reformatting from another uncommitted PR are also present in this PR


GustavoAngulo:
Please mark as in progress

test/optimizer/optimizer_rule_test.cpp
```diff
@@ -62,6 +65,199 @@ TEST_F(OptimizerRuleTests, SimpleRuleApplyTest) {
...
+  optimizer.metadata_.memo.InsertExpression(
+      optimizer.metadata_.MakeGroupExpression(child_join), true);
+
+  // Make Parent join
+  std::vector<AnnotatedExpression> parent_join_predicates;
+  auto dummy_expr = std::shared_ptr<expression::AbstractExpression>{
+      new expression::OperatorExpression(ExpressionType::COMPARE_EQUAL,
+                                         type::TypeId::INTEGER)};
+  std::unordered_set<std::string> pred1_tables({"test1", "test3"});
+  AnnotatedExpression pred1 = {dummy_expr, pred1_tables};
```
malin1993ml: I guess I forgot the implementation details here, but there's no column information in the AnnotatedExpression? Where is column information (which column to compare with) stored then?

src/optimizer/rule_impls.cpp
```diff
@@ -61,13 +61,116 @@ void InnerJoinCommutativity::Transform(
...
+  auto child_join = children[0]->Op().As<LogicalInnerJoin>();
+  auto left = children[0]->Children()[0];
+  auto middle = children[0]->Children()[1];
+  auto right = children[1];
+
+  // Get Alias sets
+  auto &memo = context->metadata->memo;
+
+  auto middle_group_id =
+      children[0]->Children()[1]->Op().As<LeafOperator>()->origin_group;
```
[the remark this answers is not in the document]
GustavoAngulo: Fixed.

test/optimizer/optimizer_rule_test.cpp
GustavoAngulo: I am unsure, I realize the tests should also account for this. @chenboy do you have an idea?

chenboy: The annotated expression should store the predicate, which contains the column information, and the table alias set. I think it should be fine to not construct the expression in this test because join order enumeration will only look at the table of the used columns, not the expression. For now It's hard for a unit test to actually build & bind an expression. We may want to build some testing infrastructure to make life easier.

test/optimizer/optimizer_rule_test.cpp
```diff
@@ -62,6 +65,199 @@ TEST_F(OptimizerRuleTests, SimpleRuleApplyTest) {
...
+  auto left_get = std::make_shared<OperatorExpression>(
+      LogicalGet::make(0, {}, nullptr, "test1"));
+  auto middle_get = std::make_shared<OperatorExpression>(
+      LogicalGet::make(1, {}, nullptr, "test2"));
+  auto right_get = std::make_shared<OperatorExpression>(
+      LogicalGet::make(2, {}, nullptr, "test3"));
+
+  // Create Groups for Get Operators
+  auto left_get_group = memo.InsertExpression(
+      optimizer.metadata_.MakeGroupExpression(left_get), true);
```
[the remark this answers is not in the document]
malin1993ml: I actually don't know whether it should be enforced/not enforced... I just want to make sure. @chenboy Can you explain this one a bit as well?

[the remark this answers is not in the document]
chenboy: It should not. We only enforce physical properties, e.g. we treat`PhysicalOrderBy` as enforced operator when adding sort property. `LogicalGet` should not be created as enforced here.

test/optimizer/optimizer_rule_test.cpp
malin1993ml: I see. That's fine for now.

test/optimizer/optimizer_rule_test.cpp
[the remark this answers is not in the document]
malin1993ml: Do we ever use the enforced/not enforced information in the later stage of the optimization pipeline?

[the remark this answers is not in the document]
chenboy: Yes, we will not enumerate enforced expressions when finding the best expression in a group. This is because we will enforce properties after this step anyway. Another reason is that enforced expressions have its own group as its child. When costing the expression, we may need to cost its child group first, this may cause an endless loop since it would cost itself again and again. We think separating enforced expressions is the clearest way.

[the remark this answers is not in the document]
malin1993ml: I see. Thanks for the clarification!

src/include/optimizer/rule_impls.h
```diff
@@ -38,6 +38,22 @@ class InnerJoinCommutativity : public Rule {
                  OptimizeContext *context) const override;
 };
 
+/**
+ * @brief (A join B) join C -> A join (B join C)
+ */
+
+class InnerJoinAssociativity : public Rule {
```
chenboy: I think we need symmetry rule to perform this to perform A join (B join C) -> (A join B) join C.

malin1993ml: I’ve thought about this as well. I think with associativity and commutativity rule, you can achieve this by a series of transformations. It’s just we need to do those many steps instead of directly get there with another rule. But I think we’re going to exhaust the search space anyway, right?

chenboy: I see. Let's just keep it as is.

PR #1152, opened 2018-02-14
Doxygen comments for files in concurrency

aaron-tian:


src/concurrency/timestamp_ordering_transaction_manager.cpp
```diff
@@ -25,25 +25,48 @@
...
-// transaction
-// that reads the tuple.
+/**
+ * In timestamp ordering, the last_reader_cid records the timestamp of the last
+ * transaction that reads the tuple.
+ *
+ * @param[in]  tile_group_header  The tile group header
+ * @param[in]  tuple_id           The tuple identifier
+ *
+ * @return     True if success, False otherwise
```
pervazea: Return is a cid_t, not true or false.

src/concurrency/timestamp_ordering_transaction_manager.cpp
```diff
@@ -25,25 +25,48 @@
 namespace peloton {
 namespace concurrency {
 
-// timestamp ordering requires a spinlock field for protecting the atomic access
-// to txn_id field and last_reader_cid field.
+/**
```
pervazea: I'm not seeing how to insert these at the right line numbers. 
Shall we also add Doxygen comments for the following functions in this file? Line numbers on the left. 

218: AcquireOwnership - missing header
271: PerformRead - missing header
535: PerformInsert
695: PerformDelete
781: PerformDelete
1000: AbortTransaction

src/include/concurrency/epoch_manager.h
```diff
@@ -21,28 +21,53 @@
...
+   */
   virtual void StartEpoch() = 0;
 
+  /**
+   * @brief      Stops an epoch.
+   */
   virtual void StopEpoch() = 0;
 
-  //====================================================
+  //----------------------------------------------------------------------------
```
pervazea: Why change just one? Should be consistent.

src/concurrency/timestamp_ordering_transaction_manager.cpp
aaron-tian: Yes, I didn't comment them since I thought the function names were clear, but we should do it for clarification and consistency. Will add these and fix the one below.

PR #1155, opened 2018-02-16
ensure that the dependencies for jdbc_validator.py are installed

crd477:
These packages have also been installed in the prepopulated Docker images in the Jenkins server.


crd477:
On 2018-02-17 12:10, Marcel Kost wrote:
> A general question, because I didn't follow the jenkins conversations: 
> why do we not run the tests on all configurations?
> 

There's no strong reason other than I was working through a basic matrix 
with @pervazea at one point and was roughly mimicking what was in 
```travis.yml``` at that time.  The matrix can be anything the group 
wants and is all driven by the Jenkinsfile now.

crd477:
Actually the apt-get problem you noticed was a real one.  I've fixed it in c643158.

PR #1156, opened 2018-02-16
Avoid data copy when constructing catalog object cache

### LINKED ISSUE #1137: Avoid data copy when constructing catalog object cache
When creating a [catalog object cache](https://github.com/cmu-db/peloton/blob/master/src/catalog/table_catalog.cpp#L27) (I'll use table catalog object as an example), we pass in an `executor::LogicalTile` in the constructor and use the `GetValue()` method to extract result for specific column. However, `GetValue()` method involves tons of unnecessary data copy. What we want is to [reinterpret_cast](http://en.cppreference.com/w/cpp/language/reinterpret_cast)
 a pointer that points to the starting address of a tuple directly into catalog object cache .

 Here is the list of files that we may need to take a look at:
 1. src/catalog/table_catalog.cpp
 2. src/planner/index_scan_plan.cpp
 3. src/executor/plan_executor.cpp

mbutrovich:
This PR changes the behavior when creating a catalog object cache: currently we pass in an executor::LogicalTile in the constructor and use the GetValue() method to extract result for specific column. However, GetValue() method involves tons of unnecessary data copy. Instead, this fix reinterpret_casts a pointer that points to the starting address of a tuple directly into catalog object cache.

This resolves #1137.


src/catalog/index_catalog.cpp
```diff
@@ -175,7 +201,8 @@ bool IndexCatalog::InsertIndex(oid_t index_oid, const std::string &index_name,
   return InsertTuple(std::move(tuple), txn);
 }
 
-bool IndexCatalog::DeleteIndex(oid_t index_oid, concurrency::TransactionContext *txn) {
```
poojanilangekar: Please undo the formatting change.

src/catalog/index_catalog.cpp
```diff
@@ -273,7 +300,8 @@ std::shared_ptr<IndexCatalogObject> IndexCatalog::GetIndexObject(
  * @return  a vector of index catalog objects
  */
 const std::unordered_map<oid_t, std::shared_ptr<IndexCatalogObject>>
-IndexCatalog::GetIndexObjects(oid_t table_oid, concurrency::TransactionContext *txn) {
```
poojanilangekar: Same as above.

src/catalog/language_catalog.cpp
```diff
@@ -20,11 +20,22 @@
...
+
+  field_location = tile->GetTuplePointer(0, LanguageCatalog::ColumnId::LANNAME);
+  PL_ASSERT(field_location != nullptr);
+  auto varchar = *(reinterpret_cast<const char *const *>(field_location));
+  auto size = reinterpret_cast<const uint32_t *>(varchar);
+  auto string = (reinterpret_cast<const char *>(varchar) + sizeof(uint32_t));
+  lang_name_ = std::string(string, *size - 1);
+}
 
-LanguageCatalog &LanguageCatalog::GetInstance(concurrency::TransactionContext *txn) {
```
poojanilangekar: Same as above.

mbutrovich:
I've made the requested changes, but regarding
> I have another question, what happens to the calls to the function in the LLVM engine? I don't think you added a proxy function in the LLVM engine. Would this still work after we completely deprecate the interpreted engine?
I'm afraid I don't know enough about the LLVM engine enough to say. Would Prashanth be the best one to talk to?

apavlo:
Yes. @pmenon can you chime in here?

src/catalog/column_catalog.cpp
```diff
@@ -21,29 +21,57 @@ namespace peloton {
...
+                                         int tupleId) {
+  auto field_location =
+      tile->GetTuplePointer(tupleId, ColumnCatalog::ColumnId::TABLE_OID);
+  PL_ASSERT(field_location != nullptr);
+  table_oid = *(reinterpret_cast<const oid_t *>(field_location));
+
+  field_location =
+      tile->GetTuplePointer(tupleId, ColumnCatalog::ColumnId::COLUMN_NAME);
+  PL_ASSERT(field_location != nullptr);
+  auto varchar = *(reinterpret_cast<const char *const *>(field_location));
```
mengranwo: Quick question. 
Does Peloton store all the varchar type value in place or using a char * pointer as a placeholder or both?

mengranwo:
I have reviewed Matt's PR and at first I thought it’s not exactly what I expected. What I expect would be something like this: 
**TableCatalogObject object = reinterpret_cast<TableCatalogObject*>(ptr); (where ptr is the physical starting address of some tuple)**
 
If we have string type as member variable in TableCatalogObject(e.g table_name), there is no way to do the reinterpret cast as the way I described above. Because in Peloton, for varchar type, we have two mechanism for underly storage. 1.store length of the string and its content in place  2.use a pointer as a placeholder where the pointer points to a different memory space.   
I guess you have to reinterpret cast every attribute/column (However, if Peloton supports char type with fixed length, then what I suggest would be feasible)

mbutrovich:
Closing and will recreate with a new branch when ready.

PR #1158, opened 2018-02-17
Support for optimizer task execution timeouts

nappelson:
Support configurable optimizer task execution timeouts. Not sure what the appropriate task timeout default should be so we can discuss that here and I can change it. 

Wrote a somewhat naive test but in the future, we can have more appropriate integration tests once join reordering lands as that will increase the optimization time. 

I am also interested in discussing how often we should check for timeouts. This can come in a later PR or this one if we feel that a change is necessary.


src/include/optimizer/optimize_context.h
```diff
@@ -12,27 +12,37 @@
 
 #pragma once
 
+#include "common/timer.h"
+
 #include "optimizer/property_set.h"
 #include "optimizer/optimizer_task.h"
 #include "optimizer/optimizer_task_pool.h"
 
+#include "settings/settings_manager.h"
```
chenboy: Put all include files next to each other.

src/include/optimizer/group.h
```diff
@@ -73,9 +73,11 @@ class Group {
 
   std::shared_ptr<ColumnStats> GetStats(std::string column_name);
 
-  void AddStats(std::string column_name, std::shared_ptr<ColumnStats> stats);  
+  void AddStats(std::string column_name, std::shared_ptr<ColumnStats> stats);
 
-  bool HasColumnStats(std::string column_name);  
+  bool HasColumnStats(std::string column_name);
+
+  bool HasExpressions(const std::shared_ptr<PropertySet> &properties) const;
```
chenboy: Please add Doxygen comments for this function.

src/optimizer/optimizer.cpp
```diff
@@ -336,5 +339,30 @@ unique_ptr<planner::AbstractPlan> Optimizer::ChooseBestPlan(
   LOG_TRACE("Finish Choosing best plan for group %d", id);
   return plan;
 }
+
+void Optimizer::ExecuteTaskStack(
+    std::unique_ptr<OptimizerTaskStack> &task_stack, const Group *root_group,
```
chenboy: A reference to a unique_ptr seems a little bit odd to me. Could we just use a raw pointer?

src/include/optimizer/optimize_context.h
```diff
@@ -12,27 +12,37 @@
...
         required_prop(required_prop),
-        cost_upper_bound(cost_upper_bound) {}
+        cost_upper_bound(cost_upper_bound),
+        timeout_limit(timeout_limit),
+        timer(Timer<std::milli>()) {}
 
   OptimizerMetadata *metadata;
   std::shared_ptr<PropertySet> required_prop;
   double cost_upper_bound;
+  unsigned int timeout_limit;
```
chenboy: Could we put members related to timeout into `OptimizerMetadata`? `OptimizerMetadata` is kept per query, while `OptimizeContext` is kept for each operator.

src/optimizer/optimizer.cpp
```diff
@@ -336,5 +339,30 @@ unique_ptr<planner::AbstractPlan> Optimizer::ChooseBestPlan(
...
+    // Check to see if we have at least one plan, and if we have exceeded our
+    // timeout limit
+    if (timer.GetDuration() >= timeout_limit &&
+        root_group->HasExpressions(required_props)) {
+      throw OptimizerException("Optimizer task execution duration " +
+                               std::to_string(timer.GetDuration()) +
+                               " exceeds timeout limit " +
+                               std::to_string(timeout_limit));
+    }
+    timer.Start();
```
chenboy: Looks like we reset the timer for each task? I think we should start the timer once for each query.

nappelson: we aren't resetting the timer here. we are simply stopping it after a given task execution and checking if the total timer duration exceeds the limit.

src/include/optimizer/optimize_context.h
nappelson: done

nappelson:
@chenboy just addressed your changes

src/optimizer/optimizer.cpp
chenboy: I see what you mean. But this does not seem to be measuring the actual execution time of the optimizer because some part of the execution hasn't been counted. Could we only call `Start()` once and after calling `Stop()` to capture the total duration, we call `Reset()` to zero `elapse_`?

nappelson: I think I am confused sorry - wouldn't this make it so that we can't stop in the middle of task execution if we went over the time limit?

mengranwo:
@chenboy 
Can you take look at the latest commit? If this PR is ready to merge, please change the tag into **"accept"** 

Thanks

chenboy:
@mengranwo Nate and I are still discussing this PR. It should take another day or two. Thanks!

mengranwo:
@chenboy
Any updates?

nappelson:
We discussed and came to a conclusion. I am currently out of town and hope
to have the PR updated tonight or tomorrow.

chenboy:
The update looks good. Seems like there are some duplicated commits though. After fixing these commits, this PR is good to go.

chenboy:
LGTM. This could be merged when build passes.

PR #1159, opened 2018-02-18
Tensorflow Integration - Step 1

saatviks:
As part of this first step of Tensorflow integration, I'm adding the following:
1. Installation of Python(where necessary), Python3-Pip, Tensorflow binaries for C/Python. Also manual installation of Protobuf 3.4.0+
2. CMake changes for integrating tensorflow and automatically generating model files.
3. A simple test to check if tensorflow has imported correctly.

It first needs to build correctly on all platforms - We then incrementally add in our other developed modules for `peloton-brain`.

### DIFF
--- a/script/installation/packages.sh
+++ b/script/installation/packages.sh
@@ -17,6 +17,7 @@
 ##  * OSX
 ## =================================================================
 
+
 # Determine OS platform
 UNAME=$(uname | tr "[:upper:]" "[:lower:]")
 # If Linux, try to determine specific distribution
@@ -36,6 +37,50 @@ fi
 unset UNAME
 DISTRO=$(echo $DISTRO | tr "[:lower:]" "[:upper:]")
 TMPDIR=/tmp
+TF_VERSION="1.4.0"
+TF_TYPE="cpu"
+
+
+function install_protobuf3.4.0() {
+ # Install Relevant tooling
+ # Remove any old versions of protobuf
+ DISTRIB=$1 # ubuntu/fedora
+ if [ "$DISTRIB" == "ubuntu" ]
+ then
+    sudo apt-get --yes --force-yes remove --purge libprotobuf-dev protobuf-compiler
+ elif [ "$DISTRIB" == "fedora" ] 
+ then
+    sudo dnf -q remove -y protobuf protobuf-devel protobuf-compiler
+ else 
+    echo "Only Ubuntu and Fedora is supported currently!"
+    return 0
+ fi
+ CWD=`pwd`
+ cd $TMPDIR
+ wget -O protobuf-cpp-3.4.0.tar.gz https://github.com/google/protobuf/releases/download/v3.4.0/protobuf-cpp-3.4.0.tar.gz
+ tar -xzf protobuf-cpp-3.4.0.tar.gz
+ cd protobuf-3.4.0
+ ./autogen.sh && ./configure && make -j4 && sudo make install && sudo ldconfig
+ # Do cleanup
+ cd $CWD
+}
+
+# Utility function for installing tensorflow components of python/C++
+function install_tf() {
+ TFCApiFile=$1
+ TFBinaryURL=$2
+ LinkerConfigCmd=$3
+ TARGET_DIRECTORY="/usr/local"
+ # Install Tensorflow Python Binary
+ sudo pip3 install --upgrade ${TFBinaryURL}
+
+ # Install C-API
+ TFCApiURL="https://storage.googleapis.com/tensorflow/libtensorflow/${TFCApiFile}"
+ wget -O $TFCApiFile $TFCApiURL
+ sudo tar -C $TARGET_DIRECTORY -xzf $TFCApiFile
+ # Configure the Linker
+ eval $LinkerConfigCmd
+}
 
 ## ------------------------------------------------
 ## UBUNTU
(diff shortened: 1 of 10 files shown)


script/installation/packages.sh
```diff
@@ -88,13 +133,14 @@ if [ "$DISTRO" = "UBUNTU" ]; then
         $PKG_CLANG \
         git \
         g++ \
+        clang-3.7 \
+        $CMAKE_NAME \
+        libgflags-dev \
```
malin1993ml: I think we already have $PKG_CMAKE, $PKG_CLANG, and libgflags-dev out there in the list. Those three here seem duplicated?

src/CMakeLists.txt
```diff
@@ -34,6 +34,34 @@ peloton_default_properties(peloton-proto)
...
 
+##################################################################################
+
+# TODO: Guess this should be moved into a module?
+# --[ tensorflow
+# Add library to be linked
+find_library(TFlowC
+        NAMES tensorflow
+        PATHS "/usr/local/lib")
+set(Peloton_LINKER_LIBS "${TFlowC}" ${Peloton_LINKER_LIBS}) # note, crucial to prepend!
```
malin1993ml: Yes it's better to create a separate module file for this. But even though you don't want to do that because it's short, at least you should move this to cmake/Dependencies.cmake instead of here.

src/CMakeLists.txt
```diff
@@ -34,6 +34,34 @@ peloton_default_properties(peloton-proto)
...
+
+# Generate essential model files
+set(MODEL_GEN_DIR ${CMAKE_CURRENT_LIST_DIR}/brain/modelgen)
+set(MODEL_GEN_PY ${MODEL_GEN_DIR}/model_generator.py)
+set(MODEL_GEN_SETTINGS ${MODEL_GEN_DIR}/settings.json)
+set(MODEL_GEN_OUT ${MODEL_GEN_DIR}/LSTM.pb)
```
malin1993ml: I don't think you want to check for a specific model file here. I think either you check for each individual pb file for each model, or just simply recreate all the pb files when calling cmake anyway.

src/brain/modelgen/LSTM_Model.py
```diff
@@ -0,0 +1,133 @@
...
+        abs_path = os.path.join(dir, fname)
+        if not os.path.exists(abs_path):
+            tf.train.write_graph(tf.get_default_graph(),
+                                 dir, fname, False)
+
+
+    def __repr__(self):
+        return "LSTM"
+
+
```
malin1993ml: Remove extra spaces.

src/include/brain/tf_util.h
```diff
@@ -0,0 +1,30 @@
+//===----------------------------------------------------------------------===//
+//
+//                         Peloton
+//
+// tf_util.h
+//
+// Identification: src/include/brain/tf_util.h
+//
+// Copyright (c) 2015-16, Carnegie Mellon University Database Group
```
malin1993ml: Copyright should be 2015-2018. I think you can just run the [formatter](https://github.com/cmu-db/peloton/blob/master/script/formatting/formatter.py) to get all those things right.

test/brain/tensorflow_util_test.cpp
```diff
@@ -0,0 +1,32 @@
+//===----------------------------------------------------------------------===//
+//
+//                         Peloton
+//
+// tensorflow_test.cpp
+//
+// Identification: test/brain/tensorflow_test.cpp
+//
+// Copyright (c) 2015-16, Carnegie Mellon University Database Group
```
malin1993ml: Same here.

test/include/common/harness.h
```diff
@@ -10,30 +10,29 @@
 //
 //===----------------------------------------------------------------------===//
 
-
```
malin1993ml: Revert the format change if you didn't change this file.

src/brain/modelgen/LSTM_Model.py
```diff
@@ -0,0 +1,133 @@
+import tensorflow as tf
```
malin1993ml: I think we need copyright header for these model files as well.

src/brain/modelgen/model_generator.py
```diff
@@ -0,0 +1,26 @@
+import json
```
malin1993ml: Copyright header.

src/brain/modelgen/LSTM_Model.py
saatviks: The `formatter.py` script cant create a copyright header for Python files currently. Should I add in the functionality?

saatviks:
There were two more things I wanted to confirm:

1. The `protobuf-3.4.0` is being built from source which takes between 5-10 minutes. Is this ok? There are currently no ubuntu packages for this, but there are for mac. That said, I'm guessing the functionality would come in soon enough.
2. Currently I haven't checked for support on Fedora or other linux distros. Is this alright?

src/brain/modelgen/LSTM_Model.py
malin1993ml: Then you need to be watchful to only create header for py files in src dir. You don’t want to create headers for all the scripts in the repo.

malin1993ml:
1. That’s okay for now. After this is merged, remember to write an issue to switch it after the official distribution comes out.
2. We need to get it build on other platforms in our Jenkins test before it’s finally merged, unless there’s some really difficult issue.

malin1993ml:
I've looked at this PR. I think all my review comments have been addressed except two minor things left:
1. I think you can manually add the copyright headers for the Python models for now. You don't need to change the existing Python scripts in the system. Just your files are fine.
2. We're adding the Fedora test to Jenkins in #1180. So after that PR got merged, you can then test the Fedora build with your stuff by looking at your PR test results. When there's starvation on Jenkins, you can manually restart your PR build on Jenkins.

malin1993ml:
Since "make check" on Fedora cannot build anyway, we're going to merge this after the Travis build passes, and post an issue about the Fedora runtime error.

PR #1161, opened 2018-02-20
Minor Cleanup of WorkerPool + Timer

apavlo:
1. Refactored `WorkerPool` to remove `std::atomic_bool`.
2. Refactored `Timer` to add the ability to record the total time.

https://www.youtube.com/watch?v=6op6TdDdl6g


src/include/threadpool/worker_pool.h
```diff
@@ -20,9 +20,16 @@
...
+
 using TaskQueue = peloton::LockFreeQueue<std::function<void()>>;
 
-void WorkerFunc(std::atomic_bool *should_shutdown, TaskQueue *task_queue);
+/**
+ * Wrapper method for invoking a function inside of the worker pool
+ * @param should_shutdown A pointer to boolean that tells whether to shutdown
+ * @param task_queue Where to poll for tasks
+ */
+void WorkerFunc(WorkerPool *pool, TaskQueue *task_queue);
```
pmenon: Since this function isn't needed in the header, you can move it to the cpp file in an anonymous namespace.

test/executor/update_test.cpp
```diff
@@ -2,9 +2,9 @@
 //
 //                         Peloton
 //
-// delete_test.cpp
+// update_test.cpp
 //
-// Identification: test/executor/delete_test.cpp
+// Identification: test/executor/update_test.cpp
 //
 // Copyright (c) 2015-16, Carnegie Mellon University Database Group
```
pmenon: Fix the date while you're at it.

src/include/threadpool/worker_pool.h
```diff
@@ -34,24 +41,33 @@ class WorkerPool {
...
 
  private:
   std::vector<std::thread> workers_;
   size_t num_workers_;
-  std::atomic_bool should_shutdown_;
+
+  // This used to be a std::atomic_bool
+  // It doesn't need to be because don't care about updating
+  // this flag atomically.
+  bool should_shutdown_;
```
pmenon: I'm not sure this is safe. Making it a vanilla member variable allows the compiler to optimize away the loop check in `WorkerFunc`. While we don't need concurrent updates to the flag, we do need every read to go through cache/memory. Atomics provide this (and `volatile` I think, but it's probably safer to use C++ atomics).

apavlo: It doesn't matter if it misses the update to the flag. It will loop back and check it again. Every read does not need to be consistent:

https://github.com/cmu-db/peloton/pull/1161/files/f57ef48227ef3eadadc7a518aef4bbda35c4c6d7#diff-01c2e44c857eb9c764c6c1fe0f244677R23

pmenon:
@apavlo I think we agreed to revert the changes to the boolean flag and make it a `std::atomic_bool`. Let me know if you don't have time for this and I can finish it.

apavlo:
Replaced by #1391

Issue #1164, opened 2018-02-23
Peloton connects to nonexistent database


schedutron:
In the `psql` client, if I type `\c foo`, where the database `foo` has not yet been created, Peloton shows a success message:

```
binaryBoy=# \c foo;
psql (10.1, server 9.5devel)
You are now connected to database "foo" as user "binaryBoy".
foo=# 
```

Later, if we, say, try to create a table, then it shows an error:

```
foo=# create table bar(id int);
Can't find Database foo to create table
foo=# 
```

eribeiro:
Hi @schedutron, it's not my intention to step on your toes, so I am leaving this here as a patch file you can git apply and see if it solves what you need or at least guide you through a direction. I dunno the trivial test I wrote is working, for example. But it seems to have solved the issue. /cc @apavlo

[1164.txt](https://github.com/cmu-db/peloton/files/1825796/1164.txt)

Please, let me know what you both think, because it's not my intention to create a PR to address this at first, as you are in charge.

Cheers!

eribeiro:
Oh, and the "3D000" code you see in the patch came from here: https://www.postgresql.org/docs/10/static/errcodes-appendix.html and I could confirm via Wireshark on a psql connection to Postgres.

pmenon:
I think #1224 supersedes this. Closing for now.

PR #1165, opened 2018-02-23
Check database existence before connecting (in progress)

schedutron:
For #1164. @pmenon Can you please add the `in progress` label? Also, to check for database existence in `traffic_cop.h`, we need access to the `Catalog` object, right? Can you please hint me on how do we do that?

### DIFF
--- a/src/include/traffic_cop/traffic_cop.h
+++ b/src/include/traffic_cop/traffic_cop.h
@@ -142,6 +142,7 @@ class TrafficCop {
   executor::ExecutionResult p_status_;
 
   void SetDefaultDatabaseName(std::string default_database_name) {
+    // check before changing the value!
     default_database_name_ = std::move(default_database_name);
   }
 


src/include/traffic_cop/traffic_cop.h
```diff
@@ -142,6 +142,7 @@ class TrafficCop {
   executor::ExecutionResult p_status_;
 
   void SetDefaultDatabaseName(std::string default_database_name) {
+    // check before changing the value!
```
pranavk: It looks like you didn't push the actual code. It's just a comment. Surely that's not supposed to fix any problem :)

pranavk:
But yeah, alright. I see now that it says 'in progress'. Sorry for the noise.

schedutron:
@pmenon Can you please hint me on how do I accomplish this check? I basically need to check for membership in a set of available databases' names.

apavlo:
@ChTimTsubasa Can you help @pranavk with this?

pranavk:
@apavlo Sorry, I just have a stray comment on this issue. I am not working on it. @schedutron is on it.

ChTimTsubasa:
The place you might want to insert the code to check the database existence is in https://github.com/cmu-db/peloton/blob/master/src/network/postgres_protocol_handler.cpp#L1118, where the protocol handler is checking the existence of the packet and send back a response. You might want to check what would be the correct response to send back. Using Wireshark on psql and Postgres would help.

edwardoliveira:
Hi @schedutron, it's not my intention to step on your toes, so I am leaving this here as a patch file you can git apply and see if it solves what you need or at least guide you through a direction. It lacks tests, for example. I dunno the trivial test I wrote is working, for example. But it seems to have solved the issue.

[1164.txt](https://github.com/cmu-db/peloton/files/1825312/1164.txt)


Please, let me know what you think, I am not able to create a PR now nor it is my intention as you are in charge. /cc @apavlo 

Cheers!

Issue #1114, opened 2018-01-31
Create HashIndex Class


apavlo:
The catalog team needs a fast index for looking object objects based on their OIDs. Right now we only support the Bw-Tree for indexes. We need to switch to use a hash table index for these. Peloton already links in libcuckoo in its build chain, so we can just use that:

https://github.com/cmu-db/peloton/tree/master/third_party/libcuckoo

There are three parts to this task:

1. Create a new class called `HashIndex` in 'src/include/index'. This class should extend the base `Index` class API: 

   https://github.com/cmu-db/peloton/blob/master/src/include/index/index.h#L213

   The `HashIndex` should use a `cuckoohash_map` as its underlying data structure. You can reference `CuckooMap` to see how it works:

   https://github.com/cmu-db/peloton/blob/master/src/include/common/container/cuckoo_map.h

   The `HashIndex` will not be able to support any of the scan operations in `Index`. It should throw a `NotImplementedException` if somebody tries to invoke those methods. One thing that might be tricky is supporting conditional inserts (`Index::CondInsertEntry`). I don't know whether `cuckoohash_map` supports that.

2. Modify `IndexFactory::GetIndex` to support creating `HashIndex` objects. It should use the existing `IndexType::HASH` type id. Make sure you write test cases to to ensure that this works as expected.

3. Modify all of the catalog initialization code to use the new `HashIndex` object now instead of a Bw-tree:

   https://github.com/cmu-db/peloton/blob/master/src/catalog/catalog.cpp#L81

MohamedBouzid:
hi, I am interested to contribute

MohamedBouzid:
@apavlo  Ok i will be happy to do so. How the meeting will be held ?

dasdipanjan04:
Hi,

Is anyone working on this issue? If not I would like to work on this. @MohamedBouzid  @apavlo please let me know.

Thanks and Regards
Dipanjan Das

apavlo:
@MohamedBouzid Tuesdays @ 12pm (GHC 7501)

@dasdipanjan04 @qwkai is working on this.

PR #1167, opened 2018-02-25
Create hash index class

qwkai:
This PR creates a `HashIndex` class (mentioned in #1114) which provides a fast index to look objects based on their OIDs for the catalog team.


poojanilangekar:
@qwkai, your tests are failing. Please fix them so I can review your code.

qwkai:
Having tests failing on `ConditionalInsert`. It throws unimplemented exception because CuckooHash does not support.

tcm-marcel:
@qwkai Are you still working on this or do you need help?

qwkai:
@tcm-marcel I am having bugs in some tests saying that `Automatic expansion triggered when load factor was below minimum threshold`. It's a variable set in `cuckoohash_config.h`. I haven't figured out the reason leading this. Some help would be really appreciated!

poojanilangekar:
@qwkai I know that error. It has something to do with the quality of the hash function you use. What is the hash function you are using on the key? You can try and use city hash. Also, your PR is old and is making changes which are already in the master branch. So I would suggest you pull the master and create a new branch with the relevant changes in them.

qwkai:
Hey @poojanilangekar I am currently using CompactIntsHasher and GenericHasher on the index. I tried to include `city_hash.hh` and use CityHasher, but it seems it includes some file called `city.h` which I cannot locate. I searched codes using CityHasher but there is little useful info. Do you have any idea how to use CityHasher correctly? Thanks very much!

poojanilangekar:
@qwkai I am looking at it. I will get back to you in a day or two.

poojanilangekar:
@qwkai I think there is some issue with the way we compute the hash over a tuple. I changed the seed but that didn't help either. I think there is a problem with the way we combine the hashes from multiple columns. Can you try to use the `tbb::concurrent_unordered_map` and try to raise a new PR?

PR #1174, opened 2018-02-28
Making the ReadWriteSet ThreadSafe

poojanilangekar:
First step of changing the ReadWriteSet. Currently the unordered_map is changed to tbb::concurrent_unordered_map. If this passes the Jenkins build, the next step would be to modify the ReadWriteSet type to tbb::concurrent_unordered_map<ItemPointer, RWType>.


poojanilangekar:
I had opened this PR to check if the bug was in the cuckoohash PR. But now that we are able to reproduce the same bug on the master branch, I am closing this PR.

PR #1177, opened 2018-03-01
Fix the -fPIC flag setting for external projects.

### LINKED ISSUE #1173: Jenkins build fails: PIC error
Jenkins fails in build `Build / Ubuntu Trusty/gcc-4.8.4/llvm-3.7.1 (Debug/Test/LOG_LEVEL_TRACE)` with errors of the following format:

```
/usr/bin/ld.gold: error: <file> requires dynamic R_X86_64_PC32 reloc against <symbol> which may overflow at runtime; recompile with -fPIC
/usr/bin/ld.gold: error: <file> requires unsupported dynamic reloc 11; recompile with -fPIC
```

* The other builds (gcc4.8 Release and gcc5 Debug + Release) seem not affected. 
* The error occurs on all branches.

malin1993ml:
This fixes #1173. The way we were setting the -fPIC flag for external projects cannot propagate those flags appropriately. I don't know why we were able to pass the build before either...


malin1993ml:
@pervazea Please merge this if Marcel is okay with it and the build passes. Without this PR the Jenkins will fail even when no "starvation" occurs.

tcm-marcel:
@crd477 has the answer on why it didn't occur before: https://github.com/cmu-db/peloton/issues/1173#issuecomment-369615102

>  I think it only occurs now after the switch to the gold linker in #1109. I can confirm that Fedora 27 also needs the -fPIC change as hinted at in #1105 (comment)

tcm-marcel:
A question just because I am curious: I thought that Jenkins always takes the configuration from the *master* branch, so how is it possible that the build for this PR succeeds on Jenkins?

malin1993ml:
@tcm-marcel What do you mean by "configuration" here? Also just curious, what is this so-called "gold linker"...?

tcm-marcel:
> What do you mean by "configuration" here? 

Sorry, I mixed up the CMake files with the actual Jenkins configuration in `/Jenkinsfile`. It makes sense now. 

To be honest, I have never heard of the *gold* linker before, but it just seems to be a better linker that doesn't use the BFD library anymore. 
@pmenon Is there a special reason we need the *gold* linker for #1109 , or just because it's better/faster?

pmenon:
It's a faster linker. It is used if it's available on the system. Otherwise, we fall back to the default GNU linker.

Issue #1169, opened 2018-02-26
Jenkins pipeline configurations


pervazea:
We'd like to have more consistency in the Jenkins jobs, so the following is proposed for pipelines:

For debug and release pipeline, which would be similar to current debug/test pipeline (not enumerating every detailed step below):
1. packages.sh
1. source validator
1. build with Debug (or Release), Coveralls false, USE_SANITIZER=Address (sanitizer only on Debug builds)
1. make check
1. (omit make benchmark)
1. make install
1. psql_test.sh
1. jdbc_validator.py

So, we'd have the above for:
1. Ubuntu Xenial gcc 5.4, Debug
1. Ubuntu Xenial gcc 5.4 Release
1. Ubuntu Trusty gcc 4.8 Debug
1. Ubuntu Trusty gcc 4.8 Release

In addition to this, add a Trace build, but without test execution. Execution with the trace build produces excessive output. We would however like to ensure builds with trace enabled don't break.
So the trace build proposed is:
1. Ubuntu Xenial, gcc 5.4, Debug, with relevant settings for LOG_LEVEL_TRACE. Step "make check" and all subsequent steps would be omitted.

PR #1180, opened 2018-03-01
changes requested in #1169

crd477:


pervazea:
The Fedora 27 debug build it failing. libasan.... which I'd guess is associated with the address sanitizer.

pmenon:
@pervazea Looks like these sanitizer libraries, which normally come bundled with GCC, must be installed separately on Fedora/RHEL ([see here](https://bugzilla.redhat.com/show_bug.cgi?id=1303766))

@crd477 Can you modify `packages.sh` to also install `libasan`, `libubsan` and `libtasn` on Fedora/RHEL?

crd477:
@pmenon OK, I've done this for Fedora but apparently the CentOS packages don't follow the same packaging conventions.  Even after installing ```libasan``` and ```libtsan``` there, the compiler fails with:
```
/usr/include/llvm3.9/llvm/Support/Compiler.h:389:39: fatal error: sanitizer/asan_interface.h: No suc
h file or directory
 # include <sanitizer/asan_interface.h>
                                       ^
compilation terminated.
```
In order to get those headers it seems to me like you have to install the "Software collections from the CentOS SCLo SIG" repository which then installs a new version of LLVM, clang, and all of its dependencies under ```/opt```.  Instead of messing around with this, I did the easier thing in https://github.com/cmu-db/peloton/pull/1180#commits-pushed-d49b278, which is to leave out ```-DUSE_SANITIZER=Address``` from the CentOS debug build stage (which is still commented out in any case).  If you want me to incorporate the SCLo SIG repository stuff, let me know but I'd suggest just relying on ASan results from other platforms.

@pervazea I've also commented out the unit test collection method in the Jenkinsfile that was probably wrong in the Jenkins master node case and **definitely** wrong on remote Jenkins worker VMs.  If you merge it in this state, we can see if it supports my theory about it being the cause of the executor stalls.

pervazea:
That all the jobs for this PR are progressing on Jenkins is very encouraging.

PR #1189, opened 2018-03-05
Fixing BwTree non-unique key insert problem

wangziqi2013:


pervazea:
Changing the LOG_TRACE to LOG_DEBUG in the test output is causing the Travis debug test invocation to fail. That change in log output should not be in the PR. I assume we can find a machine on which to run with the extra logging, if this is needed to track down the issue.

pervazea:
Reminder: revert the LOG_DEBUG back to LOG_TRACE please.

mengranwo:
@wangziqi2013 
Hi Ziqi, can you take a look at the lin's and pervazea's comment and eliminate unnecessary log info?

apavlo:
@wangziqi2013 We need to talk.

wangziqi2013:
Please discard this PR. I opened a new one that should fix the problem.

apavlo:
Replaced with #1276

PR #1190, opened 2018-03-05
Mapping table fix to avoid preallocating memory

wangziqi2013:


mengranwo:
@wangziqi2013 
Hi Ziqi, the failing test cases with gcc 4.8 on Travis is
	**145 - index_intskey_test (Failed)
	175 - index_performance_test (Failed)**
Please take a look at the log information and change the tag into "ready to review" once you're done!

Thanks.

tcm-marcel:
@wangziqi2013 Are you still working on this or do you need help?

wangziqi2013:
@malin1993ml It uses mmap()'ed memory chunk without physical memory backup to replace the fixed sized array as the mapping table.

The advantage of using mmap() is that the OS does not allocate physical pages until the page is written into for the first time. Memory consumption can therefore be reduced.

wangziqi2013:
@tcm-marcel Fixed

src/include/index/bwtree.h
```diff
@@ -2864,6 +2866,17 @@ class BwTree : public BwTreeBase {
    * the mapping table rather than CAS with nullptr
    */
   void InitMappingTable() {
+    mapping_table = (std::atomic<const BaseNode *> *) \
+                    mmap(NULL, 1024 * 1024 * 1024, 
+                         PROT_READ | PROT_WRITE, 
+                         MAP_ANONYMOUS | MAP_PRIVATE,
+                         -1, 0);
+    if(mapping_table == (void *)-1) {
```
pervazea: This looks incorrect. If the mmap fails, we log a trace message. Most of the time we don't run with tracing, so the mmap failure is silently ignored. Sometime later, things will fail, but the original failure will not be clear.

Not sure on the correct fix, but:
- LOG_ERROR here, not trace.
- First choice would be to gracefully reject the request to create a BWTree (via an exception or failure of the constructor), OR kill the process via an exception or explicitly.

src/include/index/bwtree.h
```diff
@@ -7599,6 +7614,14 @@ class BwTree : public BwTreeBase {
                 epoch_leave.load());
 #endif
 
+      // NOTE: Only unmap memory here because we need to access the mapping
+      // table in the above routine. If it was unmapped in ~BwTree() then this
+      // function will invoke illegal memory access
+      int munmap_ret = munmap(tree_p->mapping_table, 1024 * 1024 * 1024);
+      LOG_DEBUG("munmap() returns with %d", munmap_ret);
```
pervazea: While not fatal, munmap failure is an error, not a debug event.

wangziqi2013:
@pervazea Can you please take a look at the Jenkins build error? It seems from the log that git checkout failed.

malin1993ml:
@pervazea It looks like this PR is ready to be merged?

Issue #1154, opened 2018-02-16
Jenkins tests with the address sanitizer are failing 


poojanilangekar:
Even though Jenkins says all test have passed, we have a total of 59 tests failing due to memory leaks. This is something @pervazea noticed on my PR. But I just checked that all the other PR and the master repo is also leaking the same memory. The [master](http://jenkins.db.cs.cmu.edu:8080/blue/organizations/jenkins/peloton/detail/master/5/pipeline/18) also leaks memory. Strangely, I wasn't able to reproduce it on my vagrant machine. Pervaze was unable to reproduce this with the same configuration. 

This happens on `Build / Ubuntu Xenial/gcc-5.4.0/llvm-3.7.1 (Debug/Test)` for the `cd build && make check -j4 || true` test.

pervazea:
I'll talk to Chad tomorrow. We should be paying attention to the return value from the tests. Jenkins seems to have no mechanism to run a set of things, then report whether one of N items failed. You either stop as soon as one invocation fails, or ignore the error code and continue running. We should be able to write our own wrapper to address this, in due course.


At the moment, we've been trying to resolve the issue that Jenkins stalls waiting for "executors", even though resources are idle. Until this is fixed, we can't scale things up properly.


________________________________

pervazea:
>Also, I don't think the build command should be make check -j4 || true since that hides the errors.

Completely agree.


________________________________

crd477:
Disregard what I said previously about ```make -k```; it was nonsensical.

The purpose of the ```||true``` is so that later steps, e.g., ```psql_test.sh``` and ```jdbc_validator.py``` can still run even if one of the unit tests fails.  I talked with @pervazea a bit about it this morning and I'm no longer even convinced that the current test invocation and collection is doing the right thing because the test results bar graph reports no failed tests even though the console logs clearly do.  It's on my short list to look into.

pervazea:
On the status of PR1134

We see in the Jenkins builds:
[](url)http://jenkins.db.cs.cmu.edu:8080/blue/organizations/jenkins/peloton/activity?branch=PR-1134
Failures:
1 - Xenial, Debug - Peloton failed with assertion failure
4 - same as 1.
6 - Trusty, Release - failed psql_test.sh, with data being inconsistent with reference output
12 - failed in communicating with Github. (Ignore this, nothing to do with Peloton)

I've tried to reproduce the failure(s) on my laptop, with no success, although I've not run the test a large number of times yet.

[...]

PR #1200, opened 2018-03-12
Modify StatsStorage::GetTableStats() to take a txn as a parameter instead creating a new one

chenboy:
As @poojanilangekar and @yingjunwu mentioned in previous emails, when visiting `ColumnStatsCatalog`, the stats storage use a `concurrency::TransactionContext*`  passed in as a parameter instead of creating a new one. This is a quick fix for that.

### DIFF
--- a/src/binder/binder_context.cpp
+++ b/src/binder/binder_context.cpp
@@ -37,10 +37,8 @@ void BinderContext::AddRegularTable(const std::string db_name,
                                     const std::string table_alias,
                                     concurrency::TransactionContext *txn) {
   // using catalog object to retrieve meta-data
-  // PL_ASSERT(txn->catalog_cache.GetDatabaseObject(db_name) != nullptr);
   auto table_object =
     catalog::Catalog::GetInstance()->GetTableObject(db_name, table_name, txn);
-    // txn->catalog_cache.GetDatabaseObject(db_name)->GetTableObject(table_name, true);
 
   if (regular_table_alias_map_.find(table_alias) !=
           regular_table_alias_map_.end() ||
--- a/src/include/optimizer/cost_calculator.h
+++ b/src/include/optimizer/cost_calculator.h
@@ -21,7 +21,8 @@ class Memo;
 // Derive cost for a physical group expressionh
 class CostCalculator : public OperatorVisitor {
  public:
-  double CalculateCost(GroupExpression *gexpr, Memo *memo);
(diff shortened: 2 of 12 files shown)


pervazea:
This PR was a result of investigating excess transaction in #1154

mengranwo:
@poojanilangekar @yingjunwu 
Hi guys,
Can you take a look at this PR, it seems like a quick fix(100 lines modified in total)
If nothing is wrong, please change the tag into **"accept"**

Thanks!

yingjunwu:
looks good to me. but I was wondering why the previous version creates so many transactions? The major update in this PR is to pass in transaction manager pointer. But I am not sure why it fixes the transaction creation bug.

yingjunwu:
build failed. please fix it.

chenboy:
Jenkins seems to be failing for all PR now. I think it’s caused by another issue in optimizer_rule_test. This PR builds on Travis so I think it should be good.

yingjunwu:
okay. I wonder whether you could answer the following question: `looks good to me. but I was wondering why the previous version creates so many transactions? The major update in this PR is to pass in transaction manager pointer. But I am not sure why it fixes the transaction creation bug.`

chenboy:
As @poojanilangekar  mentioned in an earlier email, the previous version would create (and commit) a txn in `StatsStorage::GetTableStats()` because it needs to access `ColumnStatsCatalog`. This function may be called multiple times in the optimizer so we may create multiple transactions. Now we pass in the current transaction and do not create a txn each time the function is called, so it could hopefully solve the issue. I'm not sure if this is the answer you're looking for.

yingjunwu:
okay the PR is ready to go.

mengranwo:
@pervazea
Hi Pervaze, can you help merge in this PR? both @yingjunwu and @poojanilangekar has confirmed that it is ready to go!

Thanks!

apavlo:
@mengranwo I will give you permissions.

ChTimTsubasa:
@chenboy Do you also want to change the begin txn here?
https://github.com/cmu-db/peloton/blob/master/src/optimizer/stats/stats_storage.cpp#L47

chenboy:
This function should only be called when the system initializes, so I think its fine here. But right now we do not initialize the singleton `StatsStorage` so it may cause the first query that requires stats to initialize `StatsStorage`, and thus create another txn. @poojanilangekar @yingjunwu are you currently blocked by this? If not, the cost model team will work on this in the near future, we can merge this PR first.

PR #1202, opened 2018-03-16
Improvements to 'packages.sh'

saatviks:
These are tensorflow specific improvements to the `packages.sh` script.
Added the following:
1. Fix for Mac `tar -C` error occuring on recent OSX versions.
2. Fix for `pip install` within a proxy environment.
3. Cleanup of bloated downloads post installation.
4. `wget` brew installation.

### DIFF
--- a/script/installation/packages.sh
+++ b/script/installation/packages.sh
@@ -55,14 +55,13 @@ function install_protobuf3.4.0() {
     echo "Only Ubuntu and Fedora is supported currently!"
     return 0
  fi
- CWD=`pwd`
- cd $TMPDIR
  wget -O protobuf-cpp-3.4.0.tar.gz https://github.com/google/protobuf/releases/download/v3.4.0/protobuf-cpp-3.4.0.tar.gz
  tar -xzf protobuf-cpp-3.4.0.tar.gz
  cd protobuf-3.4.0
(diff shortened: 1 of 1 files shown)


pmenon:
@saatviks With the changes you've made, is it safe to run the script multiple times? I've also noticed some hard-coded links - how stable are they? Are there alternates that can be used if they're down?

saatviks:
@pmenon Yes, it should be safe to run the script multiple times. There would be some redundant installation, but it would be safe.
As for the **hardcoded links**, these are mainly for a specific version of Tensorflow python binaries and C header files.
The C header files will need to be updated only when Google stops supporting a particular version of Tensorflow, which should be relatively rare. There's no way to have a package installer install it directly so it will have to be hardcoded.
The python binaries are relatively safe for the supported OSes(Mac, Ubuntu 14/16/17). That said, there is a way to make them non-hardcoded. The problem is that Jenkins builds seem to be hitting some unrelated SSL certificate permission errors after installing it directly via pip like: `pip install tensorflow`. This leads to a failing SSL test during `make check`.
I'll write up a detailed issue about all of this.

mbutrovich:
I think you need to add “brew install wget” otherwise this script will fail on a clean macOS installation.

PR #1206, opened 2018-03-16
Query Clustering

sivaprasadsudhir:
This PR integrates a header only third-party library ANNOY to the code base and implements the clustering algorithm for workload forecasting.


src/include/brain/cluster.h
```diff
@@ -0,0 +1,62 @@
...
+
+  int GetIndex() { return index_; }
+
+  void SetIndex(int index) { index_ = index; }
+
+  int GetSize() { return templates_.size(); }
+
+  std::vector<double> GetCentroid() { return centroid_; }
+
+  const std::set<std::string> &GetTemplates() { return templates_; }
```
malin1993ml: I know what those functions are, but we still need at least @brief comments on them (Java style).

src/include/brain/kd_tree.h
```diff
@@ -0,0 +1,88 @@
...
+    }
+    size_ = clusters_.size();
+    Build();
+  }
+
+ private:
+  int size_;
+  int num_features_;
+  AnnoyIndex<int, double, Angular, Kiss32Random> index_;
+  vector<Cluster *> clusters_;
```
malin1993ml: Again, comments on member variables and functions.

src/include/brain/kd_tree.h
```diff
@@ -0,0 +1,88 @@
...
+                             &distances);
+    cluster = clusters_[closest[0]];
+    similarity = distances[0];
+  }
+
+  void Build() {
+    for (int i = 0; i < size_; i++) {
+      index_.add_item(i, clusters_[i]->GetCentroid().data());
+    }
+    // n_trees = 2 * num_features, the more the faster, but requires more memory
```
malin1993ml: What is n_trees?

src/include/brain/kd_tree.h
```diff
@@ -0,0 +1,88 @@
...
+
+class KDTree {
+ public:
+  KDTree(int num_features)
+      : num_features_(num_features), index_(num_features) {}
+
+  void Insert(Cluster *cluster) {
+    index_.unbuild();
+    index_.add_item(size_, cluster->GetCentroid().data());
+    index_.build(2 * num_features_);
```
malin1993ml: So even with insert we still need to completely rebuild the kd-tree? I thought it's only that we cannot update or delete... This still may not be the bottleneck, but I just want to confirm.

src/include/brain/query_clusterer.h
```diff
@@ -0,0 +1,65 @@
...
+
+  void UpdateCluster();
+
+  ~QueryClusterer();
+
+ private:
+  // Map from the fingerprint of the template query to its feature vector
+  std::map<std::string, std::vector<double>> features_;
+  std::map<std::string, Cluster *> template_cluster_;
+  std::set<Cluster *> clusters_;
```
malin1993ml: Are the clusters owned by QueryClusterer? If so, you can use unique_ptr here to manage the objects. In KDTree as long as you don't dereference the pointers you can use get() to get the raw pointer from the unique_ptr as the look-up key.

src/include/brain/query_clusterer.h
```diff
@@ -0,0 +1,65 @@
...
+
+  int num_features_;
+  double threshold_;
+
+  // Similarity search structure for finding the nearest centroid;
+  KDTree kd_tree_;
```
malin1993ml: Still, comments for the variables and functions.

src/include/brain/kd_tree.h
```diff
@@ -0,0 +1,88 @@
...
+#include "brain/cluster.h"
+
+namespace peloton {
+namespace brain {
+
+//===--------------------------------------------------------------------===//
+// KDTree
+//===--------------------------------------------------------------------===//
+
+class KDTree {
```
malin1993ml: We should still create a separate .cpp file for this class. The length of the functions in this file exceeds what should be included in the header file. It's not only about reducing compilation dependency, but also about reducing instruction cache misses.

src/brain/query_clusterer.cpp
```diff
@@ -0,0 +1,122 @@
...
+    clusters_.insert(cluster);
+  }
+  template_cluster_[fingerprint] = cluster;
+}
+
+void QueryClusterer::UpdateExistingTemplates() {
+  // for each template check the similarity with the cluster
+  // if the distance is less than the threshold, then remove it and insert into
+  // the next nearest cluster. Don't update the centroids
+
```
malin1993ml: Remove the extra blank line...

src/brain/query_clusterer.cpp
```diff
@@ -0,0 +1,122 @@
...
+   * For new templates - insert into templates_ and
+   * call UpdateTemplate(fingerprint, true)
+   */
+}
+
+void QueryClusterer::UpdateTemplate(std::string fingerprint, bool is_new) {
+  auto feature = features_[fingerprint];
+  double similarity = 0.0;
+  Cluster *cluster = nullptr;
+  // TODO[Siva]: Ensure that this doesn't fail when the KDTree is empty
```
malin1993ml: We should fix this.

src/brain/query_clusterer.cpp
```diff
@@ -0,0 +1,122 @@
...
+  kd_tree_.GetNN(feature, cluster, similarity);
+
+  if (cluster == nullptr) {
+    cluster = new Cluster(num_features_);
+    cluster->AddTemplateAndUpdateCentroid(fingerprint, feature);
+    kd_tree_.Insert(cluster);
```
malin1993ml: If this is going to rebuild the tree as well, please add a TODO here. We may need to come back to optimize it. For now I think we can move on.

src/brain/query_clusterer.cpp
```diff
@@ -0,0 +1,122 @@
...
+    auto fingerprint = feature.first;
+    auto *cluster = template_cluster_[fingerprint];
+    auto similarity = cluster->CosineSimilarity(feature.second);
+
+    if (similarity < threshold_) {
+      cluster->RemoveTemplate(fingerprint);
+
+      if (cluster->GetSize() == 0) {
+        clusters_.erase(cluster);
+        // TODO[Siva]: This line can be avoided
```
malin1993ml: Are we going to fix this?

src/include/brain/kd_tree.h
```diff
@@ -0,0 +1,88 @@
...
+  void Insert(Cluster *cluster) {
+    index_.unbuild();
+    index_.add_item(size_, cluster->GetCentroid().data());
+    index_.build(2 * num_features_);
+    cluster->SetIndex(size_);
+    clusters_[size_] = cluster;
+    size_++;
+  }
+
+  void Update(UNUSED_ATTRIBUTE Cluster *cluster) {
```
malin1993ml: I understand that the parameter is unused because the tree is static, but we should add comments and TODO for this.

malin1993ml:
@sivaprasadsudhir Oh you also need to rebase your PR against the master.

test/brain/query_clusterer_test.cpp
```diff
@@ -0,0 +1,176 @@
...
+      clusters, {{fingerprints[0], fingerprints[1], fingerprints[3]},
+                 {fingerprints[2], fingerprints[4], fingerprints[5]}}));
+
+  query_clusterer.MergeClusters();
+
+  clusters = query_clusterer.GetClusters();
+  EXPECT_EQ(clusters.size(), 1);
+  EXPECT_TRUE(ClusterCorrectness(
+      clusters, {{fingerprints[0], fingerprints[1], fingerprints[2],
+                  fingerprints[3], fingerprints[4], fingerprints[5]}}));
```
malin1993ml: The tests look pretty good!

src/include/brain/query_clusterer.h
```diff
@@ -0,0 +1,103 @@
...
+   */
+  ~QueryClusterer();
+
+ private:
+  // Map from the fingerprint of the template query to its feature vector
+  std::map<std::string, std::vector<double>> features_;
+  // Map from the fingerprint of the template query to its cluster
+  std::map<std::string, Cluster *> template_cluster_;
+  // Set of all clusters
+  std::set<Cluster *> clusters_;
```
malin1993ml: Are the clusters owned by QueryClusterer? If so, you can use unique_ptr here to manage the objects. In KDTree as long as you don't dereference the pointers you can use get() to get the raw pointer from the unique_ptr as the look-up key.

sivaprasadsudhir: Oops! I missed answering this. In the current implementation, it's not. We have to dereference it to get the centroid of each cluster during the build phase.

malin1993ml: Okay that's fine.

PR #1207, opened 2018-03-17
Tensorflow Integration: Step 2

saatviks:
As part of the second step of tensorflow integration within Peloton the following are introduced:
1. The `TFSessionEntity` class which supports importing graphs along with feeding and evaluating graph ops.
2. Additional input/output classes which allow for preparing native/container types to be fed into a tensorflow graph op.
3. Minor test to check whether graph import works correctly.


src/include/brain/tf_session_entity/tf_session_entity.h
```diff
@@ -0,0 +1,90 @@
...
+  /**
+   * Evaluation/Session.Run() options.
+   * Provides a set of options to evaluate an Op either with
+   * 1. No input/outputs eg. Global variable initialization
+   * 2. Inputs, but no outputs eg. Backpropagation
+   * 3. Inputs and Outputs eg. Loss calculation, Predictions.
+   */
+  void Eval(const std::string &opName);
+  void Eval(std::vector<TfSessionEntityInput<InputType>> helper_inputs,
+            const std::string &op_name);
```
malin1993ml: Are we ever going to change the value of `helper_inputs`? If not, we can use const reference here to avoid copying.

src/include/brain/tf_session_entity/tf_session_entity.h
```diff
@@ -0,0 +1,90 @@
...
+  OutputType *Eval(
+      std::vector<TfSessionEntityInput<InputType>> helper_inputs,
+      std::vector<TfSessionEntityOutput<OutputType>> helper_outputs);
```
malin1993ml: Const reference. If we replace all these references, I think we can also remove the includes for the input and output types, and move them to the .cpp file.

src/include/brain/tf_session_entity/tf_session_entity_input.h
```diff
@@ -0,0 +1,54 @@
...
+  // 2d vector
+  TfSessionEntityInput(std::vector<std::vector<InputType>> input,
+                       const std::string &op);
+  // raw flattened input
+  TfSessionEntityInput(InputType *input, std::vector<int64_t> dims,
+                       const std::string &op);
+
+ private:
+  // Flattens 2d inputs
+  InputType *Flatten(std::vector<std::vector<InputType>> elems);
```
malin1993ml: Maybe replace the vector copying with const references for all the functions above?

src/include/brain/tf_session_entity/tf_session_entity_output.h
```diff
@@ -0,0 +1,53 @@
...
+   * Output may be a scalar const in which case only an opname would be needed
+   * Alternatively it might be a multidimensional output in which case the
+   * constructor
+   * with `dims` can be used.
+   */
+  // const output
+  explicit TfSessionEntityOutput(const std::string &op);
+  // n-d output
+  explicit TfSessionEntityOutput(std::vector<int64_t> dims,
+                                 const std::string &op);
```
malin1993ml: Vector const reference.

src/include/brain/tf_session_entity/tf_session_entity.h
saatviks: I've fixed everything else. I'm not sure what you mean here - Move the includes from line 15 and 16 to the .cpp?

malin1993ml: Yes. I think that's doable if you replace the vector arguments here with their const references.

saatviks: Then I'll have to import `#include <vector>` and similarly for string. 

Also, I thought the convention was to keep includes in the .h files only.

Issue #1209, opened 2018-03-20
packages.sh fails on Jenkins Xenial, gcc 5.4 


pervazea:
Packages.sh is no longer working correctly on Jenkins Xenial builds. Likely cause is recent incorporation of #1202.

PR #1210, opened 2018-03-20
Fix for #1209, packages.sh failures on Ubuntu Xenial

pervazea:
1. Fix so we do apt update for all Ubuntu variants
2. Add -e option to script, to exit on any failure, rather than
   ignoring errors

Fixes courtesy of crd477

### DIFF
--- a/script/installation/packages.sh
+++ b/script/installation/packages.sh
@@ -1,4 +1,4 @@
-#!/bin/bash -x
+#!/bin/bash -ex
 
 ## =================================================================
 ## PELOTON PACKAGE INSTALLATION
@@ -104,7 +104,6 @@ if [ "$DISTRO" = "UBUNTU" ]; then
             echo -e "\n# Added by Peloton 'packages.sh' script on $(date)\ndeb $LLVM_PKG_URL $LLVM_PKG_TARGET" | sudo tee -a /etc/apt/sources.list > /dev/null
         fi
         sudo apt-key adv --keyserver keyserver.ubuntu.com --recv-keys 15CF4D18AF4F7421
-        sudo apt-get update -qq
         CMAKE_NAME="cmake3"
         FORCE_Y="--force-yes"
     else
@@ -112,6 +111,7 @@ if [ "$DISTRO" = "UBUNTU" ]; then
         FORCE_Y=""
     fi
 
+    sudo apt-get update -qq
     FORCE_Y=""
     PKG_CMAKE="cmake"
     PKG_LLVM="llvm-3.7"
--- a/script/installation/packages.sh
+++ b/script/installation/packages.sh
@@ -1,4 +1,4 @@
-#!/bin/bash -ex
+#!/bin/bash -x
 
 ## =================================================================
 ## PELOTON PACKAGE INSTALLATION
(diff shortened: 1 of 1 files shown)


script/installation/packages.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash -x
+#!/bin/bash -ex
```
tcm-marcel: I am not familiar with passing flags in the shebang. Does this add the intended behavior, even if both `-e` and `-x` are given?

script/installation/packages.sh
```diff
@@ -104,14 +104,14 @@ if [ "$DISTRO" = "UBUNTU" ]; then
...
         sudo apt-key adv --keyserver keyserver.ubuntu.com --recv-keys 15CF4D18AF4F7421
-        sudo apt-get update -qq
         CMAKE_NAME="cmake3"
         FORCE_Y="--force-yes"
     else
         CMAKE_NAME="cmake"
         FORCE_Y=""
     fi
 
+    sudo apt-get update -qq
```
tcm-marcel: The manpage of `apt` says

> you should never use -qq without a no-action modifier such as -d, --print-uris or -s as APT may decide to do something you did not expect. Configuration Item: quiet.

This was actually a problem in #1155, but I think we didn't fix/change this. If we have a better idea for this, we should change it everywhere.

tcm-marcel: We can try `-qy` instead of `-qq`. 
`-q` single q: a little bit more output
`-y` still answering all questions with yes

tcm-marcel:
I just saw that the travis build for Mac is failing: https://travis-ci.org/cmu-db/peloton/jobs/355887894
I think the script now returns an error that we ignored before.

script/installation/packages.sh
crd477: I don't think ```apt-get update``` will ever prompt for input, will it?

tcm-marcel: @crd477 I think you are right.

crd477: A single ```q``` should be fine then.  I only added the second one to squeeze even more useless information out of the output.
Which reminds me - in my opinion, it would be good to change the ```wget``` invocations added as part of the TF integration to ```wget -q``` for the same reason.

pervazea:
The updates:
- strict status checking on all platforms except OSX. For OSX revert to permissive. 
- remove redundant value setting
- tone down apt-get params to -q

For a future PR / Issue
- make OSX status checking strict. Brew install invocations need changing. Possibly other things too.
- change apt-get invocations not to force yes. I'm expecting turning off force-yes to require fixing some underlying install issue.

crd477:
Regarding:

> change apt-get invocations not to force yes. I'm expecting turning off force-yes to require fixing some underlying install issue

it is only the ```apt-get update``` that doesn't need ```-y```.  You should leave the ```-y``` for the ```apt-get install``` commands.

pervazea:
I have left -y on apt-get install commands.
I should have worded more clearly, that my objection is to use of --force-yes in conjunction with -y, as the man page discourages use of --force-yes. For now, it remains as is, but we should see if we can get rid of it.

Issue #1217, opened 2018-03-22
ssl_test fails on Jenkins


pervazea:
This was on a release build, Ubuntu Xenial 16.04 on Jenkins.

The same test passed on Ubuntu Trusty (14.04) on Jenkins. The difference seems to be that in this case, the logs include:
**could not open certificate file "/home/jenkins/.postgresql/postgresql.crt": Permission denied**

```
41: Test command: /tmp/workspace/peloton_PR-1201-OT4KEIEOGLSE5FTNR72ZYTXKVRBNBJMAPG7KAP65MNYFC724H2ZQ@2/build/test/ssl_test "--gtest_color=yes" "--gtest_output=xml:/tmp/workspace/peloton_PR-1201-OT4KEIEOGLSE5FTNR72ZYTXKVRBNBJMAPG7KAP65MNYFC724H2ZQ@2/build/test/ssl_test.xml"
41: Test timeout computed to be: 9.99988e+06
41: Running main() from gmock_main.cc
41: [0;32m[==========] [mRunning 1 test from 1 test case.
41: [0;32m[----------] [mGlobal test environment set-up.
41: [0;32m[----------] [m1 test from SSLTests
41: [0;32m[ RUN      ] [mSSLTests.BasicTest
41: 2018-03-22 13:14:55 [test/network/ssl_test.cpp:103:TestBody] INFO  - Server initialized
41: 2018-03-22 13:14:55 [src/network/peloton_server.cpp:132:SSLInit] INFO  - certificate file path /tmp/workspace/peloton_PR-1201-OT4KEIEOGLSE5FTNR72ZYTXKVRBNBJMAPG7KAP65MNYFC724H2ZQ@2/test/network/ssl/server_test.crt
41: 2018-03-22 13:14:55 [src/network/peloton_server.cpp:142:SSLInit] INFO  - private key file path /tmp/workspace/peloton_PR-1201-OT4KEIEOGLSE5FTNR72ZYTXKVRBNBJMAPG7KAP65MNYFC724H2ZQ@2/test/network/ssl/server_test.key
41: 2018-03-22 13:14:55 [src/network/peloton_server.cpp:265:SetupServer] INFO  - Listening on port 15721
41: 2018-03-22 13:14:55 [src/network/postgres_protocol_handler.cpp:1062:ProcessInitialPackets] INFO  - protocol version: 80877103
[...]

PR #1218, opened 2018-03-23
SSL handshake refactor

ChTimTsubasa:
This PR refactored the code of SSL handshaking for several goals:

1. Enhance modularity by decoupling logic of SSL request & response, which should be handled by protocol handler, and actual SSL handshake code, which is processed by OpenSSL's SSL_accept(). Now the Network reading and Packet parsing code should be quite established and extendable. Packet Sending would be fixed in later PRs.

2. Better CPU utilization could be achieved by making SSL_accept() wrapped in a separate state to allow network multiplexing. Previously it is wrapped in a while loop.


ChTimTsubasa:
There can be two solutions to #1217 .
1. Change the permission of default certificates and key.
2. Use custom certificate and key.

I tried 2 for a while by using Haoxiang's create_certificates.sh script or manually generate certificates and keys, but since I am not very familiar to correct step for certificate and key generation, the ssl_test throws an exception for either self-signed certificates or certificate verification failed.

I would suggest merging this PR first and then fix this problem later in another PR.

PR #1221, opened 2018-03-24
Clean up Brain framework code

tli2:
Some cleanup for the pre-spring break hack job


mengranwo:
@malin1993ml 
Hi Lin, can you please review this PR when you're free. It seems like a simple clean up and does not involve major change.

Thanks

src/include/brain/brain.h
```diff
@@ -15,52 +15,88 @@
...
+    auto *job = new BrainJob(&env_, args...);
+    jobs_[name] = job;
+    auto callback = [](int, short, void *arg) {
+      reinterpret_cast<BrainJob *>(arg)->Invoke();
     };
-    // TODO(tianyu) Deal with this memory
-    auto *pair = new std::pair<BrainJob *, BrainEnvironment *>(&job, &env_);
     job_handles_[name] =
-        scheduler_.RegisterPeriodicEvent(period, callback, pair);
+        scheduler_.RegisterPeriodicEvent(period, callback, job);
```
malin1993ml: As far as I understood, this "callback" here is just a wrapper to invoke the brain job, right? Then why is called a "callback"? The naming is just a bit confusing to me because usually, the callback contains the extra work you need to do after the invocation returns.

src/main/peloton/peloton.cpp
```diff
@@ -49,8 +49,18 @@ void RunPelotonBrain() {
...
-  brain.RegisterJob(&one_second, "test", job);
+  one_second.tv_usec = 0;
+
+  auto example_task = [](peloton::brain::BrainEnvironment *) {
+    // TODO(tianyu): Replace with real address
+    capnp::EzRpcClient client("localhost:15445");
+    PelotonService::Client peloton_service = client.getMain<PelotonService>();
+    auto request = peloton_service.createIndexRequest();
+    request.getRequest().setIndexKeys(42);
+    auto response = request.send().wait(client.getWaitScope());
```
malin1993ml: I think we talked about adding a shim layer to wrap around the RPC implementation, right? Right now the capnp proto stuff is exposed to other parts of the system.

src/include/brain/brain.h
tli2: This is libevent parlance. (I agree that really we should call them handlers, but since they call it callback...)

malin1993ml:
As I mentioned in the last comment, can you add the namespace closure in brain.h here:
https://github.com/cmu-db/peloton/pull/1221/files#diff-af2a401370217814d0cb15d58828c367R116

PR #1225, opened 2018-03-26
Add statistics counter to record memory usage

wangziqi2013:
Currently the counter is a shared member variable. Use GetMemAlloc() to obtain aggregated number of bytes allocated for Inner, Leaf and Delta nodes.

### DIFF
--- a/src/include/index/bwtree.h
+++ b/src/include/index/bwtree.h
@@ -2986,7 +2986,8 @@ class BwTree : public BwTreeBase {
    * If both are nullptr then we just traverse and do not do anything
    */
   const KeyValuePair *Traverse(Context *context_p, const ValueType *value_p,
-                               std::pair<int, bool> *index_pair_p) {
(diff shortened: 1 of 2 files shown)


wangziqi2013:
OK, I need to think more on this.

@mengranwo Do you need a statistics counter per-instance, or just a global counter for all BwTree? I ask because these two require quite different implementations

eribeiro:
Hi @wangziqi2013, 

Does it make sense to add a method to reset or decrement the counter?

ChTimTsubasa:
We can keep this PR as a reference for me to do memory stats without merging it.
@malin1993ml I have added some code to do memory metric stats in backend_stats_context.

malin1993ml:
@ChTimTsubasa Are you saying that you're going to implement the feature in this PR in your own branch, so we can just close this one? You can still see the code even if we close it.

ChTimTsubasa:
@malin1993ml Sure.

PR #1229, opened 2018-03-27
Suppress Postgres parser mem leaks

pervazea:
Add a leak suppression file and pass it to make check. On other than sanitizer builds, it will be a no-op.


pervazea:
Correct, if you run manually, the suppression does not apply. (TBD, document on wiki).

If you run one or more tests using ctest, e.g. ctest -I ...  suppression does apply.

PR #1242, opened 2018-03-29
Fix 4 table join bug in binding iterator

### LINKED ISSUE #1195: Joins with more than three join partners segfault
On 0db1a2a478ce2102451a47c0be14ecb5ecb45e00 (current master), this causes a segfault:

```
test=# CREATE DATABASE test;
CREATE DATABASE
test=# CREATE TABLE a (x INTEGER);
CREATE TABLE
test=# CREATE TABLE b (x INTEGER);
CREATE TABLE
test=# CREATE TABLE c (x INTEGER);
CREATE TABLE
test=# CREATE TABLE d (x INTEGER);
CREATE TABLE
```

```
test=# SELECT * FROM a JOIN b ON a.x = b.x JOIN d ON a.x = d.x;
 x | x | x
---+---+---
(0 rows)
```

```
test=# SELECT * FROM a JOIN b ON a.x = b.x JOIN C ON a.x = c.x JOIN d ON a.x = d.x;
server closed the connection unexpectedly
	This probably means the server terminated abnormally
	before or while processing the request.
The connection to the server was lost. Attempting reset: Failed.
```

In 5878dc7, this still works.

chenboy:
This PR fixes #1195 and also added some JSON style GetInfo() method for debugging purpose. The bug was just a silly bug in the legacy pattern matching framework. Basically it's just a one-line change  in `binding.cpp` line -149 -> +157. Previously when advancing the iterator, we remove an incorrect number of children pattern. 

As I take a deeper look at the pattern matching code, it seems to me the current implementation is not that space-efficient, it materializes all the matching patterns in all child groups and does a permutation to create matching patterns. I'm concerning the number of the matching pattern is going to exponentially grow as the number of table to join gets larger. We could merge this first and discuss this later.


chenboy:
Address some comments:

We need to get all group expressions from the target group object which is initialized by the memo. These iterators recursively create their child iterators so we need to pass down the memo.
https://github.com/cmu-db/peloton/blob/master/src/optimizer/binding.cpp#L31 

I'll add a test case for it and modify these assert soon.

malin1993ml:
I see. I missed that line.

And after double-think about those iterators, I actually think the current structure is fine. I won't create the issue for now but we should indeed discuss your comment on next Wednesday to make sure that we're on the same page.

Issue #1234, opened 2018-03-28
False positive in rule checking in optimizer


GustavoAngulo:
It seems that by having the following enum start at 0, the optimizer can result in a false-positive, where when checking if the INNER_JOIN_COMMUTE rule has been applied to a group, it will return true when it hasn't been applied.

https://github.com/cmu-db/peloton/blob/fd2cf84a637a70e8033dc9c6bf3fb7fa68881cef/src/include/common/internal_types.h#L1312-L1314

I believe this may be due to the mask that is used to check, but I'm not certain. A fix is to offset the enum at the by inserting a type such as `INVALID_RULE_TYPE = 0`, however, this messes the following type, as it seems to be a count of the number of rules:

https://github.com/cmu-db/peloton/blob/fd2cf84a637a70e8033dc9c6bf3fb7fa68881cef/src/include/common/internal_types.h#L1350

PR #1245, opened 2018-03-29
Fixed bug in optimizer rule bitmap

GustavoAngulo:
This PR introduces a fix for #1234. The `bitset.set()` function was being used incorrectly, and not correctly updating the bitmap, leading to false positives. I've also included a test case to check correctness, which without the fix, would highlight the false positive.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 language: c++
+python: 3.6
 matrix:
     fast_finish: true
 
@@ -141,7 +142,7 @@ install:
 
 before_script:
     # first, run source_validator
-    - python ./script/validators/source_validator.py
+    - python script/validators/source_validator.py
 
 # build peloton (override this value to execute tests)
 script:
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -14,7 +14,7 @@ pipeline {
                     }
                     steps {
                         sh 'sudo /bin/bash -c "source ./script/installation/packages.sh"'
-                        sh 'python ./script/validators/source_validator.py'
+                        sh 'python script/validators/source_validator.py'
                         sh 'mkdir build'
                         sh 'cd build && cmake -DCMAKE_BUILD_TYPE=Debug -DUSE_SANITIZER=Address -DCOVERALLS=False .. && make -j4'
                         sh 'cd build && make check -j4 || true'
@@ -30,7 +30,7 @@ pipeline {
                     agent { docker { image 'ubuntu:xenial' } }
                     steps {
                         sh 'sudo /bin/bash -c "source ./script/installation/packages.sh"'
-                        sh 'python ./script/validators/source_validator.py'
+                        sh 'python script/validators/source_validator.py'
                         sh 'mkdir build'
                         sh 'cd build && cmake -DCMAKE_BUILD_TYPE=Release -DCOVERALLS=False .. && make -j4'
                         sh 'cd build && make check -j4 || true'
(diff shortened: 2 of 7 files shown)


malin1993ml:
@GustavoAngulo @chenboy This PR is not passing the Travis test. Please fix the failed test cases before merging.

GustavoAngulo:
Bugs should be fixed by #1242, that PR should be merged first.

malin1993ml:
Okay I'm reviewing that PR.

src/optimizer/memo.cpp
```diff
@@ -43,8 +43,6 @@ GroupExpression *Memo::InsertExpression(std::shared_ptr<GroupExpression> gexpr,
   auto it = group_expressions_.find(gexpr.get());
 
   if (it != group_expressions_.end()) {
-    PL_ASSERT(target_group == UNDEFINED_GROUP ||
-           target_group == (*it)->GetGroupID());
```
GustavoAngulo: I debugged for a while and reached the conclusion that this assert was incorrect. From my understanding, it doesn't make sense that if we create a group expression, we search for it in the memo table, and we find an existing equal expression, that we should have the same `GroupID`. Consider two possible join trees (AxB)x(CxD) and Ax(Bx(CxD)). In these two cases, both of the cases will generate a group expression for (CxD) independently.

src/optimizer/memo.cpp
```diff
@@ -43,8 +43,6 @@ GroupExpression *Memo::InsertExpression(std::shared_ptr<GroupExpression> gexpr,
   auto it = group_expressions_.find(gexpr.get());
 
   if (it != group_expressions_.end()) {
-    PELOTON_ASSERT(target_group == UNDEFINED_GROUP ||
-           target_group == (*it)->GetGroupID());
```
GustavoAngulo: I debugged for a while and reached the conclusion that this assert was incorrect. From my understanding, it doesn't make sense that if we create a group expression, we search for it in the memo table, and we find an existing equal expression, that we should already know it's `GroupID` when its passed in as `target_group`. Consider two possible join trees (AxB)x(CxD) and Ax(Bx(CxD)). In these two cases, both of the cases will generate a group expression for (CxD) independently. The first will generate it from the group Bx(CxD), but the second will generate it from the group (CxD).

chenboy: Yes after thinking I think you're correct. A case would be from some transformation we get `Ax(BxC)` and insert it into the memo, then from another transformation we get `(AxB)xC` and decide it should go into another group but actually they should be in the same group. Notice that we do not generate all logical expressions at once now so this is possible. Later we'll get this assert failure after applying`Ax(BxC) -> (AxB)xC`. So to solve this we need to merge groups in the memo.

malin1993ml: I actually think the assertion is fine... I think the fundamental problem is what you have discussed in Slack: the logical and physical join for the same children should be evaluated to be equal. But maybe i’m missing something. Let’s discuss today.

src/optimizer/group_expression.cpp
```diff
@@ -87,15 +87,21 @@ hash_t GroupExpression::Hash() const {
...
+
+  for (size_t i = 0; i < left_groups.size(); ++i) {
+    eq = eq && (left_groups[i] == right_groups[i]);
   }
 
   return eq;
 }
 
 void GroupExpression::SetRuleExplored(Rule *rule) {
-  rule_mask_.set(rule->GetRuleIdx()) = true;
```
zhaoguoquan94: In #1285 , you mentioned that  
>We created and added a handful of "plan_selection_tests" that test if the optimizer does the appropriate join ordering (it currently does not).

Is this line the reason that the optimizer is not doing it? Or there are other rules that cause this problem?  
For simple join, say 2 tables, is it possible that the "logical to physical" rule is applied first and the rest of the possible plan are pruned off, like early stopping?

GustavoAngulo: I am not sure about possible early pruning, however, without this change, rules such as Join Commutativity were never being applied

Issue #1232, opened 2018-03-28
Tensorflow background thread problem


saatviks:
Tensorflow has a currently open [issue](https://github.com/tensorflow/tensorflow/issues/17739). Basically they don't expose any API to shut down background threads when closing a Tensorflow session. As a result Valgrind(and possibly the Address Sanitizer) is bound to throw errors related to this.

There's nothing much we can do until Tensorflow releases a relevant API for the same.

saatviks:
@pervazea I'm curious about the correct way to handle suppressions to ensure that I supress the correct set of leaks(and future genuine memory leaks dont go unnoticed). I think the current way of supressing`tf_session_entity.cpp` is incorrect since future genuine user introduced memory leaks might go unnoticed. 

It seems that a regex like `tensorflow::thread` would be more suitable? I'm not confident though that this would cover all possible cases.

PFA the Valgrind(from my machine) log and Address Sanitizer(from Travis) log. 
I can confirm that the leak should introduce 2 Valgrind errors in the No-op case due to the linked bug.

pervazea:
I don't know that I have a good answer, without studying the issue. Even after that, this is going to be imprecise.

In general, one wants to suppress only specific leaks, so that suppression should identify the leak location as precisely as possible. One could:
1. Go as far down the stack as possible (which is is what we've done with tf_session_entity.cpp). Going even further down is possible, but transitions to pre-compiled library rather than source code. In some of the cases we've fixed, this (the lowest level) has not been the real leak location.

1. Go up the stack, examining source code, until one finds the real place where something is allocated (but never freed). This requires reading and understanding the code.

1. If the leaks end up being from binaries, then it will be difficult. Will likely have to suppress broadly in the binary.

1. For a tensorflow::thread suppression, I'd be concerned that it is too general a suppression. 

So, looking just at the reported stack traces, is going to be insufficient. Someone familiar with the code (or should study the code enough to be familiar) should examine the stack locations. If we can identify the real location of the leak, then we can define a reliable suppression regex.

saatviks:
Thanks @pervazea and @malin1993ml - based on your suggestions have set the following as the new leak supression for this bug:
```
+leak:tensorflow::thread::EigenEnvironment::CreateThread
+leak:tensorflow::thread::ThreadPool::ThreadPool
+leak:__cxa_thread_atexit_impl
```

PR #1247, opened 2018-03-30
Tensorflow memory leak fix

saatviks:
This should fix most of the memory leaks from Tensorflow.
Valgrind still throws 2 errors when doing a leak check which corresponds to #1232.


saatviks:
Seems like I'll need to add the leak supression bit to deal with #1232.

malin1993ml:
Yes, I think we need to do so. You can use something like `tensorflow::(anonymous namespace)::PosixEnv::StartThread` to identify the stack trace. I remember that's how we do it in Valgrind. We also needed to add both the "real" function name and the mangled name in Valgrind (for debug and release build). Address Sanitizer may work differently.

malin1993ml:
Looks good!

Issue #1214, opened 2018-03-22
Hardcoded Tensorflow Install


saatviks:
The ideal way to be installing tensorflow python binaries are as: `pip3 install --upgrade tensorflow==1.4.0`, but this causes the problem as described in the discussion #1202:

The python binaries are relatively safe for the supported OSes(Mac, Ubuntu 14/16/17). That said, there is a way to make them non-hardcoded. The problem is that Jenkins builds seem to be hitting some unrelated SSL certificate permission errors after installing it directly via pip like: `pip install tensorflow`. This leads to a failing SSL test during make check.

@pervazea this is what we spoke about.

saatviks:
Resolved with #1251.

PR #1251, opened 2018-04-01
fix 'not a supported wheel on this platform' error when running package.sh on Ubuntu 17.10

Nov11:
what happened?
I encountered this error when I executed package.sh on my machine.
'tensorflow-1.5.0-cp35-cp35m-linux_x86_64.whl is not a supported wheel on this platform.'
my system info :
OS version:Ubuntu 17.10
python3 version: Python 3.6.3
what's changed ?
version number after 'cp' is changed corresponding to version of python3.  35 to 36.
does this fix the problem?
I tried on my machine. the error is gone. the script works as expected.


saatviks:
Thanks for the PR @Nov11.

I actually think that with the SSL test being fixed in a previous PR, the problem described in #1214 should be resolved. Instead of using a hardcoded tensorflow binary URL everywhere for python-tensorflow, it would be better to do something like `pip install tensorflow==$TF_VERSION`.

I think it would be more helpful to set this up instead.

Nov11:
you're right. I added a if clause so that the script can get the correct binary url on 17.04.

I tried to specify version number after package name. It seems that this is not working for the moment.
Here's error message:
+sudo -E pip3 install tensorflow==1.4.0
Downloading/unpacking tensorflow==1.4.0
  Could not find any downloads that satisfy the requirement tensorflow==1.4.0
guess i need to revert this modification

Nov11:
those urls have been removed. the script should work as expected for now

saatviks:
Looks good - You'll need to update your branch to resolve the jenkins failiure.

script/installation/packages.sh
```diff
@@ -120,19 +120,16 @@ if [ "$DISTRO" = "UBUNTU" ]; then
     if [ "$MAJOR_VER" == "14" ]; then
         PKG_CMAKE="cmake3"
         FORCE_Y="--force-yes"
-	TF_VERSION="1.4.0"
-        TFBinaryURL="https://storage.googleapis.com/tensorflow/linux/${TF_TYPE}/tensorflow-${TF_VERSION}-cp34-cp34m-linux_x86_64.whl"
+	    TF_VERSION="1.4.0"
```
saatviks: Very minor change - just fixing the indentation on all the `TF_VERSION` lines.

Nov11: Oh, I should have changed indentation setting in my editor earlier.

saatviks:
This LGTM.

saatviks:
@malin1993ml @pervazea  I think we should merge this in if its passing Jenkins too.

pervazea:
Sure. Will try to do this next.

 Is this holding you up?

saatviks:
No - was just asking since it would clean up the `packages.sh`.

Issue #973, opened 2017-12-20
Replacing/Improving LockFreeArray


rohit-cmu:
I am documenting this issue so that there is a record of it:
Currently we maintain a lock free array for the following : 
* LockFreeArray<std::shared_ptr<index::Index>>;
* LockFreeArray<std::shared_ptr<storage::TileGroup>>;
* LockFreeArray<std::shared_ptr<storage::Database>>;
* LockFreeArray<std::shared_ptr<storage::IndirectionArray>>;

All uses initialize from container/lock_free_array.h which initializes the lock free array of size 1 million. This results in a massive heap usage. We need to either replace this with some other concurrent structure or work on this implementation.

apavlo:
@pervazea Can you look to find a replacement data structure for this? You can look at libcds as a starting point:

https://github.com/khizmax/libcds

rohit-cmu:
Hi Andy, I started work on this just yesterday. Looked at some other options but the iterable list in libcds seems relevant to us.

apavlo:
@rohit-cmu Oh that's great! Keep everyone updated

zhaoguoquan94:
I and Vamshi discussed this issue. We don't think that libcds has a lock-free ARRAY data structure. We can use a KV map in libcds to replace it, but this will add a lot of lookup overhead.
If we'd like to reduce the memory usage asap, we could just **lower the default array size**. Currently, it is a fixed size array. It will crash after adding 1048576 values anyway. And only two files are using it. If memory is a bigger issue, a smaller number could be a quick fix. (1024 is enough for passing all the test case).

Replacing it without changing the current use cases is not very intuitive. In fact, the user of LockFreeArray uses it in different manners. In DataTable, people append value first and do a full array lookup.  In Manager, people update a value before adding the value first(looks like a javascript array). If we change it to a map-like data structure, the current use cases becomes: 
* DataTable do a lookup among all entries in the map. (Current version: DataTable do a look up in the first few continuous memory address.)
* Manager update the element given the index as a key, search it in the map. (Current version: Manager update the element given an array index).

Seems that a map could be slower. 
What we need is a lock-free dynamically resizable arrays like this [one](www.stroustrup.com/lock-free-vector.pdf). But we didn't find any existing implementation. Intel's tbb has a concurrent_vector implementation, but I'm not sure whether it's lock-free or not.
We could try libcds first and see how it performs, meanwhile any ideas/help are welcomed!

PR #1253, opened 2018-04-02
Lockfreearray

zhaoguoquan94:
#1168 #973 
Use tbb::concurrent_vector to implement the LockFreeArray.
Use tbb::concurrent_unorded_map to replace the old lockfree array.

### DIFF
--- a/src/include/common/container/lock_free_array.h
+++ b/src/include/common/container/lock_free_array.h
@@ -10,9 +10,9 @@
 //
 //===----------------------------------------------------------------------===//
 
-
 #pragma once
-
+#include "tbb/concurrent_vector.h"
+#include "tbb/tbb_allocator.h"
 #include <cstdlib>
 #include <cstring>
 #include <cstdio>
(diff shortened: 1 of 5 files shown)


zhaoguoquan94:
> Is it possible that the existing element gets deleted between line 38 and line 40, and find(oid) returns end()? https://github.com/cmu-db/peloton/pull/1168/files/dfde397fdd49ae74f62125c68ba3a520fd70d5a2#diff-e63c398b482ad11164d943fe24f98df0

We now store an empty place holder object to avoid this problem. Deleting in the map is setting the value to the empty object. The elements in the map only get inserted when key is not in the map, updated when key is in the map. Deletion just set the value to the placeholder object.

src/include/common/container/lock_free_array.h
```diff
@@ -22,7 +22,7 @@
 
 namespace peloton {
 
-#define LOCK_FREE_ARRAY_MAX_SIZE 1024 * 1024
+static const size_t kLockfreeArrayInitSize = 256;
 
 // LOCK_FREE_ARRAY_TEMPLATE_ARGUMENTS
 #define LOCK_FREE_ARRAY_TEMPLATE_ARGUMENTS template <typename ValueType>
```
tcm-marcel: https://github.com/cmu-db/peloton/pull/1168#discussion_r177144365 from old PR:
`#define LOCK_FREE_ARRAY_TEMPLATE_ARGUMENTS template <typename ValueType>`
(Not your changes) I personally would avoid definitions here, especially because the def-name is longer than the actual content. Every modern IDE is able refactor the types if something changes.

tcm-marcel:
@zhaoguoquan94 Thanks for making the changes. I think you missed the template instantiations I mentioned  [here](https://github.com/cmu-db/peloton/pull/1253#pullrequestreview-108964503).

PR #1255, opened 2018-04-03
Adapt Readme: Remove Support for Fedora

tcm-marcel:
Remove Fedora as a "known to work platform" because of missing support for Tensorflow.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,11 +32,10 @@ Check out the [installation instructions](https://github.com/cmu-db/peloton/wiki
 
 Peloton has been tested to work on the following platforms:
 
-* Ubuntu Linux 14.04+ (64-bit) [gcc4, gcc5]
-* Fedora Linux 24+ (64-bit) [gcc4, gcc5]
+* Ubuntu Linux 14.04+ (64-bit) [gcc4.8, gcc5]
 * Mac OS X 10.9+ (64-bit) [XCode v8 Only]
 
- Please note that it will not compile on 32-bit systems.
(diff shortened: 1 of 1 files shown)


tcm-marcel:
@mbutrovich I can remember you had/solved/whatever problems with new XCode Versions. Please let me know if the version in the README is still correct or if we can add another one.

mbutrovich:
I only updated the Wiki regarding my findings on the latest Xcode version:

https://github.com/cmu-db/peloton/wiki/Installation#macos-quick-setup

I think it's safe to update the readme to say Xcode <= 9.2.

tcm-marcel:
Asking @apavlo and @pmenon:
@mbutrovich just pointed out, that the [list of supported platforms](https://github.com/cmu-db/peloton/wiki/Installation#supported-platforms) can also be found in the wiki. What about simply replacing the part in Readme with a link to the wiki? This way we don't have the problem of duplicated information that may differ.

pmenon:
@tcm-marcel I'm okay with that.

zhaoguoquan94:
> Command Line Tools for Xcode 9.3 manifests [this](https://reviews.llvm.org/D33526) LLVM bug and fails to build Peloton.
@mbutrovich 
I think we don't actually use the captured value and it is probably not a compiler bug. Correct me if I'm wrong. I tried to remove the unused value in #1265 and it seems to work.

mbutrovich:
@zhaoguoquan94 You're probably right. At first glance it looked like a spurious warning to me which aligned with false positives for that warning in recent LLVM releases. I'll modify the wiki accordingly and remove the comment about 9.3.

PR #1257, opened 2018-04-04
Static Code Checks for Test Files (in progress)

schedutron:
I've added a check for test class name. Do I also have to create another `paths` variable via `glob` module, which contains all the test files' paths (which end with `*_test.cpp`)?

Also, I'll add other checks soon.


tcm-marcel:
I would check in `validate_file(file_path)` [1] if the file matches `/test/*/*.cpp` and then run the check in addition to the other checks. This way every test file that the validator is called with gets checked. The pre-commit hook calls the validator with the stages files, to this happens automatically. 

[1] https://github.com/schedutron/peloton/blob/dda18900b9c5efafbea1d4ada73b2903ecd4dc9a/script/validators/source_validator.py#L237

schedutron:
I think I should add a path regex for checking whether we have a test file instead of checking membership in the path set returned by `glob.glob()` for faster execution. Will do so soon.

schedutron:
@tcm-marcel How do I check the methods? They don't follow a prefix/suffix pattern.

tcm-marcel:
> How do I check the methods? They don't follow a prefix/suffix pattern.

You can use a regex to find these lines and extract the information. (see example: https://regexr.com/3nnro)

> Also, to resolve the issue you mentioned above, should I omit the files that end with a `_util` suffix?

Sounds good.

schedutron:
Alright, but do we have to check after we extract the identifiers?

tcm-marcel:
Ah sorry, I got you wrong. The [naming conventions](https://github.com/cmu-db/peloton/blob/master/test/README.md#naming-conventions) say, that every method (test case) should have the suffix `Test`.

schedutron:
Yeah, but a lot of test cases across different test files do not. Are we going to use the validator script to fix all these files?

script/validators/source_validator.py
```diff
@@ -202,6 +204,39 @@ def check_includes(file_path):
...
+            line_num += 1
+            if line.startswith('class '):
+                class_name = line.split(' ')[1]
+                if not class_name.endswith('Tests'):
+                    if class_status:
+                        LOG.info("Invalid class name in %s", file_path)
+                        class_status = False
+                    LOG.info("Line %s: %s", line_num, line.strip())
+
+            else:
```
schedutron: Should there be a separate loop for this? That will prevent interleaved output.

script/validators/source_validator.py
```diff
@@ -66,6 +66,8 @@
     "src/codegen/util/cc_hash_table.cpp"
 ]
 
+TEST_CASE_PATT = re.compile(r'TEST_F\(([a-zA-Z]+), ([a-zA-Z]+)\)')
```
schedutron: Is this the right place to define the regex?

tcm-marcel: I think so.

script/validators/source_validator.py
tcm-marcel: I am ok with this, we hope not to have thousands of those errors.

script/validators/source_validator.py
```diff
@@ -224,6 +259,11 @@ def validate_file(file_path):
         if not validator(file_path):
             file_status = False
 
+    relative_path = file_path.replace(PELOTON_DIR, '')
```
tcm-marcel: One more thing: Can you use [`os.path.relpath`](https://docs.python.org/3/library/os.path.html#os.path.relpath) here instead of string replace?

schedutron:
As a first iteration, I will simply add the `Test` suffix to the incorrect test class name and the `Tests` suffix to the incorrect test method names.

tcm-marcel:
@schedutron No, we don't provide automated fixing. We expect people to fix their violations manually ;)

test/binder/binder_test.cpp
```diff
@@ -36,7 +36,7 @@ using std::make_tuple;
 namespace peloton {
 namespace test {
 
-class BinderCorrectnessTest : public PelotonTest {
+class BinderCorrectnessTestTests : public PelotonTest {
```
schedutron: This class should've been names `BinderCorrectnessTests`!

tcm-marcel: @schedutron Can you apply this change and rebase the PR? How many changes of the test files are still missing?

It's time that we merge it in and finish that part.

schedutron: Sorry for getting this delayed. I'll apply the changes soon.

test/binder/binder_test.cpp
```diff
@@ -36,7 +36,7 @@ using std::make_tuple;
 namespace peloton {
 namespace test {
 
-class BinderCorrectnessTest : public PelotonTest {
+class BinderCorrectnessTests : public PelotonTest {
```
schedutron: @tcm-marcel At this stage, there are parent classes which still have the `Test` suffix instead of the `Tests` suffix. They are defined in `.h` files, and thus were not changed by the name-fixing script I ran. If I change them, I would also have to change their name in inheritance statements, e.g., here the statement will look like `class BinderCorrectnessTests : public PelotonTests {`. Should I do that?

schedutron:
@tcm-marcel Have a look now :)

test/binder/binder_test.cpp
apavlo: @schedutron Yes.

tcm-marcel:
Hey, I fear you screwed up the code when resolving the merge conflicts in 2f28acc. This is also the reason the tests fail. 

I think the easiest way to solve this is to
1. make a temporary copy of your branch `git checkout -b temp`
2. reset the old branch (the one with the PR) to upstream master `git branch -f ...` (don't forget to fetch and rebase)
3. cherry pick the commits you want to have (all except the resolve merge conflicts one and the last one*)
4. force push the new version of this branch, the PR will update accordingly

*Please remove the changes where you format the headers. It's a good idea, but it should be a separate PR.

If you need help, let me know.

schedutron:
@tcm-marcel I was afraid that could happen. Will fix this soon. Thanks for the tips!

schedutron:
@tcm-marcel The checks are passing now!

tcm-marcel:
Hey, sorry for looking at this only now. 
1) something is still wrong from rebasing, there are several changes in the test files, for example in ` test/codegen/oa_hash_table_test.cpp`
2) I think the two classes `PelotonTest` and `PelotonCodeGenTest` should not be changed. I just created a commit to revert this, but I now realized that you will have to touch all files again anyway. 

Let me try if I can fix this quickly.

schedutron:
What exactly are the changes in, say, `test/codegen/oa_hash_table_test.cpp`?

tcm-marcel:
I can't find it anymore... seems like I looked at a wrong version. I am sorry! Let me take another look.

tcm-marcel:
Ok sorry, I messed this up a little bit:
1) I was wrong, I must have looked at the wrong diff
2) I thought about this again and decided that we leave it that way

I will try to rebase your PR and make it ready to merge.

saatviks:
@schedutron can you fix the merge conflicts so this can be merged in?

tli2:
What is the status of this PR right now? I am putting In Progress tag until merge conflicts are fixed.

schedutron:
Lets finalize this!

PR #1260, opened 2018-04-04
Modify UpdateSqlTest to add HalloweenTest and MultiTileGroupTest

poojanilangekar:
This PR should not be merged before the Sequential Scan bug is fixed. It WILL break all builds.


poojanilangekar:
I think the Halloween Problem has been fixed. Closing this one to open a new PR.

PR #1263, opened 2018-04-04
Sequential Scan + Halloween Problem Fix 

poojanilangekar:
This PR fixes the Halloween Problem along with the sequential scan problem.

It builds on #1240 and checks for an entry in each WriteSet for each tuple (irrespective of whether or not it updates the Primary Key). 

I have tested this with varying tile_group_count. I have also tested the fix with and without codegen. It appears to have solved the Halloween problem for both the engines.

### DIFF
--- a/src/codegen/updater.cpp
+++ b/src/codegen/updater.cpp
@@ -37,6 +37,8 @@ void Updater::Init(storage::DataTable *table,
   // Target list is kept since it is required at a new version update
   target_list_ =
       new TargetList(target_vector, target_vector + target_vector_size);
+
+  statement_write_set_ = new ReadWriteSet();
 }
 
 char *Updater::GetDataPtr(uint32_t tile_group_id, uint32_t tuple_offset) {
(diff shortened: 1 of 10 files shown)


yingjunwu:
@poojanilangekar Your PR looks interesting. However, I wonder whether you could add the test case on top of my PR (or after merging my PR)? Or you could follow @pmenon's style to directly commit to my branch. Creating two PRs for a single problem usually confuses the developers, and that's what we don't want to do as discussed long time ago (1-2 years back). Sorry for the inconvenience.

yingjunwu:
Hi @pervazea! I found that you have merged this branch to master. However, I just did a quick code review and it has bugs. The way to add statement-level write set in update executor/updater is wong. We could talk in person to discuss what's going wrong. I think you may want to revert this merge. Thank you!

pervazea:
Reviewed this morning, prior to merging.

Will be collecting yingjunwu's comments and feedback.

yingjunwu:
@pervazea The solution provided in this PR does not handle snapshot isolation and in-place updates. I am not sure whether you would like to revert this merge or I fix the bug in my PR?

Issue #1269, opened 2018-04-06
Network layer soft errors


pervazea:
These don't seem to cause any visible failures, but imply that something is not 100% correct.

Occur when running the smallbank benchmark. There are more of these logged. Only 2 samples are extracted below.

```
work/connection_handle.cpp:538:CloseSocket] DEBUG - Attempt to close the connection 63
....
/connection_handle.cpp:569:CloseSocket] DEBUG - Already Closed the connection 58
```

tcm-marcel:
I see this line all the time, I think is is usual behavior. But I have no idea what exactly is happening.

Nov11:
the only piece of code print debug message in connection_handle.cpp
```c++
  while (true) {
    int status = close(sock_fd_);
    if (status < 0) {
      // failed close
      if (errno == EINTR) {
        // interrupted, try closing again
        continue;
      }
    }
    LOG_DEBUG("Already Closed the connection %d", sock_fd_);
    return Transition::NONE;
  }
```
I think retrying on close failure should be removed. 
1. The fd is closed on linux regardless of close function call returns an error or not. No need to worry about leaking fd.
[https://stackoverflow.com/questions/33114152/what-to-do-if-a-posix-close-call-fails](url)
[http://man7.org/linux/man-pages/man2/close.2.html](url)
2. As the man page ( Dealing with error returns from close() section) suggests, it is harmful to close again on failure. This may result in closing a reused fd. For example, It may close a file descriptor opened by the smallbank test and cause an hard-to-reproduce test failure. 

I'd like to fix this by remove while loop around close.

pervazea:
I agree on correctness of removing the while.

However, if the close fails, it should be logged, something like:

Close failed on connection %d, errno %d

(the current log message is wrong... you don't know it was "already closed", it could be EIO)

If a string description of errno is readily available, use it instead of the number, otherwise the numeric form is ok.


Maybe, if we find that the close failures are legitimate, we can suppress some of the logging later. If there are real failures occurring, this will give us more data to diagnose and fix them.


________________________________

PR #1271, opened 2018-04-07
Remove retry on close failure

Nov11:
a fix for issue #1269 
1.no retry on close failure
2.provide more detailed debug message on close failure


Nov11:
It seems that jenkins server runs out out disk space.
Here is one of the build error:
```
[ 55%] Building CXX object test/CMakeFiles/peloton-test-common.dir/statistics/testing_stats_util.cpp.o

/tmp/workspace/peloton_PR-1271-OTJTL66AGB2C26ZL6NIM4XVKJGQQTMZYVGTE2PEWOG5PXARBHSLQ@2/third_party/gmock/gmock-gtest-all.cc:11445:1: fatal error: error writing to /tmp/ccL6f2H4.s: No space left on device

 }  // namespace testing

 ^
```

what can I do with this ?

tcm-marcel:
The Jenkins disk problem is addressed in #1273.

Nov11:
I prefer the function approach as close is a function too.
1.Any hint on the file which I should put 'peloton_close' into? Or maybe I create a header file in include/common/ named utility.h and put the function in it.
2.what about using 'int peloton_close(int fd, int failure_log_level = LOG_LEVEL_DEBUG)'  as signature?

pervazea:
I would put the function in the common directory, and have a header file but put the real code in a cpp file. Named something suggesting file operations. along the lines of file_ops?

I would do 'int peloton_close(int fd); so that it is a direct replacement for close(..).

src/common/utility.cpp
```diff
@@ -0,0 +1,72 @@
...
+    (void)strerror_r(errno, buffer.data(), buffer.size() - 1);
+    error_message = buffer.data();
+#else
+    error_message = strerror_r(saved_errno, buffer.data(), buffer.size() - 1);
+#endif
+    (void) error_message;
+
+    const char *format_string = "Close failed on fd: %d, errno: %d [%s]";
+
+    switch (failure_log_level) {
```
tcm-marcel: What is this `switch` needed for? Is there actually the need for calling the function and having the output on different log levels?
If I understand the problem correctly, #1269 is about whether we expect the connection to  be closed already or not. In that case I would suggest to put exactly this semantic in the function, and only write the message, if something happened that we didn't expect. (e.g. a flag `bool expect_still_open=true`)

Nov11: 1.switch-case has been removed.
2.I think the intention is to call close on a fd once and only once. On return of 'close', the fd is guaranteed to be closed anyway by OS even if return value is -1. Leave a hint message on failure for debug purpose.

src/network/peloton_server.cpp
```diff
@@ -277,10 +278,9 @@ void PelotonServer::ServerLoop() {
...
   dispatcher_task_->EventLoop();
   LOG_INFO("Closing server");
-  int status;
-  do {
-    status = close(listen_fd_);
-  } while (status < 0 && errno == EINTR);
+
+  peloton_close(listen_fd_);
+
   LOG_DEBUG("Already Closed the connection %d", listen_fd_);
```
pervazea: This LOG_DEBUG should be removed.

src/network/peloton_server.cpp
```diff
@@ -277,10 +278,9 @@ void PelotonServer::ServerLoop() {
   }
   dispatcher_task_->EventLoop();
   LOG_INFO("Closing server");
-  int status;
-  do {
-    status = close(listen_fd_);
-  } while (status < 0 && errno == EINTR);
+
+  peloton_close(listen_fd_);
+
```
pervazea: IMO the "Closing server" and "Server Closed" messages are bit too much now.
I would suggest removing the former ("Closing server") also.

src/network/connection_handle.cpp
```diff
@@ -99,23 +100,23 @@ DEF_TRANSITION_GRAPH
...
-  DEFINE_STATE(WRITE) 
+  DEFINE_STATE(WRITE)
     ON(WAKEUP) SET_STATE_TO(WRITE) AND_INVOKE(ProcessWrite)
     ON(NEED_DATA) SET_STATE_TO(PROCESS) AND_INVOKE(Process)
     ON(PROCEED) SET_STATE_TO(PROCESS) AND_INVOKE(Process)
   END_DEF
 
-  DEFINE_STATE(GET_RESULT) 
+  DEFINE_STATE(GET_RESULT)
     ON(WAKEUP) SET_STATE_TO(GET_RESULT) AND_INVOKE(GetResult)
```
tli2: These look like unintentional changes, can you revert them?

pervazea: These are ok, they can stay. They remove extra spaces at the end of lines and in otherwise blank lines.

@Nov11 for future work though, I would recommend and request that you keep functional changes separate from cleanup or cosmetic changes. Stick to just the changes. This makes it easier to review the changes.

Code cleanup is desirable, but it should be a separate PR. There is a code formatter that we'll eventually run across the code base, to clean up and standardize the style.

Nov11: Got it.  I'll stick to this rule.

PR #1274, opened 2018-04-09
Peloton Brain Recommended Indexes

Blade-Lee:
This is the PR for Peloton Brain Recommended Indexes task. The primary function is `PlanUtil::GetAffectedColumns()`, and I used #1105 as reference. 

The purpose of this function: given a `SQLStatement` (a query), return all the affected columns in a vector.

I completed the code for `UPDATE`/`INSERT`/`DELETE`/`SELECT`:

* `INSERT`/`DELETE`: all columns are affected
* `UPDATE`: columns in `UpdateClause` are affected
* `SELECT`: all scanned columns are affected (in or not in predicates)

The returned vector is actually a vector of triplets: each element of the vector is a tuple containing three values, representing database oid, table oid and column oid, respectively.

For `SELECT` queries, the elements of the returned vector are "sorted" in such a way that columns in predicates (having higher priority) are put in the front of the vector, and output columns (having lower priority) are put in the end of the vector.

This is the initial version of the code, still need suggestions on improvement/test cases.

### DIFF
--- a/src/planner/plan_util.cpp
+++ b/src/planner/plan_util.cpp
@@ -100,12 +100,13 @@ const std::set<oid_t> PlanUtil::GetAffectedIndexes(
   return (index_oids);
 }
 
-const std::set<oid_t> PlanUtil::GetAffectedColumns(
+const std::set<col_triplet> PlanUtil::GetAffectedColumns(
     catalog::CatalogCache &catalog_cache,
     std::unique_ptr<parser::SQLStatementList> sql_stmt_list,
     const std::string &db_name) {
-  std::set<oid_t> column_oids;
+  std::set<col_triplet> column_oids;
   std::string table_name;
+  oid_t database_id, table_id;
   auto sql_stmt = sql_stmt_list->GetStatement(0);
   switch (sql_stmt->GetType()) {
     // For INSERT, DELETE, all columns are affected
@@ -119,11 +120,15 @@ const std::set<oid_t> PlanUtil::GetAffectedColumns(
         auto &delete_stmt = static_cast<parser::DeleteStatement &>(*sql_stmt);
         table_name = delete_stmt.GetTableName();
       }
+      database_id = catalog_cache.GetDatabaseObject(db_name)->GetDatabaseOid();
+      table_id = catalog_cache.GetDatabaseObject(db_name)
+                     ->GetTableObject(table_name)
+                     ->GetTableOid();
       auto column_map = catalog_cache.GetDatabaseObject(db_name)
                             ->GetTableObject(table_name)
                             ->GetColumnObjects();
       for (auto &column : column_map) {
-        column_oids.insert(column.first);
+        column_oids.emplace(database_id, table_id, column.first);
       }
     } break;
     case StatementType::UPDATE: {
@@ -131,6 +136,8 @@ const std::set<oid_t> PlanUtil::GetAffectedColumns(
       table_name = update_stmt.table->GetTableName();
       auto db_object = catalog_cache.GetDatabaseObject(db_name);
       auto table_object = db_object->GetTableObject(table_name);
+      database_id = db_object->GetDatabaseOid();
+      table_id = table_object->GetTableOid();
 
       auto &update_clauses = update_stmt.updates;
       std::set<oid_t> update_oids;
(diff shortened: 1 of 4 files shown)


src/planner/plan_util.cpp
```diff
@@ -96,5 +100,148 @@ const std::set<oid_t> PlanUtil::GetAffectedIndexes(
...
+      std::set<oid_t> update_oids;
+      for (const auto &update_clause : update_clauses) {
+        LOG_TRACE("Affected column name for table(%s) in UPDATE query: %s",
+                  table_name.c_str(), update_clause->column.c_str());
+        column_oids.emplace_back(
+            database_id, table_id,
+            table_object->GetColumnObject(update_clause->column)
+                ->GetColumnId());
+      }
+    } break;
```
malin1993ml: We're not really looking for "affected" columns here. Probably `GetIndexableColumns` is a better name. For our purpose here, we don't need to include columns to change in INSERT/UPDATE/DELETE statements (if we ignore index update cost for now). Instead, we should include the columns in the where clause of UPDATE and DELETE statements, and return nothing for INSERT statements.

src/planner/plan_util.cpp
```diff
@@ -96,5 +100,148 @@ const std::set<oid_t> PlanUtil::GetAffectedIndexes(
...
+        for (auto &triplet : low_col) {
+          column_oids.push_back(std::move(triplet));
+        }
+
+      } catch (Exception &e) {
+        LOG_ERROR("Error in BuildPelotonPlanTree: %s", e.what());
+      }
+
+      // TODO: should transaction commit or not?
+      txn_manager.CommitTransaction(txn);
```
malin1993ml: It should abort.

test/planner/plan_util_test.cpp
```diff
@@ -163,7 +164,139 @@ TEST_F(PlanUtilTests, GetAffectedIndexesTest) {
...
+  sql_stmt_list = peloton_parser.BuildParseTree(query_string);
+  affected_cols_vector = planner::PlanUtil::GetAffectedColumns(
+      txn->catalog_cache, std::move(sql_stmt_list), TEST_DB_COLUMNS);
+  affected_cols = std::set<planner::col_triplet>(affected_cols_vector.begin(),
+                                                 affected_cols_vector.end());
+
+  EXPECT_EQ(1, static_cast<int>(affected_cols.size()));
+  expected_oids.clear();
+  expected_oids.emplace(database_id, table_id, fname_col_oid);
+  EXPECT_EQ(expected_oids, affected_cols);
```
malin1993ml: Add some predicates to the UPDATE statements. We should return the columns in those predicates.

test/planner/plan_util_test.cpp
```diff
@@ -163,7 +164,139 @@ TEST_F(PlanUtilTests, GetAffectedIndexesTest) {
...
+  affected_cols = std::set<planner::col_triplet>(affected_cols_vector.begin(),
+                                                 affected_cols_vector.end());
+
+  // all columns are affected
+  EXPECT_EQ(3, static_cast<int>(affected_cols.size()));
+  expected_oids.clear();
+  expected_oids.emplace(database_id, table_id, lname_col_oid);
+  expected_oids.emplace(database_id, table_id, id_col_oid);
+  expected_oids.emplace(database_id, table_id, fname_col_oid);
+  EXPECT_EQ(expected_oids, affected_cols);
```
malin1993ml: Again, we should add some predicates to the DELETE statement.

test/planner/plan_util_test.cpp
```diff
@@ -163,7 +164,139 @@ TEST_F(PlanUtilTests, GetAffectedIndexesTest) {
...
```
malin1993ml: Insert should returns nothing.

test/planner/plan_util_test.cpp
```diff
@@ -163,7 +164,139 @@ TEST_F(PlanUtilTests, GetAffectedIndexesTest) {
...
+  txn_manager.CommitTransaction(txn);
```
malin1993ml: Maybe add another select with a more complex predicate.

test/planner/plan_util_test.cpp
```diff
@@ -163,7 +164,139 @@ TEST_F(PlanUtilTests, GetAffectedIndexesTest) {
...
 }
```
malin1993ml: I actually think the tests here are fine after you fix the above things. Eventually, we want to use the suggestion from the AutoAdmin algorithm so we don't need to overthink it for now.

PR #1280, opened 2018-04-11
[15721] Support for Outer Join

zhaoguoquan94:
This PR adds support for [Left | Right | Outer] JOIN queries in peloton's query optimizer. 

Major changes:
* Replace InnerJoin* Rules with Join* Rules which support not only inner join transformation rules but also outer join rules. [rule_impls.h, rule.cpp]
* Add new Join operators [query_to_operator_transformer.cpp, operators.h,operator_node.h, internal_types.h]
* New physical plan generators[plan_generator.cpp]. 
* Bug fixes.
* Test cases.

## Current Status
* two test cases still fails because old [bug](https://github.com/cmu-db/peloton/blame/master/src/optimizer/group_expression.cpp#L98)  in optimizer. [planner_equality_test, optimizer_sql_test]
* Unknown bug in executor which generate incorrect results given correct plan.


src/optimizer/operators.cpp
```diff
@@ -553,6 +527,104 @@ Operator PhysicalLimit::make(int64_t offset, int64_t limit) {
...
+  join->left_keys = std::move(left_keys);
+  join->right_keys = std::move(right_keys);
+  join->type = _type;
+
+  return Operator(join);
+}
+
+hash_t PhysicalNLJoin::Hash() const {
+  hash_t hash = BaseOperatorNode::Hash();
+  for (auto &expr : left_keys)
```
nappelson: can these be for `(cost auto &...`)? Generally good to prefer using const where possible. Based on an initial glance looks like it could be but i may have missed something

src/optimizer/operators.cpp
```diff
@@ -553,6 +527,104 @@ Operator PhysicalLimit::make(int64_t offset, int64_t limit) {
...
+  join->join_predicates = std::move(conditions);
+  join->left_keys = std::move(left_keys);
+  join->right_keys = std::move(right_keys);
+  join->type = _type;
+  return Operator(join);
+}
+
+hash_t PhysicalHashJoin::Hash() const {
+  hash_t hash = BaseOperatorNode::Hash();
+  for (auto &expr : left_keys)
```
nappelson: same as above regarding constness

src/optimizer/operators.cpp
```diff
@@ -553,6 +527,104 @@ Operator PhysicalLimit::make(int64_t offset, int64_t limit) {
...
+
+bool PhysicalHashJoin::operator==(const BaseOperatorNode &r) {
+  if (r.GetType() != OpType::HashJoin) return false;
+
+  const PhysicalHashJoin &node = *static_cast<const PhysicalHashJoin *>(&r);
+  if (join_predicates.size() != node.join_predicates.size() ||
+      left_keys.size() != node.left_keys.size() ||
+      right_keys.size() != node.right_keys.size())
+    return false;
+  for (size_t i = 0; i < left_keys.size(); i++) {
```
nappelson: see other related comment about using stl iterators

src/include/optimizer/rule_impls.h
```diff
@@ -41,17 +41,19 @@ class InnerJoinCommutativity : public Rule {
...
+  JoinAssociativity();
 
   bool Check(std::shared_ptr<OperatorExpression> plan,
              OptimizeContext *context) const override;
 
   void Transform(std::shared_ptr<OperatorExpression> input,
                  std::vector<std::shared_ptr<OperatorExpression>> &transformed,
                  OptimizeContext *context) const override;
+ private:
+  bool StrongPredicate(std::shared_ptr<OperatorExpression> plan,
```
GustavoAngulo: There should be some documentation as to why a strong predicate check is needed for outer join associativity.

src/optimizer/plan_generator.cpp
```diff
@@ -177,6 +177,88 @@ void PlanGenerator::Visit(const PhysicalDistinct *) {
...
+      op->type, move(join_predicate), move(proj_info), proj_schema,
+      left_keys, right_keys, settings::SettingsManager::GetBool(
+                                 settings::SettingId::hash_join_bloom_filter)));
+
+  join_plan->AddChild(move(children_plans_[0]));
+  join_plan->AddChild(move(hash_plan));
+  output_plan_ = move(join_plan);
+}
+
 void PlanGenerator::Visit(const PhysicalInnerNLJoin *op) {
```
GustavoAngulo: Delete the methods for `PhysicalInnerNLJoin` and `PhysicalInnerHashJoin` if we are consolidating Inner+Outer joins

src/optimizer/rule_impls.cpp
```diff
@@ -630,19 +678,19 @@ InnerJoinToInnerNLJoin::InnerJoinToInnerNLJoin() {
...
   (void)plan;
   return true;
 }
 
-void InnerJoinToInnerNLJoin::Transform(
+void JoinToNLJoin::Transform(
     std::shared_ptr<OperatorExpression> input,
     std::vector<std::shared_ptr<OperatorExpression>> &transformed,
     UNUSED_ATTRIBUTE OptimizeContext *context) const {
   // first build an expression representing hash join
```
GustavoAngulo: Should be "representing nested loop join"

script/testing/junit/OptimizerTest.java
```diff
@@ -0,0 +1,191 @@
...
+            assertTrue(resultSet.next());
+            assertEquals(1, resultSet.getInt(1));
+            assertTrue(resultSet.next());
+            assertEquals(2, resultSet.getInt(1));
+            assertFalse(resultSet.next());
+        } catch (Exception e) {
+            e.printStackTrace();
+            fail();
+        }
+        try (
```
dasteere: I would suggest splitting these into different tests. That way if it fails it will be quicker to identify where it went wrong.

script/testing/junit/OptimizerTest.java
```diff
@@ -0,0 +1,191 @@
...
+            assertTrue(resultSet.next());
+            assertEquals(null, resultSet.getObject(1));
+            assertEquals(null, resultSet.getObject(2));
+            assertEquals(null, resultSet.getObject(3));
+            assertFalse(resultSet.next());
+        } catch (Exception e) {
+            e.printStackTrace();
+            fail();
+        }
+        try (
```
dasteere: Same here

src/include/optimizer/operators.h
```diff
@@ -407,6 +384,48 @@ class PhysicalLimit : public OperatorNode<PhysicalLimit> {
   int64_t limit;
 };
 
+//===--------------------------------------------------------------------===//
+// NLJoin (Inner + Outer Joins)
+//===--------------------------------------------------------------------===//
+class PhysicalNLJoin : public OperatorNode<PhysicalNLJoin> {
+ public:
+  static Operator make(
+      JoinType _type, std::vector<AnnotatedExpression> conditions,
```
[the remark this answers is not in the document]
apavlo: I don't think we want to use leading underscores for this. AFAIK, we don't use that anywhere else in the code.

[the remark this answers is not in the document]
apavlo: Agreed.

src/optimizer/plan_generator.cpp
apavlo: Agreed.

src/include/optimizer/rule_impls.h
apavlo: @GustavoAngulo Where do you want them to add that?

script/testing/junit/OptimizerTest.java
apavlo: Agreed.

src/optimizer/operators.cpp
```diff
@@ -553,6 +527,104 @@ Operator PhysicalLimit::make(int64_t offset, int64_t limit) {
...
+  for (auto &expr : left_keys)
+    hash = HashUtil::CombineHashes(hash, expr->Hash());
+  for (auto &expr : right_keys)
+    hash = HashUtil::CombineHashes(hash, expr->Hash());
+  for (auto &pred : join_predicates)
+    hash = HashUtil::CombineHashes(hash, pred.expr->Hash());
+  return hash;
+}
+
+bool PhysicalHashJoin::operator==(const BaseOperatorNode &r) {
```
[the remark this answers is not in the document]
apavlo: Agreed.

apavlo:
Good job. I am closing this PR. You can resubmit it at the end of the semester.

@nappelson @GustavoAngulo @dasteere Nice work.

src/include/optimizer/rule_impls.h
GustavoAngulo: We don't really have a central documentation for the optimizer yet, so above the implementation of the `StrongPredicate` function should be good.

chenboy: Hey, can we add this in https://github.com/cmu-db/peloton/blob/master/src/include/optimizer/util.h? It makes more sense add it as a helper function instead of a member function of the rule.

tli2:
Who is currently responsible for merging this in? Are code reviews here outdated or still being addressed?

apavlo:
@saatviks is the PR czar this month.

saatviks:
@apavlo I was last month - For June its @tli2

PR #1283, opened 2018-04-12
[15721] Peloton Self-Driving Infrastructure

ChTimTsubasa:
This PR is working toward an infrastructure for self-driving components on Peloton.

It includes:
1. Brain - Peloton communication and periodic replication of Catalog on the brain side;
2. Memory Statistics: in-line tile memory allocation is recorded and stored in the table metric catalog;
3. Query Latency Statistics: query execution latency is measured and recorded.


src/executor/plan_executor.cpp
```diff
@@ -147,6 +147,13 @@ void PlanExecutor::ExecutePlan(
   PELOTON_ASSERT(plan != nullptr && txn != nullptr);
   LOG_TRACE("PlanExecutor Start (Txn ID=%" PRId64 ")", txn->GetTransactionId());
 
+  if (static_cast<StatsType>(settings::SettingsManager::GetInt(
+      settings::SettingId::stats_mode)) != StatsType::INVALID) {
+    stats::BackendStatsContext::GetInstance()
+        ->GetQueryLatencyMetric()
+        .StartTimer();
```
pbollimp: The Execute function in query.h has a comment saying that it is asynchronous. I think your code looks right, but please double check if this is correct.

src/include/settings/settings.h
```diff
@@ -171,10 +171,11 @@ SETTING_bool(brain,
             false,
             true, true)
 
-SETTING_string(peloton_address,
-               "ip and port of the peloton rpc service, address:port",
-               "127.0.0.1:15445",
-               false, false)
+// Enable or disable data collection
+SETTING_bool(brain_data_collection,
```
sivaprasadsudhir: The default value should be false?

src/include/statistics/memory_metric.h
```diff
@@ -0,0 +1,83 @@
...
+        usage_(MetricType::COUNTER) {}
+
+  //===--------------------------------------------------------------------===//
+  // ACCESSORS
+  //===--------------------------------------------------------------------===//
+  inline void IncreaseAllocation(int64_t bytes) { alloc_.Increment(bytes); }
+
+  inline void IncreaseUsage(int64_t bytes) { usage_.Increment(bytes); }
+
+  inline void DecreaseAllocation(int64_t bytes) { alloc_.Increment(bytes); }
```
vkonagar: alloc_.Decrement(bytes)?

src/storage/tile.cpp
```diff
@@ -67,6 +67,14 @@ Tile::Tile(BackendType backend_type, TileGroupHeader *tile_header,
   // if (schema.IsInlined() == false) {
   pool = new type::EphemeralPool();
   //}
+
+  // Record memory deallocation
+  if (table_id != INVALID_OID &&
+      static_cast<StatsType>(settings::SettingsManager::GetInt(
```
pbollimp: I see that this condition is used in many places in your code. Can this logic be moved to a function somewhere?

src/include/network/peloton_rpc_handler_task.h
```diff
@@ -23,7 +25,6 @@ namespace network {
 class PelotonRpcServerImpl final : public PelotonService::Server {
  protected:
   kj::Promise<void> createIndex(CreateIndexContext) override {
-    // TODO(tianyu) Write actual index code
```
vkonagar: Why is this removed?

src/catalog/catalog.cpp
```diff
@@ -145,17 +149,18 @@ void Catalog::Bootstrap() {
...
-  ProcCatalog::GetInstance(txn);
+  catalogs_.push_back(DatabaseMetricsCatalog::GetInstance(txn));
+  catalogs_.push_back(TableMetricsCatalog::GetInstance(txn));
+  catalogs_.push_back(IndexMetricsCatalog::GetInstance(txn));
+  catalogs_.push_back(QueryMetricsCatalog::GetInstance(txn));
+  // TODO(tianyu): WTF?
+  catalogs_.push_back(&SettingsCatalog::GetInstance(txn));
+  catalogs_.push_back(&TriggerCatalog::GetInstance(txn));
+  catalogs_.push_back(&LanguageCatalog::GetInstance(txn));
+  catalogs_.push_back(&ProcCatalog::GetInstance(txn));
```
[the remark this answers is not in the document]
tli2: This is larger in scale than this PR probably. Will post issue and leave until later.

src/include/brain/brain.h
```diff
@@ -28,7 +31,26 @@ namespace brain {
...
+    pqxx::work w(sql_connection_);
+    pqxx::result result =  w.exec(query);
+    w.commit();
+    return result;
+  }
+
+ private:
+  capnp::EzRpcClient rpc_client_;
+  // TODO(tianyu): eventually change this into rpc
+  pqxx::connection sql_connection_;
```
malin1993ml: We need some comments for the variables and methods in this class.

src/include/brain/brain.h
```diff
@@ -16,6 +16,9 @@
 #include <cstdio>
 #include <utility>
 #include <functional>
+#include <pqxx/pqxx>
+#include "settings/setting_id.h"
+#include "settings/settings_manager.h"
```
malin1993ml: This .h file is going to be used in many places. So I think we should move the implementation of `BrainEnvironment()` and these two includes into .cpp file to reduce dependency.

src/include/brain/brain.h
```diff
@@ -28,7 +31,26 @@ namespace brain {
...
 class BrainEnvironment {
-  // TODO(tianyu): fill in as needed
+ public:
+  BrainEnvironment() :
+      rpc_client_{settings::SettingsManager::GetString(settings::SettingId::peloton_rpc_address)},
+      sql_connection_{settings::SettingsManager::GetString(settings::SettingId::peloton_address)} {}
+
+  inline PelotonService::Client GetPelotonService() {
+    return rpc_client_.getMain<PelotonService>();
+  }
```
malin1993ml: Is this how we're going to communicate with Peloton server? It does not seem used anywhere.

src/include/brain/catalog_sync_brain_job.h
```diff
@@ -0,0 +1,124 @@
...
+#include "catalog/table_catalog.h"
+#include "storage/data_table.h"
+#include "catalog/schema.h"
+#include "type/value_factory.h"
+#include "catalog/query_history_catalog.h"
+#include "catalog/query_metrics_catalog.h"
+
+namespace peloton {
+namespace brain {
+class CatalogSyncBrainJob : public BrainJob {
```
malin1993ml: This class overall looks good! Two small issues:
1. Some brief comments would be good.
2. I would suggest moving the definitions of all the functions except `CatalogSyncBrainJob(BrainEnvironment *env)` into the cpp file and reduce includes as much as possible.

It is a hacky way to do syncing, but it will stay here for a while...

src/include/statistics/abstract_metric_new.h
```diff
@@ -0,0 +1,58 @@
+//===----------------------------------------------------------------------===//
+//
+//                         Peloton
+//
+// abstract_metric.h
+//
+// Identification: src/statistics/abstract_metric.h
+//
+// Copyright (c) 2015-16, Carnegie Mellon University Database Group
```
malin1993ml: Copy right date

src/include/statistics/abstract_metric_new.h
```diff
@@ -0,0 +1,58 @@
+//===----------------------------------------------------------------------===//
+//
+//                         Peloton
+//
+// abstract_metric.h
+//
+// Identification: src/statistics/abstract_metric.h
```
malin1993ml: You're calling this file abstract_metric_new.h...

Actually, I'm not sure that's a good name. Do we still need the old abstract_metric.h?

src/include/statistics/backend_stats_context_new.h
```diff
@@ -0,0 +1,115 @@
+//===----------------------------------------------------------------------===//
+//
+//                         Peloton
+//
+// backend_stats_context.h
+//
+// Identification: src/statistics/backend_stats_context.h
```
malin1993ml: Eh this is again confusing. I think we only need to keep one of this class, either the old one or the new one. Can you explain what's going on here?

src/include/storage/tuple.h
```diff
@@ -252,7 +252,7 @@ inline Tuple::Tuple(char *data, catalog::Schema *schema) {
 
   tuple_data_ = data;
   tuple_schema_ = schema;
-  allocated_ = false;  // ???
```
malin1993ml: This comment was funny... Thanks for the new explanation!

src/common/init.cpp
```diff
@@ -90,9 +91,20 @@ void PelotonInit::Initialize() {
 
   // Initialize the Statement Cache Manager
   StatementCacheManager::Init();
+
+  // Start Statistic Aggregator
+  if (static_cast<StatsType>(settings::SettingsManager::GetInt(
+          settings::SettingId::stats_mode)) != StatsType::INVALID) {
```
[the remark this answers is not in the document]
apavlo: Should be `StatsModeType`

PR #1284, opened 2018-04-12
[15721] Add/Drop Index

alphalzh:
Group: Add Drop Index
Member: Xingyu Jin, Rong Huang, Ziheng Liao
Please check the following files:
/src/concurrency/lock_manager.cpp (lock manager, please check the whole file)
/src/include/concurrency/lock_manager.h (lock manager, please check the whole file)
/src/executor/create_executor.cpp (please check the function CreateIndex() )
/src/executor/drop_executor.cpp (please check the function DropIndex() )
/src/executor/*_executor.cpp (some other additions to various executors, no single functions. Please comment if you want to see those changes.)
/test/concurrency/lock_manager_test.cpp (tests for lock manager, please check the whole file)
/test/concurrency/testing_transaction_util.cpp (support functions for index create and drop)
/test/include/concurrency/testing_transaction_util.h (header for support functions for index create and drop)
/test/concurrency/transaction_index_test.cpp (tests for index create and drop, both in one thread and multiple threads)


src/executor/create_executor.cpp
```diff
@@ -211,20 +222,39 @@ bool CreateExecutor::CreateIndex(const planner::CreatePlan &node) {
...
+      database_name, table_name, txn);
+
+  // Get table oid
+  oid_t table_oid = table_object->GetTableOid();
+
+  // Lock the table being indexed
+  concurrency::LockManager *lm = concurrency::LockManager::GetInstance();
+  bool lock_success = lm->LockExclusive(table_oid);
+  if (!lock_success) {
+    LOG_TRACE("Cannot obtain lock for the table, abort!");
```
DeanChensj: Where is the abort logic?

src/executor/create_executor.cpp
```diff
@@ -211,20 +222,39 @@ bool CreateExecutor::CreateIndex(const planner::CreatePlan &node) {
...
   ResultType result = catalog::Catalog::GetInstance()->CreateIndex(
-      database_name, table_name, key_attrs, index_name, unique_flag,
-      index_type, txn);
+      database_name, table_name, key_attrs, index_name, unique_flag, index_type,
+      txn);
+
+  // Unlock the table being indexed
+  bool unlock_success = lm->UnlockExclusive(table_oid);
+  if (!unlock_success) {
+    LOG_TRACE("Cannot unlock the table, abort!");
```
DeanChensj: See above.

src/executor/drop_executor.cpp
```diff
@@ -178,16 +183,29 @@ bool DropExecutor::DropTrigger(const planner::DropPlan &node,
...
+ * @return  bool    always false
+ */
 bool DropExecutor::DropIndex(const planner::DropPlan &node,
                              concurrency::TransactionContext *txn) {
+  // Get the name of index and the index in catalog.
   std::string index_name = node.GetIndexName();
   auto index_object =
       catalog::IndexCatalog::GetInstance()->GetIndexObject(index_name, txn);
 
+  // Drop index in catalog.
```
DeanChensj: Do we need the lock when drop the index?

src/concurrency/lock_manager.cpp
```diff
@@ -0,0 +1,254 @@
...
+//
+// lock_manager.cpp
+//
+// Identification: src/concurrency/lock_manager.cpp
+//
+// Copyright (c) 2015-18, Carnegie Mellon University Database Group
+//
+//===----------------------------------------------------------------------===//
+
+#include <map>
```
DeanChensj: It's already in the header?

src/concurrency/lock_manager.cpp
```diff
@@ -0,0 +1,254 @@
...
+  rw_lock->lock();
+
+  // Unlock internal lock
+  internal_rw_lock_.unlock_shared();
+  LOG_TRACE("Exclusive lock success.");
+  return true;
+}
+
+/**
+ * @brief    Acquire exclusive lock for RW_LOCK(blocking)
```
DeanChensj: @brief Converts the ownership from shared to exclusive for the calling thread.

test/concurrency/lock_manager_test.cpp
```diff
@@ -0,0 +1,133 @@
...
+  EXPECT_TRUE(result3);
+}
+
+/**
+ * @brief Test shared lock behaviour
+ */
+TEST_F(LockManagerTests, LockSharedTest) {
+  // Initialize lock
+  concurrency::LockManager *lm = concurrency::LockManager::GetInstance();
+  bool result0 = lm->InitLock((oid_t)1, concurrency::LockManager::RW_LOCK);
```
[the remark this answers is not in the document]
alphalzh: Well I don't think this could be a problem. The point to add comment here is, this feature is newly added, and I think it's necessary to remind other developers of its existence.

src/executor/create_executor.cpp
alphalzh: Abort here will cause problems, for there are other operations calling the executor of create and delete, and such operations don't init locks. My bad here, should change the log message.

src/executor/drop_executor.cpp
alphalzh: No, we don't. It's first logically deleted and then garbage collected, so no lock is required since no other new txns will be able to see the index.

src/concurrency/lock_manager.cpp
alphalzh: Sorry, will fix :0

src/concurrency/lock_manager.cpp
alphalzh: Should be. Will delete this include

test/executor/create_test.cpp
```diff
@@ -80,6 +80,7 @@ TEST_F(CreateTests, CreatingDB) {
   txn_manager.CommitTransaction(txn);
 }
 
+
```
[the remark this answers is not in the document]
alphalzh: Well.. It's an empty line, and it does no harm

test/include/concurrency/testing_transaction_util.h
```diff
@@ -68,7 +68,10 @@
 #include "storage/tile_group_header.h"
 #include "storage/tuple.h"
 #include "catalog/catalog_defaults.h"
+#include "catalog/catalog.h"
 #include "common/internal_types.h"
+
+#include <assert.h>
```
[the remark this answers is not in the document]
alphalzh: According to my teammates, PL_ASSERT seems to not function, and he used assert.h instead. Will follow up on this issue.

test/index/skiplist_index_test.cpp
```diff
@@ -1,70 +0,0 @@
-//===----------------------------------------------------------------------===//
```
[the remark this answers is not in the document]
alphalzh: IDK... But it doesn't do anything anyway.

test/include/concurrency/testing_transaction_util.h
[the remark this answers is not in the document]
apavlo: `PL_ASSERT` has been renamed to `PELOTON_ASSERT`

test/index/skiplist_index_test.cpp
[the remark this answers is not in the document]
alphalzh: We should add it back in our next commit.

test/concurrency/lock_manager_test.cpp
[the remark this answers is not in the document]
DeanChensj: IMO, adding some block comments about the test functionality is enough...  We don't need to comment every single statement...

src/executor/create_executor.cpp
DeanChensj: What will happen if the `lock_success` is false?

src/executor/drop_executor.cpp
DeanChensj: What if there's another index scan going on?

alphalzh: It'll see the index anyway, since it still has a pointer to the index. The index itself will be garbage collected only after no thread is holding any pointer to the index, as it utilizes the smart pointer in C++. 
It's a little bit complicated here, actually. If you have further questions, I can explain to you in person about how this works.

src/executor/create_executor.cpp
alphalzh: Basically, that means the mutex that we want to lock does not exist. We only initialize locks when we call the create table executor, but there are some internal tables (e.g. the catalog tables) that are created without calling the create table executor. However, we may still call other executors on these tables, and in such cases it will throw an error. If we abort here, the system will fail. So our current solution is to leave it there, but I agree that this should be handled in future. The potential problem here is that this may cause deadlocks if a locked mutex failed to unlock silently, but this should be ultra-rare since every lock in the lock manager's table is unlocked before any return condition - that is, unless the boost::upgrade_mutex itself failed for some reason, we should not expect an out-of-scope hanging locked mutex.

src/catalog/catalog.cpp
```diff
@@ -433,6 +433,7 @@ ResultType Catalog::CreateIndex(
             table_oid);
   auto pg_index = IndexCatalog::GetInstance();
   oid_t index_oid = pg_index->GetNextOid();
+
```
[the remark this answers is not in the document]
alphalzh: We changed this file in earlier commits and changed it back later, so there might be some small changes which are trivial.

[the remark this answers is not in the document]
DeanChensj: I understand, but these minor changes may cause trouble for the future merge..

src/executor/drop_executor.cpp
DeanChensj: Thanks, the explanation is pretty clear!

DeanChensj: Thanks, the explanation is pretty clear!

src/catalog/catalog.cpp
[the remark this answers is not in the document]
alphalzh: Sure, will fix.

test/CMakeLists.txt
```diff
@@ -47,7 +47,7 @@ set(TESTING_UTIL_TXN        ${PROJECT_SOURCE_DIR}/test/concurrency/testing_trans
 set(TESTING_UTIL_STATS      ${PROJECT_SOURCE_DIR}/test/statistics/testing_stats_util.cpp)
 set(TESTING_UTIL_SQL        ${PROJECT_SOURCE_DIR}/test/sql/testing_sql_util.cpp)
 set(TESTING_UTIL_INDEX      ${PROJECT_SOURCE_DIR}/test/index/testing_index_util.cpp)
-set(TESTING_UTIL_CODEGEN ${PROJECT_SOURCE_DIR}/test/codegen/testing_codegen_util.cpp)
+set(TESTING_UTIL_CODEGEN ${PROJECT_SOURCE_DIR}/test/codegen/testing_codegen_util.cpp concurrency/transaction_index_test.cpp)
```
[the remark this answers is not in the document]
alphalzh: @eloiseh @xingyu Any comments on this? I think we shouldn't change the cmakelists.txt file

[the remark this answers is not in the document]
eloiseh: Yeah, I think we don't need to change it.

src/concurrency/lock_manager.cpp
```diff
@@ -0,0 +1,254 @@
...
+  boost::upgrade_mutex *rw_lock = GetLock(oid);
+  if (rw_lock == nullptr) {
+    internal_rw_lock_.unlock();
+    LOG_TRACE("Remove lock failed.");
+    return false;
+  }
+
+  // Remove from the lock map
+  lock_map_.erase(oid);
+  delete (rw_lock);
```
Dingshilun: Do we need to check whether there are still transactions holding the lock pointer here?
I think using a deprecated way to delete a lock is better as when removing a lock there might still be transacations holding the lock pointer. Maybe add a TODO here and we can help handle this.

test/concurrency/transaction_index_test.cpp
```diff
@@ -0,0 +1,186 @@
...
+
+      scheduler.Run();
+
+      EXPECT_EQ(ResultType::SUCCESS, scheduler.schedules[0].txn_result);
+      EXPECT_EQ(ResultType::ABORTED, scheduler.schedules[1].txn_result);
+      EXPECT_EQ(ResultType::SUCCESS, scheduler.schedules[2].txn_result);
+
+      EXPECT_EQ(1, scheduler.schedules[0].create_index_results[0]);
+      // TODO: maybe should return 0?
+      EXPECT_EQ(1, scheduler.schedules[1].create_index_results[0]);
```
Dingshilun: As the transaction aborts, I think it should return 0.

src/concurrency/lock_manager.cpp
alphalzh: Good idea. Maybe make it a shared_ptr, as it will automatically clean itself after the reference count decreased to 0. Thanks for the suggestion!

alphalzh: About the automatic lock release.. Yes I know the method and it can ensure the lock be released if some error occurs in the middle of function execution. But it's indeed a little tricky and not standard c++ code style, so I don't know if it's applicable.. Maybe we need to ask Andy about this?

test/index/skiplist_index_test.cpp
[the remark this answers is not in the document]
apavlo: @alphalzh Don't add it back as a commit. Expunge the delete from your commit log.

test/executor/create_test.cpp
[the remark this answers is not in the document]
apavlo: @alphalzh Avoid unnecessary changes.

src/concurrency/lock_manager.cpp
```diff
@@ -0,0 +1,254 @@
...
+ */
+bool LockManager::InitLock(oid_t oid, LockManager::LockType /*type*/) {
+  // Need to lock internal lock to protect internal lock map
+  internal_rw_lock_.lock();
+
+  // Try to find the lock in table
+  boost::upgrade_mutex *lock = GetLock(oid);
+  if (lock != nullptr) {
+    internal_rw_lock_.unlock();
+    LOG_TRACE("Init lock failed.");
```
sxzh93: Maybe add more information in log trace? Such as value of oid.

src/concurrency/lock_manager.cpp
```diff
@@ -0,0 +1,254 @@
...
+
+  std::pair<oid_t, boost::upgrade_mutex *> p;
+  p.first = oid;
+  p.second = rw_lock;
+  // Insert lock into map
+  lock_map_.insert(p);
+
+  // Unlock internal lock
+  internal_rw_lock_.unlock();
+  LOG_TRACE("Init lock success.");
```
sxzh93: Same with line 50.

src/concurrency/lock_manager.cpp
```diff
@@ -0,0 +1,254 @@
...
+
+  // Try to access the lock
+  boost::upgrade_mutex *rw_lock = GetLock(oid);
+  if (rw_lock == nullptr) {
+    internal_rw_lock_.unlock_shared();
+    LOG_TRACE("Exclusive lock failed.");
+    return false;
+  }
+
+  // Read (shared) lock
```
sxzh93: Still for read (shared) lock? This comment is confusing.

test/include/concurrency/testing_transaction_util.h
```diff
@@ -324,6 +335,30 @@ class TransactionThread {
...
+      case TXN_OP_DROP_INDEX: {
+        auto catalog = catalog::Catalog::GetInstance();
+        auto tmp = catalog->DropIndex(index_name1, txn);
+        if (tmp == ResultType::SUCCESS)
+          schedule->drop_index_results.push_back(1);
+        else
+          schedule->drop_index_results.push_back(0);
+        LOG_INFO("Txn %d Drop Index", schedule->schedule_id);
+        break;
+      }
```
sxzh93: Add default case?

src/executor/create_executor.cpp
```diff
@@ -111,6 +110,14 @@ bool CreateExecutor::CreateTable(const planner::CreatePlan &node) {
   if (current_txn->GetResult() == ResultType::SUCCESS) {
     LOG_TRACE("Creating table succeeded!");
 
+    // Initialize table lock
+    auto table_object = catalog::Catalog::GetInstance()->GetTableObject(
+        database_name, table_name, current_txn);
+
+    oid_t table_oid = table_object->GetTableOid();
+    concurrency::LockManager *lm = concurrency::LockManager::GetInstance();
+    lm->InitLock(table_oid, concurrency::LockManager::RW_LOCK);
```
sxzh93: Should we handle negative case here? What if InitLock is failed.

src/executor/delete_executor.cpp
```diff
@@ -85,6 +86,14 @@ bool DeleteExecutor::DExecute() {
 
   auto current_txn = executor_context_->GetTransaction();
 
+  oid_t table_oid = target_table_->GetOid();
+  // Lock the table (reader lock)
+  concurrency::LockManager *lm = concurrency::LockManager::GetInstance();
+  bool lock_success = lm->LockShared(table_oid);
+  if (!lock_success) {
+    LOG_WARN("Cannot obtain lock for the table, abort!");
```
sxzh93: Make log level consist for same scenario, we use LOG_TRACE in some places.

src/executor/insert_executor.cpp
```diff
@@ -299,9 +328,19 @@ bool InsertExecutor::DExecute() {
...
+    // Unlock the table
+    bool unlock_success = lm->UnlockShared(table_oid);
+    if (!unlock_success) {
+      LOG_TRACE("Cannot unlock the table, abort!");
+    }
     done_ = true;
     return true;
   }
+  // Unlock the table
+  bool unlock_success = lm->UnlockShared(table_oid);
```
sxzh93: If negative case for unlock is quite same, maybe we could write some wrapper to do this instead of write similar logic.

src/executor/insert_executor.cpp
```diff
@@ -74,6 +75,14 @@ bool InsertExecutor::DExecute() {
     return false;
   }
 
+  oid_t table_oid = target_table->GetOid();
```
sxzh93: My comments may same stupid, why we need lock in insert_executor and update_executor? Are they related to add/drop index?

src/concurrency/lock_manager.cpp
alphalzh: Typo, already fixed in local branch. Thanks for pointing out

src/executor/create_executor.cpp
alphalzh: Currently it won't do anything, but yes I agree with you.

src/executor/delete_executor.cpp
alphalzh: Sorry, we used warn to debug and I missed this when changing everything back before PR. Fixed in local branch already

src/executor/insert_executor.cpp
alphalzh: The reason to use locks in add drop index is to prevent index changes during the add/drop index period, so we need to lock operations that modifies the index.

apavlo: You don't need to aquire locks to drop an index.

apavlo:
Replaced by #1364

PR #1285, opened 2018-04-12
[15721] Cost Model Tests

nappelson:
We refactored the optimizer to take in an abstract cost calculator as input so that we can test different cost models. We created and added a handful of "plan_selection_tests" that test if the optimizer does the appropriate join ordering (it currently does not). We also created an example "bad cost model" and showed how using the optimizer with this cost model chooses an improper join. 

Our 75% goal was a testing framework that would allow us to tests that when the optimizer produced a good and a bad plan, that when the plans were executed, we would see a difference in runtimes. These tests and the cost model refactor are steps towards that.

Note: This PR does not include the fix from #1245 , we thought to keep them separate for clarity.


zhaoguoquan94:
https://github.com/cmu-db/peloton/pull/1285/files#diff-5b6d7a5972a8d33624e21f5ccdd3eabcR41
https://github.com/cmu-db/peloton/pull/1285/files#diff-5b6d7a5972a8d33624e21f5ccdd3eabcR51
Not your changes. Just curious why you did a dynamic pointer cast when the return type is already the same as the the pointer that you cast to?

test/optimizer/plan_selection_test.cpp
```diff
@@ -0,0 +1,607 @@
...
+    if (plan->GetPlanNodeType() == PlanNodeType::SEQSCAN) {
+      auto scan = dynamic_cast<planner::AbstractScan *>(plan);
+      LOG_DEBUG("%s%s(%s)", spacing.c_str(), scan->GetInfo().c_str(),
+                scan->GetTable()->GetName().c_str());
+    } else {
+      LOG_DEBUG("%s%s", spacing.c_str(), plan->GetInfo().c_str());
+    }
+
+    for (size_t i = 0; i < plan->GetChildren().size(); i++) {
+      PrintPlan(plan->GetChildren()[0].get(), level + 1);
```
zhaoguoquan94: Seems to be bug here? 
`PrintPlan(plan->GetChildren()[0].get(), level + 1);` should be

 `PrintPlan(plan->GetChildren()[i].get(), level + 1);` right?

test/optimizer/plan_selection_test.cpp
```diff
@@ -0,0 +1,607 @@
...
+
+  EXPECT_EQ(0, plan->GetChildren()[0]->GetChildren().size());
+  EXPECT_EQ(1, plan->GetChildren()[1]->GetChildren().size());
+
+  auto left_scan =
+      dynamic_cast<planner::AbstractScan *>(plan->GetChildren()[0].get());
+  auto right_scan = dynamic_cast<planner::AbstractScan *>(
+      plan->GetChildren()[1]->GetChildren()[0].get());
+
+  // TODO: This should actually be reversed, setting it to this now so that the
```
zhaoguoquan94: Can you provide more comments here? Line 455 failed in my local build.

src/include/optimizer/optimizer.h
```diff
@@ -38,9 +39,9 @@ class TransactionContext;
...
 
 namespace test {
-  class OptimizerRuleTests_SimpleAssociativeRuleTest_Test;
-  class OptimizerRuleTests_SimpleAssociativeRuleTest2_Test;
-} 
+class OptimizerRuleTests_SimpleAssociativeRuleTest_Test;
+class OptimizerRuleTests_SimpleAssociativeRuleTest2_Test;
+}
 
 namespace optimizer {
```
qwkai: Not actually your changes. Just noticed there are two namespaces optimizer defined. Could these two be combined into one?

test/optimizer/plan_selection_test.cpp
```diff
@@ -0,0 +1,607 @@
...
+  void PrintPlan(std::shared_ptr<planner::AbstractPlan> plan, int level = 0) {
+    PrintPlan(plan.get(), level);
+  }
+
+  void PrintPlan(planner::AbstractPlan *plan, int level = 0) {
+    auto spacing = std::string(level, '\t');
+
+    if (plan->GetPlanNodeType() == PlanNodeType::SEQSCAN) {
+      auto scan = dynamic_cast<planner::AbstractScan *>(plan);
+      LOG_DEBUG("%s%s(%s)", spacing.c_str(), scan->GetInfo().c_str(),
```
qwkai: Noe sure if we need to consider this, but `scan` variable will be throwing unused variable error if the build mode is set to Release.

test/optimizer/plan_selection_test.cpp
```diff
@@ -0,0 +1,607 @@
...
+  InsertDataHelper(table_1_name, test1_table_size);
+  InsertDataHelper(table_2_name, test2_table_size);
+
+  txn_manager.CommitTransaction(txn);
+
+  AnalyzeTable(table_1_name);
+  AnalyzeTable(table_2_name);
+
+  auto plan = PerformTransactionAndGetPlan(CreateTwoWayJoinQuery(
+      table_2_name, table_1_name, column_1_name, column_1_name));
```
qwkai: Since you are using the same column_1_name for both table 1 and table 2, it seems there is no difference in switching the order of table in the `CreateTwoWayJoinQuery` function, that said I find LargeTest4 is the same as LargeTest2? The same with SmallTest2 and SmallTest4.

GustavoAngulo:
Closing this PR to allow us to submit new PR for second code review.

src/include/optimizer/optimizer.h
```diff
@@ -83,14 +87,20 @@ class Optimizer : public AbstractOptimizer {
 
   OptimizerMetadata &GetMetadata() { return metadata_; }
 
+  void SetWorstCase(bool flag) { worst_case_ = flag; }
+
+  bool DoWorstCase() { return worst_case_; }
+
```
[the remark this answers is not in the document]
GustavoAngulo: These were left by mistake. Deleted now.

test/optimizer/plan_selection_test.cpp
```diff
@@ -0,0 +1,607 @@
...
+      dynamic_cast<planner::AbstractScan *>(plan->GetChildren()[1].get());
+
+  // TODO: Join order seems to follow order of join in query. This should really
+  // be swapped. Set
+  // to this way so the tests pass
+  ASSERT_STREQ(left_scan->GetTable()->GetName().c_str(), table_2_name);
+  ASSERT_STREQ(right_scan->GetTable()->GetName().c_str(), table_1_name);
+}
+
+TEST_F(PlanSelectionTest, SimpleJoinOrderSmallTest4) {
```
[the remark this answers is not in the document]
GustavoAngulo: I believe that we wanted to test that the optimizer wasn't just picking joins in the order they were written in the SQL query.

test/optimizer/plan_selection_test.cpp
```diff
@@ -0,0 +1,607 @@
...
+      for (int i = 1; i <= tuple_count; i++) {
+        ss << "(" << i << ", 1.1, 'abcd')";
+        if (i < tuple_count) {
+          ss << ",";
+        }
+      }
+      ss << ";";
+      TestingSQLUtil::ExecuteSQLQuery(ss.str());
+    }
+    LOG_INFO("COUNT: %d", count);
```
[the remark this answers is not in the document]
GustavoAngulo: I think the count was just not updated in the loop for the else clause. Fixed.

test/optimizer/plan_selection_test.cpp
GustavoAngulo: Before we started, the optimizer was picking joins in the order they were written in the query. These tests do this to test that we don't do that with our improvements, but rather pick the joins based on the cost estimates.

PR #1286, opened 2018-04-12
[15721] Temporary Table

yjc408:
This branch contains the work to develop temporary tables feature for peloton.

We designed to support temporary tables in the following way:
Peloton contains a pg_table inside catalog, which stores each table's oid, name, database oid. Upon getting a table, searching will be performed on those attributes. As a result, we added another column in the pg_table called "table_namespace".  It can help to create tables with the same table name inside the same database but under different namespace.

In order to do so, we need to add extra indexes into pg_table when the catalog is initialized. And we need to modify the getTableObject inside catalog.cpp, table_catalog.cpp, and database_catalog.cpp.
For storage manager, there is also a need to add namespace to the database.cpp and also data_table.cpp to get storage table with the name.

So far, our temporary table only supports two kinds of namespace: one is the default public, and the other one is the namespace created for the current network session temporarily.  When the namespace is not specified, searching will be performed under the temporary session first. If nothing found, searching will then be performed under the default public namespace.

One important thing for temporary table management is to create temporary tables only visible for the current session, and also drop all the temporary tables created inside session. We achieved this management by adding the namespace as "pg_temp_sockid" when the session is created, and drop all the tables with that namespace at the end of the session (inside connection_handle).

In order to support namespace, we also modified parser(postgresparser, sql_statement), binder(BindNodeVisitor), and planner (abstract_plan and optimizer, query_to_operator_transformer). 

How things work in peloton: when a connection arrives, it is dispatched to connection_handle, and passed into postgres_sql_protocol to process messages and execute. Traffic cop is used to parse, prepare, and execute statement. Progressparser is used for parsing string to sql statement, optimizer is used to prepare plan based on the sql statement, and also executors are used to execute the plan and generate the result.

### DIFF
--- a/script/testing/junit/TempTableTest.java
+++ b/script/testing/junit/TempTableTest.java
@@ -1,0 +1,179 @@
+//===----------------------------------------------------------------------===//
+//
+//                         Peloton
+//
+// TempTableTest.java
+//
+// Identification: script/testing/junit/TempTableTest.java
+//
+// Copyright (c) 2015-2018, Carnegie Mellon University Database Group
+//
+//===----------------------------------------------------------------------===//
+
+import java.sql.*;
+import org.junit.*;
+import org.postgresql.util.PSQLException;
+
+import static junit.framework.TestCase.fail;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+
+public class TempTableTest extends PLTestBase {
+
+    private static final String SQL_DROP_TABLE =
+            "DROP TABLE IF EXISTS tbl;";
+
+    // Create a permanent table called "tbl"
+    private static final String SQL_CREATE_PERM_TABLE =
+            "CREATE TABLE tbl (id integer, year integer);";
+    // Create temp table using "TEMP" keyword
+    private static final String SQL_CREATE_TEMP_TABLE =
+            "CREATE TEMP TABLE tbl (id integer, year integer);";
+    // Create temp table using "TEMPORARY" keyword
+    private static final String SQL_CREATE_TEMPORARY_TABLE =
+            "CREATE TEMPORARY TABLE tbl (id integer, year integer);";
+
+    private static final String SQL_INSERT =
+            "INSERT INTO tbl VALUES (10, 1995);";
+
+    private static final String SQL_SELECT = "SELECT * FROM tbl;";
+            
+    /**
+     * Check the temp table created by one session is visible to itself, 
+     * but not visible to other sessions
+     */
+    @Test
+    public void test_Visibility() throws SQLException {
+        Connection conn1 = makeDefaultConnection();
+        conn1.setAutoCommit(true);
+        Statement stmt1 = conn1.createStatement();
+
+        Connection conn2 = makeDefaultConnection();
+        conn2.setAutoCommit(true);
+        Statement stmt2 = conn2.createStatement();
+        
+        // Part 1: Create temp table using "TEMP" keyword
+        stmt1.execute(SQL_DROP_TABLE);
+        // Create a temp table called "tbl"
+        stmt1.execute(SQL_CREATE_TEMP_TABLE);
+        // Insert a tuple into the temp table
+        stmt1.execute(SQL_INSERT);
+        
+        // Check the temp table is visible to the session created it
+        ResultSet res1 = stmt1.executeQuery(SQL_SELECT);
+        res1.next();
+        checkRow(res1,
+            new String [] {"id", "year"},
+            new int [] {10, 1995});
+        assertNoMoreRows(res1);
+
+        // Check the temp table is invisible to another session started 
+        // before the temp table was created
+        // Expect an exception: "Table tbl is not found"
+        try {
+            stmt2.execute(SQL_SELECT);
+            fail();
+        } catch (PSQLException e) { }
+
+        // Check the temp table is invisible to another session started 
+        // after the temp table was created
+        // Expect an exception: "Table tbl is not found"
+        conn2 = makeDefaultConnection();
+        conn2.setAutoCommit(true);
+        stmt2 = conn2.createStatement();
+        try {
+            stmt2.execute(SQL_SELECT);
+            fail();
+        } catch (PSQLException e) { }
+        stmt2.close();
+        conn2.close();
+
+        // Check the temp table is invisible to another session started 
+        // after the session which created it has closed
+        // Expect an exception: "Table tbl is not found"
+        stmt1.close();
+        conn1.close();
+        conn2 = makeDefaultConnection();
+        conn2.setAutoCommit(true);
+        stmt2 = conn2.createStatement();
+        try {
+            stmt2.execute(SQL_SELECT);
+            fail();
+        } catch (PSQLException e) { }
+        stmt2.close();
+        conn2.close();
+
+        // Part 2: Create temp table using "TEMPORARY" keyword
+        conn1 = makeDefaultConnection();
+        conn1.setAutoCommit(true);
+        stmt1 = conn1.createStatement();
+        stmt1.execute(SQL_DROP_TABLE);
+        stmt1.execute(SQL_CREATE_TEMPORARY_TABLE);
+
+        // Check the temp table is visible to the session created it
+        ResultSet res2 = stmt1.executeQuery(SQL_SELECT);
+        res2.next();
+        assertNoMoreRows(res2);
+        
+        // Check the temp table is invisible to another session started 
+        // before the table was created
+        // Expect an exception: "Table tbl is not found"
+        try {
+            stmt2.execute(SQL_SELECT);
+            fail();
+        } catch (PSQLException e) { }
+
+        stmt1.close();
+        conn1.close();
+        stmt2.close();
+        conn2.close();
+    }
+
+    /**
+     * Check that during the lifetime of a temp table, the permanent table 
+     * with the same name is invisible
+     */
+    @Test
+    public void test_Temp_Table_Hides_Perm_Table() throws SQLException {
+        Connection conn = makeDefaultConnection();
+        conn.setAutoCommit(true);
+        Statement stmt = conn.createStatement();
+
+        stmt.execute(SQL_DROP_TABLE);
+        // Create a permanent table called "tbl"
+        stmt.execute(SQL_CREATE_PERM_TABLE);
+        // Create a temp table called "tbl"
+        stmt.execute(SQL_CREATE_TEMP_TABLE);
+        // Insert a tuple into the temp table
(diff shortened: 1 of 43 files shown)


src/optimizer/optimizer.cpp
```diff
@@ -94,27 +94,63 @@ void Optimizer::OptimizeLoop(int root_group_id,
...
-    concurrency::TransactionContext *txn) {
+    concurrency::TransactionContext *txn, const std::string session_namespace) {
   // Base Case
   if (parse_tree_list->GetStatements().size() == 0) return nullptr;
 
   unique_ptr<planner::AbstractPlan> child_plan = nullptr;
 
   auto parse_tree = parse_tree_list->GetStatements().at(0).get();
 
+  auto stmt_type = parse_tree->GetType();
```
HenryZhou0333: It seems like it is possible to merge some branches in the select statement.

src/catalog/database_catalog.cpp
```diff
@@ -136,17 +140,52 @@ std::shared_ptr<TableCatalogObject> DatabaseCatalogObject::GetTableObject(
...
  */
 std::shared_ptr<TableCatalogObject> DatabaseCatalogObject::GetTableObject(
-    const std::string &table_name, bool cached_only) {
-  auto it = table_name_cache.find(table_name);
+    const std::string &table_name, const std::string &session_namespace,
+    const std::string &table_namespace, bool cached_only) {
+  // no namespace specified.
+  if (table_namespace.empty()) {
+    // search session first.
+    std::string table_key = session_namespace + std::string("#") + table_name;
```
HenryZhou0333: It seems like the logic from 148-168 has some overlaps. Create a function might improve readability/

src/parser/postgresparser.cpp
```diff
@@ -1016,6 +1016,11 @@ parser::SQLStatement *PostgresParser::CreateTransform(CreateStmt *root) {
         }
         // Update Reference Table
         col->fk_sink_table_name = constraint->pktable->relname;
+        if (constraint->pktable->schemaname) {
```
danae-s: Maybe more comments over here specifically about temporary tables in addition to the general description. Not clear about why adding temporary tables needs this.

src/network/connection_handle.cpp
```diff
@@ -557,6 +561,9 @@ Transition ConnectionHandle::CloseSocket() {
     conn_SSL_context = nullptr;
   }
 
+  // Drop all the temporary tables for the current session
+  traffic_cop_.DropTempTables();
```
HenryZhou0333: Discussed with the team about whether it is the right time to drop temp table.

src/catalog/database_catalog.cpp
```diff
@@ -136,17 +140,52 @@ std::shared_ptr<TableCatalogObject> DatabaseCatalogObject::GetTableObject(
...
+      }
+    }
+    // cache miss return empty object
+    if (cached_only) {
+      return nullptr;
+    }
+    return table_object;
+  }
+  // not empty. - get exact.
+  std::string table_key = table_namespace + std::string("#") + table_name;
```
HenryZhou0333: It is worth thinking a more general way of assigning table_key.

script/testing/junit/TempTableTest.java
```diff
@@ -0,0 +1,179 @@
...
+        stmt.execute(SQL_DROP_TABLE);
+
+        // Check the "tbl" visible now is the permanent table
+        ResultSet res2 = stmt.executeQuery(SQL_SELECT);
+        res2.next();
+        assertNoMoreRows(res2);
+
+        // Drop the permanent table
+        stmt.execute(SQL_DROP_TABLE);
+
```
danae-s: What about the visibility of other transactions? Would they be able to edit the permanent table? Should be covered by the tests.

src/catalog/database_catalog.cpp
```diff
@@ -136,17 +140,52 @@ std::shared_ptr<TableCatalogObject> DatabaseCatalogObject::GetTableObject(
...
+    // fetch from table_catalog under session_namespace.
+    auto table_object = TableCatalog::GetInstance()->GetTableObject(
+        table_name, database_oid, txn, session_namespace);
+    if (table_object == nullptr) {
+      // search under public namespace
+      std::string table_key = DEFAULT_NAMESPACE + std::string("#") + table_name;
+      auto it = table_name_cache.find(table_key);
+      if (it != table_name_cache.end()) return it->second;
+
+      if (cached_only) {
```
danae-s: This may be checked at the very first and all the previous steps can be saved.

script/testing/junit/TempTableTest.java
```diff
@@ -0,0 +1,179 @@
...
+        conn2.setAutoCommit(true);
+        stmt2 = conn2.createStatement();
+        try {
+            stmt2.execute(SQL_SELECT);
+            fail();
+        } catch (PSQLException e) { }
+        stmt2.close();
+        conn2.close();
+
+        // Part 2: Create temp table using "TEMPORARY" keyword
```
bdengNE: More comprehensive tests should be written in Part 2, just like Part 1

test/parser/parser_test.cpp
```diff
@@ -65,6 +65,15 @@ TEST_F(ParserTests, BasicTest) {
       "CREATE TABLE students (name TEXT, student_number INTEGER, city TEXT, "
       "grade DOUBLE)");
 
+  // CREATE temporary table
```
bdengNE: Is it better to write more C++ tests to test the functionalities apart from parser?

src/catalog/table_catalog.cpp
```diff
@@ -341,20 +344,30 @@ std::unique_ptr<catalog::Schema> TableCatalog::InitializeSchema() {
...
  * @param   table_oid
+ * @param   table_namespace
  * @param   table_name
  * @param   database_oid
  * @param   txn     TransactionContext
  * @return  Whether insertion is Successful
  */
-bool TableCatalog::InsertTable(oid_t table_oid, const std::string &table_name,
+bool TableCatalog::InsertTable(oid_t table_oid,
+                               const std::string &table_namespace,
```
danae-s: New arguments should be added last and assigned default values in order to avoid misbehaviors in previous tests.

src/binder/binder_context.cpp
```diff
@@ -29,16 +29,19 @@ void BinderContext::AddRegularTable(parser::TableRef *table_ref,
   table_ref->TryBindDatabaseName(default_database_name);
   auto table_alias = table_ref->GetTableAlias();
   AddRegularTable(table_ref->GetDatabaseName(), table_ref->GetTableName(),
+                  table_ref->GetSessionNamespace(), table_ref->GetNamespace(),
                   table_alias, txn);
 }
 
 void BinderContext::AddRegularTable(const std::string db_name,
                                     const std::string table_name,
+                                    const std::string session_namespace,
```
danae-s: New arguments should be added last and assigned default values to keep previous tests working well.

src/catalog/database_catalog.cpp
```diff
@@ -25,8 +24,8 @@
 namespace peloton {
 namespace catalog {
 
-DatabaseCatalogObject::DatabaseCatalogObject(executor::LogicalTile *tile,
```
bdengNE: Should keep the format consistent among all files

src/catalog/table_catalog.cpp
```diff
@@ -364,10 +377,12 @@ bool TableCatalog::InsertTable(oid_t table_oid, const std::string &table_name,
   auto val0 = type::ValueFactory::GetIntegerValue(table_oid);
```
bdengNE: Variable names should be more meaningful.

src/include/catalog/trigger_catalog.h
```diff
@@ -63,6 +64,8 @@ class TriggerCatalog : public AbstractCatalog {
 
   ResultType DropTrigger(const std::string &database_name,
                          const std::string &table_name,
+                         const std::string &session_namespace,
```
danae-s: Same here with adding new arguments.

src/parser/postgresparser.cpp
```diff
@@ -196,7 +196,6 @@ bool IsTargetListWithVariable(List *target_list) {
       case T_BoolExpr:
         continue;
       default:
-        LOG_DEBUG("HERE");
```
bdengNE: should delete this log message

src/network/postgres_protocol_handler.cpp
```diff
@@ -899,6 +901,7 @@ void PostgresProtocolHandler::ExecCloseMessage(InputPacket *pkt) {
       // do nothing, simply send close complete
       break;
   }
+  LOG_DEBUG("CLOSE MESSAGE EXECUTED");
```
bdengNE: should delete this log message ?

src/optimizer/optimizer.cpp
apavlo: @HenryZhou0333 What do you mean?

src/catalog/database_catalog.cpp
```diff
@@ -53,17 +52,18 @@ bool DatabaseCatalogObject::InsertTableObject(
     return false;
   }
 
-  if (table_name_cache.find(table_object->GetTableName()) !=
-      table_name_cache.end()) {
+  // table key
+  std::string table_key = table_object->GetTableNamespace() + std::string("#") +
```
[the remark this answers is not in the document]
apavlo: Agreed. Maybe we should have a naming util class for tables and indexes?

src/catalog/table_catalog.cpp
apavlo: I disagree. `namespace` needs to come before `table_name`. It also should **not** have a default value.

src/include/catalog/trigger_catalog.h
apavlo: Make `namespace` come before `table_name`

test/parser/parser_test.cpp
apavlo: We should have C++ tests and Java tests for this.

src/catalog/table_catalog.cpp
apavlo: Agreed. This is very important. Please fix.

src/catalog/table_catalog.cpp
yjc408: The reason why I put namespace (and session name for some other commits) after all the arguments and with a default value is because that some tests will call the function directly without going through connection. In that case, namespace should be default and session name should be empty. Is that not appropriate? Should I make those two values without default values and explicitly set them in the tests?

apavlo:
I am closing this PR. The session namespace handling is too messy. We will replace it with `SessionContext` from #1397

PR #1287, opened 2018-04-12
[15721] Logging and Recovery

aaron-tian:
Currently support logging for inserts, deletes and updates.


src/logging/wal_logger.cpp
```diff
@@ -0,0 +1,101 @@
...
+//                         Peloton
+//
+// wal_logger.cpp
+//
+// Identification: src/logging/wal_logger.cpp
+//
+// Copyright (c) 2015-18, Carnegie Mellon University Database Group
+//
+//===----------------------------------------------------------------------===//
+
```
nwang57: Probably need tests to make sure that logger thread is actually writing logs to the disk before sending the commit and abort message back to the transaction.

src/logging/wal_logger.cpp
```diff
@@ -0,0 +1,101 @@
...
+}
+
+
+void WalLogger::FlushToDisk(){
+
+  if(disk_buffer_->GetSize()==0)
+    return;
+
+  std::ofstream *stream = LogManager::GetInstance().GetFileStream();
+  stream->write(disk_buffer_->GetData(), disk_buffer_->GetSize());
```
nwang57: Multiple logger threads connecting with the same output stream may result in race condition and corrupt the log files. Maybe consider locking the file while writing and flush.

src/include/logging/checkpoint_manager.h
```diff
@@ -1,69 +0,0 @@
-//===----------------------------------------------------------------------===//
```
Onmysofa: Is checkpointing included in this project?

src/include/logging/wal_log_manager.h
```diff
@@ -0,0 +1,73 @@
...
+  inline std::ofstream *GetFileStream() { return &logger_ofstream_; }
+
+//  TODO(gandeevan): support StartLogging and StopLogging
+//  static void StartLogging() {}
+//  static void StopLogging() {}
+
+
+private:
+
+  const static size_t logger_buffer_threshold_ = 1024 * 512;  // 512 KB
```
Onmysofa: I think these could be declared with other database toggles?

src/traffic_cop/traffic_cop.cpp
```diff
@@ -104,12 +104,13 @@ ResultType TrafficCop::CommitQueryHelper() {
...
   // 'ROLLBACK' After receive 'COMMIT', see if it is rollback or really commit.
+
   if (curr_state.second != ResultType::ABORTED) {
     // txn committed
-    return txn_manager.CommitTransaction(txn);
+    return txn_manager.CommitTransaction(txn, GetOnCompleteCallback());
   } else {
     // otherwise, rollback
-    return txn_manager.AbortTransaction(txn);
+    return txn_manager.AbortTransaction(txn, GetOnCompleteCallback());
```
Onmysofa: I'm wondering if a callback is necessary for abort.

apavlo:
How hard would it be to make `LoggingManager` not be a static singleton here?

aaron-tian:
I think it's doable and will take a look tonight. But I'm wondering why we want it not to be a static singleton? Do we need more than one `LoggingManager`?

src/include/logging/checkpoint_manager.h
aaron-tian: Yes, but we will take care of it in another PR.

src/include/logging/log_record.h
```diff
@@ -40,48 +42,105 @@ class LogRecord {
...
+//    PELOTON_ASSERT(log_type == LogRecordType::TRANSACTION_BEGIN ||
+//              log_type == LogRecordType::TRANSACTION_COMMIT);
+//    return LogRecord(log_type, INVALID_ITEMPOINTER, INVALID_EID, commit_id);
+//  }
+//
+//  static LogRecord CreateEpochRecord(const LogRecordType log_type, const eid_t epoch_id) {
+//    PELOTON_ASSERT(log_type == LogRecordType::EPOCH_BEGIN ||
+//              log_type == LogRecordType::EPOCH_END);
+//    return LogRecord(log_type, INVALID_ITEMPOINTER, epoch_id, INVALID_CID);
+//  }
```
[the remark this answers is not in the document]
aaron-tian: We are not using Epoch for logging currently.

src/include/logging/log_record.h
```diff
@@ -6,15 +6,15 @@
 //
 // Identification: src/logging/log_record.h
 //
-// Copyright (c) 2015-16, Carnegie Mellon University Database Group
```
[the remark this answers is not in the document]
aaron-tian: For historical reasons, it exists in the code but can not work as we expect. We can talk next time we meet if you want to know.

src/include/logging/log_record.h
```diff
@@ -40,48 +42,105 @@ class LogRecord {
...
-              log_type == LogRecordType::EPOCH_END);
-    return LogRecord(log_type, INVALID_ITEMPOINTER, epoch_id, INVALID_CID);
+  static LogRecord CreateTupleRecord(const LogRecordType log_type, const ItemPointer &old_pos,
+                                     const ItemPointer &pos, eid_t current_eid,
+                                     txn_id_t txn_id, cid_t current_cid) {
+
+    PELOTON_ASSERT(log_type == LogRecordType::TUPLE_UPDATE);
+
+    return LogRecord(log_type, old_pos, pos, current_eid, txn_id, current_cid);
   }
```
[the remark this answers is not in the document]
apavlo: @nwang57 Excellent point.

@aaron-tian You should look into this.

src/concurrency/timestamp_ordering_transaction_manager.cpp
```diff
@@ -544,6 +573,33 @@ void TimestampOrderingTransactionManager::PerformUpdate(
   // Add the old tuple into the update set
   current_txn->RecordUpdate(old_location);
 
+  if(logging::LogManager::GetInstance().IsLoggingEnabled()) {
+    if (values_buf != nullptr) {
```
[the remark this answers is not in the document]
apavlo: Also, fix formatting

src/include/logging/wal_log_manager.h
apavlo: @Onmysofa is correct. We should move this into `pg_settings`

src/traffic_cop/traffic_cop.cpp
apavlo: Yes, we changed it do this.

test/logging/log_buffer_test.cpp
```diff
@@ -23,56 +26,136 @@ namespace test {
...
+  schemas.push_back(*schema1);
+
+  std::map<oid_t, std::pair<oid_t, oid_t>> column_map;
+  column_map[0] = std::make_pair(0, 0);
+
+  std::shared_ptr<storage::TileGroup> tile_group(storage::TileGroupFactory::GetTileGroup(
+        database_id, table_id, tile_group_id, nullptr, schemas, column_map, 3));
+  catalog::Manager::GetInstance().AddTileGroup(block, tile_group);
+
+  logging::LogRecord insert_record = logging::LogRecordFactory::CreateTupleRecord(
```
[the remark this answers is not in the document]
aaron-tian: It makes more sense to do it in the logging_test. In log_buffer_test, we focus on the functionality of  log_buffer, different operations just mean different operation_id and operation formats, which will not influence how the buffer works. It's good for logic classification though. I will do it.

src/include/logging/log_record.h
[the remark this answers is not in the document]
aaron-tian: Sounds good. I will change it.

src/include/logging/wal_log_manager.h
aaron-tian: Yes, we will move it as well as some others into `pg_settings` in the end.

src/concurrency/timestamp_ordering_transaction_manager.cpp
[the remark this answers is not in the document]
latelatif: There can be cases when the buffer is still nullptr as we have not added logging to the old engine. This function can still be called when codegen is not being used and in those cases, values_buf might be null. An assert will lead to the code asserting every time the old engine is used with logging turned on.

src/include/concurrency/timestamp_ordering_transaction_manager.h
```diff
@@ -206,7 +214,10 @@ class TimestampOrderingTransactionManager : public TransactionManager {
    * @param[in]  location     The location
    */
   virtual void PerformUpdate(TransactionContext *const current_txn,
-                             const ItemPointer &location);
+                             const ItemPointer &location,
+                             char *values_buf = nullptr,
+                             uint32_t values_size = 0,
+                             TargetList *offsets = nullptr);
```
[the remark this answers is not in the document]
latelatif: One does an in-place update while other one creates a new version. It is based on whether the given transaction already has a write lock on it. I think the reason is that If we modify a tuple twice in the same transaction, we can just overwrite the tuple in-place instead of creating a new version.

src/include/logging/wal_log_manager.h
latelatif: Done

src/include/logging/log_record.h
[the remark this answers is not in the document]
latelatif: Done

src/logging/wal_logger.cpp
latelatif: The plan as of now and in the foreseeable future is to just use one logger thread. The locking changes can be made as a part of another PR. For this PR, we are only using one logger thread.

src/include/threadpool/logger_queue_pool.h
```diff
@@ -0,0 +1,107 @@
...
+* One should use this if possible.
+*/
+class LoggerQueuePool {
+  public:
+    LoggerQueuePool(size_t num_workers)
+            : logger_queue_(kDefaultLoggerQueueSize),
+              num_workers_(num_workers),
+              is_running_(false) {
+
+      logger_queue_.GenerateTokens(log_tokens_, kDefaultNumTokens);
```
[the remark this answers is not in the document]
latelatif: Done

Issue #1300, opened 2018-04-15
IndexScanDesc and IndexScanPlan Refactor


malin1993ml:
We now store a physical pointer of the used index in the `IndexScanDesc` and `IndexScanPlan`. This won't work for "what-if" API because the "fake" index will only have a catalog entry and an OID, but not a pointer to a physical object. The optimizer will crash when generating the `IndexScanPlan` using "what-if" API.

The refactor needs to take the pointer out of the `IndexScanDesc` and `IndexScanPlan`, and replace them with the index OID.

apavlo:
@malin1993ml This is fixed in #1301, right?

malin1993ml:
This issue is fixed in #1301 so I'm closing it. I would add a final comment that we're doing bunch of initialization work in `IndexScanExecutor::DInit()` right now, including generating the `IndexScanPredicate`. We need to do similar work in the LLVM engine as well when we're suppporting index scan in the LLVM engine.

PR #1301, opened 2018-04-15
IndexScanDesc and IndexScanPlan Refactoring

malin1993ml:
This PR fixes the issue in #1300. It takes the pointer out of the `IndexScanDesc` and `IndexScanPlan`, and replace them with the index OID. The `IndexScanExecutor` will get the physical pointer to the index object in its `DInit()` function.

This PR also adjusts the calls to the `IndexScanDesc` constructor in many test cases.


src/planner/index_scan_plan.cpp
```diff
@@ -55,8 +55,8 @@ IndexScanPlan::IndexScanPlan(storage::DataTable *table,
   //
   // Values that are left blank will be recorded for future binding
   // and their offset inside the value array will be remembered
-  index_predicate_.AddConjunctionScanPredicate(index_.get(), values_,
-                                               key_column_ids_, expr_types_);
+  //index_predicate_.AddConjunctionScanPredicate(index_.get(), values_,
+  //                                             key_column_ids_, expr_types_);
```
sivaprasadsudhir: Why don't we delete these lines?

malin1993ml: Goot catch! I forgot to do delete them. I'll do it.

vkonagar: If we are shifting this from the planner to the execution engine, we won't be able to see the details of these predicates at the optimizer side, right? The plan returned by the optimizer to 'What-if index' will not have the required index plan details?

apavlo:

PR #1248, opened 2018-03-30
Workload Prediction

saatviks:
This is the work on reproducing @malin1993ml 's paper on Workload prediction within Peloton. It adds the following:
1. The Eigen3 Linear Algebra library and some utilities to work with it. 
2. Slight reorganization of all the different Tensorflow utility functions.
3. Rewriting of the Tensorflow `LSTM` python code in a cleaner more intuitive way.
4. A test which attempts to predict a sine wave like workload.

PR #1303, opened 2018-04-16
Fix for python-six upgrade on Ubuntu 14.04

pervazea:


malin1993ml:
@pervazea I think #1248 also fixes this issue by doing this:
https://github.com/cmu-db/peloton/pull/1248/files#diff-46454d00238f98e83096b2d44703df7eR76

That PR is very close to be ready except a minor namespace closure thing. I'm not sure which way is better though.

malin1993ml:
@pervazea Yes, we're using Python3.

@saatviks Can you answer the third one, and adjust your PR accordingly if it does not uninstall the old version?

pervazea:
Thanks.


You should be able to easily tell by looking in

/usr/lib/python3/dist-packages

ls -l six*

should show you only the updated version of six, if it removed the old one.

saatviks:
@pervazea the new PR is installing tensorflow as `pip3 install tensorflow==${TF_VERSION} --upgrade --ignore-installed six`. This way even in the future if `python3-six` is already installed it will not install  or update it. Does this look alright?

saatviks:
Yes your right I had misunderstood what it meant. We can then replace my fix with this PR instead. I dont think you need to check the version during `remove` - You can directly run:
`sudo apt-get remove --purge python3-six`?

pervazea:
Yes, your fix below sounds good.

Issue #1299, opened 2018-04-14
Invalid Index Scan Plan Selection with Multi-column Indexes


vkonagar:
As part of our class project "Index Selection", we are trying to generate multi-column hypothetical indexes and checking the cost of using them for a query. 

The optimizer does not select a valid plan with multi-column indexes. If we create a multiple-column index say Index1 on columns say (a, b, c) on a table T, and if a query has a where clause on only (b), then the optimizer chooses Index1 in its best plan for the query. Same goes with (c) as well. Ideally, Index1 should only help queries with (a) or (a, b) or (a, b, c) as the where clause predicates. Surprisingly, even with the wrong plan, the query executed successfully.

Steps to reproduce:
- CREATE table T(a int, b int, c int, d int, e int);
- INSERT into T VALUES(1, 2, 3, 4, 5);
- INSERT into T VALUES(2, 3, 4, 5, 6);
- INSERT into T VALUES(3, 4, 5, 6, 7);
- INSERT into T VALUES(4, 5, 6, 7, 8);
- INSERT into T VALUES(5, 6, 7, 8, 9);
- CREATE INDEX Index1 on T(a, c, d, e);
- EXPLAIN SELECT * FROM T where e = 8;
                          Query plan
    Plan Type: INDEXSCAN
    Info: IndexScan
    NumChildren: 0
(5 rows)
- SELECT * FROM T where e = 8;
 a | b | c | d | e
---+---+---+---+---
 4 | 5 | 6 | 7 | 8
(1 row)

GustavoAngulo:
https://github.com/cmu-db/peloton/blob/78b85dc21d5f1d7b09b150380ecec22b92ea804e/src/optimizer/rule_impls.cpp#L241-L253

This is probably the cause of the issue. The `Check` function essentially checks if we can use an index scan for a "Get" operator. We're only checking the existence of an index, not if we can actually use then index. I can probably take a look.

vkonagar:
Gustavo, Thanks! I looked at the code you pointed and I think the issue is here.

https://github.com/cmu-db/peloton/blob/78b85dc21d5f1d7b09b150380ecec22b92ea804e/src/optimizer/rule_impls.cpp#L394

Here, we are just checking if the given column is present in the column set of the index object but not enforcing any order in the matches. Instead, we should take the vector of column oids present in index catalog object (this preserves multi-column order as specified by the user) and check in the given order if there is any match with the columns given in the query predicates.

Something like this?


[...]

PR #1305, opened 2018-04-17
Multi-column IndexScan plan selection fix

vkonagar:
This PR fixes the issue https://github.com/cmu-db/peloton/issues/1299. This changes the way we find index match for predicate columns in IndexScan rule implementation. Specifically, this change makes sure that the optimizer picks a multi-column index only if the predicate columns match the index columns in order.

### DIFF
--- a/src/optimizer/rule_impls.cpp
+++ b/src/optimizer/rule_impls.cpp
@@ -6,7 +6,7 @@
 //
 // Identification: src/optimizer/rule_impls.cpp
 //
-// Copyright (c) 2015-16, Carnegie Mellon University Database Group
+// Copyright (c) 2015-2018, Carnegie Mellon University Database Group
 //
 //===----------------------------------------------------------------------===//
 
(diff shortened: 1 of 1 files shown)


apavlo:
@vkonagar Can you provide test cases for this to know what you fixed? Thanks!

vkonagar:
@apavlo Andy, I have added a test to verify the query plan correctness with respect to multi-column indexes. 

This doesn't fix the cost model for multi-column indexes, which is not currently supported in the optimizer. I have talked to bowei and we will look into that.

chenboy:
As discussed in today's meeting, we want to fix the cost model to consider multi-column indices. Let me see if I can fix it. @GustavoAngulo @nappelson I'm wondering if we have a testing infrastructure for cost model correctness right now?

src/optimizer/rule_impls.cpp
```diff
@@ -386,17 +386,22 @@ void GetToIndexScan::Transform(
...
+      // the same order.
+      auto index_id_list = index_object->GetKeyAttrs();
+      for (size_t offset = 0; (offset < key_column_id_list.size()) &&
+                              (offset < index_id_list.size());
+           offset++) {
+        if (index_id_list[offset] == key_column_id_list[offset]) {
+          index_key_column_id_list.push_back(key_column_id_list[offset]);
           index_expr_type_list.push_back(expr_type_list[offset]);
           index_value_list.push_back(value_list[offset]);
+        } else {
```
malin1993ml: I don't think we should check for an exact same ordering here. For example, if you have an index on column (a, b), and your predicates are "b = 5 and a = 1", then we should be able to use the index scan. However, the check here won't identify that because it requires the order in the predicates to be exactly the same as in the index.

After thinking about this, I actually think that you should just keep the old `index_key_column_id_list`. You just need to add a flag about whether the lead (highest) column in the index has been referenced in the index. As long as that is true, we should be able to use the index for the scan. Thoughts? @chenboy @vkonagar

chenboy: Agree. I also think we don't need to consider order here. The way to fix this issue is letting the cost model compute the correct cost for these indices.

malin1993ml: Hi, @chenboy! @vkonagar and I are discussing some implementation details for this on Slack. We've added you into the channel. There's some cost model related issue we think you probably have a better idea on what's going on. Can you take a look at Slack? Thanks!

apavlo:
This is another important fix that we are going to need for TPC-C.

PR #1310, opened 2018-04-18
Checkpointing

aaron-tian:
Checkpointing as part of the resistance mechanism.


aaron-tian:
@ksaito7 Tests are failing due to newly added files (e. g. -Werror=maybe-uninitialized). Could you please take a look at them?

test/logging/timestamp_checkpointing_test.cpp
```diff
@@ -0,0 +1,161 @@
...
+
+  EXPECT_TRUE(checkpoint_manager.GetStatus());
+
+  std::this_thread::sleep_for(std::chrono::seconds(3));
+  checkpoint_manager.StopCheckpointing();
+
+  TestingSQLUtil::ExecuteSQLQuery("COMMIT;");
+
+  EXPECT_FALSE(checkpoint_manager.GetStatus());
+
```
aaron-tian: It would be better if we can read the file back and see if it is what we expect.

src/logging/timestamp_checkpoint_manager.cpp
```diff
@@ -0,0 +1,1084 @@
...
+        value.SerializeTo(output_buffer);
+        LOG_TRACE("%s(column %d, tuple %d):%s\n", table->GetName().c_str(),
+                  column_id, tuple_id, value.ToString().c_str());
+      }
+    }
+
+    // write down tuple data to file
+    int ret = fwrite((void *)output_buffer.Data(), output_buffer.Size(), 1,
+                     file_handle.file);
+    if (ret != 1 && ret != 0) {
```
aaron-tian: Why do we need (ret != 0)? We should give an error when ret == 0?

src/logging/timestamp_checkpoint_manager.cpp
```diff
@@ -0,0 +1,1084 @@
...
+    for (oid_t tg_offset = 0; tg_offset < table->GetTileGroupCount();
+         tg_offset++) {
+      auto tile_group = table->GetTileGroup(tg_offset);
+      auto max_tuple_count = tile_group->GetNextTupleSlot();
+      for (oid_t tuple_id = 0; tuple_id < max_tuple_count; tuple_id++) {
+        // check all primary key columns
+        bool check_all_pk_values_same = true;
+        for (auto pk_column : pk_columns) {
+          if (tile_group->GetValue(tuple_id, pk_column)
+                  .CompareNotEquals(tuple->GetValue(pk_column)) ==
```
aaron-tian: Should the duplicate check iterates tuples in the current table and compare the tuple recovered from the file with each of them for primary key columns? The tuple recovered shouldn't be stored in the table as long as there is one current tuple whose pk columns match its corresponding ones. It doesn't necessarily have to be no current tuples mismatching.

src/logging/timestamp_checkpoint_manager.cpp
```diff
@@ -0,0 +1,1084 @@
...
+        auto column_itr = columns.begin();
+        for (oid_t idx_count = START_OID; idx_count < column_oid; idx_count++) {
+        	if (column_itr == columns.end() ||
+        			column_itr->GetOffset() > column.GetOffset()) {
+        		break;
+        	} else {
+        		column_itr++;
+        	}
+        }
+      	columns.insert(column_itr, column);
```
aaron-tian: Are we skipping any specific columns? I'm wondering why we can't just insert by `push_back` now.

test/logging/timestamp_checkpoint_recovery_test.cpp
```diff
@@ -0,0 +1,600 @@
...
+                                        "9|10|11|12|2|11|12"};
+  TestingSQLUtil::ExecuteSQLQueryAndCheckResult(sql3, expected3, false);
+
+  // make sure the constraints are working
+  // PRIMARY KEY (1 column: pid)
+  LOG_INFO("PRIMARY KEY (1 column) check");
+  std::string primary_key_dml1 =
+      "INSERT INTO checkpoint_table_test VALUES (0, 5.5, 'eee');";
+  ResultType primary_key_result1 =
+      TestingSQLUtil::ExecuteSQLQuery(primary_key_dml1);
```
aaron-tian: All the tests before this constraint `PRIMARY KEY (1 column) check` work well while a `Segmentation fault (core dumped)` is encountered here.

src/include/logging/timestamp_checkpoint_manager.h
```diff
@@ -0,0 +1,295 @@
...
+
+    // Get list of checkpoint directories
+    if (LoggingUtil::GetDirectoryList(checkpoint_base_dir_.c_str(),
+                                      dir_name_list) == false) {
+      LOG_ERROR("Failed to get directory list in %s",
+                checkpoint_base_dir_.c_str());
+    }
+
+    // Remove old checkpoint directory
+    for (auto dir_name : dir_name_list) {
```
aaron-tian: `dir_name` is only read in this case. Make it a const reference to save copy overheads.

src/logging/timestamp_checkpoint_manager.cpp
```diff
@@ -0,0 +1,1084 @@
...
+    return false;
+  }
+
+  if (tuple_txn_id == INITIAL_TXN_ID) {
+    // this tuple is not owned by any other transaction.
+    if (activated && !invalidated) {
+      return true;
+    } else {
+      return false;
+    }
```
aaron-tian: Simplify the if-else statement to `return activated && !invalidated;`

src/logging/timestamp_checkpoint_manager.cpp
```diff
@@ -0,0 +1,1084 @@
...
+    // this tuple is owned by othre transactions.
+    if (tuple_begin_cid == MAX_CID) {
+      // this tuple is an uncommitted version.
+      return false;
+    } else {
+      if (activated && !invalidated) {
+        return true;
+      } else {
+        return false;
+      }
```
aaron-tian: Same as above. Simplify the if-else statement to `return activated && !invalidated;`

src/logging/timestamp_checkpoint_manager.cpp
ksaito7: To reduce processes of the check, if a value in the existed table's tuple in a primary key column differs from a recovered value, then the loop of the pk loop would break. This means it finds these tuples mismatching, and reduces the check processes of the rest of primary keys.

src/logging/timestamp_checkpoint_manager.cpp
ksaito7: Here is that it arranges the order of columns for column storage objects. Because column catalog objects obtained from column catalog are not ordered by column id. And the column storage object is managed column id by the vector iterator. This means there is no column id in column storage object. This is the reason why we doesn't use push_back. To use push_back, we need to implement the column id into column storage object.

test/logging/timestamp_checkpoint_recovery_test.cpp
```diff
@@ -0,0 +1,600 @@
...
+  // make sure data structures created in checkpointing test and recovered
+  // above are correct.
+  auto &txn_manager = concurrency::TransactionManagerFactory::GetInstance();
+  auto txn = txn_manager.BeginTransaction();
+  auto catalog = catalog::Catalog::GetInstance();
+  auto storage = storage::StorageManager::GetInstance();
+
+  // check an uncommitted table does not exist
+  EXPECT_FALSE(
+      catalog->ExistTableByName(DEFAULT_DB_NAME, "out_of_checkpoint", txn));
```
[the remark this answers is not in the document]
ksaito7: OK, I will fix it.

src/logging/timestamp_checkpoint_manager.cpp
ksaito7: OK, I will fix it.

test/logging/timestamp_checkpointing_test.cpp
ksaito7: I added the test case to make sure the created table information and the inserted values are correct.

test/logging/timestamp_checkpoint_recovery_test.cpp
ksaito7: This error occurs because `active_tile_groups` in DataTable object is not set correctly when tile groups are recovered.  The `active_tile_groups` is used by data insertion to find a tile group where the insertion can write the data. The number of tile groups in `active_tile_groups` depends on the number of CPU cores. (This is a reason why my local environment didn't make this error. If the number is just one, then the error didn't occur.) If there are multiple tile groups in `active_tile_groups`, a new tuple is inserted into a tile group decided by the number of existed tuples in the table like below:

```
  size_t active_tile_group_id = number_of_tuples_ % active_tilegroup_count_;
  tile_group = active_tile_groups_[active_tile_group_id];
  tuple_slot = tile_group->InsertTuple(tuple);
```

A new tile group is also set into `active_tile_groups` by the number of existed tuples. In recovery, when the second tile group is recovered, the recovered table has already had some tuples in the first recovered tile group. In this time, a place in `active_tile_groups` where the second tile group will be set might be in the same place where the first tile group is set. 

In Jenkins, the first tile group in `active_tile_groups` was overwritten by the second tile group. And the error occurred when the insertion tried to insert a tuple into another empty entry of `active_tile_groups` where the second tile group should exist originally.

I fixed the tile group insertion in our checkpoint recovery to avoid this overwriting.

ksaito7:
Some modifications are necessary due to `catalog::LayoutCatalog` and `storage::Layout` added in current master branch.

ksaito7:
1. Change for `catalog::SystemCatalogs`
 - Use SystemCatalogs in catalog checkpointing and recovery in addition to catalogs in catalog DB.
 - Eliminate basic catalogs in catalog DB from checkpoint target to avoid an table collision when a catalog table changes.
 - Add bootstrapping of SystemCatalogs when each database storage object is recovered.

2. Change for `catalog::SchemaCatalog`
 - Add SchemaCatalog checkpointing and recovery
 - Fix to use schema name when getting TableCatalogObject
 - Use CATALOG_SCHEMA_NAME to distinguish between user table and catalog table.

3. Chagen for `catalog::LayoutCatalog` and `storage::Layout`
 - Add LayoutCatalog checkpointing and recovery
 - Add layout serialize/deserialize function for tile groups recovery

src/logging/timestamp_checkpoint_manager.cpp
ksaito7: I simplify this insertion method by vector iterator using [].

ksaito7:
AddressSanitizer error on only release build in travis is resolved by deleting DataTable GetInfo function call in recovery test. I didn't know eventually why this error was occurred because AddressSanitizer didn't output the place.

ksaito7:
Update for PR #1383

PR #1311, opened 2018-04-19
Jenkins update to gcc5

pervazea:
Also update cmake to use c++11


pervazea:
Same intermittent memory leak failure on the query_logger_test. Otherwise passed on Jenkins.
I'm going to re-run the Jenkins build, just to see how reproducible the query_logger_test issue is.

pmenon:
Actually, @pervazea do we also want to build and run the test suite for macOS on Travis?

pervazea:
Test suite for macOS on Travis, sure. I'll do it as a follow up though. Get this in and unblock progress.

Issue #1295, opened 2018-04-12
Jenkins build / checkout failure


pervazea:
This is from PR-1278 (which is mine)

We have seen this possibly 3 times in the past week, on different PRs. It seems to require additional pushes to be done to the PR to get past this. 

```
hudson.plugins.git.GitException: Command "git checkout -f 2fc7e14cba60d473c06325e4c8501d9e81b62cf7" returned status code 128:
stdout: 
stderr: fatal: reference is not a tree: 2fc7e14cba60d473c06325e4c8501d9e81b62cf7

	at org.jenkinsci.plugins.gitclient.CliGitAPIImpl.launchCommandIn(CliGitAPIImpl.java:1996)
	at org.jenkinsci.plugins.gitclient.CliGitAPIImpl.access$800(CliGitAPIImpl.java:72)
	at org.jenkinsci.plugins.gitclient.CliGitAPIImpl$9.execute(CliGitAPIImpl.java:2295)
Caused: hudson.plugins.git.GitException: Could not checkout 2fc7e14cba60d473c06325e4c8501d9e81b62cf7
	at org.jenkinsci.plugins.gitclient.CliGitAPIImpl$9.execute(CliGitAPIImpl.java:2319)
	at jenkins.plugins.git.MergeWithGitSCMExtension.checkout(MergeWithGitSCMExtension.java:146)
	at jenkins.plugins.git.MergeWithGitSCMExtension.decorateRevisionToBuild(MergeWithGitSCMExtension.java:112)
	at hudson.plugins.git.GitSCM.determineRevisionToBuild(GitSCM.java:1068)
	at hudson.plugins.git.GitSCM.checkout(GitSCM.java:1161)
	at org.jenkinsci.plugins.workflow.steps.scm.SCMStep.checkout(SCMStep.java:113)
	at org.jenkinsci.plugins.workflow.cps.CpsScmFlowDefinition.create(CpsScmFlowDefinition.java:143)
	at org.jenkinsci.plugins.workflow.multibranch.SCMBinder.create(SCMBinder.java:120)
	at org.jenkinsci.plugins.workflow.job.WorkflowRun.run(WorkflowRun.java:295)
	at hudson.model.ResourceController.execute(ResourceController.java:97)
	at hudson.model.Executor.run(Executor.java:429)
Finished: FAILURE
```

crd477:
This was related to just one of the worker nodes that didn't have the correct default network route. I've fixed it.
[...]

PR #1322, opened 2018-04-21
fix #1313  Remove read only isolation level

### LINKED ISSUE #1313: Remove IsolationLevelType::READ_ONLY
The way that we are marking that a txn is read-only is by setting it's isolation level to `READ_ONLY`.

https://github.com/cmu-db/peloton/blob/master/src/include/common/internal_types.h#L425

We currently don't support setting a txn to be read-only from the SQL interface, but the `TransactionManager` is checking for this `IsolationLevel` all over the place:

https://github.com/cmu-db/peloton/blob/master/src/concurrency/transaction_manager.cpp#L37

This is incorrect logic. `READ_ONLY` is not a valid isolation level. A txn can be different isolation levels and also read-only.

We need to remove  `IsolationLevelType::READ_ONLY` and then add a `readonly_` flag to the `TransactionContext` class. The `TransactionManager` should check this flag not the isolation type.

Nov11:
this is to #1313 

Details:
1.remove IsolationLevelType::READ_ONLY
2.add a flag in transactioncontext 
3.re-implement 'IsReadOnly' function. This is safe because the function has not been call anywhere before this modification.
4.substitute every usage of 'READ_ONLY' with IsReadOnly function call
5.add a new parameter to BeginTransaction and provide a short cut 'BeginTransactionReadOnly'
6.one invocation of BeginTransaction(READ_ONLY) is changed to 'BeginTransactionReadOnly(IsolationLevelType::READ_COMMITTED);'

### DIFF
--- a/src/concurrency/transaction_manager.cpp
+++ b/src/concurrency/transaction_manager.cpp
@@ -60,8 +60,8 @@ TransactionContext *TransactionManager::BeginTransaction(
     txn = new TransactionContext(thread_id, type, read_id);
   }
 
-  if(read_only){
-    txn->setReadOnly();
+  if (read_only) {
+    txn->SetReadOnly();
   }
 
   if (static_cast<StatsType>(settings::SettingsManager::GetInt(
--- a/src/concurrency/transaction_manager.cpp
+++ b/src/concurrency/transaction_manager.cpp
@@ -39,7 +39,7 @@ TransactionContext *TransactionManager::BeginTransaction(
     cid_t read_id = EpochManagerFactory::GetInstance().EnterEpoch(
         thread_id, TimestampType::SNAPSHOT_READ);
     txn = new TransactionContext(thread_id, type, read_id);
-
+    txn->setReadOnly();
   }
 
   if (type == IsolationLevelType::SNAPSHOT) {
--- a/src/concurrency/transaction_manager.cpp
+++ b/src/concurrency/transaction_manager.cpp
@@ -31,16 +31,18 @@ ConflictAvoidanceType TransactionManager::conflict_avoidance_ =
     ConflictAvoidanceType::ABORT;
 
 TransactionContext *TransactionManager::BeginTransaction(
-    const size_t thread_id, const IsolationLevelType type) {
+    const size_t thread_id, const IsolationLevelType type, bool read_only) {
   TransactionContext *txn = nullptr;
 
-  if (type == IsolationLevelType::READ_ONLY) {
+  if (read_only) {
     // transaction processing with decentralized epoch manager
     cid_t read_id = EpochManagerFactory::GetInstance().EnterEpoch(
         thread_id, TimestampType::SNAPSHOT_READ);
     txn = new TransactionContext(thread_id, type, read_id);
 
-  } else if (type == IsolationLevelType::SNAPSHOT) {
+  }
+
+  if (type == IsolationLevelType::SNAPSHOT) {
     // transaction processing with decentralized epoch manager
     // the DBMS must acquire
     cid_t read_id = EpochManagerFactory::GetInstance().EnterEpoch(
--- a/src/concurrency/transaction_manager.cpp
+++ b/src/concurrency/transaction_manager.cpp
@@ -34,14 +34,6 @@ TransactionContext *TransactionManager::BeginTransaction(
     const size_t thread_id, const IsolationLevelType type, bool read_only) {
   TransactionContext *txn = nullptr;
 
-  if (read_only) {
-    // transaction processing with decentralized epoch manager
-    cid_t read_id = EpochManagerFactory::GetInstance().EnterEpoch(
-        thread_id, TimestampType::SNAPSHOT_READ);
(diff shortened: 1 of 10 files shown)


poojanilangekar:
@Nov11 Your change is breaking the `serializable_transaction_test `. Please fix it and let me know so I can review the change.

Nov11:
It won't break tests now.
jenkins failed because of network error : 'stderr: fatal: unable to access 'https://github.com/cmu-db/peloton.git/': Failed to connect to github.com port 443: No route to host
'

crd477:
On 2018-04-21 13:08, Lin Ma wrote:
> Hi @crd477 <https://github.com/crd477> @pervazea 
> <https://github.com/pervazea>, it looks like some of our PRs including 
> this one cannot download the repo with the Jenkins build again. For example:
> http://jenkins.db.cs.cmu.edu:8080/blue/organizations/jenkins/peloton/detail/PR-1322/1/pipeline/
> http://jenkins.db.cs.cmu.edu:8080/blue/organizations/jenkins/peloton/detail/PR-1321/1/pipeline/
> http://jenkins.db.cs.cmu.edu:8080/blue/organizations/jenkins/peloton/detail/PR-1133/86/pipeline/
> http://jenkins.db.cs.cmu.edu:8080/blue/organizations/jenkins/peloton/detail/PR-1310/7/pipeline/
> 
> I'm wondering is this because we have another machine that has 
> networking problem? Would really appreciate your help.
> 

Ugh...

Yes, this was the default route problem again that originally cropped up 
in #1295.  I thought I'd fixed it the other day but obviously it 
happened again.  I've implemented a more permanent fix now and it should 
not happen again.

Sorry (again) for the inconvenience...

src/concurrency/transaction_manager.cpp
```diff
@@ -66,6 +60,10 @@ TransactionContext *TransactionManager::BeginTransaction(
     txn = new TransactionContext(thread_id, type, read_id);
   }
 
+  if(read_only){
```
apavlo: Fix spacing

src/include/concurrency/transaction_context.h
```diff
@@ -269,7 +269,15 @@ class TransactionContext : public Printable {
...
   inline bool IsReadOnly() const {
-    return is_written_ == false && insert_count_ == 0;
+    return read_only_;
+  }
+
+  /**
+   * @brief      mark this context as read only
+   *
+   */
+  inline void setReadOnly() {
```
apavlo: Remove inline

src/include/concurrency/transaction_context.h
```diff
@@ -269,7 +269,15 @@ class TransactionContext : public Printable {
    * @return     True if read only, False otherwise.
    */
   inline bool IsReadOnly() const {
```
apavlo: Remove inline

src/include/concurrency/transaction_context.h
```diff
@@ -341,6 +349,8 @@ class TransactionContext : public Printable {
   IsolationLevelType isolation_level_;
 
   std::unique_ptr<trigger::TriggerSet> on_commit_triggers_;
+
+  bool read_only_ = false;
```
apavlo: Add comment

src/include/concurrency/transaction_manager.h
```diff
@@ -219,11 +219,16 @@ class TransactionManager {
   }
 
   TransactionContext *BeginTransaction(const IsolationLevelType type) {
-    return BeginTransaction(0, type);
+    return BeginTransaction(0, type, false);
+  }
+
+  TransactionContext *BeginTransactionReadOnly(const IsolationLevelType type) {
```
apavlo: This is unnecessary bloat. I would remove this.

Nov11:
As there's already a read only transaction test prior to this change(SerializableTransactionTests->ReadOnlyTransactionTest), I thought it's adequate and didn't add new ones.
I broke that test case once.  poojanilangekar  reminded me of it. And it's been fixed.
If that test is not suffice to peloton, I'd like to do the addition.

apavlo:
@Nov11 What does the test do? We need to test the following scenario:

```
Txn #1 | Txn #2
-------------------
BEGIN  |
W(X)   |
       | BEGIN R/O
       | R(X)
W(X)   |
COMMIT | 
       | R(X)
       | COMMIT
```

apavlo:
@Nov11 Your comment about the broken test case looks to be correct. We want to test the scenario when there are two active txns running simultaneously. The code in `ReadOnlyTransactionTest` looks to have only one txn at a time.

Don't worry about any isolation level other than snapshot isolation.

Nov11:
There's a concurrent test case now.

poojanilangekar:
@apavlo @pervazea Can you please review this PR? 
@mengranwo has added support for change column name but we currently unable to test it because we need concurrent transactions to check the correctness.

apavlo:
I will look at this today.

apavlo:
I am going to merge this when the build passes. There are some stylistic things that I want to change but I am just going to fix that myself.

@Nov11 Thanks for your help!

PR #1323, opened 2018-04-21
[oltpbenchmark] Add scripts for updating website

lmwnshn:
Cleaned up and merged the bash scripts that are currently updating the website daily, into a single python script + init.sql script.

Some code duplication in config_xml_file with benchmark.py, though slightly different:

- we keep default weights
- support for upload code, different config paths

I haven't been able to find out why TPCC doesn't always report daily yet.


apavlo:
@pervazea These are the nightly testing scripts that we want to deploy.

script/oltpbenchmark/website/run.py
```diff
@@ -0,0 +1,198 @@
...
+        work.find('time').text = str(time)
+        work.find('rate').text = "unlimited"
+    outfile = config_out + benchmark + '_config.xml'
+    xml.write(outfile, encoding='UTF-8')
+    print("Wrote {}".format(outfile))
+
+
+def run_commands(commands):
+    for command in commands:
+        x = command.split(' ')
```
pervazea: This is ugly. Recommendation further down.

script/oltpbenchmark/website/run.py
```diff
@@ -0,0 +1,198 @@
...
+
+
+def peloton_build():
+    peloton_path = PELOTON_BASE_DIR
+    install_script_path = os.path.join(peloton_path, 'script', 'installation')
+    build_path = os.path.join(peloton_path, 'build')
+    bin_path = os.path.join(peloton_path, 'bin')
+
+    commands = [
+        'cd ' + peloton_path,
```
pervazea: I would convert the command list into a set of small functions (class methods).  The directories where things run shouldn't be explicitly stated for each step.
e.g.
def run_cmake_(self):
  os.chdir(self.build_path)
  ret_val = subprocess.call(... some command)
  check return value and report error if needed

script/oltpbenchmark/website/run.py
```diff
@@ -0,0 +1,198 @@
...
+
+    run_commands(commands)
+
+
+def peloton_init():
+    peloton_binary = os.path.join(PELOTON_BASE_DIR, 'build', 'bin', 'peloton')
+    script_dir = SCRIPT_DIR
+
+    commands = [
+        'cd ' + SCRIPT_DIR,
```
pervazea: and again here...

script/oltpbenchmark/website/run.py
```diff
@@ -0,0 +1,198 @@
...
+        'killall peloton',
+        peloton_binary + ' -port 15721 &',
+        'sleep 3',
+        'psql -U postgres -h localhost -p 15721 default_database < init.sql',
```
pervazea: 1. As per above comment, don't use init.sql. Construct the necessary sql commands and pass in here via the -c argument to psql.  

2. There are several places where the 15721 port number is hard coded in this script. It should not be. It should be a command line parameter, with a default value of 15721 if desired.  Pass it down through the layers.

3. Sleep as a synchronization mechanism works poorly and is best avoided unless there is no viable alternative. Here, It recommend lifting or adapting _wait_for_peloton from script/testing/junit/run_junit.py

4. killall is also best used only as a last resort. If you chose to run multiple copies of peloton, this would indiscriminately terminate all of them. Run peloton using subprocess, save the pid, then kill your process only. Again, see run_junit.py for code that can be re-used.

script/oltpbenchmark/website/run.py
```diff
@@ -0,0 +1,198 @@
...
+        'port': '15721',
+        'config_out': PELOTON_CONFIG_DIR,
+        'username': '',
+        'password': '',
+        'upload_code': ''
+    }
+
+    postgres = {
+        'host': 'localhost',
+        'port': '5432',
```
pervazea: This too should probably be configurable with a default.

apavlo:
@lmwnshn We should get this cleaned up and merged once you are finished this semester. I will send you an email about scheduling a time to meet.

lmwnshn:
A lot of changes have been made, still finalizing what we need. Will reopen when ready.

PR #1324, opened 2018-04-21
Delete the unsed index_predicate_ in index scan plans.

malin1993ml:
Apparently I forgot to delete the now unused index_predicate_ in IndexScanPlan and HybridScanPlan in #1301. This PR fixes that.

### DIFF
--- a/src/include/planner/index_scan_plan.h
+++ b/src/include/planner/index_scan_plan.h
@@ -16,11 +16,11 @@
 #include <string>
 #include <vector>
 
+#include "common/internal_types.h"
 #include "expression/abstract_expression.h"
 #include "index/scan_optimizer.h"
 #include "planner/abstract_scan_plan.h"
 #include "storage/tuple.h"
-#include "common/internal_types.h"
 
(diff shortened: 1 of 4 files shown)


src/include/planner/index_scan_plan.h
```diff
@@ -56,8 +56,7 @@ class IndexScanPlan : public AbstractScan {
      * be called to notify later procedures of the absense of an index
      */
     IndexScanDesc(
-        oid_t p_index_id,
-        const std::vector<oid_t> &p_tuple_column_id_list,
+        oid_t p_index_id, const std::vector<oid_t> &p_tuple_column_id_list,
```
pervazea: Is putting both declarations on the same line intentional? I didn't see anything in our coding style.
I would have expected that we prefer separate lines.

malin1993ml: The clang-format on my machine just automatically adjusted it that way... I run on-save formatting automatically. I've adjusted the format manually.

Issue #1307, opened 2018-04-18
query_logger_test failure


pervazea:
On Jenkins, has leaks when running under the address sanitizer
Seen in several recent builds. Extract of logs from PR-1271 appended.

This seems to be reporting that the leaks are from:
.../src/brain/query_logger.cpp:37

The symptom is as if the transaction is not committed,  but the code in this file does terminate the transaction. 

I am unable to reproduce this on my machine.

    Start  13: query_logger_test

13: Test command: /tmp/workspace/peloton_PR-1271-OTJTL66AGB2C26ZL6NIM4XVKJGQQTMZYVGTE2PEWOG5PXARBHSLQ/build/test/query_logger_test "--gtest_color=yes" "--gtest_output=xml:/tmp/workspace/peloton_PR-1271-OTJTL66AGB2C26ZL6NIM4XVKJGQQTMZYVGTE2PEWOG5PXARBHSLQ/build/test/query_logger_test.xml"
13: Environment variables: 
13:  LSAN_OPTIONS=suppressions=/tmp/workspace/peloton_PR-1271-OTJTL66AGB2C26ZL6NIM4XVKJGQQTMZYVGTE2PEWOG5PXARBHSLQ/test/leak_suppr.txt
13: Test timeout computed to be: 9.99988e+06
13: Running main() from gmock_main.cc
13: [0;32m[==========] [mRunning 1 test from 1 test case.
13: [0;32m[----------] [mGlobal test environment set-up.
13: [0;32m[----------] [m1 test from QueryLoggerTests
13: [0;32m[ RUN      ] [mQueryLoggerTests.QueriesTest
13: 2018-04-18 14:27:17 [test/sql/testing_sql_util.cpp:58:ExecuteSQLQuery] INFO  - Query: SELECT query_string, fingerprint FROM pg_catalog.pg_query_history;
13: 2018-04-18 14:27:18 [src/codegen/operator/table_scan_translator.cpp:40:TableScanTranslator] DEBUG - Constructing TableScanTranslator ...
13: 2018-04-18 14:27:18 [src/codegen/operator/table_scan_translator.cpp:53:TableScanTranslator] DEBUG - Finished constructing TableScanTranslator ...
13: 2018-04-18 14:27:18 [test/sql/testing_sql_util.cpp:93:ExecuteSQLQuery] INFO  - Statement executed. Result: SUCCESS
[...]

PR #1327, opened 2018-04-24
Schema - Layout decoupling 

poojanilangekar:
This PR decouples the logical schema from the physical layout of the data. Previously, each TileGroup contained a vector of schemas for which violates the notion of the storage layer being a physical store. We also had cases where the `LayoutTuner` modifies the `column_map` which ended up modifying the `schemas` vector  via the `DataTable`. There were multiple other instances where the `Schema` was used to get the layout of a `TileGroup` while all we needed was the physical mapping between column_ids and a <tile_id, column_offset> pair. This PR now makes it possible to make schema changes like changing column name, adding & dropping columns without making any calls to the storage layer (i.e., It can now be done entirely in the catalog). 

Changes:
- Added `Layout` and `LayoutCatalog` classes to store the layout in a physical object and the `pg_layout`.
- Added a `default_layotut_` object to each table (replacing `default_partition`) and `tile_group_layout_` to each `TileGroup`.
- Persist the `HYBRID` layouts to the `pg_catalog` table. The two pre-defined layouts, `ROW` and `COLUMN` stores are not persisted. 
- Added support in the LLVM engine to read the `Layout` on a per `TileGroup` while invoking the `TableScanTranslator`. 
- Modified the LLVM updater and inserter to check that we insert new versions only if the `TileGroup` has a `LayoutType::ROW` layout.
- Modified `LayoutTuner` to tune the `Layout` rather than the `column_map`. Additionally, the tuning decisions are now transactional.
- Added the `pg_catalog` to the `SystemsCatalog` to ensure that the each layout is stored in a catalog with respect its database.  
- Minor cleanup in the catalog (some typos and a couple of magic numbers which caused tests to fail).

Testing:
- Modified the `LayoutTuner` test to ensure it works as expected after the refactor. 
- Minor changes in the existing `TileGroupLayoutTest` and `SeqScanTest` which ensures that the old engine works as expected. 
- Added tests to the `TableScanTranslatorTests` to replicate the tests in `TileGroupLayoutTest` and `SeqScanTest`. The tests check that the LLVM engine can perform scans on tables different layouts and on tables where each `TileGroup` has a different `Layout`. 
- Added a test to `CatalogTests` to ensure that the `pg_layout` stores and retrieves data as expected and its state is transactional. 

Disclaimer:
- Each `Tile` still contains a `Schema` which will be used by the old engine, the `GetValue` and `SetValue` functions and (currently) the LLVM engines to determine the stride in case of a `HYBRID` layout. I believe we have to continue supporting the `GetValue` and `SetValue` functions even after the old engine is deprecated because we will be needing these interfaces in the `Catalog` and possibly the `Optimizer`. 
- In order to make the new layout change compatible with the `LogicalTile` generated by the executor engine, I had to reconstruct the `Schemas`. This isn't very efficient but the idea is that it is currently only used in tests and will soon be deprecated. 
- Unlike the other Catalogs (i.e., `ColumnCatalog`, `StatsCatalog`), the `LayoutCatalog` directly returns a `Layout` rather than a `LayoutCatalogObject` since the API are designed to be read only. In essence, the layout is immutable because if you were changing the layout, you need to change the physical representation. So it felt redundant to create a read only object to be returned by the catalog. 

@mengranwo Can you please review the catalog changes? 
@pmenon Can you please review the changes to the LLVM engine? Mainly, `RuntimeFunctions`, `Inserter`, `Updater`, `TestingCodegenUtil` and `TableScanTranslatorTests`. 
@pervazea Can you please review the overall changes? I have modified a few APIs in the old engine. The current code doesn't break the existing tests and it seems to function as expected. But please let me know if you think there is anything I need to change.
@jarulraj I know you may not have the bandwidth for this, but if possible could you please review the changes to `LayoutTuner` and `LayoutTunerTest`. Nobody else has modified that code and it would be great if you could take a look.


src/include/catalog/table_catalog.h
```diff
@@ -73,6 +79,15 @@ class TableCatalogObject {
   std::shared_ptr<ColumnCatalogObject> GetColumnObject(
       const std::string &column_name, bool cached_only = false);
 
+  // Evict all layouts from the cache
```
mengranwo: I may remember this wrong, but last time we talked you were saying add an extra column in pg_table to record the largest layout_id for each data table.

src/catalog/catalog.cpp
```diff
@@ -180,6 +181,9 @@ void Catalog::BootstrapSystemCatalogs(storage::Database *database,
   system_catalogs->GetTableCatalog()->InsertTable(
       COLUMN_CATALOG_OID, COLUMN_CATALOG_NAME, CATALOG_SCHEMA_NAME,
       database_oid, pool_.get(), txn);
+  system_catalogs->GetTableCatalog()->InsertTable(
```
mengranwo: When we bootstrap catalog tables(e.g pg_table, pg_index, etc..), did you handle the logic to insert corresponding record into pg_layout table?

test/catalog/catalog_test.cpp
```diff
@@ -303,5 +304,105 @@ TEST_F(CatalogTests, DroppingCatalog) {
...
+                              type::Type::GetTypeSize(type::TypeId::INTEGER),
+                              "val3", true);
+  std::unique_ptr<catalog::Schema> table_schema(
+      new catalog::Schema({val0, val1, val2, val3}));
+  catalog->CreateTable(db_name, DEFAULT_SCHEMA_NAME, table_name,
+                       std::move(table_schema), txn);
+  txn_manager.CommitTransaction(txn);
+
+  txn = txn_manager.BeginTransaction();
+  auto table =
```
mengranwo: Try to use **table catalog object** instead of **data table** to obtain information like database oid and table oid.

mengranwo:
@poojanilangekar 
Hi, this PR looks good for me except for some minor changes involved and some questions I posted in the comment area. I mainly focus on the logic of catalog & its corresponding test cases and I also take a look at the layout.h/layout.cpp, it stands.

src/catalog/catalog.cpp
```diff
@@ -555,6 +561,63 @@ ResultType Catalog::CreateIndex(
...
+  // If the layout creation was successful, set it as the default
+  if (new_layout != nullptr) {
+    auto storage_manager = storage::StorageManager::GetInstance();
+    auto database = storage_manager->GetDatabaseWithOid(database_oid);
+    auto table = database->GetTableWithOid(table_oid);
+    table->SetDefaultLayout(new_layout);
+  }
+  return new_layout;
+}
+
```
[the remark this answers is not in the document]
mengranwo: And from my understanding, only **HYBRID** layout type will be persist in pg_layout catalog table, right?

test/catalog/catalog_test.cpp
```diff
@@ -303,5 +304,105 @@ TEST_F(CatalogTests, DroppingCatalog) {
...
+      catalog->CreateLayout(database_oid, table_oid, non_default_map, txn);
+  txn_manager.CommitTransaction(txn);
+
+  // Check that the default layout is still the same.
+  EXPECT_NE(*other_layout.get(), table->GetDefaultLayout());
+
+  // Drop the default layout.
+  auto default_layout_oid = default_layout->GetOid();
+  txn = txn_manager.BeginTransaction();
+  EXPECT_EQ(ResultType::SUCCESS, catalog->DropLayout(database_oid, table_oid,
```
mengranwo: You said Persist the **HYBRID** layouts to the pg_catalog table. The two pre-defined layouts, **ROW** and **COLUMN** stores are not persisted. 
I may understand this wrong, the word ** Persist** means store this information in catalog table, right? If that is true, why `DropLayout(default_layout_oid)` returns SUCCESS? Since `default_layout` is **ROW** stored, it should not be inserted into pg_layout?

test/catalog/catalog_test.cpp
```diff
@@ -234,23 +235,23 @@ TEST_F(CatalogTests, DroppingTable) {
   auto &txn_manager = concurrency::TransactionManagerFactory::GetInstance();
   auto txn = txn_manager.BeginTransaction();
   auto catalog = catalog::Catalog::GetInstance();
-  // NOTE: everytime we create a database, there will be 8 catalog tables inside
+  // NOTE: everytime we create a database, there will be 9 catalog tables inside
   EXPECT_EQ(
-      11,
+      CATALOG_TABLES_COUNT + 3,
```
pervazea: The +3 hardcoded number... improvement that now using CATALOG_TABLES_COUNT. Can we fix it all the way by using define constant instead of the 3 or updating CATALOG_TABLES_COUNT if that is the right solution?

test/tuning/layout_tuner_test.cpp
```diff
@@ -6,7 +6,7 @@
 //
 // Identification: test/tuning/layout_tuner_test.cpp
 //
-// Copyright (c) 2015-16, Carnegie Mellon University Database Group
+// Copyright (c) 2015-16, Carnegie Mellon University Databa e Group
```
pervazea: Accidental deletion of s.

src/include/catalog/layout_catalog.h
```diff
@@ -0,0 +1,95 @@
...
+
+namespace storage {
+class Layout;
+}  // namespace storage
+
+namespace catalog {
+
+class LayoutCatalog : public AbstractCatalog {
+ public:
+  // Global Singleton, only the first call requires passing parameters.
```
pervazea: Aren't we trying to eliminate our use of singletons?

src/tuning/layout_tuner.cpp
```diff
@@ -142,12 +106,26 @@ void LayoutTuner::UpdateDefaultPartition(storage::DataTable* table) {
...
-  table->SetDefaultLayout(layout);
+  // Since we need the layout to be updated in the catalog,
+  // Start a transaction before updating the default layout.
+  auto &txn_manager = concurrency::TransactionManagerFactory::GetInstance();
+  auto *txn = txn_manager.BeginTransaction();
+  auto catalog = catalog::Catalog::GetInstance();
+  if (catalog->CreateDefaultLayout(database_oid, table_oid, column_map, txn) ==
+      nullptr) {
+    txn_manager.AbortTransaction(txn);
+    LOG_DEBUG("Layout Update to failed.");
```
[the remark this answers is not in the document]
poojanilangekar: Actually, this is going to be a part of the `brain` code to `Tune` the table layout. A transaction may fail because it conflicts with other concurrent updates. In an ideal case it must fail because tuning is a background task while the other transaction would be user specified. Since this is a background process, in case of failure, it would come around in the next iteration and attempt the change again. 
I am not sure if it would be appropriate to throw an exception when the failure was due to an internal event rather than a user action. What do you think?

src/catalog/table_catalog.cpp
```diff
@@ -330,6 +331,88 @@ TableCatalog::TableCatalog(
...
+  }
+
+  oid_t layout_id = layout->GetOid();
+  // layout is already present in the cache.
+  if (layout_objects_.find(layout_id) != layout_objects_.end()) {
+    LOG_DEBUG("Layout %u already exists in cache!", layout_id);
+    return false;
+  }
+
+  valid_layout_objects_ = true;
```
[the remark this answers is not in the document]
mengranwo: From our logic within `TableCatalogObject`, we only the parameter `valid_table_objects_` when **ALL** the table objects are inserted into cache.

test/catalog/catalog_test.cpp
mengranwo: Yeah, it makes sense now. Thanks for your explanation

src/catalog/table_catalog.cpp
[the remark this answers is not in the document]
poojanilangekar: Done

src/codegen/updater.cpp
```diff
@@ -44,10 +44,11 @@ void Updater::Init(storage::DataTable *table,
...
   auto tile_group = table_->GetTileGroupById(tile_group_id);
 
-  // Get the tile offset assuming that it is still in a tuple format
-  oid_t tile_offset, tile_column_offset;
-  tile_group->LocateTileAndColumn(0, tile_offset, tile_column_offset);
-  tile_ = tile_group->GetTileReference(tile_offset);
+  // Get the tile offset assuming that it is still a row store
+  // Hence the Tile offset is 0.
+  auto layout = tile_group->GetLayout();
+  PELOTON_ASSERT(layout.IsRowStore());
```
pmenon: 1. You probably want to make this a `const` reference to avoid the copy.
2. You'll need to add `UNUSED_ATTRIBUTE` to pass Release builds.

src/include/catalog/table_catalog.h
```diff
@@ -36,6 +36,11 @@
 #include "executor/logical_tile.h"
 
 namespace peloton {
+
+namespace storage {
+class Layout;
+}
```
pmenon: Close namespace.

src/include/codegen/runtime_functions.h
```diff
@@ -17,6 +17,10 @@
 
 namespace peloton {
 
+namespace catalog {
+class Schema;
+}
```
pmenon: Close namespace.

src/include/common/internal_types.h
```diff
@@ -1396,6 +1397,20 @@ typedef std::unordered_set<expression::AbstractExpression *,
...
+
+/* column_map_type used to store the layout of a tile_group
+ * <column offset> to <tile offset, tile column offset>
+ */
+typedef std::map<oid_t, std::pair<oid_t, oid_t>> column_map_type;
+
+/* tile_map_type used to store the mapping between a tile and its columns
+ * <tile index> to vector{<original column index, tile column offset>}
+ */
+typedef std::map<oid_t, std::vector<std::pair<oid_t, oid_t>>> tile_map_type;
```
pmenon: I'm not sold that these are common enough types for the whole system that they should be here ... but I guess it's okay.

src/include/storage/data_table.h
```diff
@@ -247,9 +249,17 @@ class DataTable : public AbstractTable {
 
   void ClearLayoutSamples();
 
-  void SetDefaultLayout(const column_map_type &layout);
+  inline void SetDefaultLayout(std::shared_ptr<const Layout> new_layout) {
```
pmenon: You can remove the `inline` keyword.

src/include/storage/data_table.h
```diff
@@ -247,9 +249,17 @@ class DataTable : public AbstractTable {
...
+    PELOTON_ASSERT(new_layout->GetColumnCount() == schema->GetColumnCount());
+    default_layout_ = new_layout;
+  }
 
-  column_map_type GetDefaultLayout() const;
+  inline void ResetDefaultLayout(LayoutType type = LayoutType::ROW) {
```
pmenon: Remove `inline`.

src/include/storage/data_table.h
```diff
@@ -414,16 +437,17 @@ class DataTable : public AbstractTable {
   // dirty flag. for detecting whether the tile group has been used.
   bool dirty_ = false;
 
+  // Last used layout_oid. Used while creating new layouts
+  // Initialized to COLUMN_STORE_OID since its the highest predfeined value.
+  std::atomic<oid_t> current_layout_oid_ = ATOMIC_VAR_INIT(COLUMN_STORE_OID);
```
pmenon: Any reason this isn't set in the constructor?

src/include/storage/layout.h
```diff
@@ -0,0 +1,141 @@
...
+/** @brief  Pre-defined OIDs for ROW and COLUMN store.  */
+#define ROW_STORE_OID 0
+#define COLUMN_STORE_OID 1
+//@}
+
+namespace peloton {
+
+namespace catalog {
+class Schema;
+}
```
pmenon: Close namespace.

src/include/storage/layout.h
```diff
@@ -0,0 +1,141 @@
+//===----------------------------------------------------------------------===//
+//
+//                         Peloton
+//
+// layout.h
+//
+// Identification: src/include/storage/layout.h
+//
+// Copyright (c) 2015-18, Carnegie Mellon University Database Group
```
pmenon: If you use the `formatter.py` script, it'll add the properly formatted copyright.

src/include/storage/layout.h
```diff
@@ -0,0 +1,141 @@
...
+
+  /**
+   * @brief   Constructor for arbitrary column_maps.
+   * @param   column_map Column map of the layout to be constructed.
+   * @param   layout_oid Per-table unique OID. Generted by DataTable.
+   */
+  Layout(const column_map_type &column_map, oid_t layout_oid);
+
+  /** @brief  Check whether this layout is a row store. */
+  inline bool IsRowStore() const { return (layout_type_ == LayoutType::ROW); }
```
pmenon: Remove `inline`.

src/include/storage/layout.h
```diff
@@ -0,0 +1,141 @@
...
+
+  /** @brief  Check whether this layout is a column store. */
+  inline bool IsColumnStore() const {
```
pmenon: Remove `inline`.

test/codegen/table_scan_translator_test.cpp
```diff
@@ -35,6 +36,122 @@ class TableScanTranslatorTest : public PelotonCodeGenTest {
     CreateAndLoadAllColsTable();
   }
 
+  void ExecuteTileGroupTest(peloton::LayoutType layout_type) {
```
pmenon: Can you modify the base test class functions to accept a `LayoutType`? Then you should reuse the create/loading functions rather than creating a one-off function. This also allows other tests to use custom table layouts for free.

pmenon: Correct, the tests do not use the inserter, but instead directly insert into the table [here](https://github.com/cmu-db/peloton/blob/master/test/codegen/testing_codegen_util.cpp#L134).

src/tuning/layout_tuner.cpp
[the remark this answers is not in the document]
pervazea: Ok. I think it is fine not to throw an exception. 
This is not your issue, but I'm wondering why the tuner doesn't apply a bit more intelligence. Just re-trying an operation without keeping any stats or ensuring you don't end up re-trying it forever, seems optimistic.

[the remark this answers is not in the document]
poojanilangekar: I will change this function to return a `bool` to indicate the success. Later, @malin1993ml can determine how to handle failures in the `Tuner` as a part of the`brain`. Can you please review the rest of the changes? This needs to be merged in order to support schema changes. 

Also the `query_logger_test` failure is the same as #1307. So I am assuming we can ignore it.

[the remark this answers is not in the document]
poojanilangekar: @pervazea I have changed that part. Please let me know if there are any other changes I need to make.

test/codegen/table_scan_translator_test.cpp
poojanilangekar: @pmenon I have modified the `CreateTestTables` function to get the layout as a parameter. However, since its called from the constructor, so the entire test suit would get called with the given layout. I have however moved the `CreateAndLoad` function to the base class. Please let me know if you want me to change anything else.

src/gc/gc_manager.cpp
```diff
@@ -23,39 +23,40 @@ namespace peloton {
...
-            varlen_ptr = type::Value::GetDataFromStorage(type_id, field_location);
-            // Call the corresponding varlen pool free
-              if (varlen_ptr != nullptr) {
-                tile->pool->Free(varlen_ptr);
-              }
-          }
+void GCManager::CheckAndReclaimVarlenColumns(storage::TileGroup *tile_group,
+                                             oid_t tuple_id) {
+  oid_t tile_count = tile_group->tile_count_;
+  oid_t tile_col_count;
```
pervazea: tile_count and tile_col_count should be uint32_t not oid_t.
While use of oid_t will compile and run correctly, it is confusing and incorrect style.

src/include/catalog/catalog.h
```diff
@@ -99,7 +100,8 @@ class Catalog {
       const std::string &database_name, const std::string &schema_name,
       const std::string &table_name, std::unique_ptr<catalog::Schema>,
       concurrency::TransactionContext *txn, bool is_catalog = false,
-      oid_t tuples_per_tilegroup = DEFAULT_TUPLES_PER_TILEGROUP);
+      oid_t tuples_per_tilegroup = DEFAULT_TUPLES_PER_TILEGROUP,
```
pervazea: oid_t -> uint32_t

src/include/storage/tile_group.h
```diff
@@ -214,13 +190,12 @@ class TileGroup : public Printable {
   oid_t num_tuple_slots;
```
pervazea: not oid_t

src/include/storage/tile_group.h
```diff
@@ -214,13 +190,12 @@ class TileGroup : public Printable {
...
 
   // number of tiles
-  oid_t tile_count;
+  oid_t tile_count_;
```
pervazea: not oid_t

src/optimizer/stats/tuple_sampler.cpp
```diff
@@ -91,15 +91,14 @@ bool TupleSampler::GetTupleInTileGroup(storage::TileGroup *tile_group,
...
   size_t tile_count = tile_group->GetTileCount();
 
   LOG_TRACE("tile_count: %lu", tile_count);
   for (oid_t tile_itr = 0; tile_itr < tile_count; tile_itr++) {
-    const catalog::Schema &schema = tile_schemas[tile_itr];
-    oid_t tile_column_count = schema.GetColumnCount();
 
     storage::Tile *tile = tile_group->GetTile(tile_itr);
+    const catalog::Schema &schema = *(tile->GetSchema());
+    oid_t tile_column_count = schema.GetColumnCount();
```
pervazea: not oid_t

src/storage/data_table.cpp
```diff
@@ -1196,25 +1194,28 @@ catalog::ForeignKey *DataTable::GetForeignKeySrc(const size_t offset) const {
...
-    new_tile_column_offset = column_map_entry.second.second;
-    oid_t column_offset = column_map_entry.first;
-
-    tile_group->LocateTileAndColumn(column_offset, orig_tile_offset,
-                                    orig_tile_column_offset);
 
-    // Get the column info from original schema
-    auto orig_schema = orig_schemas[orig_tile_offset];
-    auto column_info = orig_schema.GetColumn(orig_tile_column_offset);
+  oid_t column_count = layout.GetColumnCount();
```
pervazea: not oid_t

test/codegen/table_scan_translator_test.cpp
```diff
@@ -584,5 +629,207 @@ TEST_F(TableScanTranslatorTest, ScanWithModuloPredicate) {
...
 }
 
+TEST_F(TableScanTranslatorTest, ScanRowLayout) {
+  //
+  // Creates a table with LayoutType::ROW and
+  // invokes the TableScanTranslator
+  //
+  oid_t tuples_per_tilegroup = 100;
+  oid_t tilegroup_count = 5;
+  oid_t column_count = 100;
```
pervazea: The above 3 items, not oid_t

test/codegen/table_scan_translator_test.cpp
```diff
@@ -584,5 +629,207 @@ TEST_F(TableScanTranslatorTest, ScanWithModuloPredicate) {
...
+}
+
+TEST_F(TableScanTranslatorTest, ScanColumnLayout) {
+  //
+  // Creates a table with LayoutType::COLUMN and
+  // invokes the TableScanTranslator
+  //
+  oid_t tuples_per_tilegroup = 100;
+  oid_t tilegroup_count = 5;
+  oid_t column_count = 100;
```
pervazea: not oid_t

test/codegen/table_scan_translator_test.cpp
```diff
@@ -584,5 +629,207 @@ TEST_F(TableScanTranslatorTest, ScanWithModuloPredicate) {
...
+  // Inserts tuples
+  // Sets the default_layout_ as LayoutType::HYBRID
+  // Inserts some more tuples
+  // invokes the TableScanTranslator
+  //
+
+  const int tuples_per_tilegroup = 100;
+  const int col_count = 6;
+  const bool is_inlined = true;
+  oid_t tuple_count = 100;
```
pervazea: not oid_t

test/codegen/table_scan_translator_test.cpp
```diff
@@ -584,5 +629,207 @@ TEST_F(TableScanTranslatorTest, ScanWithModuloPredicate) {
...
+  auto layout =
+      catalog->CreateDefaultLayout(database_oid, table_oid, column_map, txn);
+  EXPECT_NE(nullptr, layout);
+  txn_manager.CommitTransaction(txn);
+
+  /////////////////////////////////////////////////////////
+  // Load in 200 tuples
+  /////////////////////////////////////////////////////////
+  tuple_count = 200;
+  oid_t prev_tuple_count = 100;
```
pervazea: not oid_t

test/codegen/testing_codegen_util.cpp
```diff
@@ -175,6 +178,80 @@ void PelotonCodeGenTest::LoadTestTable(oid_t table_id, uint32_t num_rows,
   txn_manager.CommitTransaction(txn);
 }
 
+void PelotonCodeGenTest::CreateAndLoadTableWithLayout(
+    peloton::LayoutType layout_type, oid_t tuples_per_tilegroup,
+    oid_t tile_group_count, oid_t column_count, bool is_inlined) {
+  oid_t tuple_count = tuples_per_tilegroup * tile_group_count;
```
pervazea: more oid_t -> uint32_t

test/storage/tile_group_test.cpp
```diff
@@ -232,11 +238,14 @@ TEST_F(TileGroupTests, StressTest) {
 //  std::vector<catalog::Schema> schemas;
 //
 //  // SCHEMA
-//  catalog::Column column1(type::TypeId::INTEGER, type::Type::GetTypeSize(type::TypeId::INTEGER),
+//  catalog::Column column1(type::TypeId::INTEGER,
```
pervazea: I understand these are just formatting changes... but wrt to the MVCCInsertTest.
It is entirely commented out (has been for a while). Is there a reason why we are keeping it but disabling it?
Unless there is a good reason, I would delete the test.

src/include/storage/layout.h
```diff
@@ -0,0 +1,144 @@
...
+  void LocateTileAndColumn(oid_t column_offset, oid_t &tile_offset,
+                           oid_t &tile_column_offset) const;
+
+  /** @brief    Returns the layout difference w.r.t. the other Layout.
+   *  @double   The delta between the layouts. Used by LayoutTuner class.
+   */
+  double GetLayoutDifference(const storage::Layout &other) const;
+
+  /** @brief  Returns the tile_id of the given column_id. */
+  oid_t GetTileIdFromColumnId(oid_t column_id) const;
```
pervazea: not oid_t

src/include/storage/layout.h
```diff
@@ -0,0 +1,144 @@
...
+
+  /** @brief  Returns the column offset of the column_id. */
+  oid_t GetTileColumnOffset(oid_t column_id) const;
```
pervazea: not oid_t

src/include/storage/layout.h
```diff
@@ -0,0 +1,144 @@
...
+
+  /** @brief  Returns the number of columns in the layout. */
+  oid_t GetColumnCount() const { return num_columns_; }
```
pervazea: not oid_t

src/include/storage/layout.h
```diff
@@ -0,0 +1,144 @@
...
+  /** @brief  Check whether this layout is a column store. */
+  bool IsColumnStore() const { return (layout_type_ == LayoutType::COLUMN); }
+
+  /** @brief  Return the layout_oid_ of this object. */
+  oid_t GetOid() const { return layout_oid_; }
+
+  /** @brief  Sets the tile id and column id w.r.t. of the tile corresponding
+   *          to the specified tile group column id.
+   */
+  void LocateTileAndColumn(oid_t column_offset, oid_t &tile_offset,
```
pervazea: Misnamed? More correctly something more like SetTileAndColumn?

src/storage/layout.cpp
```diff
@@ -0,0 +1,354 @@
...
+
+// Sets the tile id and column id w.r.t that tile corresponding to
+// the specified tile group column id.
+void Layout::LocateTileAndColumn(oid_t column_offset, oid_t &tile_offset,
+                                 oid_t &tile_column_offset) const {
+  // Ensure that the column_offset is not out of bound
+  PELOTON_ASSERT(num_columns_ > column_offset);
+
+  // For row store layout, tile id is always 0 and the tile
+  // column_id and tile column_id is the same.
```
pervazea: Comment text is unclear...
and the tile column_id and tile column_id is the same

src/include/storage/layout.h
poojanilangekar: I am sorry! This function was previously a part of `TileGroup`, I copied the comments as is. 

I have changed it now. Essentially, we are using references here because we need to return both the `tile_id` and `tile_column_id`.

src/include/storage/layout.h
poojanilangekar: I think this should be `oid_t` because it's the `column_oid` within the `Tile`.

src/include/storage/layout.h
poojanilangekar: I think this should also be `oid_t` because the function returns a `tile_id`.

src/include/storage/tile_group.h
```diff
@@ -76,7 +75,7 @@ class TileGroup : public Printable {
   // Tile group constructor
   TileGroup(BackendType backend_type, TileGroupHeader *tile_group_header,
             AbstractTable *table, const std::vector<catalog::Schema> &schemas,
-            const column_map_type &column_map, int tuple_count);
+            std::shared_ptr<const Layout> layout, int tuple_count);
```
pmenon: We're passing in a shared pointer, does that mean this can be null? If not, should we make it a reference? 

Also, in terms of ownership, does the TileGroup own the layout?

src/catalog/catalog.cpp
```diff
@@ -555,6 +561,44 @@ ResultType Catalog::CreateIndex(
...
+  auto table = database->GetTableWithOid(table_oid);
+
+  oid_t layout_oid = table->GetNextLayoutOid();
+  // Ensure that the new layout
+  PELOTON_ASSERT(layout_oid < INVALID_OID);
+  auto new_layout = std::shared_ptr<const storage::Layout>(
+      new const storage::Layout(column_map, layout_oid));
+
+  // Add the layout the pg_layout table
+  auto pg_layout = catalog_map_[database_oid]->GetLayoutCatalog();
```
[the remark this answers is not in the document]
poojanilangekar: You are right. There are other small bugs I noticed in #1356 and #1352. I will raise a Catalog cleanup PR once this gets merged in.

src/include/storage/tile_group.h
poojanilangekar: The layout is actually shared among multiple TileGroups which are initialized with the same layout. We would want the ptr to be alive as long as we have atleast one TileGroup with the same. 

To make sure a TileGroup always get a non-null layout, I am now throwing an exception in the TileGroupFactory if it passed a `nullptr`. Let me know if you want me to handle this differently.

PR #1329, opened 2018-04-24
Run clang-format on staged hunks only instead of whole file

tcm-marcel:
This PR adapts the [formatter script](https://github.com/cmu-db/peloton/blob/master/script/formatting/formatter.py) to run clang-format 
1. on staged changes using `-f`
2. on the changes of the last _n_ commits using `-n <n>`

Only the changed lines are formatted by clang-format. This prevents clang-format to change parts of the file that haven't been touched at all. 

Implemented in a kind-of ugly way, parsing the diff output of git. However, it should work reliable. The available `git clang-format` seems not to work reliable, especially on Mac, so this is the best approach so far.


tcm-marcel:
@poojanilangekar I just realized that the git commit hook still checks with clang-format on the whole file and therefore will complain. Let me think about this...

I also accidentally found a python script in the LLVM repo that apparently does exactly the same: https://llvm.org/svn/llvm-project/cfe/trunk/tools/clang-format/git-clang-format

poojanilangekar:
@tcm-marcel I think we need exactly what the script does. Would it be possible to use it instead of our existing format script?

tcm-marcel:
It seems like this script is usually installed automatically when installing clang-format. The only problem is that similar to clang-format itself, we have different versions and the command differs. 

I suggest to change both validator and formatter to use `git clang-format` instead of `clang-format` directly and add an option for both to check/format only uncommitted/stages changes. I can do that in the next few days. 

`git clang-format` can already be used manually on every system.

pervazea:
Looking over the code, the changes look ok.
What testing has been done to verify that the modifications work as expected?

tcm-marcel:
Only manual testing. It was successfully tested on the huge PR from Matt #1326.

PR #1330, opened 2018-04-25
Fix travis build to actually use gcc5

tcm-marcel:
Travis vehemently ignores our environmental variables for gcc5, but setting it in front of the `cmake` command works. 

@pmenon gave this hint before, but it was forgotten in #1311.


pmenon:
@tcm-marcel For consistency, you should set `CC=gcc-5`, too.

Also, I wonder if you can `export` the `CXX` and `CC` variables before invoking the build? Then the `cmake` invocation should remain the same and, hopefully, make the environment setup clear.

pmenon:
@tcm-marcel Huh, then it looks like the stuff in `env` is getting overridden. Does it make sense to move everything currently into `env` into `before_script` as `export`'s?

I just found [this](https://github.com/apache/incubator-quickstep/blob/master/.travis.yml) as an example.

tcm-marcel:
The tests fail because of an incompatibility between the pre-compiled tensorflow version and protobuf, which is already discussed on the mailing list. I will leave this open, until we have a solution/workaround for this.

PR #1331, opened 2018-04-27
Catalog Synchronization

tli2:
Catalog synchronization and rpc

### DIFF
--- a/src/include/brain/brain.h
+++ b/src/include/brain/brain.h
@@ -33,16 +33,18 @@ namespace brain {
 class BrainEnvironment {
  public:
   BrainEnvironment() :
-      rpc_client_{settings::SettingsManager::GetString(settings::SettingId::peloton_rpc_address)},
-      sql_connection_{settings::SettingsManager::GetString(settings::SettingId::peloton_address)} {}
+      rpc_client_{
+          settings::SettingsManager::GetString(settings::SettingId::peloton_rpc_address)},
+      sql_connection_{
+          settings::SettingsManager::GetString(settings::SettingId::peloton_address)} {}
 
-  inline PelotonService::Client GetPelotonService() {
-    return rpc_client_.getMain<PelotonService>();
+  inline capnp::EzRpcClient &GetPelotonClient() {
+    return rpc_client_;
   }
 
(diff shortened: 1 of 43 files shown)


src/include/brain/brain.h
```diff
@@ -99,18 +133,25 @@ class Brain {
...
+   */
+  inline void Terminate() { scheduler_.ExitLoop(); }
 
  private:
+  // Main Event loop
   NotifiableTask scheduler_;
+  // collection of all the jobs registered
   std::unordered_map<std::string, BrainJob *> jobs_;
+  // mapping of jobs to their event structs (used to trigger the job)
   std::unordered_map<std::string, struct event *> job_handles_;
```
malin1993ml: I actually didn't see where this thing is used after the registration. Does this have some future use?

src/include/brain/catalog_sync_brain_job.h
```diff
@@ -0,0 +1,126 @@
...
+#include "catalog/table_catalog.h"
+#include "storage/data_table.h"
+#include "catalog/schema.h"
+#include "type/value_factory.h"
+#include "catalog/query_history_catalog.h"
+#include "catalog/query_metrics_catalog.h"
+
+namespace peloton {
+namespace brain {
+class CatalogSyncBrainJob : public BrainJob {
```
malin1993ml: Like I said in [this comment](https://github.com/cmu-db/peloton/pull/1283/files/412ebb9785bade124e610924fb2155275e373536#diff-e783af558990ff366767d25fe307af88):
1. Please add some brief comments to the functions in this class
2. I would move most of the function definitions in this class to a cpp file since they're more than the size of a function that you want to inline. And the reduce the includes.

src/include/brain/brain.h
tli2: Potentially. This is where you can drop a job or manually trigger it on demand,

src/include/brain/catalog_sync_brain_job.h
tli2: Will do

tli2:
No longer relevant. Catalog refactor makes this out of date and hard to fix. Will just implement replication with log records since nobody is currently depending on this to run experiments.

Issue #1336, opened 2018-05-01
Assertion fail when insert and delete in the same transaction


nwang57:
```
create table foo(id integer);

begin;
insert into foo values(5);
delete from foo where id = 5;
```
The assertion failed at line 618 in `src/concurrency/timestamp_ordering_transaction_manager.cpp`.  It seems that if the txn owns the tuple, it will not update the tuple's last reader id. So the commitId is 0 during the assertion.

apavlo:
Thank you for reporting this. We are having a lot of problems with our concurrency control scheme. Is this a blocker for your project?

nwang57:
Yeah, we passed all the `make check` tests except the functions_test which is related to this problem. We will talk with @pmenon on Friday about this issue.

BYW, we just merged with catalog refactoring commit in the master with our changes. So far it doesn't break any test yet.

PR #1337, opened 2018-05-01
Fix for #1336

pmenon:
It looks like the deletion logic in codegen was incomplete. 

**Problem:** Under SI, as an optimization, the transaction manager doesn't update the last reader field when reading a version created by the current transaction. Then, in the version of `TimestampOrderingTransactionManager::PerformDelete(...)` that is called in `codegen::Deleter`, there is a sanity check to ensure that the transaction actually read the version. This will be false because of the aforementioned optimization.

**Solution:** There is an existing `TimestampOrderingTransactionManager::PerformDelete(...)` that doesn't require insertion of an empty version. This is the API we should be using in `codgen::Deleter` only in the case where we delete a version we've created. This is the same logic as in the interpreted engine.

### DIFF
--- a/src/codegen/deleter.cpp
+++ b/src/codegen/deleter.cpp
@@ -6,11 +6,12 @@
 //
 // Identification: src/codegen/deleter.cpp
 //
-// Copyright (c) 2015-2017, Carnegie Mellon University Database Group
+// Copyright (c) 2015-2018, Carnegie Mellon University Database Group
 //
 //===----------------------------------------------------------------------===//
 
 #include "codegen/deleter.h"
+
 #include "codegen/transaction_runtime.h"
 #include "concurrency/transaction_manager_factory.h"
 #include "storage/data_table.h"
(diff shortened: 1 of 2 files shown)


yingjunwu:
hi @pmenon, I have reviewed your code and it looks great to me. The original codegen code for delete executor is incomplete: it only contains a simplified logic for deletion. Your code basically matches the logic in `DeleteExecutor` and checks both `is_owner` and `is_written`. I don't have any question on this part.

However, there are two questions.

1. You mentioned that `Under SI, as an optimization, the transaction manager doesn't update the last reader field when reading a version created by the current transaction.` This is not precise, as this is not an optimization for SI. If the current transaction has already acquired the ownership of a tuple, then it must has already updated the last reader field of the tuple, hence there's no need to update it again. However, it seems that the codegen version of delete executor does not check the isolation level for a transaction, am I right? We also did not check the isolation level in codegen executor.

2. #1336 mentioned that the test cases failed because of the codegen version of delete executor. It seems that when running test cases, we always set the codegen flag to true. If so, I wonder whether there's still need to maintain two versions of delete executor (the codegen version and the original version).

yingjunwu:
@pmenon yes you are right. Regarding to the isolation levels, we don't want to check the isolation level in the codegen version of update/delete executor, right? 

I think the PR is ready to go.

Issue #1298, opened 2018-04-14
QueryCache Bug


nwang57:
Currently, when comparing two plans, lets say sequential plans, it only compares if the table, the number of predicates and the predicate types are the same. 

Lets say we have plan A:` SELECT a FROM foo WHERE a = 40`.
and plan B: `SELECT b FROM foo WHERE b = 41`.

They are considered to be the same plan. See test below. When execute plan A it saves the complied plan into the query cache. Then when execute the plan B, it looks into the cache and find the complied version of plan A and the system just use the cached plan A to execute plan B, but they have different columns in the where clause. So the system will use the column a to match value 41 and doesn't find the tuple we want.

This bug is causing test failures in #1281 .

See the test case to reproduce:
In `test/codegen/query_cache_test.cpp`
```
  std::shared_ptr<planner::SeqScanPlan> GetSeqScanPlanA() {
    auto *a_col_exp =
        new expression::TupleValueExpression(type::TypeId::INTEGER, 0, 0);
    auto *const_40_exp = PelotonCodeGenTest::ConstIntExpr(40).release();
    auto *a_eq_40 = new expression::ComparisonExpression(
        ExpressionType::COMPARE_EQUAL, a_col_exp, const_40_exp);
    return std::shared_ptr<planner::SeqScanPlan>(new planner::SeqScanPlan(
        &GetTestTable(TestTableId()), a_eq_40, {0, 1}));
  }

  std::shared_ptr<planner::SeqScanPlan> GetSeqScanPlanB() {
    auto *b_col_exp =
        new expression::TupleValueExpression(type::TypeId::INTEGER, 0, 1);
    auto *const_41_exp = PelotonCodeGenTest::ConstIntExpr(41).release();
    auto *b_eq_41 = new expression::ComparisonExpression(
        ExpressionType::COMPARE_EQUAL, b_col_exp, const_41_exp);
    return std::shared_ptr<planner::SeqScanPlan>(new planner::SeqScanPlan(
        &GetTestTable(TestTableId()), b_eq_41, {0, 1}));
  }

TEST_F(QueryCacheTest, SimpleCacheWithDiffPredicate) {
  std::shared_ptr<planner::SeqScanPlan> plan_a = GetSeqScanPlanA();

  std::shared_ptr<planner::SeqScanPlan> plan_b = GetSeqScanPlanB();
  // Do binding
  planner::BindingContext context_1;
  plan_a->PerformBinding(context_1);
  planner::BindingContext context_2;
[...]

PR #1339, opened 2018-05-04
[15721] Compiled Catalog Query Access

nwang57:
We enabled complied catalog lookup in the first pull request. Based upon this, we further support complied insert and delete queries for all catalogs. We also fix bugs for the following issues:

* #1298 Need to manually bind the tuple value expression so that equality checks will work correctly.

* Sequential scan assumes that the output columns should start at offset 0 otherwise PerformBinding will not correctly find the corresponding column attributes.

* ZoneMapCatalog needs special care to avoid chicken and egg problem. Each sequential plan needs to check the zone map to know whether to scan the tile or not, but checking the zone map requires ZoneMap catalog access which leads to an infinity loop. So we ensure in the ZoneMap catalog manager that if the scanning table is ZoneMap catalog then it will just scan it without checking the ZoneMap

### DIFF
--- a/src/catalog/abstract_catalog.cpp
+++ b/src/catalog/abstract_catalog.cpp
@@ -11,7 +11,9 @@
 //===----------------------------------------------------------------------===//
 
 #include "catalog/abstract_catalog.h"
-
+#include "executor/plan_executor.h"
+#include "codegen/buffering_consumer.h"
+#include "common/internal_types.h"
 #include "common/statement.h"
 
 #include "catalog/catalog.h"
@@ -129,6 +131,58 @@ bool AbstractCatalog::InsertTuple(std::unique_ptr<storage::Tuple> tuple,
   return this_p_status.m_result == peloton::ResultType::SUCCESS;
 }
 
+/*@brief   insert tuple(reord) helper function
+* @param   tuple     tuple to be inserted
+* @param   txn       TransactionContext
+* @return  Whether insertion is Successful
+*/
+bool AbstractCatalog::InsertTupleWithCompiledPlan(const std::vector<std::vector<
+  std::unique_ptr<expression::AbstractExpression>>> *insert_values,
+                                  concurrency::TransactionContext *txn) {
+ if (txn == nullptr)
+  throw CatalogException("Insert tuple requires transaction");
+
+ std::vector<std::string> columns;
+ std::shared_ptr<planner::InsertPlan> insert_plan(
+   new planner::InsertPlan(catalog_table_, &columns, insert_values));
+
+ // Bind the plan
+ planner::BindingContext context;
+ insert_plan->PerformBinding(context);
+
+ // Prepare a consumer to collect the result
+ codegen::BufferingConsumer buffer{{}, context};
+
+
+ bool cached;
+
+ codegen::QueryParameters parameters(*insert_plan, {});
+ std::unique_ptr<executor::ExecutorContext> executor_context(
+   new executor::ExecutorContext(txn, std::move(parameters)));
+
+ // search for query
+ codegen::Query *query = codegen::QueryCache::Instance().Find(insert_plan);;
+ std::unique_ptr<codegen::Query> compiled_query(nullptr);
+ cached = (query != nullptr);
+//  cached = false;
+ // LOG_DEBUG("cache %d", cached);
+ // if not cached, compile the query and save it into cache
+ executor::ExecutionResult ret;
+ if (!cached) {
+  compiled_query = codegen::QueryCompiler().Compile(
+    *insert_plan, executor_context->GetParams().GetQueryParametersMap(),
+    buffer);
+  query = compiled_query.get();
+  codegen::QueryCache::Instance().Add(insert_plan, std::move(compiled_query));
+ }
+
+ query->Execute(std::move(executor_context), buffer,
+                [&ret](executor::ExecutionResult result) { ret = result; });
+
+ return ret.m_result == peloton::ResultType::SUCCESS;
+}
+
+
 /*@brief   Delete a tuple using index scan
  * @param   index_offset  Offset of index for scan
  * @param   values        Values for search
@@ -177,6 +231,63 @@ bool AbstractCatalog::DeleteWithIndexScan(
 
   return status;
 }
+
+
+/*@brief   Delete a tuple using index scan
+* @param   index_offset  Offset of index for scan
+* @param   values        Values for search
+* @param   txn           TransactionContext
+* @return  Whether deletion is Successful
+*/
+bool AbstractCatalog::DeleteWithCompiledSeqScan(
+  std::vector<oid_t> column_offsets,
+  expression::AbstractExpression *predicate,
+  concurrency::TransactionContext *txn) {
+ if (txn == nullptr)
+  throw CatalogException("Delete tuple requires transaction");
+
+ std::shared_ptr<planner::DeletePlan> delete_plan{
+   new planner::DeletePlan(catalog_table_)};
+
+ std::unique_ptr<planner::AbstractPlan> scan{new planner::SeqScanPlan(
+  catalog_table_, predicate, column_offsets)};
+ delete_plan->AddChild(std::move(scan));
+
+ // Do binding
+ planner::BindingContext context;
+ delete_plan->PerformBinding(context);
+
+ codegen::BufferingConsumer buffer{column_offsets, context};
+
+ bool cached;
+
+ codegen::QueryParameters parameters(*delete_plan, {});
+ std::unique_ptr<executor::ExecutorContext> executor_context(
+   new executor::ExecutorContext(txn, std::move(parameters)));
+
+ // search for query
+ codegen::Query *query = codegen::QueryCache::Instance().Find(delete_plan);;
+ std::unique_ptr<codegen::Query> compiled_query(nullptr);
+ cached = (query != nullptr);
+
+ // if not cached, compile the query and save it into cache
+ executor::ExecutionResult ret;
+ if (!cached) {
+  compiled_query = codegen::QueryCompiler().Compile(
+    *delete_plan, executor_context->GetParams().GetQueryParametersMap(),
+    buffer);
+  query = compiled_query.get();
+  codegen::QueryCache::Instance().Add(delete_plan, std::move(compiled_query));
+ }
+
+ query->Execute(std::move(executor_context), buffer,
+                [&ret](executor::ExecutionResult result) { ret = result; });
+
+ return ret.m_result == peloton::ResultType::SUCCESS;
+}
+
+
+
 
 /*@brief   Index scan helper function
  * @param   column_offsets    Column ids for search (projection)
@@ -259,6 +370,56 @@ AbstractCatalog::GetResultWithSeqScan(std::vector<oid_t> column_offsets,
   }
 
   return result_tiles;
+}
+
+/*@brief   Complied Sequential scan helper function
+* NOTE: try to use efficient index scan instead of sequential scan, but you
+* shouldn't build too many indexes on one catalog table
+* @param   column_offsets    Column ids for search (projection)
+* @param   predicate         predicate for this sequential scan query
+* @param   txn               TransactionContext
+*
+* @return  Unique pointer of vector of logical tiles
+*/
+std::vector<codegen::WrappedTuple>
+AbstractCatalog::GetResultWithCompiledSeqScan(
+    std::vector<oid_t> column_offsets,
+    expression::AbstractExpression *predicate,
+    concurrency::TransactionContext *txn) const {
+  if (txn == nullptr) throw CatalogException("Scan table requires transaction");
+
+  // Create sequential scan
+  auto plan_ptr = std::make_shared<planner::SeqScanPlan>(
+      catalog_table_, predicate, column_offsets);
+  planner::BindingContext scan_context;
+  plan_ptr->PerformBinding(scan_context);
+
+  // Create consumer
+  codegen::BufferingConsumer buffer{column_offsets, scan_context};
+  bool cached;
+
+  codegen::QueryParameters parameters(*plan_ptr, {});
+  std::unique_ptr<executor::ExecutorContext> executor_context(
+      new executor::ExecutorContext(txn, std::move(parameters)));
+
+  // search for query
+  codegen::Query *query = codegen::QueryCache::Instance().Find(plan_ptr);
+  std::unique_ptr<codegen::Query> compiled_query(nullptr);
+  cached = (query != nullptr);
+
+  // if not cached, compile the query and save it into cache
+  if (!cached) {
+    compiled_query = codegen::QueryCompiler().Compile(
+        *plan_ptr, executor_context->GetParams().GetQueryParametersMap(),
+        buffer);
+    query = compiled_query.get();
+    codegen::QueryCache::Instance().Add(plan_ptr, std::move(compiled_query));
+  }
+
+  query->Execute(std::move(executor_context), buffer,
+                 [](executor::ExecutionResult result) { return result; });
+
+  return buffer.GetOutputTuples();
 }
 
 /*@brief   Add index on catalog table
--- a/src/catalog/abstract_catalog.cpp
+++ b/src/catalog/abstract_catalog.cpp
@@ -131,8 +131,8 @@ bool AbstractCatalog::InsertTuple(std::unique_ptr<storage::Tuple> tuple,
   return this_p_status.m_result == peloton::ResultType::SUCCESS;
 }
 
-/*@brief   insert tuple(reord) helper function
-* @param   tuple     tuple to be inserted
+/*@brief   insert tuple(reord) using compiled plan
+* @param   insert_values     tuples to be inserted
 * @param   txn       TransactionContext
 * @return  Whether insertion is Successful
 */
@@ -233,9 +233,9 @@ bool AbstractCatalog::DeleteWithIndexScan(
 }
 
 
-/*@brief   Delete a tuple using index scan
-* @param   index_offset  Offset of index for scan
-* @param   values        Values for search
+/*@brief   Delete a tuple using sequential scan
+* @param   column_offsets  Offset of seq scan
+* @param   predicate        Predicate used in the seq scan
 * @param   txn           TransactionContext
 * @return  Whether deletion is Successful
 */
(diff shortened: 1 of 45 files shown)


test/catalog/catalog_test.cpp
```diff
@@ -60,7 +61,37 @@ TEST_F(CatalogTests, CreatingDatabase) {
...
+// EXPECT_EQ(table_object->GetTableName(), "emp_table");
+//
+// catalog::Catalog::GetInstance()->DropTable("emp_db", DEFUALT_SCHEMA_NAME,
+//                                            "emp_table", txn);
+// auto table_object_1 = catalog::Catalog::GetInstance()->GetTableObject(
+//   "emp_db", DEFUALT_SCHEMA_NAME, "emp_table", txn);
+// EXPECT_EQ(nullptr, table_object_1);
+// txn_manager.CommitTransaction(txn);
+//}
+//
```
aaron-tian: Do you have problems passing this test? Are you going to uncomment it later?

test/catalog/catalog_test.cpp
```diff
@@ -24,6 +24,7 @@
 #include "storage/storage_manager.h"
 #include "type/ephemeral_pool.h"
 #include "sql/testing_sql_util.h"
+#include "common/timer.h"
```
aaron-tian: Why do you need to include this header file?

src/include/catalog/trigger_catalog.h
```diff
@@ -17,7 +17,7 @@
 // 0: oid (pkey)
 // 1: tgrelid   : table_oid
 // 2: tgname    : trigger_name
-// 3: tgfoid    : function_oid
+// 3: tgfoid    : function_name
```
aaron-tian: Should you change `tgfoid` if `function_oid` was meant to be `function_name`?

test/catalog/catalog_test.cpp
nwang57: This is needed for performance testing. We plan to remove it later.

src/catalog/abstract_catalog.cpp
```diff
@@ -129,12 +126,53 @@ bool AbstractCatalog::InsertTuple(std::unique_ptr<storage::Tuple> tuple,
...
+
+ // Bind the plan
+ planner::BindingContext context;
+ insert_plan->PerformBinding(context);
+
+ // Prepare a consumer to collect the result
+ codegen::BufferingConsumer buffer{{}, context};
+
+
+ bool cached;
```
aaron-tian: Would be better to define `cached` below where it's first assigned to a value.

src/catalog/abstract_catalog.cpp
```diff
@@ -178,13 +216,53 @@ bool AbstractCatalog::DeleteWithIndexScan(
...
+  catalog_table_, predicate, column_offsets)};
+ delete_plan->AddChild(std::move(scan));
+
+ // Do binding
+ planner::BindingContext context;
+ delete_plan->PerformBinding(context);
+
+ codegen::BufferingConsumer buffer{column_offsets, context};
+
+ bool cached;
```
aaron-tian: Would be better to define `cached` below where it's first assigned to a value.

src/catalog/abstract_catalog.cpp
```diff
@@ -261,15 +331,48 @@ AbstractCatalog::GetResultWithSeqScan(std::vector<oid_t> column_offsets,
...
+
+  // Create sequential scan
+  auto plan_ptr = std::make_shared<planner::SeqScanPlan>(
+      catalog_table_, predicate, column_offsets);
+  planner::BindingContext scan_context;
+  plan_ptr->PerformBinding(scan_context);
+
+  // Create consumer
+  codegen::BufferingConsumer buffer{column_offsets, scan_context};
+  bool cached;
```
aaron-tian: Move the `cached` definition to where it's first assigned.

src/catalog/abstract_catalog.cpp
```diff
@@ -298,13 +401,77 @@ void AbstractCatalog::AddIndex(const std::vector<oid_t> &key_attrs,
...
+       catalog_table_, predicate, column_offsets)};
+     update_plan->AddChild(std::move(scan));
+
+     // Do binding
+     planner::BindingContext context;
+     update_plan->PerformBinding(context);
+
+     codegen::BufferingConsumer buffer{column_offsets, context};
+
+     bool cached;
```
aaron-tian: Move the `cached` definition to where it's first assigned.

test/codegen/query_cache_test.cpp
```diff
@@ -207,6 +230,62 @@ class QueryCacheTest : public PelotonCodeGenTest {
...
+      // check if the two plans are the same
+      auto hash_equal = (plan_a->Hash() == plan_b->Hash());
+      EXPECT_FALSE(hash_equal);
+
+      auto is_equal = (*plan_a.get() == *plan_b.get());
+     EXPECT_FALSE(is_equal);
+
+      // execute SELECT a FROM table where a == 40;
+      codegen::BufferingConsumer buffer_1{{0, 1}, context_1};
+      bool cached;
```
aaron-tian: Move the `cached` definition to where it's first assigned.

src/catalog/column_catalog.cpp
```diff
@@ -142,9 +145,9 @@ bool ColumnCatalog::InsertColumn(oid_t table_oid,
                                  const std::vector<Constraint> &constraints,
                                  type::AbstractPool *pool,
                                  concurrency::TransactionContext *txn) {
+  (void) pool;
```
gandeevan: you could probably use the  "UNUSED ATTRIBUTE" specifier instead.

src/catalog/column_catalog.cpp
```diff
@@ -165,28 +168,74 @@ bool ColumnCatalog::InsertColumn(oid_t table_oid,
...
+    val3);
+  auto constant_expr_4 = new expression::ConstantValueExpression(
+    val4);
+  auto constant_expr_5 = new expression::ConstantValueExpression(
+    val5);
+  auto constant_expr_6 = new expression::ConstantValueExpression(
+    val6);
+  auto constant_expr_7 = new expression::ConstantValueExpression(
+    val7);
+
```
gandeevan: Smart pointers could be used instead. Memory management is a pain.

gandeevan: Since the number of ConstantValueExpression objects to be allocated, is known at compile time - the seven different mallocs invocations can be replaced by a single invocation of malloc using an array of objects with dynamic initialization. 

eg: arr_const_expr = new expression::ConstantValueExpression[8]({val0}, {val2},...,{val7})

src/catalog/index_catalog.cpp
```diff
@@ -154,33 +156,71 @@ bool IndexCatalog::InsertIndex(oid_t index_oid, const std::string &index_name,
...
+    val3);
+  auto constant_expr_4 = new expression::ConstantValueExpression(
+    val4);
+  auto constant_expr_5 = new expression::ConstantValueExpression(
+    val5);
+  auto constant_expr_6 = new expression::ConstantValueExpression(
+    val6);
+  auto constant_expr_7 = new expression::ConstantValueExpression(
+    val7);
+
```
gandeevan: Same as above.

src/catalog/index_metrics_catalog.cpp
```diff
@@ -50,25 +53,57 @@ bool IndexMetricsCatalog::InsertIndexMetrics(
...
+  auto constant_expr_2 = new expression::ConstantValueExpression(
+      val2);
+  auto constant_expr_3 = new expression::ConstantValueExpression(
+      val3);
+  auto constant_expr_4 = new expression::ConstantValueExpression(
+      val4);
+  auto constant_expr_5 = new expression::ConstantValueExpression(
+      val5);
+  auto constant_expr_6 = new expression::ConstantValueExpression(
+      val6);
```
gandeevan: Same as above.

src/catalog/proc_catalog.cpp
```diff
@@ -78,32 +82,56 @@ bool ProcCatalog::InsertProc(const std::string &proname,
...
+  auto constant_expr_1 = new expression::ConstantValueExpression(
+    val1);
+  auto constant_expr_2 = new expression::ConstantValueExpression(
+    val2);
+  auto constant_expr_3 = new expression::ConstantValueExpression(
+    val3);
+  auto constant_expr_4 = new expression::ConstantValueExpression(
+    val4);
+  auto constant_expr_5 = new expression::ConstantValueExpression(
+    val5);
```
gandeevan: Same as above.

src/catalog/query_history_catalog.cpp
```diff
@@ -40,20 +41,29 @@ bool QueryHistoryCatalog::InsertQueryHistory(
...
-  tuple->SetValue(ColumnId::QUERY_STRING, val0,
-                  pool != nullptr ? pool : &pool_);
-  tuple->SetValue(ColumnId::FINGERPRINT, val1, pool != nullptr ? pool : &pool_);
-  tuple->SetValue(ColumnId::TIMESTAMP, val2, pool != nullptr ? pool : &pool_);
+  auto constant_expr_0 = new expression::ConstantValueExpression(
+      val0);
+  auto constant_expr_1 = new expression::ConstantValueExpression(
+      val1);
+  auto constant_expr_2 = new expression::ConstantValueExpression(
+      val2);
```
gandeevan: Same as above.

src/catalog/query_metrics_catalog.cpp
```diff
@@ -80,80 +84,179 @@ bool QueryMetricsCatalog::InsertQueryMetrics(
...
+  values.emplace_back(new expression::ConstantValueExpression(
+      val8));
+  values.emplace_back(new expression::ConstantValueExpression(
+      val9));
+  values.emplace_back(new expression::ConstantValueExpression(
+      val10));
+  values.emplace_back(new expression::ConstantValueExpression(
+      val11));
+  values.emplace_back(new expression::ConstantValueExpression(
+      val12));
```
gandeevan: Again, the multiple mallocs can be coalesced using an array of objects.

src/catalog/column_catalog.cpp
```diff
@@ -165,28 +168,74 @@ bool ColumnCatalog::InsertColumn(oid_t table_oid,
...
-      IndexId::PRIMARY_KEY;  // Index of table_oid & column_name
-  std::vector<type::Value> values;
-  values.push_back(type::ValueFactory::GetIntegerValue(table_oid).Copy());
-  values.push_back(
-      type::ValueFactory::GetVarcharValue(column_name, nullptr).Copy());
+  std::vector<oid_t> column_ids(all_column_ids);
+
+  auto *table_oid_expr =
+      new expression::TupleValueExpression(type::TypeId::INTEGER, 0,
+                                           ColumnId::TABLE_OID);
```
gandeevan: I don't think the allocated memory has been freed. Irrespective, you should probably use smart pointers instead of working with raw pointers.  The same applies to a bunch of other places in the code where raw pointers could be replaced with smart pointers.

src/catalog/column_stats_catalog.cpp
```diff
@@ -96,74 +98,157 @@ bool ColumnStatsCatalog::InsertColumnStats(
...
-  column_name = tile->GetValue(0, ColumnStatsOffset::COLUMN_NAME_OFF);
-  has_index = tile->GetValue(0, ColumnStatsOffset::HAS_INDEX_OFF);
+  num_rows = tuple.GetValue(ColumnId::NUM_ROWS);
+  cardinality = tuple.GetValue(ColumnId::CARDINALITY);
+  frac_null = tuple.GetValue(ColumnId::FRAC_NULL);
+  most_common_vals = tuple.GetValue(ColumnId::MOST_COMMON_VALS);
+  most_common_freqs = tuple.GetValue(ColumnId::MOST_COMMON_FREQS);
+  hist_bounds = tuple.GetValue(ColumnId::HISTOGRAM_BOUNDS);
+  column_name = tuple.GetValue(ColumnId::COLUMN_NAME);
+  has_index = tuple.GetValue(ColumnId::HAS_INDEX);
```
gandeevan: I remember Pooja mentioning in the last meeting that tuple.GetValue() is going to deprecated. However, I vaguely remember her mentioning that it could still be used with catalog tables. You might want to talk to her/Andy about this.

src/include/index/bwtree.h
```diff
@@ -2864,6 +2866,21 @@ class BwTree : public BwTreeBase {
    * the mapping table rather than CAS with nullptr
    */
   void InitMappingTable() {
+    mapping_table = (std::atomic<const BaseNode *> *) \
+                    mmap(NULL, 1024 * 1024 * 1024,
```
gandeevan: 1. The value (1024*1024*1024) shouldn't be hardcoded (the MAPPING_TABLE_SIZE is #defined to 1MB). 

2. InitMappingTable() seems to be invoked from the bwtree constructor - is 1GB mmaped for every index created?

src/include/index/bwtree.h
```diff
@@ -7605,6 +7624,20 @@ class BwTree : public BwTreeBase {
                 epoch_leave.load());
 #endif
 
+      // NOTE: Only unmap memory here because we need to access the mapping
+      // table in the above routine. If it was unmapped in ~BwTree() then this
+      // function will invoke illegal memory access
+      int munmap_ret = munmap(tree_p->mapping_table, 1024 * 1024 * 1024);
```
gandeevan: Again, a hard-coded value shouldn't be used.

src/include/catalog/column_stats_catalog.h
```diff
@@ -107,6 +107,7 @@ class ColumnStatsCatalog : public AbstractCatalog {
 
  private:
   ColumnStatsCatalog(concurrency::TransactionContext *txn);
+  std::vector<oid_t> all_column_ids = {0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10};
```
gandeevan: It seems the purpose of all_column_ids is to use it a copy constructor to construct a vector with all the column oid's. Firstly, I don't think you need to declare an all_column_ids vector, it can be easily constructed explicitly at runtime.  Moreover hardcoding the values it is not a good idea since anyone changing the schema of the catalog table must also remember to update this. 

 A better way to construct this at runtime without breaking abstractions is by using the "num_columns" method and filling the vector using std::iota.

src/include/catalog/constraint.h
```diff
@@ -65,16 +64,15 @@ class Constraint : public Printable {
...
       return;
     }
 
     default_value.reset(new peloton::type::Value(value));
   }
 
-  type::Value* getDefaultValue() {
-    return default_value.get();
-  }
+  type::Value *getDefaultValue() { return default_value.get(); }
```
gandeevan: This can probably make inline.

src/include/catalog/zone_map_catalog.h
```diff
@@ -64,11 +64,9 @@ class ZoneMapCatalog : public AbstractCatalog {
     TYPE = 6
   };
 
-  enum class ZoneMapOffset { 
-    MINIMUM_OFF = 0, 
-    MAXIMUM_OFF = 1, 
-    TYPE_OFF = 2 
-  };
+  enum ZoneMapOffset { MINIMUM_OFF = 0, MAXIMUM_OFF = 1, TYPE_OFF = 2 };
+
```
gandeevan: I'm not sure why a strongly typed enum has been replaced by an old-style enum. It's better to stick to stongly typed enums. 

The same applies to a bunch of places where old-style enums have been used.

src/catalog/column_catalog.cpp
gandeevan: Since the number of objects to be allocated is known at compile time. Multiple malloc invocations can be replaced by a single invocation by allocating an array of objects with parameterized constructor. 

eg: arr_constant_expr = new expression::ConstantValueExpression[8]{{val0},{val1}...{val7}}

src/catalog/column_catalog.cpp
```diff
@@ -15,35 +15,38 @@
...
 
 namespace peloton {
 namespace catalog {
 
-ColumnCatalogObject::ColumnCatalogObject(executor::LogicalTile *tile,
-                                         int tupleId)
-    : table_oid(tile->GetValue(tupleId, ColumnCatalog::ColumnId::TABLE_OID)
+
+ColumnCatalogObject::ColumnCatalogObject(codegen::WrappedTuple wrapped_tuple)
+    : table_oid(wrapped_tuple.GetValue(ColumnCatalog::ColumnId::TABLE_OID)
```
latelatif: As mentioned by @gandeevan, GetValue is deprecated now. Please talk to @poojanilangekar to make sure you can use this for a wrapped tuple object safely

src/catalog/column_catalog.cpp
latelatif: I am not sure why we would want to keep the parameter if we are not using it at all.

src/catalog/column_stats_catalog.cpp
```diff
@@ -63,8 +65,8 @@ bool ColumnStatsCatalog::InsertColumnStats(
     std::string most_common_freqs, std::string histogram_bounds,
     std::string column_name, bool has_index, type::AbstractPool *pool,
     concurrency::TransactionContext *txn) {
-  std::unique_ptr<storage::Tuple> tuple(
-      new storage::Tuple(catalog_table_->GetSchema(), true));
+  (void) pool;
```
latelatif: Same as before. Why do we want to keep unused parameters around?

src/catalog/column_stats_catalog.cpp
```diff
@@ -96,74 +98,157 @@ bool ColumnStatsCatalog::InsertColumnStats(
...
-  tuple->SetValue(ColumnId::NUM_ROWS, val_num_row, nullptr);
-  tuple->SetValue(ColumnId::CARDINALITY, val_cardinality, nullptr);
-  tuple->SetValue(ColumnId::FRAC_NULL, val_frac_null, nullptr);
-  tuple->SetValue(ColumnId::MOST_COMMON_VALS, val_common_val, pool);
-  tuple->SetValue(ColumnId::MOST_COMMON_FREQS, val_common_freq, pool);
-  tuple->SetValue(ColumnId::HISTOGRAM_BOUNDS, val_hist_bounds, pool);
-  tuple->SetValue(ColumnId::COLUMN_NAME, val_column_name, pool);
-  tuple->SetValue(ColumnId::HAS_INDEX, val_has_index, nullptr);
+  tuples.emplace_back();
+//  tuples.push_back(std::vector<ExpressionPtr>());
```
latelatif: dead code

src/catalog/language_catalog.cpp
```diff
@@ -16,13 +16,15 @@
...
+#include "codegen/buffering_consumer.h"
+#include "expression/expression_util.h"
 
 namespace peloton {
 namespace catalog {
 
-LanguageCatalogObject::LanguageCatalogObject(executor::LogicalTile *tuple)
-    : lang_oid_(tuple->GetValue(0, 0).GetAs<oid_t>()),
-      lang_name_(tuple->GetValue(0, 1).GetAs<const char *>()) {}
+LanguageCatalogObject::LanguageCatalogObject(codegen::WrappedTuple tuple)
```
latelatif: Do we want to make a copy of the WrappedTuple parameter? Wouldn't using a reference suffice?

src/catalog/proc_catalog.cpp
```diff
@@ -17,20 +17,23 @@
...
+#include "expression/expression_util.h"
+#include "codegen/buffering_consumer.h"
 
 namespace peloton {
 namespace catalog {
 
 #define PROC_CATALOG_NAME "pg_proc"
 
-ProcCatalogObject::ProcCatalogObject(executor::LogicalTile *tile,
+ProcCatalogObject::ProcCatalogObject(codegen::WrappedTuple wrapped_tuple,
```
latelatif: Same as above. Is a reference not good enough? This applies to all such uses in other constructors as well

src/catalog/trigger_catalog.cpp
```diff
@@ -109,122 +122,190 @@ ResultType TriggerCatalog::DropTrigger(const oid_t database_oid,
...
+  expression::AbstractExpression *predicate =
+      expression::ExpressionUtil::ConjunctionFactory(
+          ExpressionType::CONJUNCTION_AND, type_equality_expr,
+          oid_equality_expr);
+
+  std::vector<codegen::WrappedTuple> result_tuples =
+      GetResultWithCompiledSeqScan(column_ids, predicate, txn);
+
+  // carefull! the result could be null!
+  LOG_INFO("size of the result tiles = %lu", result_tuples.size());
```
latelatif: Is this still needed? Should it be LOG_INFO?

src/include/catalog/column_catalog.h
```diff
@@ -40,7 +40,7 @@ namespace catalog {
 
 class ColumnCatalogObject {
  public:
-  ColumnCatalogObject(executor::LogicalTile *tile, int tupleId = 0);
+  ColumnCatalogObject(codegen::WrappedTuple wrapped_tuple);
```
latelatif: Use a reference instead

src/catalog/abstract_catalog.cpp
```diff
@@ -298,13 +401,77 @@ void AbstractCatalog::AddIndex(const std::vector<oid_t> &key_attrs,
...
+  std::vector<oid_t> column_offsets, expression::AbstractExpression *predicate,
+  concurrency::TransactionContext *txn) {
+ if (txn == nullptr) throw CatalogException("Scan table requires transaction");
+     // Construct update executor
+     TargetList target_list;
+     DirectMapList direct_map_list;
+
+     size_t column_count = catalog_table_->GetSchema()->GetColumnCount();
+     for (size_t col_itr = 0; col_itr < column_count; col_itr++) {
+      // Skip any column for update
```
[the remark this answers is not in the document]
nwang57: I think it is only looking for the columns in the tuple which need to be updated. For more detailed explanation you can turn to @mengranwo

apavlo:
I am reviving this PR. We will need this when we get rid of the interpreted engine. It also has bug fix for #1362

PR #1340, opened 2018-05-05
[15721] Query Rewrite

zhaoguoquan94:
This PR adds support for query rewriting optimizations that are oblivious to the cost model in Peloton.
* Transitive predicates (optimized in OptimizeLoop)
* Constant folding (optimized in PostgresParser).

We implement query rewrite rules using the current optimizer framework:

 - Transitive predicates: we implemented this rule by adding a new step to the beginning OptimizerLoop using a TopDownRewriter. We added rules to take predicates from nodes, generating all possible predicates by taking a transitive closure and add the new ones;

 - Trivial predicate simplification: we used the same framework as in the transitive predicates rewrite. This method removes trivial predicates like `a = a`, some of which are generated by the transitive predicates rule;

Constant folding can transform queries like `select * from table where t.column1 = 3*(2-1)` to `select * from table where t.column1 = 3`. These transformation are done in parser level instead of optimizer level because of implementation simplicity. And we don't really lose any information.
It support 3 type of constant folding 
1. Operators like +-*/ 
2. Comparison operators like >=, NOT, is null, 
3. Conjunctions : AND OR.


src/expression/comparison_expression.cpp
```diff
@@ -80,5 +80,56 @@ const std::string ComparisonExpression::GetInfo() const {
...
 
+bool ComparisonExpression::SymmetricEquals(
+    const AbstractExpression &other) const {
+  PELOTON_ASSERT(children_.size() == 2);
+
+  if (other.GetChildrenSize() != 2) {
+    return false;
+  }
+
+  auto vcl = children_[0].get();
```
nappelson: it isn't clear exactly what `vcl` stands for. would recommend spelling it out or at least adding a comment. same applies to the other acronyms below

src/expression/comparison_expression.cpp
```diff
@@ -80,5 +80,56 @@ const std::string ComparisonExpression::GetInfo() const {
...
+               (ocl->ExactlyEquals(*vcr) && ocr->ExactlyEquals(*vcl) &&
+                oexp_type_ == ExpressionType::COMPARE_LESSTHANOREQUALTO);
+      case (ExpressionType::COMPARE_DISTINCT_FROM):
+        return exp_type_ == oexp_type_ && ocl->ExactlyEquals(*vcl) &&
+               ocr->ExactlyEquals(*vcr);
+      default:
+        throw Exception("Invalid comparison expression type.");
+    }
+  } catch (std::exception &e) {
+    return false;
```
nappelson: you should log out the exception here unless if you follow my proposal earlier

src/expression/comparison_expression.cpp
```diff
@@ -80,5 +80,56 @@ const std::string ComparisonExpression::GetInfo() const {
...
+      case (ExpressionType::COMPARE_GREATERTHANOREQUALTO):
+        return (ocl->ExactlyEquals(*vcl) && ocr->ExactlyEquals(*vcr) &&
+                oexp_type_ == ExpressionType::COMPARE_GREATERTHANOREQUALTO) ||
+               (ocl->ExactlyEquals(*vcr) && ocr->ExactlyEquals(*vcl) &&
+                oexp_type_ == ExpressionType::COMPARE_LESSTHANOREQUALTO);
+      case (ExpressionType::COMPARE_DISTINCT_FROM):
+        return exp_type_ == oexp_type_ && ocl->ExactlyEquals(*vcl) &&
+               ocr->ExactlyEquals(*vcr);
+      default:
+        throw Exception("Invalid comparison expression type.");
```
nappelson: if you are simply catching the exception you throw here, you should probably just return false in `default` and log information that there was an invalid comparison expression type. If you want the exception to cause a failure, then you can throw but would recommend not catching here. If you are expecting another exception from this `try` block, then I'd recommend picking a finer grained exception in the catch statement.

src/include/optimizer/rule_query_rewrite.h
```diff
@@ -0,0 +1,73 @@
...
+class TransitivePredicatesLogicalGet : public Rule {
+ public:
+  TransitivePredicatesLogicalGet();
+
+  bool Check(std::shared_ptr<OperatorExpression> plan,
+             OptimizeContext *context) const override;
+
+  void Transform(std::shared_ptr<OperatorExpression> input,
+                 std::vector<std::shared_ptr<OperatorExpression>> &transformed,
+                 OptimizeContext *context) const override;
```
nappelson: i think OptimizeContext is usually passed around as a shared_ptr in other parts of the optimizer. you may want to do that for consistency and to gain the benefits of it

src/include/optimizer/rule_query_rewrite.h
```diff
@@ -0,0 +1,73 @@
...
+/**
+ * Implements transitive predicates generation for
+ * LogicalGet nodes:
+ *   a = b AND b = c -> a = b AND b = c AND a = c
+ */
+class TransitivePredicatesLogicalGet : public Rule {
+ public:
+  TransitivePredicatesLogicalGet();
+
+  bool Check(std::shared_ptr<OperatorExpression> plan,
```
nappelson: as you are declaring the same methods here and below for all of these classes, i'd recommend creating an abstract class that inherits from Rule so that you don't need to duplicate the declaration of the same `Check` and `Transform` methods

src/optimizer/rule_query_rewrite.cpp
```diff
@@ -0,0 +1,191 @@
...
+
+///////////////////////////////////////////////////////////////////////////////
+/// LogicalGet
+TransitivePredicatesLogicalGet::TransitivePredicatesLogicalGet() {
+  type_ = RuleType::TRANSITIVE_PREDICATES_LOGICAL_GET;
+
+  match_pattern = std::make_shared<Pattern>(OpType::Get);
+}
+
+bool TransitivePredicatesLogicalGet::Check(std::shared_ptr<OperatorExpression> input,
```
nappelson: add a TODO for what you actually want to do here. same for the other Check methods that do this

src/optimizer/util.cpp
```diff
@@ -23,6 +26,145 @@ namespace peloton {
 namespace optimizer {
 namespace util {
 
+void FillTransitiveTable(
```
nappelson: i think this method should have some inline documentation about what is happening here

src/optimizer/util.cpp
```diff
@@ -23,6 +26,145 @@ namespace peloton {
...
+}
+
+std::vector<AnnotatedExpression> GenerateTransitivePredicates(
+    std::vector<AnnotatedExpression> predicates,
+    std::map<PredicateInfo, std::vector<expression::AbstractExpression *>> &transitive_table) {
+  std::vector<AnnotatedExpression> new_predicates;
+
+  for (auto predicate : predicates) {
+    std::string exp_type = ExpressionTypeToString(predicate.expr->GetExpressionType());
+    if (exp_type.substr(0, 7) == "COMPARE") {
```
nappelson: i'd try and avoid using magic number (`7` in this case)

src/optimizer/optimizer.cpp
```diff
@@ -68,6 +71,20 @@ void Optimizer::OptimizeLoop(int root_group_id,
...
+    // Apply query rewrite rules
+    task_stack->Push(new TopDownRewrite(root_group_id, root_context,
+					RewriteRuleSetName::TRANSITIVE_PREDICATES));
+
+    ExecuteTaskStack(*task_stack, root_group_id, root_context);
+  }
+
+  // Apply query rewrite rules
+  task_stack->Push(new TopDownRewrite(root_group_id, root_context,
+                                      RewriteRuleSetName::SIMPLIFY_PREDICATES));
```
GustavoAngulo: Why is simplifying predicates executed after transitive predicates? Will this not apply the transitive predicate rule on trivial predicates that will just get thrown out by the simplify predicate rule? Seems like we can save some work by flipping the order of these

src/optimizer/rule_query_rewrite.cpp
```diff
@@ -0,0 +1,191 @@
...
+  
+  transformed.push_back(output);
+}
+
+///////////////////////////////////////////////////////////////////////////////
+/// LogicalFilter
+TransitivePredicatesLogicalFilter::TransitivePredicatesLogicalFilter() {
+  type_ = RuleType::TRANSITIVE_PREDICATES_LOGICAL_FILTER;
+
+  std::shared_ptr<Pattern> child(std::make_shared<Pattern>(OpType::InnerJoin));
```
GustavoAngulo: Why is the child for this a join and not a leaf? Is there a reason we should only apply this to filters with joins as children?

src/optimizer/rule_query_rewrite.cpp
```diff
@@ -0,0 +1,191 @@
...
+//===--------------------------------------------------------------------===//
+// Simplify predicates rules
+//===--------------------------------------------------------------------===//
+
+///////////////////////////////////////////////////////////////////////////////
+/// LogicalFilter
+SimplifyPredicatesLogicalFilter::SimplifyPredicatesLogicalFilter() {
+  type_ = RuleType::SIMPLIFY_PREDICATES_LOGICAL_FILTER;
+
+  std::shared_ptr<Pattern> child(std::make_shared<Pattern>(OpType::InnerJoin));
```
GustavoAngulo: Same comment as in earlier LogicalFilter rule, is there a reason for this to be an InnerJoin and not a leaf?

src/optimizer/util.cpp
GustavoAngulo: Agreed, you could probably do it like how it is done [here](https://github.com/cmu-db/peloton/pull/1340/files#diff-d1982d39ac0cf4b4426e20aa7f9c0b6aR225)

src/optimizer/util.cpp
```diff
@@ -23,6 +26,145 @@ namespace peloton {
...
+		new_predicates.emplace_back(AnnotatedExpression(transitive_predicate, table_alias_set));
+	      }
+	    }
+	  }
+	}
+      }
+    }
+  }
+
+  return new_predicates;
```
GustavoAngulo: Do we ever add the original predicate to the `new_predicates` list?

apavlo:
Closing. We will have @Zeninma build this in the fall with a proper rewrite layer.

PR #1343, opened 2018-05-05
[15721] Add Drop Index PR 2

alphalzh:
PR for 2nd code review, Group Add/Drop Index

Member: Xingyu Jin, Rong Huang, Ziheng Liao

Focus in this phase: fixed problems that were pointed out in the last PR (big thank to Andy & the alter table team!); added more tests to ensure the correctness of our solutions; came up with a method for the 100% and 125% goal (still implementing as you will see more commits to the PR very soon).

Dev notes: 
- During one of the tests, Xingyu discovered that in catalog level, creating index after inserting some entries into the table will result in the index not being populated. At first I thought Peloton doesn’t have logic for populating the index and tried to write one myself in the catalog level, but I discovered later that the populate table logic was written in separate parsers and executors. We decided to follow the original code structure and try to fix the PopulateIndexExecutor together with CreateExecutor, but that requires careful attention for other developers who tries to call create index related functions in levels lower than planner, since that will result in the index not being properly populated.
- Two of the major issues originated in the previous PR was resolved, we used shared_ptr for locks to avoid freeing locks still held by other transactions when we drop tables, and we used variable life cycle to automatically unlock locks once out of scope.
- Some very minor issues brought up in the last PR may not be addressed(mainly style differences based on different personal preferences).


src/executor/insert_executor.cpp
```diff
@@ -74,6 +75,18 @@ bool InsertExecutor::DExecute() {
...
+  oid_t table_oid = target_table->GetOid();
+  // Lock the table (reader lock)
+  concurrency::LockManager *lm = concurrency::LockManager::GetInstance();
+  bool lock_success = lm->LockShared(table_oid);
+  concurrency::LockManager::SafeLock dummy;
+  if (!lock_success) {
+    LOG_TRACE("Cannot obtain lock for the table, abort!");
+  }
+  else{
+    dummy.Set(table_oid, concurrency::LockManager::SafeLock::SHARED);
```
Dingshilun: Nice way to use object to release a lock.

test/concurrency/lock_manager_test.cpp
```diff
@@ -0,0 +1,115 @@
...
+}
+
+/**
+ * @brief Test lock upgrade/downgrade behaviour
+ */
+TEST_F(LockManagerTests, LockChangeTest) {
+  concurrency::LockManager *lm = concurrency::LockManager::GetInstance();
+  bool result0 = lm->InitLock((oid_t)1, concurrency::LockManager::RW_LOCK);
+  bool result1 = lm->LockShared((oid_t)1);
+  bool result2 = lm->LockToExclusive((oid_t)1);
```
Dingshilun: Just being curious, wouldn't this call stuck and waiting for the shared lock to be released?

test/concurrency/transaction_index_test.cpp
```diff
@@ -0,0 +1,186 @@
...
+      scheduler.Run();
+      EXPECT_EQ(ResultType::SUCCESS, scheduler.schedules[0].txn_result);
+      EXPECT_EQ(ResultType::ABORTED, scheduler.schedules[1].txn_result);
+      EXPECT_EQ(ResultType::SUCCESS, scheduler.schedules[2].txn_result);
+
+      EXPECT_EQ(1, scheduler.schedules[0].create_index_results[0]);
+      EXPECT_EQ(1, scheduler.schedules[1].drop_index_results[0]);
+      EXPECT_EQ(1, scheduler.schedules[2].drop_index_results[0]);
+    }
+  }
```
Dingshilun: Test cases LGTM.

src/executor/create_executor.cpp
```diff
@@ -220,19 +239,23 @@ bool CreateExecutor::CreateIndex(const planner::CreatePlan &node) {
 
   auto key_attrs = node.GetKeyAttrs();
 
+  auto table_object = catalog::Catalog::GetInstance()->GetTableObject(
```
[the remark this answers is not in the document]
alphalzh: Sorry! Used to have related logic here, forgot to delete redundant code

test/concurrency/lock_manager_test.cpp
alphalzh: It's LockToExclusive, which is to unlock shared and lock exclusive instead

test/sql/index_scan_sql_test.cpp
```diff
@@ -99,137 +99,6 @@ TEST_F(IndexScanSQLTests, CreateIndexAfterInsertOnMultipleColumnsTest) {
   catalog::Catalog::GetInstance()->DropDatabaseWithName(DEFAULT_DB_NAME, txn);
   txn_manager.CommitTransaction(txn);
 }
-TEST_F(IndexScanSQLTests, SQLTest) {
```
DeanChensj: Does these test corrupt?

test/concurrency/testing_transaction_util.cpp
```diff
@@ -212,6 +214,50 @@ storage::DataTable *TestingTransactionUtil::CreateTable(
...
+  txn_manager.CommitTransaction(txn);
+  LOG_INFO("create database %s", database_name.c_str());
+
+  txn = txn_manager.BeginTransaction();
+  catalog->CreateSchema(database_name, schema_name, txn);
+  txn_manager.CommitTransaction(txn);
+  LOG_INFO("create schema %s", schema_name.c_str());
+
+  txn = txn_manager.BeginTransaction();
+  catalog->CreateTable(database_name, schema_name, table_name, std::move(table_schema), txn);
```
DeanChensj: Should call CreateTableWithoutIndex() ?

test/include/concurrency/testing_transaction_util.h
```diff
@@ -68,7 +68,11 @@
 #include "storage/tile_group_header.h"
 #include "storage/tuple.h"
 #include "catalog/catalog_defaults.h"
+#include "catalog/catalog.h"
+#include "catalog/database_catalog.h"
 #include "common/internal_types.h"
+
+#include <assert.h>
```
DeanChensj: Do we need this header?

test/concurrency/testing_transaction_util.cpp
star013: No. It is just CreateTable. Although CreateTable creates primary index for that table, we want to build other index explicitly for the table and test insert, update, and delete on that index. There is no CreateTableWithoutIndex in the catalog.cpp

test/sql/index_scan_sql_test.cpp
star013: Actually we are moving to build junit test for SQL. Thus most tests here will be abandoned in the final pull request.

src/catalog/system_catalogs.cpp
```diff
@@ -80,11 +80,13 @@ SystemCatalogs::~SystemCatalogs() {
  * @param   txn              TransactionContext
  */
 void SystemCatalogs::Bootstrap(const std::string &database_name,
-                               concurrency::TransactionContext *txn) {
+                              UNUSED_ATTRIBUTE concurrency::TransactionContext *txn) {
```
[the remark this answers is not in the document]
star013: You are right. It is a legacy due to debugging procedure. We should delete that later.

src/catalog/catalog.cpp
```diff
@@ -239,13 +242,55 @@ ResultType Catalog::CreateDatabase(const std::string &database_name,
...
+  // put database object into rw_object_set
+  txn->RecordCreate(database_oid, INVALID_OID, INVALID_OID);
+  // Insert database record into pg_db
+  pg_database->InsertDatabase(database_oid, database_name, pool_.get(), txn);
+
+  // add core & non-core system catalog tables into database
+  LOG_INFO("begin bootstrap %s", database_name.c_str());
+  BootstrapSystemCatalogs(database, txn);
+  LOG_INFO("end bootstrap %s", database_name.c_str());
+//  catalog_map_[database_oid]->Bootstrap(database_name, txn);
```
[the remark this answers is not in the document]
star013: It is used in transaction_index_test. Someone changed logic in catalog after checkpoint 1, and we have to bypass those effects to allow our test work as previous.

src/catalog/catalog.cpp
```diff
@@ -764,6 +809,48 @@ ResultType Catalog::DropIndex(oid_t database_oid, oid_t index_oid,
...
+  // index_oid
+  auto database_object =
+      DatabaseCatalog::GetInstance()->GetDatabaseObject(database_name, txn);
+  if (database_object == nullptr)
+    throw CatalogException("Drop Index: database " + database_name +
+                           " does not exist");
+  auto database_oid = database_object->GetDatabaseOid();
+  auto pg_index = catalog_map_[database_oid]->GetIndexCatalog();
+  auto index_object = pg_index->GetIndexObject(index_name, schema_name, txn);
+  auto index_oid = index_object->GetIndexOid();
```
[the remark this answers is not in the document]
star013: Yes, it should be moved to line 838.

src/catalog/system_catalogs.cpp
```diff
@@ -80,11 +80,13 @@ SystemCatalogs::~SystemCatalogs() {
...
   LOG_DEBUG("Bootstrapping database: %s", database_name.c_str());
 
   if (!pg_trigger_) {
+    LOG_INFO("TriggerCatalog begin");
```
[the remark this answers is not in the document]
star013: No. They are legacy for debugging. During this project, we found several bugs in previous system and they were difficult to hunt.

src/catalog/catalog.cpp
```diff
@@ -239,13 +242,55 @@ ResultType Catalog::CreateDatabase(const std::string &database_name,
...
             database_name.c_str());
   return ResultType::SUCCESS;
 }
 
+
+ResultType Catalog::CreateDatabaseWithoutIndex(const std::string &database_name,
+                                   concurrency::TransactionContext *txn) {
+  if (txn == nullptr)
+    throw CatalogException("Do not have transaction to create database " +
+                           database_name);
```
sxzh93: [Good], good defensive coding

src/catalog/catalog.cpp
```diff
@@ -764,6 +809,48 @@ ResultType Catalog::DropIndex(oid_t database_oid, oid_t index_oid,
...
+  auto database_object =
+      DatabaseCatalog::GetInstance()->GetDatabaseObject(database_name, txn);
+  if (database_object == nullptr)
+    throw CatalogException("Drop Index: database " + database_name +
+                           " does not exist");
+  auto database_oid = database_object->GetDatabaseOid();
+  auto pg_index = catalog_map_[database_oid]->GetIndexCatalog();
+  auto index_object = pg_index->GetIndexObject(index_name, schema_name, txn);
+  auto index_oid = index_object->GetIndexOid();
+  if (index_object == nullptr) {
```
sxzh93: [Trivial], bracket usage style better to be consistent.

src/include/concurrency/lock_manager.h
```diff
@@ -0,0 +1,147 @@
...
+    }
+
+    // Set the oid and type for the lock
+    void Set(oid_t oid, RWLockType type){
+      oid_ = oid;
+      type_ = type;
+    }
+
+    // Unlock the given lock in the constructor
+    ~SafeLock(){
```
sxzh93: [Good], exploit object life cycle to release lock automatically. Maybe better to add a log trace here for debugging?

src/catalog/catalog.cpp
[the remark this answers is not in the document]
DeanChensj: It seems it's the only difference between CreateDatabase() and this function, maybe you should elaborate more on this and why the test crush.

src/include/concurrency/lock_manager.h
```diff
@@ -0,0 +1,147 @@
...
+   private:
+    oid_t oid_;
+    RWLockType type_;
+  };
+
+  // Type of locks. Currently only support RWLOCK
+  enum LockType { RW_LOCK };
+
+  // Constructor
+  LockManager() {}
```
sxzh93: If we use global singleton, it's better to put constructor into private region.

src/include/concurrency/lock_manager.h
```diff
@@ -0,0 +1,147 @@
...
+    // Try to access the lock
+    boost::upgrade_mutex *rw_lock;
+    try {
+      rw_lock = lock_map_.at(oid).get();
+    } catch (const std::out_of_range &oor) {
+      return nullptr;
+    }
+    return rw_lock;
+  }
+};
```
sxzh93: [Good], excellent lock manager implementation!

src/include/concurrency/lock_manager.h
```diff
@@ -0,0 +1,147 @@
...
+ private:
+  // Local RW lock to protect lock dict
+  boost::upgrade_mutex internal_rw_lock_;
+
+  // Map to store RW_LOCK for different objects
+  // Changed to shared_ptr for safety reasons
+  std::map<oid_t, std::shared_ptr<boost::upgrade_mutex> > lock_map_;
+
+  // Get RW lock by oid
+  boost::upgrade_mutex *GetLock(oid_t oid) {
```
DeanChensj: Why not return a shared ptr?

src/catalog/catalog.cpp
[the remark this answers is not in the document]
star013: You have good observations. Actually Bootstrap will refer many other variables like pg_table_metrics_ which is an instance of class TableMetricsCatalog. These variables are metadata about tables and they are not related to correctness of index. 
The correct configuration for those metadata is not related to our index implementation and is beyond our team's responsibility. Because of these two reasons, we simply ignored that function to accelerate our test and debug procedure in order to save time at stressful final weeks.

test/concurrency/lock_manager_test.cpp
```diff
@@ -0,0 +1,115 @@
...
+  bool result0 = lm->InitLock((oid_t)1, concurrency::LockManager::RW_LOCK);
+  // Try to remove lock
+  bool result1 = lm->RemoveLock((oid_t)1);
+  // Try to remove lock again
+  bool result2 = lm->RemoveLock((oid_t)1);
+  EXPECT_TRUE(result0);
+  EXPECT_TRUE(result1);
+  EXPECT_FALSE(result2);
+}
+
```
sxzh93: Add a test case to cover lock/unlock on a non-existent oid?

src/executor/insert_executor.cpp
```diff
@@ -74,6 +75,18 @@ bool InsertExecutor::DExecute() {
...
 
+  oid_t table_oid = target_table->GetOid();
+  // Lock the table (reader lock)
+  concurrency::LockManager *lm = concurrency::LockManager::GetInstance();
+  bool lock_success = lm->LockShared(table_oid);
+  concurrency::LockManager::SafeLock dummy;
+  if (!lock_success) {
+    LOG_TRACE("Cannot obtain lock for the table, abort!");
+  }
+  else{
```
DeanChensj: [Trivial] combine line 85 and 86

src/include/index/index.h
```diff
@@ -442,6 +442,8 @@ class Index : public Printable {
     indexed_tile_group_offset++;
   }
 
+  // TODO: add logic to protect index from being read during index build
```
DeanChensj: [Good] Add todo for future implementation.

test/include/concurrency/testing_transaction_util.h
```diff
@@ -175,6 +186,8 @@ struct TransactionSchedule {
   ResultType txn_result;
   std::vector<TransactionOperation> operations;
   std::vector<int> results;
```
DeanChensj: Name it to read_write_results to avoid confusion?

src/catalog/catalog.cpp
star013: Good suggestion, the files are written by different teammates, we would use consistent style in future.

test/include/concurrency/testing_transaction_util.h
star013: Yeah, the name is a little bit ambiguous. We will rename that variable in future.

src/include/concurrency/lock_manager.h
star013: Could you specify what is a singleton in your opinion?

PR #1347, opened 2018-05-05
[15721] Index Suggestion

sivaprasadsudhir:
This PR implements a search based algorithm for suggesting the best set of indexes for a given workload based on the AutoAdmin paper from Microsoft. 

This involves 
- WhatIfAPI - that returns the cost of a query for a hypothetical index configuration (brain/what_if_index, optimizer/optimizer)

- IndexSelection - Runs the search based algorithm. This involves 3 steps in each iteration - candidate index selection to choose indexes that benefit at least one query in the workload, configuration enumeration to explore the search space of the indexes and choose the top k indexes for the best configuration, and multi-column index generation to generate the indexes for the next iteration. (brain/index_selection*)


src/include/network/peloton_rpc_handler_task.h
```diff
@@ -13,22 +13,135 @@
...
 namespace network {
 class PelotonRpcServerImpl final : public PelotonService::Server {
  protected:
-  kj::Promise<void> createIndex(CreateIndexContext) override {
-    // TODO(tianyu) Write actual index code
-    LOG_DEBUG("Received rpc to create index");
+  kj::Promise<void> dropIndex(DropIndexContext request) override {
+    auto database_oid = request.getParams().getRequest().getDatabaseOid();
+    auto index_oid = request.getParams().getRequest().getIndexOid();
+    LOG_DEBUG("Database oid: %d", database_oid);
```
ChTimTsubasa: Change to LOG_TRACE
-- Tianyu, Justin & Tianyi

src/include/network/peloton_rpc_handler_task.h
```diff
@@ -13,22 +13,135 @@
...
+    LOG_DEBUG("Received RPC to create index");
+    auto database_oid = request.getParams().getRequest().getDatabaseOid();
+    auto table_oid = request.getParams().getRequest().getTableOid();
+    auto col_oids = request.getParams().getRequest().getKeyAttrOids();
+    auto is_unique = request.getParams().getRequest().getUniqueKeys();
+    auto index_name = request.getParams().getRequest().getIndexName();
+    std::vector<oid_t> col_oid_vector;
+    LOG_DEBUG("Database oid: %d", database_oid);
+    LOG_DEBUG("Table oid: %d", table_oid);
+    for (auto col : col_oids) {
```
ChTimTsubasa: Wrap this for loop in a IF_DEFINE DEBUG macro

-- Tianyu, Justin & Tianyi

src/include/network/peloton_rpc_handler_task.h
```diff
@@ -13,22 +13,135 @@
...
+      std::atomic_int *count = static_cast<std::atomic_int *>(arg);
+      count->store(0);
+    };
+
+    // Set the callback and context state.
+    auto &traffic_cop = tcop::TrafficCop::GetInstance();
+    traffic_cop.SetTaskCallback(callback, &counter);
+    traffic_cop.SetTcopTxnState(txn);
+
+    // Execute the plan through the traffic cop so that it runs on a separate
```
ChTimTsubasa: You are actually still waiting for the output here because you are waiting for the queueing flag to be reset.

-- Tianyu, Justin & Tianyi

src/include/network/peloton_rpc_handler_task.h
```diff
@@ -13,22 +13,135 @@
...
+    std::atomic_int counter;
+    std::vector<int> result_format;
+
+    auto callback = [](void *arg) {
+      std::atomic_int *count = static_cast<std::atomic_int *>(arg);
+      count->store(0);
+    };
+
+    // Set the callback and context state.
+    auto &traffic_cop = tcop::TrafficCop::GetInstance();
```
ChTimTsubasa: Singleton of traffic_cop is used in the test only and should not be in the engine's path. Actually, we should remove those from tests too.

-- Tianyu, Justin & Tianyi

src/include/catalog/abstract_catalog.h
```diff
@@ -67,6 +67,12 @@ class AbstractCatalog {
                          std::vector<type::Value> values,
                          concurrency::TransactionContext *txn) const;
 
+  std::unique_ptr<std::vector<std::unique_ptr<executor::LogicalTile>>>
+  GetResultWithIndexScan(std::vector<oid_t> column_offsets, oid_t index_offset,
```
Blade-Lee: nit: is it possible to pass `vector &` or `const vector &` instead?

src/include/network/peloton_rpc_handler_task.h
```diff
@@ -13,22 +13,135 @@
...
+      catalog->CreateIndex(database_oid, table_oid, col_oid_vector,
+                           DEFUALT_SCHEMA_NAME, index_name, IndexType::BWTREE,
+                           IndexConstraintType::DEFAULT, is_unique, txn);
+    } catch (CatalogException e) {
+      LOG_ERROR("Create Index Failed");
+      txn_manager.AbortTransaction(txn);
+      return kj::NEVER_DONE;
+    }
+
+    // Index created. Populate it.
```
ChTimTsubasa: Wouldn't create_index() in Catalog also populate tuples into it?  @mengranwo 

-- Tianyu, Justin & Tianyi

src/main/peloton/peloton.cpp
```diff
@@ -83,20 +79,41 @@ int main(int argc, char *argv[]) {
...
   int exit_code = 0;
   if (peloton::settings::SettingsManager::GetBool(
-      peloton::settings::SettingId::brain))
-    exit_code =  RunPelotonBrain();
+          peloton::settings::SettingId::brain))
+    exit_code = RunPelotonBrain();
   else
     exit_code = RunPelotonServer();
+
+  // TODO[Siva]: Remove this from the final PR. Uncomment this to run brain
```
Blade-Lee: remove this maybe?

src/brain/index_selection_job.cpp
```diff
@@ -0,0 +1,174 @@
...
+    is.GetBestIndexes(best_config);
+
+    if (best_config.IsEmpty()) {
+      LOG_INFO("Best config is empty");
+    }
+
+    // Get the existing indexes and drop them.
+    // TODO: Handle multiple databases
+    auto database_object = catalog::Catalog::GetInstance()->GetDatabaseObject(
+      DEFAULT_DB_NAME, txn);
```
ChTimTsubasa: This is should be addressed at least by having a wrapper function with database name as argument pass in. Multiple database handling is important especially after the catalog refactor.

-- Tianyu, Justin & Tianyi

src/include/network/peloton_rpc_handler_task.h
```diff
@@ -13,22 +13,135 @@
...
+    std::vector<ResultValue> result;
+    std::atomic_int counter;
+    std::vector<int> result_format;
+
+    auto callback = [](void *arg) {
+      std::atomic_int *count = static_cast<std::atomic_int *>(arg);
+      count->store(0);
+    };
+
+    // Set the callback and context state.
```
ChTimTsubasa: This is quite a hacky way to use Traffic_Cop. Probably should handout some general framework to handle database change for RPC request in the summer. At lease add a TODO here.

-- Tianyu, Justin & Tianyi

src/include/network/peloton_rpc_handler_task.h
sivaprasadsudhir: No, it doesn't.

test/brain/testing_index_selection_util.cpp
```diff
@@ -0,0 +1,335 @@
...
+
+TestingIndexSelectionUtil::~TestingIndexSelectionUtil() {
+  for (auto it = tables_created_.begin(); it != tables_created_.end(); it++) {
+    DropTable(it->first);
+  }
+  DropDatabase();
+}
+
+std::pair<std::vector<TableSchema>, std::vector<std::string>>
+TestingIndexSelectionUtil::GetQueryStringsWorkload(
```
saatviks: I think this is an awesome utility given that theres a lot of self-driving work that can use something like this now(like ours :)) and in the future. 
A few suggestions here would be:
1. Adding a metric utility here - which given the query and hypothetical configuration can provide the "what-if" cost directly. I know we can call your internal classes and do ourselves - but exposing an API would just be so much easier for everyone. Other metrics can then be added over time like the one we discussed today(number of tuples read.)
2. Is there any specific reason these specific workloads have been chosen - does each one have some specific characteristic - it might be worthwhile to appropriately name and add comments about these workloads.

test/brain/what_if_index_test.cpp
```diff
@@ -0,0 +1,488 @@
...
+  // The cost of using one index {1} should be greater than the cost
+  // of using both the indexes {1, 2} for the query.
+  EXPECT_EQ(result->plan->GetPlanNodeType(), PlanNodeType::INDEXSCAN);
+  LOG_DEBUG("Cost of the query with index {'b'}: %lf", cost_with_index_4);
+  LOG_DEBUG("%s", result->plan->GetInfo().c_str());
+
+  txn_manager.CommitTransaction(txn);
+}
+
+TEST_F(WhatIfIndexTests, MultiColumnTest2) {
```
saatviks: What is the exact difference between the different MultiColumnTests - it seems like its testing different workloads with different table setups and configurations. Is this the case or is it something specific - Might be useful to have brief comments?

src/main/peloton/peloton.cpp
```diff
@@ -83,20 +79,41 @@ int main(int argc, char *argv[]) {
...
+  // and server in the same process for testing. This is a temporary to way to
+  // run both peloton server and the brain together to test the index suggestion
+  // at the brain end without catalog replication between the server and the
+  // brain
+  // peloton::settings::SettingsManager::SetBool(
```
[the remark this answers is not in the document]
pbollimp: We will remove this in the final PR. We currently have it because we have different things to test for performance and this is a quick way to test the 'online' index suggestion (see IndexSelectionJob)

src/optimizer/cost_calculator.cpp
```diff
@@ -50,14 +53,73 @@ void CostCalculator::Visit(UNUSED_ATTRIBUTE const PhysicalIndexScan *op) {
   auto table_stats = std::dynamic_pointer_cast<TableStats>(
       StatsStorage::GetInstance()->GetTableStats(
           op->table_->GetDatabaseOid(), op->table_->GetTableOid(), txn_));
-  if (table_stats->GetColumnCount() == 0 || table_stats->num_rows == 0) {
+  auto index_scan_rows = (double)table_stats->num_rows;
+  if (table_stats->GetColumnCount() == 0 || index_scan_rows == 0) {
     output_cost_ = 0.f;
     return;
   }
+  auto index_object = op->table_->GetIndexObject(op->index_id);
```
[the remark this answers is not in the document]
pbollimp: Yes, it was not. @chenboy fixed it for us.

test/brain/index_selection_test.cpp
```diff
@@ -0,0 +1,710 @@
...
+  index_selection.GenerateCandidateIndexes(candidate_config, admissible_config,
+                                           workload);
+
+  LOG_DEBUG("Admissible Index Count: %ld", admissible_config.GetIndexCount());
+  LOG_DEBUG("Admissible Indexes: %s", admissible_config.ToString().c_str());
+  LOG_DEBUG("Candidate Indexes: %s", candidate_config.ToString().c_str());
+
+  EXPECT_EQ(admissible_config.GetIndexCount(), 2);
+  // TODO: There is no data in the table. Indexes should not help. Should return
+  // 0. But currently, the cost with index for a query if 0.0 if there are no
```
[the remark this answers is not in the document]
sivaprasadsudhir: What exactly do you mean by this?

test/brain/testing_index_selection_util.cpp
sivaprasadsudhir: 1. When we designed the utility, we thought of one that can be used for testing the What-If API as well. But if everyone needs it such an interface we will add it later.

2. These are some simple workloads which can expose all the (logical) branches in the index selection algorithm.

test/brain/what_if_index_test.cpp
sivaprasadsudhir: One of them is a very basic test. The second one is a more complex test with just one query in the workload. The third one uses workload with multiple queries.

src/catalog/index_catalog.cpp
```diff
@@ -270,6 +283,28 @@ std::shared_ptr<IndexCatalogObject> IndexCatalog::GetIndexObject(
...
+std::unordered_map<oid_t, std::shared_ptr<IndexCatalogObject>>
+IndexCatalog::GetIndexObjects(concurrency::TransactionContext *txn) {
+  std::unordered_map<oid_t, std::shared_ptr<IndexCatalogObject>> result_indexes;
+  if (txn == nullptr) {
+    throw CatalogException("Transaction is invalid!");
+  }
+  // try get from cache
+  auto pg_table = Catalog::GetInstance()
+                      ->GetSystemCatalogs(database_oid)
+                      ->GetTableCatalog();
```
malin1993ml: This is not the correct way to do this. We should not directly access the catalog instance. We should get the database catalog object from `txn->catalog_cache_`, then get table objects, and then index objects. Everything should go through the local catalog cache of the transaction but no the global catalog instance.

src/include/network/peloton_rpc_handler_task.h
```diff
@@ -11,24 +11,199 @@
...
+  }
+
+  static void UtilTestTaskCallback(void *arg) {
+    std::atomic_int *count = static_cast<std::atomic_int *>(arg);
+    count->store(0);
+  }
+
+  // TODO: Avoid using this function.
+  // Copied from SQL testing util.
+  // Execute a SQL query end-to-end
```
malin1993ml: Yeah this is a hack.... Is there a way to fix this?

src/optimizer/cost_calculator.cpp
```diff
@@ -50,14 +53,73 @@ void CostCalculator::Visit(UNUSED_ATTRIBUTE const PhysicalIndexScan *op) {
...
+  // i.e. till the first column with no equality predicate on it
+  // index cols (a, b, c)
+  // example1 : predicates(a=1 AND b=2 AND c=3) index helps on both a, b and c
+  // example2 : predicates(a<1 AND b<=2 and c<3) index helps on only a
+  // example3 : predicates(a=1 AND b>2 AND c>3) index helps on a and b
+  bool has_non_equality_pred = false;
+  for (size_t idx = 0; idx < key_attr_list.size(); ++idx) {
+    // If index cannot further reduce scan range, break
+    if (idx == op->key_column_id_list.size() ||
+        key_attr_list[idx] != op->key_column_id_list[idx]) {
```
malin1993ml: @chenboy It looks like this thing requires the `key_column_id_list` has exactly the same order as `key_attr_list`? But in fact, an index with col(a, b, c) can also help a query with predicates(b=2 AND a=1 AND c=3), right?

PR #1348, opened 2018-05-05
[15721] Logging and Recovery

aaron-tian:
TBA


src/logging/wal_recovery.cpp
```diff
@@ -0,0 +1,589 @@
...
+
+    else if(record_type==LogRecordType::TRANSACTION_BEGIN) {
+      LOG_INFO("Replaying TXN_BEGIN");
+    }
+
+    else if(record_type==LogRecordType::TRANSACTION_COMMIT) {
+      LOG_INFO("Replaying TXN_COMMIT");
+    }
+
+    else if(record_type==LogRecordType::TRANSACTION_ABORT) {
```
Zeninma: This check might need to be moved to ```ParseFromDisk``` and files an error there.
If a ```TRANSACTION_ABORT``` record were found at then at this time, then the previous changes in this transaction would have been made (which should not be, if I got this correctly).
On the other hand, if the transaction could reach here, then it should also have a ```TRANSACTION_COMMIT``` record.  I wonder if it is possible for a transaction to have both records. If not, then such check can be redundant.

src/logging/wal_recovery.cpp
```diff
@@ -0,0 +1,589 @@
...
+      continue;
+
+    auto offset_pair = std::make_pair(curr_txn_offset, curr_txn_offset);
+    commited_txns_.insert(std::make_pair(it->first, offset_pair));
+
+    curr_txn_offset += it->second.second;
+  }
+
+  // Pass 2
+  log_buffer_  = new char[log_buffer_size_];
```
Zeninma: I wonder if the recovery needs to read the log file twice from the disk. Since ```log_buffer_``` is constructed here after the first pass.
If this is the case, will it be better to populate the ```log_buffer_``` in the first phase and save the second disk read in the second phase?

src/logging/wal_recovery.cpp
```diff
@@ -0,0 +1,589 @@
...
+#include "concurrency/transaction_manager_factory.h"
+#include "catalog/catalog_defaults.h"
+#include "type/ephemeral_pool.h"
+#include "catalog/manager.h"
+#include "concurrency/epoch_manager_factory.h"
+#include "logging/wal_recovery.h"
+#include "logging/log_util.h"
+#include "common/logger.h"
+
+
```
nwang57: Please add some tests for the recovery code.

Issue #1325, opened 2018-04-22
Garbage collection doesn't reclaim in all scenarios


mbutrovich:
David and I completed our first comprehensive evaluation of the garbage collector. Based on the test suite we wrote [here](https://github.com/mbutrovich/peloton/blob/gc_fixes/test/gc/transaction_level_gc_manager_test.cpp), we've identified the following tuple-related problem scenarios:

1. Primary key constraint violation (FailedInsertPrimaryKeyTest)
- Insert tuple (0,0), commit
- Insert tuple (0,1), commit (primary index abort due to non-unique primary key)
- **Aborted insert still occupies a tuple slot in the table**
2. Secondary index constraint violation (FailedInsertSecondaryKeyTest)
- Insert tuple (0,1), commit
- Insert tuple (1,1), commit (secondary index abort due to non-unique value)
- **Aborted insert still occupies a tuple slot in the table**
- **Aborted insert present in primary index**
3. Committed update (CommitUpdateSecondaryKeyTest)
- Insert tuple (5,1), commit
- Update tuple to (5,2), commit
- **Old version still present in secondary index**
4. Aborted update (AbortUpdateSecondaryKeyTest)
- Insert tuple (0,1), commit
- Update tuple to (0,2), abort
- **New version still present in secondary index**
5. Insert and update in same transaction, commit (CommitInsertUpdateTest)
- Insert tuple (0,1), update to (0,2), commit
- **Old version present in both indexes, new version only present in one**
6. Insert and update in same transaction, abort (AbortInsertUpdateTest)
- Insert tuple (0,1), update to (0,2), abort
- **Old version present in one index**
7. Delete doesn't recycle tombstone, pollutes index (CommitDeleteTest)
- Insert tuple (0,1), commit
- Delete tuple (0,1), commit
- **Number of recycled slots only 1** (tombstone still in table)
- **Deleted tuple still present in both indexes**
8. Update and delete in same transaction, commit (CommitUpdateDeleteTest)
- Insert tuple (0,1), commit
- Update to (0,2), delete, commit
- **Number of recycled slots only 1**
- **Both versions of tuple present in both indexes**
9. Update and delete in same transaction, abort (AbortUpdateDeleteTest)
- Insert tuple (0,1), commit
- Update to (0,2), delete, abort
- **Number of recycled slots only 1**
- **Aborted update present in both indexes**
10. Update primary key (CommitUpdatePrimaryKeyTest)
- Insert tuple (3, 30), commit
[...]

PR #1349, opened 2018-05-05
[15721] TileGroup Compaction and GC Fixes

dqrs:
## Project: TileGroup Compaction, TileGroup Freeing, and Garbage Collection Fixes

Our project has 3 main aspects:

1. Enabling the Garbage Collector to free empty tile groups and reclaim their memory (Done)
2. Merging sparsely occupied tile groups to save memory (Done)
3. Fixing several important bugs in the Garbage Collector and Transaction Manager (Done)

# Status
**Completed 75% 100% and 125% goals**

# Summary

**1) TileGroup Freeing**
This PR enhances the Garbage Collector to free empty TileGroups when all of their tuple slots have been recycled. We also scanned the codebase for all points (outside of codegen) that fetch tile groups from the catalog without checking for a nullptr. TileGroups can no longer be assumed to live forever, so these checks must be done. We need to follow up with Prashanth to do a similar check in the codegen engine. It also created a new class, RecycleStack to replace  the Garbage Collector’s recycle queues. This concurrent data structure offers a non-blocking, constant-time TryPop() to retrieve a recycled tuple slot, a blocking, constant-time Push(), and a concurrent RemoveAllWithKey() function that uses hand-over-hand locking. The performance of TryPop() is very important because TryPop() lies on the critical path for DataTable::Insert(). We had to write a new data structure instead of using the existing concurrent queue because we need to be able to iterate through and remove recycled slots from TileGroups that become immutable. None of the existing concurrent structures support concurrent iteration, insertion, and removal.

**2) TileGroup Compaction**
This PR adds a class called TileGroupCompactor that performs compaction of tile groups. Compaction is triggered by the GarbageCollector, which submits a CompactTileGroup() task to the MonoQueuePool when the majority of a TileGroup is recycled garbage. The exact fraction of garbage required to trigger compaction is determined by a global setting, settings:SettingId::compaction_threshold. Compaction can also be enabled/disabled via another setting, global setting settings:SettingId::compaction_enabled. 

**3) Garbage Collection Fixes**
This PR resolves almost all of the issues identified in issue #1325. 
It includes the results of a thorough correctness audit we performed on Peleton's garbage collection system. It includes a whole new test suite for the Transaction-Level Garbage Collector and several important bug fixes to the Garbage Collector and Transaction Manager.  

**GC Fixes Summary**:

* Added 14 tests to transaction_level_gc_manager_test.cpp to handle more complex GC scenarios. 
* Currently 4 of these tests still fail for polluting indexes with old keys, but we believe this will require more significant changes at the execution layer to resolve.
*  We believe we have resolved all of the tuple-level GC bugs and most of the index bugs.
* We have disabled the 4 check that fail and will open a new issue describing those scenarios only.
* GCManager::RecycleTupleSlot allows unused ItemPointers to be returned without going through the entire Unlink and Reclaim process.
* Modified TOTransactionManager to pass tombstones created by deletes to the GCManager.
*Modified DataTable's Insert to return the ItemPointer to the GCManager in the case of a failed insert.
* Modified DataTable's InsertIntoIndexes to iterate through indexes and remove inserted keys in the event of a failure.
* Modified GCManager's Unlink function to clean indexes from garbage created by COMMIT_DELETE, COMMIT_UPDATE, and ABORT_UPDATE.


src/gc/tile_group_compactor.cpp
```diff
@@ -0,0 +1,194 @@
...
+
+    storage::DataTable *table =
+        dynamic_cast<storage::DataTable *>(tile_group->GetAbstractTable());
+
+    if (table == nullptr) {
+      return; // this table no longer exists
+    }
+
+    bool success = MoveTuplesOutOfTileGroup(table, tile_group);
+
```
suafeng: I find there is only one usage of function MoveTuplesOutOfTileGroup() and it is here. It seems that the tuples in the given tile group is always moved to the tile group's table. Then I think there may not be necessary to have a data table as argument in MoveTuplesOutOfTileGroup(), since we know that this table is this tile group's table.

src/gc/transaction_level_gc_manager.cpp
```diff
@@ -339,139 +278,127 @@ void TransactionLevelGCManager::AddToRecycleMap(
...
-  // Search for a slot that can be recycled
-  // TODO: We're relying on GetRecycledTupleSlot to clean the recycle queue. Fix this later.
-  while (recycle_queue->Dequeue(location) == true) {
-    auto tile_group_id = location.block;
-    auto tile_group = table->GetTileGroupById(tile_group_id);
+  // Try to get a slot that can be recycled
+  ItemPointer location = recycle_stack->TryPop();
+  if (location.IsNull()) {
+    return INVALID_ITEMPOINTER;
+  }
```
suafeng: I think it may be better for this function GetRecycledTupleSlot() to have a while around TryPop() instead of having while loop around GetRecycledTupleSlot, since this function name seems it guarantees to get a tuple slot.

src/gc/transaction_level_gc_manager.cpp
```diff
@@ -642,5 +569,66 @@ void TransactionLevelGCManager::UnlinkVersion(const ItemPointer location,
...
+    num_processed++;
+  }
+
+  return num_processed;
+}
 }  // namespace gc
 }  // namespace peloton
+
+
+
```
suafeng: You may not need so many blank lines here.

suafeng (commented):
Overall good job and code's quality seems to be good, I only add some comments about style and the design of functions.

src/gc/tile_group_compactor.cpp
```diff
@@ -0,0 +1,194 @@
...
+      LOG_TRACE("Failed to move tuple. Could not acquire ownership of tuple.");
+      txn_manager.SetTransactionResult(txn, ResultType::FAILURE);
+      txn_manager.AbortTransaction(txn);
+      return false;
+    }
+
+    // ensure that this is the latest version
+    bool is_latest_version = tile_group_header->GetPrevItemPointer(physical_tuple_id).IsNull();
+    if (is_latest_version == false) {
+      LOG_TRACE("Skipping tuple, not latest version.");
```
ssyssy: Why skip the tuple here? Why not just abort the transaction?

src/gc/tile_group_compactor.cpp
```diff
@@ -0,0 +1,194 @@
...
+        txn, tile_group_header, physical_tuple_id,
+        VisibilityIdType::COMMIT_ID);
+    if (visibility != VisibilityType::OK) {
+      // ignore garbage tuples because they don't prevent tile group freeing
+      continue;
+    }
+
+    LOG_TRACE("Moving Visible Tuple id : %u, Physical Tuple id : %u ",
+              visible_tuple_id, physical_tuple_id);
+
```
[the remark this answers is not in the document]
ssyssy: Same here.

src/include/gc/transaction_level_gc_manager.h
```diff
@@ -21,51 +21,66 @@
...
+          LockFreeQueue<concurrency::TransactionContext* >>(INITIAL_UNLINK_QUEUE_LENGTH));
+
     }
 
-    reclaim_maps_.clear();
-    reclaim_maps_.resize(gc_thread_count_);
-    recycle_queue_map_.clear();
+    immutable_tile_group_queue_ = std::make_shared<LockFreeQueue<oid_t>>(INITIAL_TG_QUEUE_LENGTH);
+    recycle_stacks_ = std::make_shared<peloton::CuckooMap<
+        oid_t, std::shared_ptr<RecycleStack>>>(INITIAL_MAP_SIZE);
```
ssyssy: I think recycle_stack_map is better for understanding.

test/concurrency/testing_transaction_util.cpp
```diff
@@ -212,6 +212,24 @@ storage::DataTable *TestingTransactionUtil::CreateTable(
...
+  std::vector<oid_t> key_attrs = {1};
+  auto tuple_schema = table->GetSchema();
+  bool unique = false;
+  auto key_schema = catalog::Schema::CopySchema(tuple_schema, key_attrs);
+  key_schema->SetIndexedColumns(key_attrs);
+  auto index_metadata2 = new index::IndexMetadata(
+      "unique_btree_index", 1235, TEST_TABLE_OID, CATALOG_DATABASE_OID,
+      IndexType::BWTREE, IndexConstraintType::UNIQUE, tuple_schema,
+      key_schema, key_attrs, unique);
+
```
ssyssy: What is the 1235 here? Some random number?

src/include/storage/tile_group_header.h
```diff
@@ -229,9 +230,15 @@ class TileGroupHeader : public Printable {
...
+  immutable flag to be true. GC must be notified in order to stop recycling
+   slots from it
   */
-  inline bool SetImmutability() {
+  bool SetImmutability();
+
+  /*
+  * @brief Set's Immutable Flag to True. Only used by the Garbage Collector
+  */
+  inline bool SetImmutabilityWithoutNotifyingGC() {
```
ssyssy: Where would this function be called?

ssyssy (commented):
Nice job and really solid code here. I only find some small issues here. Hope that my comments can help.

bohanjason:
As my teammates said,  Nice Job ! !

src/codegen/inserter.cpp
```diff
@@ -34,6 +34,8 @@ void Inserter::Init(storage::DataTable *table,
 char *Inserter::AllocateTupleStorage() {
   location_ = table_->GetEmptyTupleSlot(nullptr);
 
+  PELOTON_ASSERT(location_.IsNull() == false);
+
   // Get the tile offset assuming that it is in a tuple format
   auto tile_group = table_->GetTileGroupById(location_.block);
```
bohanjason: assert tilegroup != nullptr ?

src/gc/tile_group_compactor.cpp
```diff
@@ -0,0 +1,160 @@
...
+        txn_manager.AcquireOwnership(txn, tile_group_header, physical_tuple_id);
+    if (!acquired_ownership) {
+      LOG_TRACE("Failed to move tuple. Could not acquire ownership of tuple.");
+      txn_manager.SetTransactionResult(txn, ResultType::FAILURE);
+      txn_manager.AbortTransaction(txn);
+      return false;
+    }
+
+    // check again now that we have ownsership
+    // to ensure that this is stil the latest version
```
tcm-marcel: Typo: *ownership* and *still*

src/gc/transaction_level_gc_manager.cpp
```diff
@@ -153,13 +265,12 @@ int TransactionLevelGCManager::Unlink(const int &thread_id,
     }
 
     if (txn_ctx->GetEpochId() <= expired_eid) {
-      // as the global expired epoch id is no less than the garbage version's
-      // epoch id, it means that no active transactions can read the version. As
-      // a result, we can delete all the tuples from the indexes to which it
-      // belongs.
+      // since this txn's epochId is <= the global expired epoch id
+      // no active transactions can read the version. Asa result,
```
tcm-marcel: Typo: *As a*

src/include/gc/tile_group_compactor.h
```diff
@@ -0,0 +1,77 @@
...
+#include "concurrency/transaction_context.h"
+#include "concurrency/transaction_manager_factory.h"
+#include "executor/executor_context.h"
+#include "index/index.h"
+#include "settings/settings_manager.h"
+#include "storage/database.h"
+#include "storage/storage_manager.h"
+#include "storage/tile_group.h"
+#include "storage/tuple.h"
+#include "threadpool/mono_queue_pool.h"
```
tcm-marcel: Can you move all includes that are not needed here to the implementation file?

test/concurrency/testing_transaction_util.cpp
```diff
@@ -456,5 +474,83 @@ bool TestingTransactionUtil::ExecuteScan(
   }
   return true;
 }
+
+ResultType TestingTransactionUtil::UpdateTuple(storage::DataTable *table,
+                                               const int key) {
+  srand(15721);
```
tcm-marcel: Is it ok to have non-deterministic stuff in the test cases? Shouldn't they be reproducible at any chance?

tcm-marcel (dismissed):
I didn't dig deep into this, since I am not very familiar with the Transaction and Garbage Collector code. I can go over it in more detail if requested.

src/codegen/deleter.cpp
```diff
@@ -38,6 +38,7 @@ void Deleter::Delete(uint32_t tile_group_id, uint32_t tuple_offset) {
 
   auto *txn = executor_context_->GetTransaction();
   auto tile_group = table_->GetTileGroupById(tile_group_id);
+  PELOTON_ASSERT(tile_group != nullptr);
```
poojanilangekar: Why was this line added? Is this necessary? Or was it for debugging? 
Can it cause undefined behavior in Release mode?

src/codegen/inserter.cpp
```diff
@@ -34,6 +34,8 @@ void Inserter::Init(storage::DataTable *table,
 char *Inserter::AllocateTupleStorage() {
   location_ = table_->GetEmptyTupleSlot(nullptr);
 
+  PELOTON_ASSERT(location_.IsNull() == false);
```
poojanilangekar: Same question as above. Why was this added? Ideally, I believe these checks should be handled at a higher layer, probably the `DataTable`.

src/codegen/inserter.cpp
```diff
@@ -54,8 +56,11 @@ void Inserter::Insert() {
   auto *txn = executor_context_->GetTransaction();
   auto &txn_manager = concurrency::TransactionManagerFactory::GetInstance();
 
-  ContainerTuple<storage::TileGroup> tuple(
-      table_->GetTileGroupById(location_.block).get(), location_.offset);
+  auto tile_group = table_->GetTileGroupById(location_.block).get();
+  PELOTON_ASSERT(tile_group != nullptr);
```
poojanilangekar: Same question. Why was this added?

src/common/container/cuckoo_map.cpp
```diff
@@ -18,6 +18,9 @@
 #include "common/item_pointer.h"
 #include "common/logger.h"
 #include "common/macros.h"
+#include "common/container/lock_free_queue.h"
```
poojanilangekar: I don't think you need to include this file and `storage/data_table.h`.

src/common/container/lock_free_array.cpp
```diff
@@ -53,6 +53,7 @@ template <typename ValueType>
 void LOCK_FREE_ARRAY_TYPE::Erase(const std::size_t &offset,
                                  const ValueType &invalid_value) {
   LOG_TRACE("Erase at %lu", offset);
+  PELOTON_ASSERT(lock_free_array.size() > offset);
```
poojanilangekar: Again, was this for debugging? Or have you introduced some new logic that could cause this condition?

src/common/container/lock_free_array.cpp
```diff
@@ -107,6 +108,20 @@ bool LOCK_FREE_ARRAY_TYPE::Contains(const ValueType &value) const {
   return exists;
 }
 
+template <typename ValueType>
```
poojanilangekar: Two main concerns: 
1. What happens if another thread modifies the value between line 115 and 117?
2. This seems inefficient. What is the use case? Would using a `CuckooHash` be more efficient?

src/common/init.cpp
```diff
@@ -52,7 +52,7 @@ void PelotonInit::Initialize() {
   threadpool::MonoQueuePool::GetExecutionInstance().Startup();
 
   int parallelism = (CONNECTION_THREAD_COUNT + 3) / 4;
-  storage::DataTable::SetActiveTileGroupCount(parallelism);
+  storage::DataTable::SetDefaultActiveTileGroupCount(parallelism);
```
poojanilangekar: Nice job!

src/concurrency/timestamp_ordering_transaction_manager.cpp
```diff
@@ -580,6 +580,9 @@ void TimestampOrderingTransactionManager::PerformDelete(
   current_txn->RecordDelete(old_location);
 }
 
+// Performs Delete on a tuple that was created by the current transaction, and
+// never
```
poojanilangekar: Combine line 584 and 585.

poojanilangekar (changes requested):
Hey @mbutrovich! I am a little confused/concerned about your changes in the execution layer. 

Ideally, the GC tasks should be transparent to the execution layer. The storage layer might access a nullptr because it is at the physical level. But the execution layer should never see the effects of GC. The whole point of epoch based GC is that it only removes those blocks/tiles/data stores that aren't visible to the executor.  (Think of something along the lines of GC in the SkipList index, you never checked the pointers during traversal right?)

There are a bunch of changes in the `updater`, `inserter`, `deleter`, `hybrid_scan_executor`, `index_scan_executor` and `seq_scan_executor` which
shouldn't be there. 

I am yet to review the rest of the code but I think we should discuss this before the rest.

src/codegen/deleter.cpp
mbutrovich: GetTileGroupById can now return a nullptr, this is just a check that we don't hit that during make check and other Debug builds. That's what most of the ASSERTs added to the execution layer are looking for.

If you're suggesting that we make it impossible to return a nullptr to the execution level, then we need to revisit how we access TileGroups entirely and go only through an iterator that will always be valid.

src/common/container/lock_free_array.cpp
mbutrovich: No new logic. .at() will throw an exception that we never catch if it's out of bounds. Seems more correct to me to have a PELOTON_ASSERT first. Also, see Find() and Update() which already have that logic. Erase() not having it just looks like an oversight.

src/codegen/deleter.cpp
poojanilangekar: I don't think you need to touch the execution layer. In a correct implementation, the execution layer should be completely oblivious to the GC. 

_**GetTileGroupById can now return a nullptr if the TileGroup has been dropped.**_ -> That is my biggest concern. So this thing is checked in debug which is great. But what happens in release mode? If this function gets a nullptr, it hits a segfault and peloton crashes? That's not expected behavior.

Like I said, in my overall comment, the whole point of the epoch based garbage collection is that the `TileGroup` references are removed only when you are certain that no other transaction is accessing it.  This is very similar to GC in the SkipListIndex.  The storage may get these `nullptr` references, but the execution layer should never get such a reference. 

I would suggest that we can either do a hangouts call over the weekend. Or you talk to @apavlo about this.

PR #1352, opened 2018-05-09
Modified pg_column_stats initialization

poojanilangekar:
This PR modifies the `ColumnStatsCatalog` constructor to use a predefined schema instead of using DDL. Additionally, it creates the `pg_column_stats` table to a per database basis. Initially this was not done because of the dependencies of the `StatsStorage` on the view of "Global Stats". 

I think this needs to be changed because right now, the catalog table are no longer stored in a separate database. 

@GustavoAngulo @chenboy Currently, `AnalyzeStatsForAllTables` is only used in tests. If that's the case, I think it should be deprecated.  Also can you please check the schema and the indexes of the `pg_column_stats` table. It looks like documentation and definitions are inconsistent. 

@gandeevan @aaron-tian You can use this patch. But it will create conflicts with the Layout PR.  So you may want to use it for the class and then change it because we might not merge this change.


apavlo:
@poojanilangekar Nice!

poojanilangekar:
@mengranwo @GustavoAngulo We need to merge this in ASAP as its blocking the performance benchmarks. Please take a look at it when you get the time.

GustavoAngulo:
I think it might be good to keep `AnalyzeStatsForAllTables ` for now, my group in the class this semester noticed that we don't support the default `ANALYZE` behavior like [Postgres does](https://www.postgresql.org/docs/9.1/static/sql-analyze.html), where not specifying a table analyzes all tables. We might want to support this in the future, so that function/logic might come in handy.

As for the the schema, I'm not sure which inconsistency you're referring to @poojanilangekar ? @chenboy do you see an inconsistency?

poojanilangekar:
@GustavoAngulo I have modified the function to `AnalyzeStatsForAllTablesWithDatabaseOid`. I checked the Postgres documentation, the `ANALYZE` command is run on a per database basis. So I think this should solve the problem. 

Regarding the inconsistency, the declaration [creates two SKEY indexes ](https://github.com/cmu-db/peloton/blob/d052644e7922480ea03bceada59527997b0b9ca5/src/catalog/column_stats_catalog.cpp#L46)and the table has no primary key.  While the header file states that the table should contain only one [index with the primary key](https://github.com/cmu-db/peloton/blob/d052644e7922480ea03bceada59527997b0b9ca5/src/include/catalog/column_stats_catalog.h#L28).

chenboy:
The `ColumnStatsCatalog` use both indexes.

[](https://github.com/cmu-db/peloton/blob/d68ab719fdd9499f23ac5887f907ed07ce4d2644/src/catalog/column_stats_catalog.cpp#L135)

[](https://github.com/cmu-db/peloton/blob/d68ab719fdd9499f23ac5887f907ed07ce4d2644/src/catalog/column_stats_catalog.cpp#L186)

So I think we should stick to the declaration, not the header.

poojanilangekar:
@GustavoAngulo @camellyx Can you please review these changes?

tli2:
@pervazea Can you take a look since these other people are out of town?

apavlo:
This is an important PR that we should merge. We don't want to lose track of this.

src/include/catalog/column_stats_catalog.h
```diff
@@ -47,50 +45,52 @@ namespace catalog {
...
-  static ColumnStatsCatalog *GetInstance(
-      concurrency::TransactionContext *txn = nullptr);
+  ~ColumnStatsCatalog();
 
   //===--------------------------------------------------------------------===//
   // write Related API
   //===--------------------------------------------------------------------===//
-  bool InsertColumnStats(oid_t database_id, oid_t table_id, oid_t column_id,
-                         int num_rows, double cardinality, double frac_null,
+  bool InsertColumnStats(oid_t table_id, oid_t column_id, int num_rows,
```
pervazea: Add comment documenting function purpose and arguments.

src/include/catalog/column_stats_catalog.h
```diff
@@ -47,50 +45,52 @@ namespace catalog {
...
+                         double cardinality, double frac_null,
                          std::string most_common_vals,
                          std::string most_common_freqs,
                          std::string histogram_bounds, std::string column_name,
                          bool has_index, type::AbstractPool *pool,
                          concurrency::TransactionContext *txn);
-  bool DeleteColumnStats(oid_t database_id, oid_t table_id, oid_t column_id,
+  bool DeleteColumnStats(oid_t table_id, oid_t column_id,
```
pervazea: Add a comment documenting function purpose and arguments.

src/include/catalog/column_stats_catalog.h
```diff
@@ -47,50 +45,52 @@ namespace catalog {
...
                          concurrency::TransactionContext *txn);
 
   //===--------------------------------------------------------------------===//
   // Read-only Related API
   //===--------------------------------------------------------------------===//
   std::unique_ptr<std::vector<type::Value>> GetColumnStats(
-      oid_t database_id, oid_t table_id, oid_t column_id,
-      concurrency::TransactionContext *txn);
+      oid_t table_id, oid_t column_id, concurrency::TransactionContext *txn);
```
pervazea: Add comment document function, args and return information.

src/include/catalog/system_catalogs.h
```diff
@@ -59,6 +60,13 @@ class SystemCatalogs {
     return pg_attribute_;
   }
 
+  ColumnStatsCatalog *GetColumnStatsCatalog() {
+    if (!pg_column_stats_) {
```
pervazea: Is this a recoverable error? If not, this should be a PELOTON_ASSERT.

src/include/optimizer/stats/stats_storage.h
```diff
@@ -71,8 +67,8 @@ class StatsStorage {
 
   /* Functions for triggerring stats collection */
 
-  ResultType AnalyzeStatsForAllTables(
-      concurrency::TransactionContext *txn = nullptr);
+  ResultType AnalyzeStatsForAllTablesWithDatabaseOid(
```
pervazea: Add function header describing this function, args, etc.

src/optimizer/stats/stats_storage.cpp
```diff
@@ -109,20 +97,21 @@ void StatsStorage::InsertOrUpdateColumnStats(
...
     single_statement_txn = true;
     txn = txn_manager.BeginTransaction();
   }
-  column_stats_catalog->DeleteColumnStats(database_id, table_id, column_id,
-                                          txn);
-  column_stats_catalog->InsertColumnStats(
-      database_id, table_id, column_id, num_rows, cardinality, frac_null,
-      most_common_vals, most_common_freqs, histogram_bounds, column_name,
-      has_index, pool_.get(), txn);
+  pg_column_stats->DeleteColumnStats(table_id, column_id, txn);
```
pervazea: DeleteColumnStats if they exist, I assume. May be helpful to add comment stating that, if correct.

test/optimizer/stats_storage_test.cpp
```diff
@@ -108,14 +110,18 @@ TEST_F(StatsStorageTests, InsertAndGetTableStatsTest) {
                                           table_stats_collector.get());
 
   VerifyAndPrintColumnStats(data_table.get(), 4);
```
pervazea: If I understand correctly, this function VerifyAndPrint... does print but very little verify. Acknowledge that this wasn't modified, but it does mean some of the old tests are not particularly useful.

Issue #1350, opened 2018-05-06
EXPLAIN handle push down


ChTimTsubasa:
The way #1108 add support to EXPLAIN is to extract the explained statement directly from query string at protocol handler layer, which is not the best place to handle EXPLAIN for 2 reasons:

1. Pointed out by @nappelson, #1108 only support EXPLAIN for Postgres Simple Protocol and would fail Postgres Extended Protocol used by JDBC. This is because Optimizer does not support handling EXPLAIN.

2. Protocol handler should only handle logic to protocol-specific things such as packet format serializing and deserializing. This is also true for PREPARE and EXECUTE, which would be pushed down too in a later fix.

The correct way to do this is to push down the logic on optimizer and let it identify the special statements and report to traffic_cop and traffic_cop should schedule special treatments to those plans.

apavlo:
What if we just made a special `ExplainExecutor` that just generates the plan string and then outputs that as a single tuple?

chenboy:
We just looked at Postgres and found they handle all utility commands in the traffic cop, I'm not quite sure why they do it this way though.

ChTimTsubasa:
I would say one reason they do that in this way is that utility queries would touch some of the frontend components during execution. For example, EXPLAIN would need optimizer, PREPARE and EXECUTE would need plan cache (Statement Cache in our system)

ChTimTsubasa:
Utility commands are all executed in https://github.com/postgres/postgres/blob/master/src/backend/tcop/utility.c#L337
While generally executed in https://github.com/postgres/postgres/blob/master/src/backend/executor/execMain.c

PR #1357, opened 2018-05-12
Handle EXPLAIN in the optimizer and executor

chenboy:
This PR moves the handling logic of `EXPLAIN` from the network layer to the execution layer. One reason for this is we don't want to duplicate the logic for different protocols, e.g. psql and JDBC, as it is mentioned in #1350.

Two test cases are added, one of which is a SQL test for the functionality of the `EXPLAIN` executor. Since we also touch the network layer, an end-to-end test is also added to make sure we didn't break anything in the network layer.


chenboy:
Blocked by #1368

PR #1371, opened 2018-05-17
Support COPY for CSV files

pmenon:
### Summary 
This PR adds support for psql's `COPY` command for bulk loading CSV files into the database. Using `COPY`, one can quickly load millions of rows into a table in a few seconds. For example, I was able to load 20M rows into a table with four integer columns in under three seconds; I also loaded an SF-1 `lineitem` table from TPC-H in about 20 seconds (which isn't great, but faster than through oltpbench).  I find it very convenient to use this to quickly load a crap-tonne of data into the database fast to do benchmarks.

Right now, we only support CSV files, but the quoting, escaping, and delimiter characters can be configured. I tried to make the parser fairly robust to erroneous files, but we're not as generous as Postgres (which goes to great lengths to try to understand your CSV).

### Modifications
1. Changes in the parser to look for `format`, `delimiter`, `escape`, and `quote` characters.
2. Added logical and physical `ExternalFileScan` and `ExportExternalFile` operators to optimizer for copy-from and copy-to.
3. Added `CSVScanPlan` and `ExportExternalFilePlan` to planner.
4. Added `CSVScanTranslator` to codegen. Also added runtime helper `CSVScanner` that accepts a callback function to invoke per-row in the CSV.
5. Added integer, decimal, string, and date parsing logic into the `functions` namespace.
6. Lots of tests.

### Reviewers

- @chenboy Can you take a look at the changes to the optimizer? I didn't add any costs (though one probably could use the schema and file size to estimate the number of rows).
- @tcm-marcel Want to take a stab at this?


src/codegen/buffering_consumer.cpp
```diff
@@ -40,6 +40,15 @@ WrappedTuple &WrappedTuple::operator=(const WrappedTuple &o) {
   return *this;
 }
 
+std::string WrappedTuple::ToCSV() const {
+  std::string ret;
+  for (uint32_t i = 0; i < tuple_.size(); i++) {
+    if (i != 0) ret.append(",");
+    ret.append(tuple_[i].ToString());
```
apavlo: The problem with this is that we are not escaping commas. Is there a CSV library that we could use?

src/codegen/codegen.cpp
```diff
@@ -127,26 +151,52 @@ llvm::Value *CodeGen::CallFunc(llvm::Value *fn,
   return GetBuilder().CreateCall(fn, args);
 }
 
-llvm::Value *CodeGen::CallPrintf(const std::string &format,
-                                 const std::vector<llvm::Value *> &args) {
+llvm::Value *CodeGen::Printf(const std::string &format,
+                             const std::vector<llvm::Value *> &args) {
   auto *printf_fn = LookupBuiltin("printf");
   if (printf_fn == nullptr) {
+#if GCC_AT_LEAST_6
```
apavlo: Can you add a comment to explain what this does?

src/codegen/codegen.cpp
```diff
@@ -127,26 +151,52 @@ llvm::Value *CodeGen::CallFunc(llvm::Value *fn,
...
+  // Call printf()
   return CallFunc(printf_fn, printf_args);
 }
 
+llvm::Value *CodeGen::Memcmp(llvm::Value *ptr1, llvm::Value *ptr2,
+                             llvm::Value *len) {
+  static constexpr char kMemcmpFnName[] = "memcmp";
+  auto *memcmp_fn = LookupBuiltin(kMemcmpFnName);
+  if (memcmp_fn == nullptr) {
+#if GCC_AT_LEAST_6
```
apavlo: Again, please explain why we need this.

src/codegen/operator/csv_scan_translator.cpp
```diff
@@ -0,0 +1,307 @@
...
+  // Call CSVScanner::Init()
+  codegen.Call(CSVScannerProxy::Init,
+               {scanner_ptr, exec_ctx_ptr, file_path, output_col_types,
+                codegen.Const32(num_cols), consumer_func, query_state_ptr,
+                codegen.Const8(scan.GetDelimiterChar()),
+                codegen.Const8(scan.GetQuoteChar()),
+                codegen.Const8(scan.GetEscapeChar())});
+}
+
+namespace {
```
apavlo: ?

src/codegen/type/type.cpp
```diff
@@ -18,15 +18,20 @@ namespace peloton {
...
+  aux_info.numeric_info.scale = 0;
+}
+
+Type::Type() : Type(peloton::type::TypeId::INVALID, false) {}
 
 Type::Type(const SqlType &sql_type, bool _nullable)
     : Type(sql_type.TypeId(), _nullable) {}
 
 bool Type::operator==(const Type &other) const {
+  // TODO(pmenon): This isn't correct; we need to check all other fields ...
```
apavlo: Shouldn't this be an easy fix?

src/codegen/codegen.cpp
pmenon: I've added a short comment explaining this.

TL;DR; GCC 6+ complains when it doesn't see you using all function attributes (i.e., nothrow, throw, notnull etc.). We use `decltype` to get the signature for some functions, so we don't need the attributes. This fails compilation. We ignore this warning for this section of code.

src/codegen/codegen.cpp
pmenon: I've added a short comment explaining this.

TL;DR; GCC 6+ complains when it doesn't see you using all function attributes (i.e., nothrow, throw, notnull etc.). We use `decltype` to get the signature for some functions, so we don't need the attributes. This fails compilation. We ignore this warning for this section of code.

src/codegen/buffering_consumer.cpp
pmenon: This is just a utility function to pretty print tuples for debugging purposes.  It isn't used in execution.

src/codegen/type/type.cpp
pmenon: I tried this, but it's a little more involved because of some assumptions we make in the code. We can probably address this in a separate PR.

src/codegen/operator/csv_scan_translator.cpp
pmenon: This is just an anonymous namespace. I use it so I can create classes in `.cpp` files without worrying about name collisions.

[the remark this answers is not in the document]
pmenon: This is just an anonymous namespace. I use it so I can create classes in `.cpp` files without worrying about name collisions.

src/codegen/util/csv_scanner.cpp
```diff
@@ -0,0 +1,375 @@
...
+      last_was_escape = !last_was_escape;
+    }
+    if (c == quote && !last_was_escape) {
+      in_quote = !in_quote;
+    }
+    if (c != escape) {
+      last_was_escape = false;
+    }
+
+    // Process the new-line character. If we a new-line and we're not currently
```
tcm-marcel: Typo in comment

src/codegen/util/csv_scanner.cpp
```diff
@@ -0,0 +1,375 @@
...
+    // in a quoted section, we're done.
+    if (c == '\n' && !in_quote) {
+      break;
+    }
+  }
+
+  // Flush remaining valid bytes
```
tcm-marcel: I don't get this part, can you explain briefly? (When does this occur?)

src/codegen/util/csv_scanner.cpp
```diff
@@ -0,0 +1,375 @@
...
+              continue;
+            }
+          }
+
+          // If we see the closing quote, we're done.
+          if (c == quote) {
+            break;
+          }
+
+          *out++ = c;
```
tcm-marcel: Do I understand correctly, that this line basically writes the current character to the same position in the same buffer - except for the offset that was created by quotes and escapes.

src/codegen/util/csv_scanner.cpp
```diff
@@ -0,0 +1,375 @@
...
+        // the loop if a quote character is found. It flows into a second loop
+        // whose only purpose is to find the end of the quoted section.
+        while (true) {
+          char c = *iter++;
+
+          // If we see the delimiter character, or the end of the string,
+          // finish
+          if (c == delimiter || c == '\0') {
+            col_end = out;
+            iter--;
```
tcm-marcel: Why is `iter` decremented here? The next column loop iteration will read the delimiter or null-byte again, will it?

src/codegen/operator/csv_scan_translator.cpp
```diff
@@ -0,0 +1,307 @@
...
+
+    Vector v{nullptr, 1, nullptr};
+    RowBatch one{GetCompilationContext(), codegen.Const32(0),
+                 codegen.Const32(1), v, false};
+
+    // Load the pointer to the columns view
+    llvm::Value *cols = codegen->CreateLoad(codegen->CreateConstInBoundsGEP2_32(
+        CSVScannerProxy::GetType(codegen), LoadStatePtr(scanner_id_), 0, 1));
+
+    llvm::Value *null_str = codegen.ConstString(scan.GetNullString(), "null");
```
tcm-marcel: Is there a chance that not all types have "null" as null-string?

src/function/numeric_functions.cpp
```diff
@@ -0,0 +1,439 @@
...
+      codegen::RuntimeFunctions::ThrowOverflowException();
+      __builtin_unreachable();
+    } else {
+      codegen::RuntimeFunctions::ThrowInvalidInputStringException();
+      __builtin_unreachable();
+    }
+  }
+
+  // Eat the rest
+  while (consumed_ptr < end && *consumed_ptr == ' ') {
```
tcm-marcel: Does this mean, we basically only allow spaces after the actual parsed decimal?

src/codegen/operator/csv_scan_translator.cpp
```diff
@@ -0,0 +1,307 @@
...
+
+    // Add accessors for all columns into the row batch
+    std::vector<CSVColumnAccess> column_accessors;
+    for (uint32_t i = 0; i < output_attributes_.size(); i++) {
+      column_accessors.emplace_back(output_attributes_[i], cols,
+                                    scan.GetNullString(), null_str);
+    }
+    for (uint32_t i = 0; i < output_attributes_.size(); i++) {
```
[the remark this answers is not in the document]
pmenon: It needs to be two loops because we build up a vector in the first loop and insert pointers to vector elements in the second loop. If it's in the same loop, the pointers may be invalid due to resizing.

We **could** reserve space beforehand and use one loop. We use this pattern in other places.

src/codegen/util/csv_scanner.cpp
```diff
@@ -0,0 +1,375 @@
...
+
+void CSVScanner::Init(CSVScanner &scanner,
+                      executor::ExecutorContext &executor_context,
+                      const char *file_path,
+                      const codegen::type::Type *col_types, uint32_t num_cols,
+                      CSVScanner::Callback func, void *opaque_state,
+                      char delimiter, char quote, char escape) {
+  // Forward to constructor
+  new (&scanner)
+      CSVScanner(*executor_context.GetPool(), file_path, col_types, num_cols,
```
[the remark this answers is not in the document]
pmenon: Correct, the `CSVScanner` is created and initialized at runtime per-invocation.

In fact, all the classes in the `util` directory are runtime classes. We should rename it to `peloton::codegen::runtime` to make this clear.

src/codegen/util/csv_scanner.cpp
pmenon: Good catch, will fix.

src/codegen/operator/csv_scan_translator.cpp
pmenon: The second `null` argument is the name of the constant variable, not the value we're comparing with. A null string exists in the plan node.

src/codegen/util/csv_scanner.cpp
pmenon: Essentially, all bytes between `buffer_pos_` and `curr_buffer_pos` are valid bytes that should be part of the current line. It is possible that we exit the loop not having written all valid bytes to the line buffer. This section ensures we flush these valid bytes before retuning.

src/codegen/util/csv_scanner.cpp
pmenon: Correct, we're writing non-quote characters (unless they're escaped).

src/codegen/util/csv_scanner.cpp
pmenon: The invariant we want to hold is that when this loop exists, `iter` either points to a newline character or a delimiter. We decrement here because we incremented earlier on line 300. Happy to take suggestions on how to make this clearer. CSV parsing is tricky :sweat:

src/function/numeric_functions.cpp
pmenon: We allow spaces both before and after actually. `std::strtod` handles spaces beforehand. We handle trailing spaces.

pmenon:
@tcm-marcel I didn't use `mmap` mostly because it wouldn't help performance. This is a giant sequential read so IO prefetching will kick in. If not, we can `fadvise` a hint. We're doing `read()` which bypasses the page cache, avoiding duplicated data.

There is an argument in software engineering simplicity using `mmap`, but the interaction would be similar - instead of copying from a temporary buffer managed by us to a line buffer, we'd be copying from a temporary kernel `mmap` buffer into our line buffer.

tcm-marcel:
@pmenon Thank you for the explanation! I didn't know `read()` bypasses the page cache.

pmenon:
@tcm-marcel I misspoke. `read()` will use the page cache unless we use direct IO - we don't do this. I was thinking of `fread()` that does library-level buffering that we avoid.

PR #1379, opened 2018-05-23
Resolves Clang warnings and LLVM assertion failure on Mac

mbutrovich:
Resolves the following introduced after merging PR #1304:

/peloton/src/codegen/function_builder.cpp:137:5: error: ‘~FunctionBuilder’ has a non-throwing exception specification but can still throw [-Werror,-Wexceptions]
/peloton/src/codegen/function_builder.cpp:135:18: note: destructor has a implicit non-throwing exception specification

/peloton/src/codegen/operator/table_scan_translator.cpp:219:37: error: lambda capture ‘table’ is not used [-Werror,-Wunused-lambda-capture]

/peloton/src/codegen/util/sorter.cpp:271:56: error: lambda capture ‘comp’ is not used [-Werror,-Wunused-lambda-capture]

/peloton/test/codegen/hash_table_test.cpp:100:17: error: variable ‘c1’ cannot be implicitly captured in a lambda with no capture-default specified
/peloton/test/codegen/hash_table_test.cpp:142:17: error: variable ‘c1’ cannot be implicitly captured in a lambda with no capture-default specified
/peloton/test/codegen/hash_table_test.cpp:189:17: error: variable ‘c1’ cannot be implicitly captured in a lambda with no capture-default specified
/peloton/test/codegen/hash_table_test.cpp:222:43: error: lambda capture ‘to_insert’ is not required to be captured for this use [-Werror,-Wunused-lambda-capture]

Assertion failed: (isValidElementType(EltTy) && “Invalid type for pointer element!“), function get, file /private/tmp/llvm@3.7-20170923-4009-1tn3kci/llvm-3.7.1.src/lib/IR/Type.cpp, line 740.

### DIFF
--- a/src/codegen/function_builder.cpp
+++ b/src/codegen/function_builder.cpp
@@ -131,10 +131,13 @@ FunctionBuilder::FunctionBuilder(
           cc,
           ConstructFunction(cc, name, FunctionDeclaration::Visibility::External,
                             ret_type, args)) {}
-                            
+
 FunctionBuilder::~FunctionBuilder() {
   if (!finished_) {
-    PELOTON_ASSERT(false); // Missing call to FunctionBuilder::ReturnAndFinish()
+    LOG_ERROR(
+        "Missing call to FunctionBuilder::ReturnAndFinish() for function '%s'",
+        func_->getName().data());
+    PELOTON_ASSERT(false);
   }
 }
 
--- a/src/codegen/function_builder.cpp
+++ b/src/codegen/function_builder.cpp
@@ -137,7 +137,6 @@ FunctionBuilder::~FunctionBuilder() {
     LOG_ERROR(
         "Missing call to FunctionBuilder::ReturnAndFinish() for function '%s'",
         func_->getName().data());
-    PELOTON_ASSERT(false);
   }
 }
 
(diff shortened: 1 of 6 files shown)


src/codegen/function_builder.cpp
```diff
@@ -131,10 +131,13 @@ FunctionBuilder::FunctionBuilder(
...
                             ret_type, args)) {}
-                            
+
 FunctionBuilder::~FunctionBuilder() {
   if (!finished_) {
-    throw Exception{"Missing call to FunctionBuilder::ReturnAndFinish()"};
+    LOG_ERROR(
+        "Missing call to FunctionBuilder::ReturnAndFinish() for function '%s'",
+        func_->getName().data());
+    PELOTON_ASSERT(false);
```
pmenon: On second thought, let's remove this assertion. In `Debug` mode, the destructor gets called on the way up through the call stack when an exception is thrown. We don't want this assertion to trigger here and potentially hide any exception context.

test/codegen/hash_table_test.cpp
```diff
@@ -219,7 +219,7 @@ TEST_F(HashTableTest, ParallelMerge) {
   };
 
   // Insert function
-  auto insert_fn = [&add_key, &exec_ctx, &to_insert](uint64_t tid) {
+  auto insert_fn = [&add_key, &exec_ctx](uint64_t tid) {
```
pmenon: Tricky capture semantics :confused:

PR #1380, opened 2018-05-23
Performance Fixes: pass TileGroupHeader*, avoid RW set lookups

tcm-marcel:
* add TileGroupHeader* as argument to ReadRecord to avoid lookup with GetTileGroup()
 * rewrite access to RW set in TransactionContext::Record* functions to avoid duplicated lookups


tcm-marcel:
Callgrind call graph at `PerformVectorizedRead` before (left) and after the fix (right) when running oltpbench ycsb RecordRead. Only small changes, but it seems to be a little better. I don't know yet, why some of the invocation numbers are different. 
grafik
Be aware that the colors are different between left and right when reading.

PR #1381, opened 2018-05-24
Move stats_mode check to end of transaction

mbutrovich:
Previously, every PerformRead(), PerformInsert(), etc. in TOTM would do a check on the stats_mode settings and increment the appropriate counter. This PR modifies it to a single check in EndTransaction(), which will then iterate through the RWSet and update the appropriate counters. Note that we lose some accuracy with this method for tuples that were repeatedly accessed in a transaction.

### DIFF
--- a/src/concurrency/transaction_manager.cpp
+++ b/src/concurrency/transaction_manager.cpp
@@ -82,6 +82,58 @@ void TransactionManager::EndTransaction(TransactionContext *current_txn) {
     current_txn->ExecOnCommitTriggers();
   }
 
+  // log RWSet and result stats
+  auto stats_type = static_cast<StatsType>(settings::SettingsManager::GetInt(
+      settings::SettingId::stats_mode));
+
+  if (stats_type != StatsType::INVALID) {
+    auto stats_context = stats::BackendStatsContext::GetInstance();
+
+    // update counters for each element in the RWSet
+    for (const auto &element : current_txn->GetReadWriteSet()) {
+      const auto &tile_group_id = element.first.block;
+      const auto &rw_type = element.second;
+      switch (rw_type) {
+        case RWType::READ:
+        case RWType::READ_OWN:
+          stats_context->IncrementTableReads(tile_group_id);
+          break;
+        case RWType::UPDATE:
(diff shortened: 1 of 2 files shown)


src/concurrency/transaction_manager.cpp
```diff
@@ -82,6 +82,59 @@ void TransactionManager::EndTransaction(TransactionContext *current_txn) {
...
+        default:
+          break;
+      }
+    }
+
+    oid_t database_id = 0;
+    const auto &first_tuple = current_txn->GetReadWriteSet().cbegin();
+    if (first_tuple != current_txn->GetReadWriteSet().cend()) {
+      database_id = catalog::Manager::GetInstance().
+          GetTileGroup(first_tuple->first.block)->GetDatabaseId();
```
tcm-marcel: Can you briefly explain whats happening here?

mbutrovich: The transaction result stat is indexed on database_id, but I changed the logic from what I removed in TOTM. Let me just duplicate it and add some comments to make it more clear.

mbutrovich:
A demonstration in the difference of precision for latency: (timestamps) is the value calculated from the txn timestamp (new method) while (stats) is the old method using a Timer.

transaction latency (timestamps): 52.366000
transaction latency (stats): 52.379732
transaction latency (timestamps): 4.082000
transaction latency (stats): 4.086408
transaction latency (timestamps): 8.355000
transaction latency (stats): 8.360423
transaction latency (timestamps): 9.970000
transaction latency (stats): 9.973952
transaction latency (timestamps): 12.874000
transaction latency (stats): 12.879112
transaction latency (timestamps): 39.244000
transaction latency (stats): 39.249294
transaction latency (timestamps): 64.269000
transaction latency (stats): 64.277862
transaction latency (timestamps): 0.812000
transaction latency (stats): 0.816374
transaction latency (timestamps): 0.804000
transaction latency (stats): 0.809365
transaction latency (timestamps): 0.721000
transaction latency (stats): 0.725384
transaction latency (timestamps): 0.506000
transaction latency (stats): 0.508766

Issue #1374, opened 2018-05-18
Can't recover tile group from LayoutCatalog in checkpoint recovery


ksaito7:
Current LayoutCatalog table, pg_layout, doesn't have about ROW or COLUMN layout. So checkpoint recovery can't decide which layout was used in the recovered tile group from LayoutCatalog. To recover it from LayoutCatalog, all layout info has to insert into pg_layout.

Additionally, layout oid in case of ROW or COLUMN is not initialized when constructed by `Layout(const oid_t num_columns, LayoutType layout_type)`. Actually, a layout oid of default layout in DataTable is not set 0 or 1. It has to initialize according to layout type to insert it into pg_layout.

Issue #1373, opened 2018-05-18
Unexpected catalog object might be obtained from CatalogCache


ksaito7:
When we use a getter function in CatalogCache, an unexpected catalog object might be obtained from unexpected database. For example: 
```
std::shared_ptr<TableCatalogObject> CatalogCache::GetCachedTableObject(
    oid_t table_oid) {
  for (auto it = database_objects_cache.begin();
       it != database_objects_cache.end(); ++it) {
    auto database_object = it->second;
    auto table_object = database_object->GetTableObject(table_oid, true);
    if (table_object) return table_object;
  }
  return nullptr;
}
```
In this function, same `table_oid` might be existed in an other database unexpected, because SystemCatalogs class allows each catalog to make same oid between databases. We need to add a database_oid argument to all getter functions.

Issue #1320, opened 2018-04-20
Catalog OID collision in checkpoint recovery 


ksaito7:
Some catalog objects have OID as primary key. OID is incremented automatically when a new catalog record is inserted. The OID value never distinguish built-in catalog records (e.g. built-in function, and catalog table info in pg_table) with user created catalog records (e.g. UDF, and user table info in pg_table). 

In checkpoints, all catalog tables is kept with same values including OID. However, for support to add new built-in catalog records, the checkpoint recovery prioritize built-in catalog records initialized by catalog bootstrap. This means that only user created catalog records is recovered. If new built-in catalog records is added in this case, OIDs of the catalog is incremented as mentioned above. Then, these OIDs of the new catalog records conflict with OIDs of user created catalog records recovered from the checkpoint.

To avoid this collision, the OID should distinguish between built-in catalog records and user created catalog records.

Issue #1319, opened 2018-04-20
Add a column length in ColumnCatalogObject and pg_attribute


ksaito7:
Current catalog::ColumnCatalogObject class and pg_attribute table don't have column length information, although the catalog::Column class has it. Due to this, the checkpoints can't recover the catalog::Column from pg_attribute table.

a6802739:
@ksaito7  I'm new to this project, And I want to try to fix this. One question here is what does the `column_offset` means within `ColumnCatalogObject`?

ksaito7:
@a6802739 This issue has already been implemented in PR #1383 (now reviewing). Could you try other issues?
 `column_offset` means a head location of the column in the table. It seems to be used for storage management to recognize where the column data is in the tuple. For example, assuming that a table has two columns (id interger, value varchar(36)), then id's offset is 0, and value's offset is 4.

a6802739:
@ksaito7  I see, Thanks a lot. Any issue would be fine for me to start, could you just point out another issue which is suitable for a beginner?

ksaito7:
@a6802739 You should choose an issue tagged beginner. Sorry, I don't know which is better. If you want to work for catalog issues, how about  #1414?

PR #1383, opened 2018-05-25
Catalog updates for checkpoints

ksaito7:
This PR updates some issues related to catalog stuff for checkpoints.

**- Related Issues**
1. #1319 Add column length in ColumnCatalogObject and pg_attribute
2. #1320 Catalog OID collision in checkpoint recovery
3. #1373 Unexpected catalog object might be obtained from CatalogCache
4. #1374 Can't recover tile group from LayoutCatalog in checkpoint recovery

**- Fix Points**
1. Add `column_length` feild within pg_attribute and ColumnCatalogObject, and modify related functions in ColumnCatalog.
1. Add `UpdateOid()` function within catalogs using oid_ in order to avoid collision between catalog values added by system and users.  And add `OID_FOR_USER_OFFSET` (= 10000) variable within catalog_default.cpp to be used for `UpdateOid()` function. These are called in last of bootstraps of Catalog and SystemCatalog.
1. Add database oid/name argument within `GetChachedTableObject()` function and `GetChachedIndexObject()` function in order to avoid to acquire an unexpected table from other database. Also fix related functions, `GetTableObject()` in TableCatalog and `GetIndexObject()` in IndexCatalog.
1. Modify Layout stuff to recover layout information for default layout of DataTable and layout of TileGroup.
   1. Modify Layout constructor `Layout(num_columns, layout_type)` to set appropriate layout oid.
   1. Modify `CreateTable()` and `CreateLayout()` in Catalog class to insert all Layout information including row layout and column layout into pg_layout, LayoutCatalog. These avoid to insert same layout. And also modify `DropLayout()` in Catalog to recreate default layout in pg_catalog in case that it is deleted.
   1. Add `default_layout_oid` field within pg_table and TableCatalogObject, and modify related functions in TableCatalog. To support for updating default layout, add `UpdateDefaultLayoutOid()` function within TableCatalog. It is used in `CreateDefaultLayout()` function and `DropLayout()` function.


src/include/catalog/column_catalog.h
```diff
@@ -47,6 +47,7 @@ class ColumnCatalogObject {
   inline uint32_t GetColumnId() { return column_id; }
   inline uint32_t GetColumnOffset() { return column_offset; }
   inline type::TypeId GetColumnType() { return column_type; }
+  inline size_t GetColumnLength() { return column_length; }
```
aaron-tian: Would `unit32_t` same as `GetColumnOffset()` be sufficient?

src/include/catalog/column_catalog.h
```diff
@@ -58,6 +59,7 @@ class ColumnCatalogObject {
   uint32_t column_id;
   uint32_t column_offset;
   type::TypeId column_type;
+  size_t column_length;
```
aaron-tian: The same type question. `column_length` would be consistent with `column_offset` if also using `uint32_t`.

src/include/storage/data_table.h
```diff
@@ -265,7 +265,10 @@ class DataTable : public AbstractTable {
     default_layout_ = std::shared_ptr<const Layout>(
         new const Layout(schema->GetColumnCount(), type));
   }
-  const Layout &GetDefaultLayout() const;
+
+  const std::shared_ptr<const Layout> GetDefaultLayout() const {
+  	return default_layout_;
+  }
```
aaron-tian: I'm wondering out of what reasons we change a reference of `default_layout_` to its pointer.

src/include/storage/layout.h
```diff
@@ -60,11 +60,13 @@ class Layout : public Printable {
...
 
   /**
-   * @brief   Constructor for arbitrary column_maps.
+   * @brief   Constructor for arbitrary column_maps with layout oid.
    * @param   column_map Column map of the layout to be constructed.
+   * @param   num_columns Number of column.
    * @param   layout_oid Per-table unique OID. Generted by DataTable.
    */
-  Layout(const column_map_type &column_map, oid_t layout_oid);
+  Layout(const column_map_type &column_map, const oid_t num_columns,
```
aaron-tian: The same question as before. Why we need an additional parameter `num_columns`?

src/storage/layout.cpp
```diff
@@ -63,9 +72,10 @@ Layout::Layout(const column_map_type &column_map)
 }
 
 // Constructor for Layout class with predefined layout_oid
-Layout::Layout(const column_map_type &column_map, oid_t layout_id)
+Layout::Layout(const column_map_type &column_map, const oid_t num_columns,
+		           const oid_t layout_id)
     : layout_oid_(layout_id),
-      num_columns_(column_map.size()),
+      num_columns_(num_columns),
```
aaron-tian: The same parameter question. Why don't we just use `column_map.size()`?

test/catalog/catalog_test.cpp
```diff
@@ -120,6 +120,30 @@ TEST_F(CatalogTests, CreatingTable) {
...
+  auto catalog_db_object = catalog->GetDatabaseObject(CATALOG_DATABASE_OID, txn);
+  auto catalog_table_objects = catalog_db_object->GetTableObjects();
+  EXPECT_NE(0, catalog_table_objects.size());
+
+  auto user_db_object = catalog->GetDatabaseObject("emp_db", txn);
+  auto user_database = storage::StorageManager::GetInstance()
+                          ->GetDatabaseWithOid(user_db_object->GetDatabaseOid());
+
+  // check expected table object is acquired
+  for (oid_t table_idx = 0; table_idx < user_database->GetTableCount(); table_idx++) {
```
aaron-tian: Do we know what tables we have at this point? Can we also verify the number of tables or their specific names?

test/catalog/catalog_test.cpp
```diff
@@ -352,10 +380,25 @@ TEST_F(CatalogTests, LayoutCatalogTest) {
   auto table_oid = table_object->GetTableOid();
   auto table =
       catalog->GetTableWithName(db_name, DEFAULT_SCHEMA_NAME, table_name, txn);
+  auto pg_layout = catalog->GetSystemCatalogs(database_oid)->GetLayoutCatalog();
   txn_manager.CommitTransaction(txn);
 
-  // Change default layout.
+  // Check the first default layout
+  auto first_default_layout = table->GetDefaultLayout();
+  EXPECT_EQ(0, first_default_layout->GetOid());
```
aaron-tian: Replace `0` with its macro name `ROW_STORE_OID`.

test/catalog/catalog_test.cpp
```diff
@@ -365,8 +408,23 @@ TEST_F(CatalogTests, LayoutCatalogTest) {
...
       catalog->CreateDefaultLayout(database_oid, table_oid, default_map, txn);
-  txn_manager.CommitTransaction(txn);
   EXPECT_NE(nullptr, default_layout);
+  txn_manager.CommitTransaction(txn);
+
+  // Check the changed default layout
+  auto default_layout_oid = default_layout->GetOid();
+  EXPECT_EQ(default_layout_oid, table->GetDefaultLayout()->GetOid());
+  EXPECT_FALSE(default_layout->IsColumnStore());
+  EXPECT_FALSE(default_layout->IsRowStore());
```
aaron-tian: Can we test if its type is `HYBRID` directly?

test/catalog/catalog_test.cpp
```diff
@@ -378,30 +436,45 @@ TEST_F(CatalogTests, LayoutCatalogTest) {
   txn = txn_manager.BeginTransaction();
   auto other_layout =
       catalog->CreateLayout(database_oid, table_oid, non_default_map, txn);
-  txn_manager.CommitTransaction(txn);
+  EXPECT_NE(nullptr, other_layout);
+	txn_manager.CommitTransaction(txn);
+
+  // Check the created layout
+  EXPECT_FALSE(other_layout->IsColumnStore());
+  EXPECT_FALSE(other_layout->IsRowStore());
```
aaron-tian: The same question as before. Can we test if its type is `HYBRID` directly?

test/catalog/catalog_test.cpp
```diff
@@ -378,30 +436,45 @@ TEST_F(CatalogTests, LayoutCatalogTest) {
...
   EXPECT_EQ(ResultType::SUCCESS, catalog->DropLayout(database_oid, table_oid,
                                                      default_layout_oid, txn));
   txn_manager.CommitTransaction(txn);
 
   // Check that default layout is reset and set to row_store.
-  EXPECT_NE(*default_layout.get(), table->GetDefaultLayout());
-  EXPECT_TRUE(table->GetDefaultLayout().IsRowStore());
+  EXPECT_NE(default_layout, table->GetDefaultLayout());
+  EXPECT_TRUE(table->GetDefaultLayout()->IsRowStore());
+  EXPECT_EQ(0, table->GetDefaultLayout()->GetOid());
```
aaron-tian: Use the macro name of `0`.

test/catalog/catalog_test.cpp
```diff
@@ -378,30 +436,45 @@ TEST_F(CatalogTests, LayoutCatalogTest) {
...
 
-  // Query pg_layout to ensure that the entry is dropped
+  // Query pg_layout and pg_table to ensure that the entry is dropped
   txn = txn_manager.BeginTransaction();
-  auto pg_layout = catalog->GetSystemCatalogs(database_oid)->GetLayoutCatalog();
   EXPECT_EQ(nullptr,
             pg_layout->GetLayoutWithOid(table_oid, default_layout_oid, txn));
+  EXPECT_EQ(0,
```
aaron-tian: Replace `0` with `nullptr`.

src/catalog/catalog.cpp
```diff
@@ -169,22 +169,22 @@ void Catalog::BootstrapSystemCatalogs(storage::Database *database,
   // pg_database record is shared across different databases
   system_catalogs->GetTableCatalog()->InsertTable(
       DATABASE_CATALOG_OID, DATABASE_CATALOG_NAME, CATALOG_SCHEMA_NAME,
-      CATALOG_DATABASE_OID, pool_.get(), txn);
+      CATALOG_DATABASE_OID, 0, pool_.get(), txn);
```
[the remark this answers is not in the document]
ksaito7: I moved macros defined in `layout.h` into `catalog_defaults.h` and modify all raw value with these macros.

src/catalog/catalog.cpp
```diff
@@ -573,15 +589,15 @@ std::shared_ptr<const storage::Layout> Catalog::CreateLayout(
   // Ensure that the new layout
   PELOTON_ASSERT(layout_oid < INVALID_OID);
   auto new_layout = std::shared_ptr<const storage::Layout>(
-      new const storage::Layout(column_map, layout_oid));
+      new const storage::Layout(column_map, column_map.size(), layout_oid));
```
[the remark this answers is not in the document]
ksaito7: Because there are some cases that a size of `column_map` doesn't equal to `num_columns`. This occurs when a layout type is row store or column store. In this case, `column_map` is empty, even when `num_columns` is set to a correct number rather than 0. This problem occurs in case of recovering layout object from layout catalog.

src/catalog/column_catalog.cpp
```diff
@@ -152,7 +162,8 @@ bool ColumnCatalog::InsertColumn(oid_t table_oid,
   auto val3 = type::ValueFactory::GetIntegerValue(column_offset);
   auto val4 =
       type::ValueFactory::GetVarcharValue(TypeIdToString(column_type), nullptr);
-  auto val5 = type::ValueFactory::GetBooleanValue(is_inlined);
+  auto val5 = type::ValueFactory::GetBigIntValue(column_length);
```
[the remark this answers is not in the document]
ksaito7: That's true. I misunderstood the type. I modify it.

src/catalog/layout_catalog.cpp
```diff
@@ -197,10 +197,15 @@ LayoutCatalog::GetLayouts(oid_t table_oid,
       std::string column_map_str =
           tile->GetValue(tuple_id, LayoutCatalog::ColumnId::COLUMN_MAP)
               .ToString();
-      auto column_map =
-          storage::Layout::DeserializeColumnMap(num_columns, column_map_str);
+
+      column_map_type column_map;
+      if (column_map_str.length() != 0) {
```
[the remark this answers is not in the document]
ksaito7: This is for empty `column_map` which I explained above. The function `DeserializeColumnMap()` makes `column_map`  according to `num_columns`. If `column_map` is empty, then the function will be error by difference of a number of component in `column_map` within layout catalog.

src/catalog/system_catalogs.cpp
```diff
@@ -106,6 +107,16 @@ void SystemCatalogs::Bootstrap(const std::string &database_name,
...
     pg_query_metrics_ = new QueryMetricsCatalog(database_name, txn);
   }
+
+  // Reset oid of each catalog to avoid collisions between catalog
+  // values added by system and users when checkpoint recovery.
+	pg_attribute_->UpdateOid(OID_FOR_USER_OFFSET);
+	pg_namespace_->UpdateOid(OID_FOR_USER_OFFSET);
+	pg_table_->UpdateOid(OID_FOR_USER_OFFSET);
+	pg_index_->UpdateOid(OID_FOR_USER_OFFSET);
+	// pg_layout_->UpdateOid(OID_FOR_USER_OFFSET);
```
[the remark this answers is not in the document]
ksaito7: I removed this.

src/catalog/system_catalogs.cpp
```diff
@@ -106,6 +107,16 @@ void SystemCatalogs::Bootstrap(const std::string &database_name,
...
+	pg_trigger_->UpdateOid(OID_FOR_USER_OFFSET);
+	// pg_proc->UpdateOid(OID_FOR_USER_OFFSET);
```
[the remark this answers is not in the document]
ksaito7: This is not dead code because the pg_proc would be implemented within the system catalog according to this file's line 95-97. To avoid to forget this function call, it also keeps comment out.

src/include/catalog/column_catalog.h
ksaito7: The column length in `storage/column.h` is `size_t`.  I took same type, but I can't  decide which is better.

src/include/storage/data_table.h
ksaito7: Because other getter functions of other objects in `data_table` like  `Index` or `TileGroup` return share_ptr pointer. I didn't understand why it offered the reference of layout.

test/catalog/catalog_test.cpp
ksaito7: It just checks that the cache of table catalog belongs to a correct database about #1373. The information about creating table might be checked in previous test `CreatingTable`.

test/catalog/catalog_test.cpp
ksaito7: No, there was not a function to access the layout type directly. These two `EXPECT_FALSE` tests match HYBRID actually.

ksaito7:
@aaron-tian Thanks for a lot of comments! I modified them and reply each comments. Please check them again.

src/include/catalog/column_catalog.h
aaron-tian: Could you please change the type of `GetColumnOffset()` to `oid_t` then?

test/catalog/catalog_test.cpp
aaron-tian: I see. How about just adding a function like `IsHybridStore()` to make it more clear?

aaron-tian:
Thanks for the nice work. `IsHybridStore()` could be added and the type of `GetColumnOffset()` could be changed if you think it's a good idea, all the others LGTM.

ksaito7:
I update them you mentioned. thanks.

PR #1384, opened 2018-05-26
Move tile group related code from catalog manager to storage manager

Zeninma:
Move tile group related code from catalog manager to storage manager as what had been done in PR#1376.


apavlo:

PR #1385, opened 2018-05-27
Apply limit + order optimization in the optimizer

chenboy:
This PR applies limit + sort optimization in the optimizer as the same way `ORCA` does it. Basically, we keep an internal order in `Limit` operator when the parse tree contains both limit and order, and the optimizer logic is tweaked a bit to consider this internal sort. When the plan is generated, the limit operator will generate a `LimitPlan` on top of an `OrderByPlan` with limit and offset set correctly.

The multiple test cases in `optimizer_sql_test.cpp` already tested order by with limit, so I didn't add new ones.


apavlo:
Thanks! Let's have @malin1993ml review this.

malin1993ml:
I'll review this after I get back to hotel.

src/optimizer/input_column_deriver.cpp
```diff
@@ -82,18 +82,39 @@ void InputColumnDeriver::Visit(const QueryDerivedScan *op) {
           output_cols, {input_cols}};
 }
 
-void InputColumnDeriver::Visit(const PhysicalLimit *) { Passdown(); }
+void InputColumnDeriver::Visit(const PhysicalLimit *op) {
+  // Input columns and internal sort columns are needed by the child node
```
malin1993ml: I think you want say "aggregated columns" instead of "input columns"?

src/optimizer/plan_generator.cpp
```diff
@@ -133,9 +133,34 @@ void PlanGenerator::Visit(const QueryDerivedScan *) {
 }
 
 void PlanGenerator::Visit(const PhysicalLimit *op) {
+  // Generate order by plan when there's internal sort order
+  if (!op->sort_exprs.empty()) {
```
malin1993ml: So `sort_exps` is not empty means there's no internal sort order?

apavlo:
@malin1993ml @GustavoAngulo  Since @chenboy is now at Snowflake, I think that you guys will need to take care of this. Can you fix it by next week. We are going to need this for TPC-C (AFAIK)

chenboy:
Oh, I nearly forgot this, sorry. I'll address comments & fix the code tonight.

malin1993ml:
Thanks a lot, Bowei!

src/optimizer/child_property_deriver.cpp
```diff
@@ -130,11 +130,15 @@ void ChildPropertyDeriver::Visit(const PhysicalAggregate *) {
...
-  vector<shared_ptr<PropertySet>> child_input_properties{requirements_};
-
-  output_.push_back(make_pair(requirements_, move(child_input_properties)));
+void ChildPropertyDeriver::Visit(const PhysicalLimit *op) {
+  // Limit fulfill the internal sort property
+  vector<shared_ptr<PropertySet>> child_input_properties{
+      std::make_shared<PropertySet>()};
+  std::shared_ptr<PropertySet> provided_prop(new PropertySet);
+  provided_prop->AddProperty(
+      std::make_shared<PropertySort>(op->sort_exprs, op->sort_acsending));
```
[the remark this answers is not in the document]
chenboy: Nice catch, should check for empty here, although adding an empty property will not cause correctness issues.

src/optimizer/input_column_deriver.cpp
chenboy: I meant to say all aggregate expressions and TVEs in the required columns + internal sort columns.

src/optimizer/plan_generator.cpp
chenboy: `sort_exprs` is not empty means there is internal sort order in this case.

tli2:
@malin1993ml Jenkins has been wonky lately. Restarting fixed it.

Issue #1318, opened 2018-04-20
Update LLVM to 5.0 or 6.0


pmenon:
On our newer KabyLake servers, LLVM versions older than 4.0 seem to be generating bogus code. LLVM 4.0 and above fixes these issues.  If we're considering an upgrade, I'd like to explore LLVM 5.0 and 6.0 as upgrade targets, too.

pervazea:
Matt tried out Ubuntu 18.04 over the weekend. Packages.sh does not work as LLVM 3.7 doesn't install. 

So, likely that we'll want to upgrade LLVM in conjunction with moving to 18.04. If there are any other outdated (unsupported) components moving to 18.04, those too.

mbutrovich:
Have a [branch going](https://github.com/mbutrovich/peloton/tree/llvm6) that has been tested in the following configurations:

Ubuntu 16.04 (dev5):
LLVM 3.7 (fails, but expected due to bad Skylake support)
LLVM 5.0 (passes make check, Debug, ASAN on)
LLVM 6.0 (passes make check, Debug, ASAN on)

macOS:
LLVM 3.7 (passes make check, Debug, ASAN on)
LLVM 6.0 (passes make check, Debug, ASAN on)

I'm going to do a quick LLVM 3.7 check on an Ubuntu VM, and then open a PR for Travis and Jenkins testing.

crd477:
Cool!

PR #1389, opened 2018-05-30
LLVM 5.0 and 6.0 compatibility

mbutrovich:
Addresses #1318. Note that these are only compatibility changes. This PR does not change toolchain packages or build settings. We’ll discuss those in the future and make LLVM version changes in a separate PR.

These changes simply let us change the LLVM version in the future and are backwards compatible with the current 3.7-based toolchain. Tested configurations:

Ubuntu 16.04 (dev5):
LLVM 3.7 (fails, but expected due to bad CPU support, master fails too)
LLVM 5.0 (passes make check, Debug, ASAN on)
LLVM 6.0 (passes make check, Debug, ASAN on)

Ubuntu 16.04 (VM):
LLVM 3.7 (passes make check, Debug, ASAN on)

Ubuntu 18.04 (VM):
LLVM 6.0 (passes make check, Debug, ASAN on)

macOS:
LLVM 3.7 (passes make check, Debug, ASAN on)
LLVM 6.0 (passes make check, Debug, ASAN on)

Code modifications based on:
http://lists.llvm.org/pipermail/llvm-dev/2015-August/089706.html
https://reviews.llvm.org/rL298010
https://stackoverflow.com/questions/46367910/llvm-5-0-linking-error-with-llvmmoduledump

We'll need to discuss what we want the minimum version to be going forward (CMake, packages.sh) once these code changes go in.


src/codegen/codegen.cpp
```diff
@@ -92,9 +92,9 @@ llvm::Value *CodeGen::AllocateVariable(llvm::Type *type,
 
   auto *entry_block = code_context_.GetCurrentFunction()->GetEntryBlock();
   if (entry_block->empty()) {
-    return new llvm::AllocaInst(type, name, entry_block);
+    return new llvm::AllocaInst(type, 0, name, entry_block);
```
pmenon: LLVM 5.0+ added a parameter argument to the constructor for this class, so we're actually calling different versions of the constructor when using pre-5.0 and 5.0+. To ensure consistency, we should add version checks around these calls using the `LLVM_VERSION_GE` macros (in `macros.h`).

mbutrovich: So if LLVM_VERSION_GE(5, 0) use the new calls, otherwise use the old calls?

apavlo:
Let's merge this next.

PR #1390, opened 2018-05-30
packages.sh: Ubuntu 18.04 LTS support, removed deprecated OSes

mbutrovich:
Tested on clean install of Ubuntu 18.04 LTS in a VM. Passed make check in Debug with ASAN on. Added package is due to a linker error (-lz) experienced during build on 18.04.

Removed code related to Red Hat, Debian, and Fedora and replaced with error messages.


script/installation/packages.sh
```diff
@@ -180,149 +190,25 @@ if [ "$DISTRO" = "UBUNTU" ]; then
...
-        postgresql-client \
-        libtbb-dev \
-        libeigen3-dev
+    echo "Error: Debian is not currently supported"
+    exit 1
 
 ## ------------------------------------------------
 ## FEDORA
 ## ------------------------------------------------
 elif [[ "$DISTRO" == *"FEDORA"* ]]; then
```
apavlo: Instead of keeping Fedora and RedHat in the if clauses, why not just let it fall down to the unsupported clause and let it spit the error message.

We don't want people to think that we will support it.

script/installation/packages.sh
```diff
@@ -12,9 +12,8 @@
 ## g++ manually.
 ##
 ## Supported environments:
-##  * Ubuntu (14.04, 16.04)
-##  * Fedora
-##  * OSX
+##  * Ubuntu (14.04, 16.04, 18.04)
+##  * macOS
```
apavlo: I would add a message here:

```
Update (2018-06-08):
We are no longer able to support RedHat/Fedora because those environments are not supported by TensorFlow.
```

mbutrovich:
Not sure what happened to the history here when I did the suggested changes and then a rebase. That merge of master by @tli2 combined with a later rebase might have messed with it. I'll take a look in the AM and nuke it if needed. It's only 1 file to change if I make the branch again.

apavlo:
@mbutrovich Probably best to just extract out your one commit and make a new branch.

script/installation/packages.sh
mbutrovich: Added.

mbutrovich:
@apavlo I think I got it sorted out. If the commit history looks weird I can redo it, but it looks okay to me now.

Issue #754, opened 2017-07-17
delete_test.cpp


haoxianghua:
In this test, print much log information and it seems to test with naked eyes.

apavlo:
Please provide more information (filenames, functions, etc).

haoxianghua:
File name is delete_test.cpp, and the logic of this test is straight. There is just no Programmatically check like EXPECT_EQ

haoxianghua:
To be exact, there is only one EXPECT_EQ to check if table created. But I think the purpose of this case is to test tuple delete

apavlo:
1. Move 'delete_test.cpp' into 'test/sql'. Rename to 'delete_sql_test.cpp'
2. Break up the "VariousOperations" test into smaller sub-tests. Refactor the table loading code so that yo can re-use it for all of the new smaller tests.
3. Add SELECT statements that check the contents of the table after you delete tuples.

PR #1391, opened 2018-05-31
Minor Cleanup Timer + DeleteSQLTests

apavlo:
This is a manual merge of #1161 except that I took out the changes to WorkerPool. Everything should be the same and reviewed already.

1. Refactored Timer to add the ability to record the total time.
2. @haoxianghua -- solve issues #754, write delete_sql_test.cpp to check correctness

https://www.youtube.com/watch?v=kWYYQlQhbyE


src/include/common/timer.h
```diff
@@ -43,16 +43,33 @@ class Timer : public peloton::Printable {
...
+  /**
+   * Return the total amount of time measured by
+   * this timer after multiple invocations.
+   * @return
+   */
+  double GetTotalDuration() const { return (total_); }
+
+  /**
+   * Get number of invocations
+   * @return
```
tli2: It looks like the @return tag is not filled out. (I personally leave out the general comment and just write the return tag for simple functions like this, i.e. @return number of invocations)

Issue #1392, opened 2018-05-31
Remove LOG_INFO from update_test


apavlo:
There is a bunch of `LOG_INFO` debug statements in 'update_test.cpp'

https://github.com/cmu-db/peloton/blob/master/test/executor/update_test.cpp

We need to remove them. This should be easy

Traveler98:
I am new to this . I changed what you asked for here : https://github.com/cmu-db/peloton/compare/master...Traveler98:patch-1
Hope I've done it right .

PR #1393, opened 2018-06-07
Update update_test.cpp

Traveler98:
Changed LOG_INFO to LOG_DEBUG


tli2:
Why is this relevant?

Traveler98:
I am new to this, maybe i added this for no reason! But there was a post asking for this change. I appreciate your help if i was wrong.

tli2:
You mean #1392

tli2:
Okay, it doesn't make sense to apply this fix to this test case because there's another issue with it: It doesn't have any assertions in it. If you still would like to work on this issue, do you mind replacing all the print out messages with EXPECTs instead?  (A good example would be, say CreateTest)

PR #1400, opened 2018-06-11
Fix warning for unused lambda capture on Clang 9.1

mbutrovich:
Resolves the following warning introduced by #1371 on Clang 9.1:

../git/peloton/test/codegen/csv_scan_translator_test.cpp:39:30: error: lambda capture 'quote' is not required to be captured for this use [-Werror,-Wunused-lambda-capture]
  const auto quote_string = [quote](std::string s) {

I couldn't reproduce the warning on GCC 7.3, which makes me wonder if it's spurious on Clang's part. However, I ran make check with this proposed change under Clang 9.1 and GCC 7.3 and everything looked good. Hopefully it's the same story on Jenkins and Travis.

### DIFF
--- a/test/codegen/csv_scan_translator_test.cpp
+++ b/test/codegen/csv_scan_translator_test.cpp
@@ -36,7 +36,7 @@ class CSVScanTranslatorTest : public PelotonCodeGenTest {
 TEST_F(CSVScanTranslatorTest, IntCsvScan) {
   // The quoting character and a helper function to quote a given string
   const char quote = '"';
-  const auto quote_string = [quote](std::string s) {
+  const auto quote_string = [](std::string s) {
     return StringUtil::Format("%c%s%c", quote, s.c_str(), quote);
   };
 


pmenon:
LGTM. 

@pervazea What is the status of getting the newer Mac onto the Jenkins infrastructure?

pervazea:
We provided the Mac to Chad, so it could be installed in the PDL lab. I haven't heard an update on this... will check on it later today, to see how it is progressing.


crd477:
How important is it to isolate the build process on this platform?  The problem is that, unlike the current Docker setup for Linux builds, PRs from random users would get merged and built directly on the system.

The approach I'd like to take is to run macOS inside VirtualBox on this system and install the Jenkins agent on the VM there but this will take me a bit of time.  Is that OK?

apavlo:
@crd477 It's probably best to play it safe and run the builds in a sandbox.

But doesn't mean if we can run OSX in a VM then we don't need Mac hardware in the future?

crd477:
Not exactly.  The software license agreement for macOS states:
```
2. Permitted License Uses and Restrictions. A. Preinstalled and Single-Copy Apple Software License. Subject to the terms and conditions of this License, unless you obtained the Apple Software from the Mac App Store, through an automatic 
download or under a volume license, maintenance or other written agreement from Apple, you are granted a limited, non-exclusive license to install, use and run one (1) copy of the Apple Software on a single Apple-branded computer at any one time. For example, these single-copy license terms apply to you if you obtained the Apple Software preinstalled on Apple-branded hardware.
```
So running a single macOS VM in a type 2 hypervisor, e.g., VirtualBox, under Linux appears to be permitted.  That's the way I propose to set it up.

crd477:
OK, computing services says:

> Software Licensing just updated the ticket and said they're okay with the virtual machines.  They believe that your description of your intent is within the terms of the license agreement.  Thanks!

I'm intend to make some progress on this today.

crd477:
The macOS VM has been added to the pool.  Does [the build stage that I specified for macOS](http://jenkins.db.cs.cmu.edu:8080/blue/organizations/jenkins/crd%2Fpeloton-crd/detail/mac-jenkins/10/pipeline/16) look OK to you?
If so, I'll generate a PR.  This systems seems to perform OK with a single VM and 4 Jenkins executors.  If you want to get more macOS capacity, I think it will be necessary to buy some more minis and reuse the scripts I have in place currently.

pervazea:
Yes, macOS build stage looks ok.

Pervaze

crd477:
PR submitted in #1454

Issue #1396, opened 2018-06-08
TransactionContext Read-Only Interface


apavlo:
We need to think about how we want to set the read-only flag for a `TransactionContext`. There are currently two ways to set this flag:

1. When the txn is created in `TransactionManager::BeginTransaction()`
   https://github.com/cmu-db/peloton/blob/master/src/include/concurrency/transaction_manager.h#L226-L228 

2. After it has started using `TransactionContext::SetReadOnly()`:
   https://github.com/cmu-db/peloton/blob/master/src/include/concurrency/transaction_context.h#L273

I first note that `TransactionContext::SetReadOnly()` does not check whether a txn has already executed a query that modified the database when we set the flag. This could lead to problems if somebody calls this incorrectly. It also doesn't make sense to support this functionality because if the system doesn't know that a txn is read-only when it starts, then it won't know whether it is read-only until commit time.

**Things to fix:**

1. Remove the default value for the `read_only` flag in `TransactionManager::BeginTransaction()`.

2. Remove the `TransactionContext::SetReadOnly()` method.

3. Add a `read_only` flag to the `TransactionContext` constructors.
   https://github.com/cmu-db/peloton/blob/master/src/include/concurrency/transaction_context.h#L47-L51

4. Modify `TimestampOrderingTransactionManager::CommitTransaction()` so that it checks whether a txn has not executed any modifying queries. If so, then it can then be treated as read-only and commit right away. This should happen after we check whether the read-only flag is set.
   https://github.com/cmu-db/peloton/blob/master/src/concurrency/timestamp_ordering_transaction_manager.cpp#L610

Issue #1395, opened 2018-06-08
Mark Transaction as Read-Only When Executing Single-Statement SELECT


apavlo:
If we get `SELECT` query in a single-statement txn, we should mark it as 'read-only'. This will avoid conflicts with other txns that are running at the same time.

I think the only place that we need to change is `TrafficCop::ExecuteHelper()`:

https://github.com/cmu-db/peloton/blob/master/src/traffic_cop/traffic_cop.cpp#L163

I'm not sure how we can write a test case to prove that this is working correctly. I don't know whether we have test cases for running single-stmt txns in the `TrafficCop` and then being able to see the `TransactionContext` after the txn commits.

apavlo:
I will assign this task to @lmwnshn

PR #1402, opened 2018-06-11
Read-Only TxnContext Interface; Read-Only (single-statement select, txn) optimizations

lmwnshn:
#1396 
1. Removed default read_only=false from [TransactionManager::BeginTransaction](https://github.com/lmwnshn/peloton/commit/e506f17d043581af0e77369899edf01b0174bd94#diff-cb353f0475324594d1f1c27b15dd2247R228)
2. Removed [TransactionContext::SetReadOnly](https://github.com/lmwnshn/peloton/commit/e506f17d043581af0e77369899edf01b0174bd94#diff-4d5037ca29ea43576b5613d89b7f7f09L268)
3. Added read-only param to TransactionContext [constructors](https://github.com/lmwnshn/peloton/commit/e506f17d043581af0e77369899edf01b0174bd94#diff-4d5037ca29ea43576b5613d89b7f7f09L47)
4. Transactions with no modifying queries treated as [read-only](https://github.com/lmwnshn/peloton/commit/8b1531b98f0d40bcd13edad89de8f7fb2cac0871#diff-10269290aecfeb50f6ad7d6c0748819fR626)
  4.1 assumption: if transaction is marked read-only, then READ transactions are not added to ReadWriteSet

#1395 
Single-statement selects are marked as [read-only](https://github.com/lmwnshn/peloton/commit/467834ca5cf6ae3182755220c9a0b4906cf85cbc#diff-023fd34a4233a33e1d640a552340f34cR167)

### DIFF
--- a/src/concurrency/timestamp_ordering_transaction_manager.cpp
+++ b/src/concurrency/timestamp_ordering_transaction_manager.cpp
@@ -623,6 +623,14 @@ ResultType TimestampOrderingTransactionManager::CommitTransaction(
   //// handle other isolation levels
   //////////////////////////////////////////////////////////
 
+  auto &rw_set = current_txn->GetReadWriteSet();
+
+  // if no modifying queries, treat as read-only
+  if (rw_set.empty()) {
+    EndTransaction(current_txn);
+    return ResultType::SUCCESS;
+  }
+
   auto storage_manager = storage::StorageManager::GetInstance();
   auto &log_manager = logging::LogManager::GetInstance();
 
@@ -631,7 +639,6 @@ ResultType TimestampOrderingTransactionManager::CommitTransaction(
   // generate transaction id.
   cid_t end_commit_id = current_txn->GetCommitId();
 
-  auto &rw_set = current_txn->GetReadWriteSet();
   auto &rw_object_set = current_txn->GetCreateDropSet();
 
   auto gc_set = current_txn->GetGCSetPtr();
--- a/src/concurrency/timestamp_ordering_transaction_manager.cpp
+++ b/src/concurrency/timestamp_ordering_transaction_manager.cpp
@@ -171,10 +171,9 @@ void TimestampOrderingTransactionManager::YieldOwnership(
   tile_group_header->SetTransactionId(tuple_id, INITIAL_TXN_ID);
 }
 
-bool TimestampOrderingTransactionManager::PerformRead(TransactionContext *const current_txn,
-                                                      const ItemPointer &read_location,
-                                                      storage::TileGroupHeader *tile_group_header,
-                                                      bool acquire_ownership) {
+bool TimestampOrderingTransactionManager::PerformRead(
+    TransactionContext *const current_txn, const ItemPointer &read_location,
+    storage::TileGroupHeader *tile_group_header, bool acquire_ownership) {
   ItemPointer location = read_location;
 
   //////////////////////////////////////////////////////////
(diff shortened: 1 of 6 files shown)


lmwnshn:
Currently, I am only failing copy_test (1/181).

If I add traffic_cop.SetStatement(statement) [here](https://github.com/cmu-db/peloton/blob/94012458140dced1cba836c1d173dde0a05be7e4/test/executor/copy_test.cpp#L96), I will pass. The question is, should I do that?

I'm looking into what else is using SetStatement right now.

src/concurrency/timestamp_ordering_transaction_manager.cpp
```diff
@@ -624,6 +623,14 @@ ResultType TimestampOrderingTransactionManager::CommitTransaction(
   //// handle other isolation levels
   //////////////////////////////////////////////////////////
 
+  auto &rw_set = current_txn->GetReadWriteSet();
+
+  // if no modifying queries, treat as read-only
+  if (rw_set.empty()) {
+    EndTransaction(current_txn);
```
apavlo: I would just add a `LOG_TRACE` record here to record what is happening.

lmwnshn: Added

src/concurrency/timestamp_ordering_transaction_manager.cpp
```diff
@@ -624,6 +623,15 @@ ResultType TimestampOrderingTransactionManager::CommitTransaction(
   //// handle other isolation levels
   //////////////////////////////////////////////////////////
 
+  auto &rw_set = current_txn->GetReadWriteSet();
+
+  // if no modifying queries, treat as read-only
+  if (rw_set.empty()) {
```
mbutrovich: Based on the implementation of the is_written_ flag in #1401 could we also check the is_written_ flag here? That would tell us if the RWSet is just reads.

lmwnshn: @mbutrovich I've tried changing it to use is_written_ instead of empty(), but I keep breaking on [multi_granularity_access_test](https://github.com/cmu-db/peloton/blob/master/test/concurrency/multi_granularity_access_test.cpp#L149). Did you and/or @pervazea say something about SCAN being broken earlier?

mbutrovich:
The more I look at this, the more it seems like we're building a hack on top of an unstable foundation. We're trying to find a shortcut if the RWSet only contains Reads/is empty. However, we shouldn't even be putting Reads in the RWSet in the first place. @yingjunwu [says as much](https://github.com/cmu-db/peloton/blob/master/src/concurrency/timestamp_ordering_transaction_manager.cpp#L325). We only add READ_OWNs for the sake of releasing the SpinLatch at Commit/Abort.

Commit and Abort do nothing if an element is READ when they iterate through the RWSet, so they shouldn't really be there in the first place.

This PR looks fine on its own, but it seems like a workaround for what a larger design issue that should be addressed.

yingjunwu:
I am not sure what you guys are currently working on, but I wanna mention that the original rwset was design to be general enough -- We didn't assume the underlying CC protocol, and the rwset was supposed to be compatible with any protocol, like OCC and 2PL. While our experiments show that timestamp ordering performs best, it seems that some people (e.g., Phil Bernstein) have a strong faith in 2PL :-)

mbutrovich:
So that’s sort of emblematic of the issues I feel like I’m finding in the current system. Things like maitaining a doubly-linked version chain, reads getting added to the RWSet, etc. seem to be preventing us from making the best MVTO system possible.

I definitely see why it was necessary in the context of comparing CC systems, but if MVTO is what we’re sticking with I think we can improve it.

apavlo:
@mbutrovich Do we want to merge this PR?

mbutrovich:
@apavlo Yes. @lmwnshn might need to do a rebase first, but I still think it should be merged.

lmwnshn:
@mbutrovich Undid the formatting changes

PR #1403, opened 2018-06-12
Coveralls fix

mbutrovich:
Small fix in travis config that should reenable Coveralls testing.


tcm-marcel:
Looks good. Let's wait for travis to run through and see if it actually uploads the results to Coveralls. Sometimes environment variables in travis don't work as expected.

Issue #1397, opened 2018-06-10
Add New SessionContext Object


apavlo:
There are certain parameters that we need during execution that need to persist across transactions.

1. `current_database`: The database that the client is currently connected to. This is currently called `default_database` in the code.

2. `temp_namespace`: This is the namespace (schema in Postgres parlance) where all temp tables will get created. Anytime that we need to do a look-up in the catalog for a table or an index, we should check this namespace first. If it doesn't exist, then we will check the 'public' namespace.

In going through #1345, I found that they were passing around the `temp_namespace` all over the place. They were storing it multiple places in the traffic_cop, binder, optimizer, and execution plans. The correct place to store it is in the `TransactionContext`. But since a session can have multiple transactions, we actually want to persist it outside of the `TransactionContext` and reuse it as much as possible. 

I propose that we create a new object called `SessionObject`. This will be created by the `TrafficCop` for a new connection/session. Then this parameter is passed into `TransactionManager::BeginTransaction` as a shared-pointer and stored in the `TransactionContext`.

PR #1345, opened 2018-05-05
[15721] Sequence second PR

bdengNE:
In the second PR, we have fixed the problems found in first PR and modified codes based on the updates in the master branch. In addition, we are currently adding features such as drop sequence and currency control, etc (still implementing them and you will see more commits to the PR very soon)

PR #1407, opened 2018-06-14
Sequences

apavlo:
This PR is a cleaned up version of #1345. I removed the need for temp tables since @poojanilangekar doesn't need that. This PR is still broken. I still need to fix the basic functionality. I also debating whether to bring in the new `SessionContext` code from #1397.

https://www.youtube.com/watch?v=3mmdD0pBBck


tli2:
There seems to be a lot of refactors in this PR not necessarily related to sequences. We can probably get those into a separate PR and merge those immediately.

PR #1408, opened 2018-06-14
Fix HashTable::BuildLazy if it contains zero emelemts

tcm-marcel:
Avoids a call to `NextPowerOf2` with zero when the hash table doesn't contain any elements in the function `HashTable::BuildLazy`. Occurred when running TPC-H Q3 in Peloton.

### DIFF
--- a/src/codegen/util/hash_table.cpp
+++ b/src/codegen/util/hash_table.cpp
@@ -191,6 +191,9 @@ char *HashTable::Insert(uint64_t hash) {
 }
 
 void HashTable::BuildLazy() {
+  // Early exit if no elements have been added (hash table is still valid)
+  if (num_elems_ == 0) return;
+
   // Grab entry head
   Entry *head = directory_[0];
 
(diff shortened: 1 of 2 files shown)


src/codegen/util/hash_table.cpp
```diff
@@ -203,7 +203,7 @@ void HashTable::BuildLazy() {
   // TODO: Use sketches to estimate the real # of unique elements
   // TODO: Perhaps change probing strategy based on estimate?
 
-  directory_size_ = NextPowerOf2(num_elems_) * 2;
+  directory_size_ = (num_elems_ == 0) ? 0 : NextPowerOf2(num_elems_) * 2;
```
pmenon: The problem with this is that you've overflowed `directory_mask_` when no elements exist.

Leave this as is and put an early exit at the top of the function. At this point, the hash table is still valid (i.e., has a directory, mask, size, capacity etc.). So, probes shouldn't find anything.

tcm-marcel: I thought about that and decided the mask is valid after that - but's you are right, it is not.

tcm-marcel:
@pmenon I changed the line to an early exit and added a test case.

Issue #1296, opened 2018-04-13
Invalid SQL statement in multi-statement transaction does not abort the transaction properly


latelatif:
If an invalid SQL statement is issued in a multi-statement transaction, the state of the transaction is set to ResultType::ABORTED and the idea is that the commit/rollback should abort the transaction. The user gets a warning when trying to run other SQL statements in the same transaction saying that the transaction is aborted and they should rollback.

On getting a rollback, the code calls AbortQueryHelper which checks for the state of the transaction and does not call abort if it is already set to ResultType::ABORTED. This results in the transaction with the invalid statement never being actually aborted.

The main reason the code waits for calling abort is that in such cases, we have to return back to the user in the context of the same transaction and hence, we cannot free up transaction context related data structures. We can get rid of the RW set and other stuff so the ideal fix is to release and give the RW set to the garbage collector as soon as one of the statements in the transaction returns failure. We can keep the transaction context till later when the user either rollbacks or commits. In either of those scenarios, we should add the transaction context data structures to the GC as well ensuring the abort of this transaction.

latelatif:
An easy way to reproduce this issue.

```default_database=# create table bar(id integer PRIMARY KEY, year integer);
CREATE TABLE
default_database=# select * from bar;
 id | year 
----+------
(0 rows)

default_database=# begin;
BEGIN
default_database=# insert into bar values(5, 400);
INSERT 0 1
default_database=# select * from bar;
 id | year 
----+------
  5 |  400
(1 row)

default_database=# insrt into bar values(5, 400);
syntax error at or near "insrt" at 1
default_database=# select * from bar;
current transaction is aborted, commands ignored until end of transaction block
default_database=# select * from bar;
current transaction is aborted, commands ignored until end of transaction block
default_database=# rollback;
ROLLBACK
default_database=# select * from bar;
 id | year 
----+------
(0 rows)

default_database=# insert into bar values(5, 400);
INSERT 0 0
default_database=# select * from bar;
 id | year 
----+------
(0 rows)
```

ChTimTsubasa:
I tried these commands on the current master branch. It does not exist. The txn state stack is clean and the last two command can success.

ChTimTsubasa:
I would add test cases to address this problem.

ChTimTsubasa:
[...]

PR #1410, opened 2018-06-16
Transaction State Handler

ChTimTsubasa:
This PR try to fix the transaction bug #1296 and do some refactor for traffic_cop by adding a wrapper around transaction context and manager.


src/include/concurrency/transaction_state_handler.h
```diff
@@ -0,0 +1,45 @@
...
+namespace concurrency {
+  enum class TransactionState{
+   IDLE = 0,
+   IMPLICIT_START,
+    EXPLICIT_START,
+    ABORTING,
+    ENDED,
+ };
+
+class TransactionStateHandler {
```
tli2: Maybe instead of TransactionStateHandler, call it ClientTransactionHandle or something like that, since this component stores all relevant information in the network layer about what a client's transaction is doing, instead of performing anything on transactions.

src/include/concurrency/transaction_state_handler.h
```diff
@@ -0,0 +1,45 @@
+#pragma once
```
tli2: headers?

src/include/traffic_cop/traffic_cop.h
```diff
@@ -2,11 +2,11 @@
 //
 //                         Peloton
 //
-// traffic_cop.h
```
tli2: This doesn't look like your intended changes.

tli2:
Will be incorporated into larger network refactor PR.

Issue #1398, opened 2018-06-10
Clean-up Catalog Infrastructure


apavlo:
Our catalog code is a mess. It's inconsistent. We need to clean it up so that it is easier to understand and follows the proper logical hierarchy.

I propose the following changes:

1. Refactor the methods so that `TransactionContext` is always passed in as the first argument.

2. Refactor the `Catalog` methods so that parameters are passed in using the same order as the hierarchy (i.e., Database→Schema→Table). I think that Indexes should be under Table as well. Right now they are under Databases.

3. `DatabaseCatalogObject` should **not** allow you to get `TableCatalogObject`. `DatabaseCatalogObject` should only allow you to get `SchemaCatalogObject` and then all of the tables should be stored in `SchemaCatalogObject` instead.

tli2:
Speaking of XXXCatalogObjects, what are them exactly? From what I understand they are results from a read. Would it make more sense to rename them to "CatalogEntry" or something like that? I remember getting confused by it while doing class project and not being able to find much documentation explaining what it is

apavlo:
@tli2 I am okay with renaming them to be `CatalogEntry`.

tli2:
Since I am blocked on a weird test failure in #1404, I can quickly fix some of these today.

tli2:
5. Remove all default parameters (or document them clearly for the ones that really make sense), per  #1414

tli2:
As @pervazea pointed out, we are a long way from done on this one, as there are multiple naming issues and other inconsistencies in the code. To make matters worse very little documentation is provided so it is unclear what any of the API should behave.

PR #1414, opened 2018-06-18
Catalog code cleanup

tli2:
Addresses issue #1398.

This PR fixes 1, 2 and renames CatalogObjects to Entries. Other minor code style fixes are included as well.


src/binder/binder_context.cpp
```diff
@@ -38,8 +38,10 @@ void BinderContext::AddRegularTable(const std::string db_name,
                                     const std::string table_alias,
                                     concurrency::TransactionContext *txn) {
   // using catalog object to retrieve meta-data
-  auto table_object = catalog::Catalog::GetInstance()->GetTableObject(
-      db_name, schema_name, table_name, txn);
+  auto table_object = catalog::Catalog::GetInstance()->GetTableObject(txn,
```
apavlo: We should probably rename this as `GetTableEntry` too.

src/catalog/abstract_catalog.cpp
```diff
@@ -71,21 +71,27 @@ AbstractCatalog::AbstractCatalog(const std::string &catalog_table_ddl,
   auto catalog_database_name = create_plan->GetDatabaseName();
   PELOTON_ASSERT(catalog_schema_name == std::string(CATALOG_SCHEMA_NAME));
   // create catalog table
-  Catalog::GetInstance()->CreateTable(
-      catalog_database_name, catalog_schema_name, catalog_table_name,
-      std::unique_ptr<catalog::Schema>(catalog_table_schema), txn, true);
+  Catalog::GetInstance()->CreateTable(txn,
```
apavlo: Unless this function also creates the `DataTable` object (which it shouldn't), maybe we should rename this to `CreateTableEntry`.

src/include/catalog/database_catalog.h
```diff
@@ -104,10 +110,9 @@ class DatabaseCatalog : public AbstractCatalog {
   ~DatabaseCatalog();
 
   // Global Singleton, only the first call requires passing parameters.
-  static DatabaseCatalog *GetInstance(
-      storage::Database *pg_catalog = nullptr,
-      type::AbstractPool *pool = nullptr,
-      concurrency::TransactionContext *txn = nullptr);
+  static DatabaseCatalog *GetInstance(concurrency::TransactionContext *txn = nullptr,
```
apavlo: Can we remove the default values?

src/catalog/catalog.cpp
```diff
@@ -65,12 +65,14 @@ Catalog::Catalog() : pool_(new type::EphemeralPool()) {
...
   // Create catalog tables
-  DatabaseCatalog::GetInstance(peloton, pool_.get(), txn);
-  BootstrapSystemCatalogs(peloton, txn);
+  DatabaseCatalog::GetInstance(txn, peloton, pool_.get());
+  BootstrapSystemCatalogs(txn, peloton);
 
   // Insert peloton database into pg_database
-  DatabaseCatalog::GetInstance()->InsertDatabase(
-      CATALOG_DATABASE_OID, CATALOG_DATABASE_NAME, pool_.get(), txn);
+  DatabaseCatalog::GetInstance(nullptr, nullptr, nullptr)->InsertDatabase(txn,
```
apavlo: Why pass a null txn pointer to `DatabaseCatalog::GetInstance()` when you actually have the txn pointer?

src/catalog/catalog.cpp
```diff
@@ -674,33 +848,36 @@ ResultType Catalog::DropDatabaseWithOid(oid_t database_oid,
...
+ResultType Catalog::DropSchema(concurrency::TransactionContext *txn,
+                               const std::string &database_name,
+                               const std::string &schema_name) {
   if (txn == nullptr)
     throw CatalogException("Do not have transaction to drop schema " +
                            schema_name);
 
   auto database_object =
-      DatabaseCatalog::GetInstance()->GetDatabaseObject(database_name, txn);
+      DatabaseCatalog::GetInstance(nullptr,
```
apavlo: Same here. You actually have the txn pointer.

src/catalog/catalog.cpp
```diff
@@ -1099,23 +1316,29 @@ void Catalog::AddBuiltinFunction(
  * @param   code_context  the code_context that holds the generated LLVM
  *                        query code
  */
-void Catalog::AddPlpgsqlFunction(
-    const std::string &name, const std::vector<type::TypeId> &argument_types,
-    const type::TypeId return_type, oid_t prolang, const std::string &func_src,
-    std::shared_ptr<peloton::codegen::CodeContext> code_context,
-    concurrency::TransactionContext *txn) {
+void Catalog::AddPlpgsqlFunction(concurrency::TransactionContext *txn,
```
apavlo: I don't think that this should be called `AddPlpgsqlFunction`. You are passing in the `prolang` argument, so it should just be called `AddFunction`, right? Furthermore, we are are referring to UDFs as procedures (i.e., `pg_proc` table), so it really should be called `AddProcedure`.

src/catalog/catalog.cpp
```diff
@@ -1188,14 +1411,16 @@ void Catalog::InitializeLanguages() {
     auto &txn_manager = concurrency::TransactionManagerFactory::GetInstance();
     auto txn = txn_manager.BeginTransaction();
     // add "internal" language
-    if (!LanguageCatalog::GetInstance().InsertLanguage("internal", pool_.get(),
-                                                       txn)) {
+    if (!LanguageCatalog::GetInstance().InsertLanguage(txn,
```
apavlo: This is not your problem, but we should **not** be initializing the language table in this function.

src/catalog/catalog.cpp
tli2: I had no idea what this is supposed to do. Will change.

src/catalog/catalog.cpp
tli2: Ack

tli2: Done. #1421

src/binder/binder_context.cpp
apavlo: We should vote on this. I think `Get*CatalogEntry` would be best.

ksaito7: I vote `Get*CatalogEntry`. I think the accuracy of function name is more important than length. And I think this name is not so long.

src/include/catalog/language_catalog.h
```diff
@@ -63,28 +63,29 @@ class LanguageCatalog : public AbstractCatalog {
...
+                                                         oid_t lang_oid) const;
 
-  std::unique_ptr<LanguageCatalogObject> GetLanguageByName(
-      const std::string &lang_name, concurrency::TransactionContext *txn) const;
+  std::unique_ptr<LanguageCatalogEntry> GetLanguageByName(concurrency::TransactionContext *txn,
+                                                          const std::string &lang_name) const;
 
   enum ColumnId {
     OID = 0,
     LANNAME = 1,
```
pervazea: Poor abbreviation. Should be LANG or LANGUAGE

src/binder/binder_context.cpp
pervazea: I'd go with apavlo, ksaito7 and vote for ... Entry.

tli2: Fine, I'll rename them to GetxxxCatalogEntry

src/catalog/abstract_catalog.cpp
tli2: Ironically, I think it does create the DataTable object (Not that it should)

PR #1422, opened 2018-06-21
Forecasting Code Cleanup and Model Addition

saatviks:
This PR adds the following changes:
1. **Eigen**: Additional Eigen utility functions needed for the new models. Also added a test which checks correctness of all utility functions.
2. **Model Hierarchy Setup**: Addition of the following Abstract classes - **BaseModel, BaseTFModel, BaseForecastModel**. All models must inherit **BaseModel**. **BaseTFModel** provides ready to use APIs for importing and running the Tensorflow API and can be used across model types(meant for forecasting or otherwise) if Tensorflow functionality is needed. **BaseForecastModel** is meant only for Forecasting(like the current workload forecasting models). 
3. **Model Additions**: Added Linear Regression and Kernel Regression model.
4. **Model Utilitiles**: Set of utility functions which can be used by models or other classes using those models. Added tests for checking each function.
5. Setting things up for Ensembling - Add **Fit/Predict** as BaseModel pure virtual methods, clean up the LSTM Train/Validate Epoch. 
6. Write out the Ensembling logic and test.


src/brain/util/model_util.cpp
```diff
@@ -0,0 +1,88 @@
...
+  std::vector<matrix_eig> feat_matrix, forecast_matrix;
+  for (size_t idx = 0; idx <= timesteps - model->GetHorizon() - model->GetBPTT();
+       idx++) {
+    feat_matrix.push_back(EigenUtil::ToEigenMat(
+        {EigenUtil::Flatten(data.middleRows(idx, model->GetBPTT()))}));
+    forecast_matrix.emplace_back(
+        data.row(idx + model->GetBPTT() + model->GetHorizon() - 1));
+  }
+  processed_features = EigenUtil::VStack(feat_matrix);
+  processed_forecasts = EigenUtil::VStack(forecast_matrix);
```
malin1993ml: Will these invoke copy constructors?

src/include/brain/util/model_util.h
```diff
@@ -0,0 +1,65 @@
...
+
+  /**
+   * Given data of the form (timesteps, feat_len), this function generates
+   * a feature matrix appropriate for learning a forecast model.
+   * Useful for Linear and Kernel Regression Models
+   */
+  static void GenerateFeatureMatrix(const BaseForecastModel *model,
+                                    const matrix_eig &data,
+                                    matrix_eig &processed_features,
+                                    matrix_eig &processed_forecasts);
```
malin1993ml: Is there any special reason you're using pointers for models but references for all other parameters?

src/include/brain/workload/linear_models.h
```diff
@@ -0,0 +1,60 @@
...
+  /**
+   * @return std::string representing model object
+   */
+  std::string ToString() const override;
+
+ private:
+  std::vector<vector_eig> weights_;
+};
+
+class TimeSeriesKernelReg : public BaseForecastModel {
```
malin1993ml: Kernel regression is not really a "linear model". At least not the linear models we usually refer to. I'd prefer separate "linear_model.h" and "kernel_model.h" to avoid confusion.

src/brain/workload/linear_models.cpp
```diff
@@ -0,0 +1,135 @@
...
+  model_str_builder << ", interval = " << interval_;
+  model_str_builder << ")";
+  return model_str_builder.str();
+}
+
+void TimeSeriesLinearReg::Fit(const matrix_eig &X,
+                              const matrix_eig &y,
+                              UNUSED_ATTRIBUTE int bsz) {
+  matrix_eig XTX = X.transpose() * X;
+  XTX += matrix_eig::Identity(XTX.rows(), XTX.rows());
```
malin1993ml: I think you also want to multiply a small epsilon with the identity.

src/brain/util/model_util.cpp
saatviks: `EigenUtil::VStack` takes a const ref input - so it wont.

src/include/brain/util/model_util.h
saatviks: No specific reason - just since I wanted to use it as `GenerateFeatureMatrix(this` instead of `GenerateFeatureMatrix(*this`. I don't think there should be any additional cost though in dereferencing to a const ref.

test/brain/model_test.cpp
```diff
@@ -0,0 +1,133 @@
...
+
+  size_t LOG_INTERVAL = 20;
+  size_t NUM_SAMPLES = 1000;
+  size_t NUM_FEATS = 3;
+  float VAL_SPLIT = 0.5;
+  bool NORMALIZE = false;
+  float VAL_THESH = 0.06;
+
+  ////  auto models_vec = {std::move(model1), std::move(model2),
+  ///std::move(model3)};
```
malin1993ml: Remove the unused code.

test/brain/testing_forecast_util.cpp
```diff
@@ -0,0 +1,114 @@
...
+  train_data = n.Transform(train_data);
+  matrix_eig test_data = n.Transform(data.bottomRows(split_point));
+
+  vector_eig train_loss_avg = vector_eig::Zero(val_interval);
+  float prev_train_loss = std::numeric_limits<float>::max();
+  float val_loss = val_loss_thresh * 2;
+  std::vector<float> val_losses;
+  for (int epoch = 1; epoch <= model.GetEpochs() &&
+                      !brain::ModelUtil::EarlyStop(
+                          val_losses, early_stop_patience, early_stop_delta);
```
malin1993ml: Please remeber to add the default parameters for this.

src/include/brain/util/model_util.h
```diff
@@ -0,0 +1,97 @@
...
+   * **keras.callbacks.EarlyStopping** and based on below references
+   * @param val_losses: vector of validation losses
+   * @param patience: number of epochs to wait for a threshold change in loss
+   * @param delta: minimum needed threshold change to continue training
+   * @return: True if one should stop.
+   * Refences: (1) https://stats.stackexchange.com/q/231061/72908
+   *           (2) Check: "Early Stopping - but when?" for a more detailed
+   *           suite of methods.
+   */
+  static bool EarlyStop(vector_t val_losses, size_t patience = 10, float delta = 1e-2);
```
malin1993ml: Eh this is not straightforward and easy to confuse people. Please put these default numbers in a central parameter place. For example `workload_defaults.h ` is fine.

Issue #1424, opened 2018-06-22
Setting issues


ksaito7:
- SettingsManager controls transaction for SettingsCatalog. So checkpoints can’t recover settings due to nested transaction.
- Can’t add new settings after Peloton booting. So we can’t implement SET command.
- There is no update function in SettingsCatalog.
- Why setting control functions are static?

So I propose setting refactoring to SettingManager and SettingsCatalog. Basically, I think setting user should control all functions through SettingManager, not static like other manager. Refactor points are below:

- All setting values are controlled through SettingsManager (Set/Get value, Add setting)
- A `TransactionContext` is passed through the function argument.
- Implement update function in SettingsCatalog.

tli2:
There is a bunch of Set methods in SettingsManager. I don't know how they interact with transactions though.

ksaito7:
Set methods are just update methods for existed setting value using `setting_id`. These can not create new settings actually.

pervazea:
What new settings do you wish to add (after the system has booted)?

ksaito7:
I wanted users to be able to set original variables through SET command in psql or config files. But I found that Postgres can't set the original variables. I misunderstood that. Delete its line.

PR #1432, opened 2018-06-26
Settings refactoring

ksaito7:
This PR refactors setting stuff in order to enable checkpoint manager to recover setting values. In addition,  added some convenient functions.

**- Fix Points of SettingsManager**

1. Change All set functions to non-static, and move all set/get functions to public.
1. Add `txn` argument into set functions in order that a caller of set functions can manage a necessary transaction by itself.
1. Add `set_default` argument into set functions. If this flag argument is true, then the default value is updated in addition to updating the setting value.
1. Add `Reset` public function to reset the setting value to default value.
1. Add `UpdateSettingListFromCatalog()` public function to update setting values in the manager from catalog for checkpoints.
1. Add/Modify two catalog management private functions: `InsertCatalog()`, `UpdateCatalog()`.

**- Fix Points of SettingsCatalog**

1. Add `SettingsCatalogEntry` class to acquire all setting information from catalog
1. Add `GetSetting()` and `GetSettings()` functions.
1. Delete other get functions.
1. Add `UpdateSettingValue()` function to update a setting value and a default value.

**- Related Issues**
1. Issues #1424

### DIFF
--- a/src/catalog/catalog.cpp
+++ b/src/catalog/catalog.cpp
@@ -200,7 +200,8 @@ void Catalog::Bootstrap() {
   // TODO: change pg_proc to per database
   ProcCatalog::GetInstance(txn);
 
-  if (settings::SettingsManager::GetBool(settings::SettingId::brain)) {
+  if (settings::SettingsManager::GetInstance().GetBool(
+          settings::SettingId::brain)) {
     QueryHistoryCatalog::GetInstance(txn);
   }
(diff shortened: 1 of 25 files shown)


ksaito7:
This last commit isn't related to this PR directly, but I did it because a part of index names of constraints were weird. (related to #1415 we worked for last time)

PR #1434, opened 2018-06-26
fix #1386: PerformVectorizedRead optimization

### LINKED ISSUE #1386: PerformVectorizedRead optimization
1. We should separate the visibility check from the addition to the read-write set in TransactionRuntime::PerformVectorizedRead(). (transaction_runtime.cpp line 28)
Visibility should be checked for all tuples, but PerformRead() should only be called for tuples that pass the scan predicate. Otherwise, *all* visible tuples will be added to the read-write set, which is too conservative.  (Prashanth suggesting splitting the function into two. Check visibility. Interpose scan predicate check. Second portion that does PerformRead).

2. The last argument to TransactionManager::PerformRead() should not be false (transaction_runtime.cpp, line 62).
It should be a function argument fed from the scan plan (i.e., SeqScanPlan::IsForUpdate).

The first change should yield significant improvement for high-selectivity predicates since 54% of time is spend in PerformRead(). The second one is for correctness.

mbutrovich:
I did the fixes suggested in #1386. Be gentle, it's my first time touching codegen :)

### DIFF
--- a/src/codegen/operator/table_scan_translator.cpp
+++ b/src/codegen/operator/table_scan_translator.cpp
@@ -105,7 +105,7 @@ class TableScanTranslator::ScanConsumer : public codegen::ScanCallback {
                               llvm::Value *tid_end,
                               Vector &selection_vector) const;
 
-  void RecordReads(CodeGen &codegen, Vector &selection_vector) const;
+  void PerformReads(CodeGen &codegen, Vector &selection_vector) const;
 
   // Filter all the rows whose TIDs are in the range [tid_start, tid_end] and
   // store their TIDs in the output TID selection vector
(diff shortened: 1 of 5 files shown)


src/codegen/operator/table_scan_translator.cpp
```diff
@@ -105,6 +105,8 @@ class TableScanTranslator::ScanConsumer : public codegen::ScanCallback {
                               llvm::Value *tid_end,
                               Vector &selection_vector) const;
 
+  void RecordReads(CodeGen &codegen, Vector &selection_vector) const;
```
pervazea: There has been discussion of the transaction read-write set, recording writes and recording reads, that I was not immediately clear on what this function does. Should it be renamed to be "PerformReads" to more accurately describe what it does?

mbutrovich: That sounds reasonable.

PR #1435, opened 2018-06-27
Support limit, offset, top-k serial and parallel versions in codegen

pmenon:
After #1385 was merged in, I resurrected `LIMIT`, `OFFSET` from my old branch, and implemented Top-K optimizations. These both work for serial and parallel execution, and are needed for TPC-C.

Not quite ready for merging in, but folks can take a look at it.


src/codegen/function_builder.cpp
```diff
@@ -261,23 +293,33 @@ void FunctionBuilder::ReturnAndFinish(llvm::Value *ret) {
...
+
+  // Move to start of function
+  if (entry_bb_->empty()) {
+    codegen->SetInsertPoint(entry_bb_);
+  } else {
+    codegen->SetInsertPoint(&entry_bb_->back());
+  }
+
+  // Generate loading code and save in cache
+  val = load_func();
```
tcm-marcel: Where does the caching happen?

src/codegen/operator/limit_translator.cpp
```diff
@@ -0,0 +1,106 @@
...
+
+  llvm::Value *next_limit = nullptr;
+  if (!context.GetPipeline().IsParallel()) {
+    // Serial mode. Just add increment the count.
+    next_limit = codegen->CreateLoad(limit_count_ptr);
+    next_limit = codegen->CreateAdd(next_limit, codegen.Const64(1));
+    codegen->CreateStore(next_limit, limit_count_ptr);
+  } else {
+    // Parallel mode. Atomically increment the count.
+    next_limit = codegen->CreateAtomicRMW(
```
tcm-marcel: Do you know what this looks like in IR and what LLVM compiles it to?

src/codegen/operator/order_by_translator.cpp
```diff
@@ -124,9 +124,9 @@ OrderByTranslator::OrderByTranslator(const planner::OrderByPlan &plan,
                                      CompilationContext &context,
                                      Pipeline &pipeline)
     : OperatorTranslator(plan, context, pipeline),
-      child_pipeline_(this, Pipeline::Parallelism::Serial) {
-  // Aggregations happen serially (for now ...)
-  pipeline.SetSerial();
+      child_pipeline_(this, Pipeline::Parallelism::Flexible) {
```
tcm-marcel: Do I understand correctly: The child pipeline of the Orderby (before the hashing) can be parallel, but the order-by itself has to be serial because of the aggregations.

src/codegen/util/sorter.cpp
```diff
@@ -97,8 +131,11 @@ void Sorter::Sort() {
 
   timer.Stop();
 
-  LOG_DEBUG("Sorted %zu tuples in %.2f ms", tuples_.size(),
-            timer.GetDuration());
+#ifndef NDEBUG
```
tcm-marcel: This should be `#ifdef LOG_DEBUG_ENABLED`

src/include/codegen/util/sorter.h
```diff
@@ -143,6 +219,17 @@ class Sorter {
...
+  /**
+   * Build a (max) heap from the tuples currently stored in the sorter instance.
+   */
+  void BuildHeap();
+
+  /**
+   * Sift down the element at the root of the heap while maintaining the heap
+   * property.
+   */
+  void HeapSiftDown();
```
tcm-marcel: What is the difference of this function and STL `push_heap()`?

src/codegen/operator/order_by_translator.cpp
pmenon: This is saying is that the consumer portion of the order-by (i.e., the child pipeline) can be executed in parallel, and that the production portion (i.e., the scanning of the sorter) will be serial. It has no information about other operators in the pipeline.

src/include/codegen/util/sorter.h
pmenon: In essence, they are the same. But, `push_heap` alone doesn't give us the functionality we need. We need to remove the smallest (largest) element off the min (max) heap and insert a new element. We could use a pair of `push_heap` and `pop_heap` calls, but we've managed to achieve the same functionality with a single `O(logn)` operation.

src/codegen/operator/limit_translator.cpp
pmenon: We generate LLVM's `atomicrmw` with sequential consistency. This may lower to different instructions depending on architecture, but on the few Intel Skylake+ machines I tested, it uses `lock xadd`.

pmenon:
@tcm-marcel Want to take a last crack at this? Don't want this to get stale.

apavlo:
@tcm-marcel I think this is stuck on the Jenkins mac build?

tcm-marcel:
The Jenkins Mac build fails because no space is left on the hard drive:

```
/tmp/workspace/peloton_PR-1435-XOL5JG3LYATJV2HWKIVAIOIZQP2OCJC5FY5VVGBXQZ6BT33JVMUA@2: No space left on device
```

@apavlo I am not familiar with the Jenkins Mac machine. But on Travis the Mac build succeeds.

apavlo:
@crd477 Can you check this?

crd477:
Fixed with the quick fix of restarting the VM from a clean snapshot.  The more complete fix of how to do this cleanly and regularly was still on my TODO list before I left for vacation last week and I'll get back to it today.
I also restarted the Jenkins builds for master and #1435.  They should pass soon.

crd477:
The macOS Jenkins worker should be stable now.  I grew the VM's disk and added a workspace cleanup cronjob as the other Jenkins workers have.

Issue #1427, opened 2018-06-25
SerializableTransactionTest might be failing spuriously


tli2:
Several complaints about this failing from time to time on Travis. Might be a Mac issue, might be that the test is broken, might be that we broke something in CC.

Investigation would be appreciated.

mbutrovich:
Here's an example of how it fails, and on rerun it typically passes. Maybe some nondeterminism in how stuff is being scheduled:

```
        Start  46: serializable_transaction_test



46: Test command: /Users/travis/build/cmu-db/peloton/build/test/serializable_transaction_test "--gtest_color=yes" "--gtest_output=xml:/Users/travis/build/cmu-db/peloton/build/test/serializable_transaction_test.xml"

46: Environment variables: 

46:  LSAN_OPTIONS=suppressions=/Users/travis/build/cmu-db/peloton/test/leak_suppr.txt

46: Test timeout computed to be: 9.99988e+06

46: Running main() from gmock_main.cc

46: [0;32m[==========] [mRunning 6 tests from 1 test case.

46: [0;32m[----------] [mGlobal test environment set-up.

46: [0;32m[----------] [m6 tests from SerializableTransactionTests

46: [0;32m[ RUN      ] [mSerializableTransactionTests.TransactionTest

46: [0;32m[       OK ] [mSerializableTransactionTests.TransactionTest (36 ms)

46: [0;32m[ RUN      ] [mSerializableTransactionTests.ReadOnlyTransactionTest

46: 2018-06-25 20:34:00 [test/concurrency/testing_transaction_util.cpp:192:CreateTable] INFO  - the database_id is 16777216

46: 2018-06-25 20:34:00 [test/concurrency/serializable_transaction_test.cpp:330:ExecuteNext] INFO  - Txn 0 commits: Success

46: /Users/travis/build/cmu-db/peloton/test/concurrency/serializable_transaction_test.cpp:90: Failure

46: Value of: scheduler.schedules[0].results.size()

46:   Actual: 0

46: Expected: 10

46: [0;31m[  FAILED  ] [mSerializableTransactionTests.ReadOnlyTransactionTest (366 ms)

46: [0;32m[ RUN      ] [mSerializableTransactionTests.ConcurrentReadOnlyTransactionTest

46: 2018-06-25 20:34:01 [test/concurrency/testing_transaction_util.cpp:192:CreateTable] INFO  - the database_id is 16777216

[...]

PR #1443, opened 2018-06-29
Increase sleep time for occasionally failing ReadOnlyTest.

mbutrovich:
This hopefully addresses #1427. I still don't fully understand why this is the only test that occasionally fails in this manner. In theory a lot of our tests that sleep for exactly 1 epoch duration and expect the epoch to have advanced are making a big assumption about thread scheduling. Most should probably relax their sleep time like this change.


tli2:
I am probably also going to restart the build on this multiple times before merging it to make sure it works

PR #1445, opened 2018-06-29
Protocol Layer Refactor

tli2:
Protocol handler and traffic cop, as it stands currently, are completely intertwined, and rank very high on my "worst code ever seen" list. There is so much duct tape and coat hangers here, it pretty wrecks testability and extensibility of the code. We will have to refactor this in phases.

This PR will separate out all the components like they should be. @ChTimTsubasa  is working on another PR based on this to fix the broken transaction handling logic.


tli2:
There are still failures, I will fix this concurrently with review.

src/common/internal_types.cpp
```diff
@@ -2030,9 +2030,6 @@ std::string ResultTypeToString(ResultType type) {
     case ResultType::UNKNOWN: {
       return ("UNKNOWN");
     }
-    case ResultType::QUEUING: {
```
ChTimTsubasa: Good

src/include/network/network_types.h
```diff
@@ -41,5 +41,14 @@ enum class Transition {
   NEED_SSL_HANDSHAKE,
   NEED_WRITE
 };
+
+enum class ResponseProtocol {
+  // No response required (for intermediate messgaes such as parse, bind, etc.)
+  NO,
+  // PSQL
+  SIMPLE,
+  // JDBC, PQXX, etc.
```
ChTimTsubasa: PQXX can use both SIMPLE and EXTENDED protocol

src/include/network/network_types.h
```diff
@@ -41,5 +41,14 @@ enum class Transition {
   NEED_SSL_HANDSHAKE,
   NEED_WRITE
 };
+
+enum class ResponseProtocol {
+  // No response required (for intermediate messgaes such as parse, bind, etc.)
```
ChTimTsubasa: Parse and bind would require responses(ParseOK, BindOK etc). They just do not need responses immediately.

src/include/network/protocol_interpreter.h
```diff
@@ -0,0 +1,32 @@
...
+namespace peloton {
+namespace network {
+
+class ProtocolInterpreter {
+ public:
+  virtual Transition Process(std::shared_ptr<ReadBuffer> in,
+                             std::shared_ptr<WriteQueue> out,
+                             CallbackFunc callback) = 0;
+
+  // TODO(Tianyu): Do we really need this crap?
```
ChTimTsubasa: Any other suggestion?

src/include/network/network_types.h
tli2: I think this is a legacy from a previous refactor that is no longer used. Will remove.

Issue #1448, opened 2018-07-02
Travis failure due to tensorflow requirement not found


lmwnshn:
On the first travis box, we fail the tensorflow test because pip cannot find tensorflow 1.4.0

```
Collecting tensorflow==1.4.0
[31m  Could not find a version that satisfies the requirement tensorflow==1.4.0 (from versions: )[0m
[31mNo matching distribution found for tensorflow==1.4.0
```

This affects recent PRs such as #1443 , #1446 .

mbutrovich:
This seems to coincide with the  Python 3.7 release on Homebrew a few days ago. Looks like tensorflow hasn't updated their compatibility list to include 3.7:
https://github.com/tensorflow/tensorflow/issues/17022

Note the Programming Language on this page doesn't list 3.7 yet:
https://pypi.org/project/tensorflow/

I suspect it'll be resolved quickly.

saatviks:
Just to add to this,
Also tried `pip3 install --upgrade https://storage.googleapis.com/tensorflow/mac/${TF_TYPE}/tensorflow-${TF_VERSION}-py3-none-any.whl` which is meant to be for all python versions, but isnt compatible with python3.7 currently.

tli2:
I think it is still failing on Travis last I checked (Your PR)

tli2:
Temporarily fixed due to  #1450, we should re-enable dependency when possible.

PR #1449, opened 2018-07-02
Disable TensorFlow tests to allow Travis macOS builds to succeed.

mbutrovich:
Workaround for #1448. This change should be reverted once the tensorflow package is updated with Python 3.7 compatibility.

### DIFF
--- a/test/brain/tensorflow_test.cpp
+++ b/test/brain/tensorflow_test.cpp
@@ -28,7 +28,8 @@ namespace test {
 
 class TensorflowTests : public PelotonTest {};
 
-TEST_F(TensorflowTests, BasicTFTest) {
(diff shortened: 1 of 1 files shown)


saatviks:
Thanks for pointing out the problem @mbutrovich 
An alternative as described here would be to install directly as `pip3 install https://storage.googleapis.com/tensorflow/mac/cpu/tensorflow-1.4.0-py3-none-any.whl`.

saatviks:
Seems like the solution in #1450  also doesnt work as detailed in this [issue](https://github.com/tensorflow/tensorflow/issues/20444) here. Best to go ahead with this fix of disabling tensorflow tests. I'm closing mine. 
Also notifying @malin1993ml that we are doing this.

PR #1454, opened 2018-07-06
Mac jenkins

crd477:
add a stage for building on macOS Jenkins workers


crd477:
The Jenkins build bombed on the Trusty builds for some unrelated reason so I restarted it.  I think this can be merged now.

PR #1463, opened 2018-07-13
GetInfo (#1369)

### LINKED ISSUE #1369: No informative GetInfo() method for operator nodes and some plan nodes
This is a little bit different from what we discussed in the meeting, I'll explain in the description.

There's no GetInfo() method for operator nodes in the optimizer, which makes debugging the optimizer extremely painful.

https://github.com/cmu-db/peloton/blob/master/src/include/optimizer/operators.h#L52

There are some difficulties implementing it I can think of. First is operators are stored in the memo, so we need to construct an operator tree when we want to print it. I think printing the best operator tree in the memo should be good enough for most cases. We could take the `GetBestPlan` method as a reference.

https://github.com/cmu-db/peloton/blob/master/src/optimizer/optimizer.cpp#L288

Printing memo at arbitrary time during optimization would be helpful when we implement more complex optimizations

Another nice-to-have feature is to print output columns for each operator, e.g. `t1.a + AVG(t2.b)`, as it's also something we could easily mess up. The challenge is these output columns are not stored in the operators. They are constructed on the fly and stored in a map when constructing the plan.

https://github.com/cmu-db/peloton/blob/master/src/include/common/internal_types.h#L1392 

I previously thought we could print these in the plan node, but we only store output column offset as `oid` in the plan node, which is not the most intuitive debugging info.

I think we should print out predicates, e.g. scan predicates, join predicates, having clauses, in the plan node. I've added some already. Please check if there's anything left. 

By the way, `GetInfo()` method for expressions is also kind of crappy, it's not succinct enough, an expression `t1.a + AVG(t2.b)` spans multiple lines with a lot of redundant information. We should also fix it.

To summarize, the features needed to be implemented are:

1. `GetInfo()` for the best operator tree in the optimizer
2. Informative `GetInfo()` for plan nodes
3. Succinct `GetInfo()` method for expressions

The features that are nice to have:

1. `GetInfo()` for memo
2. Printing output columns with operator tree nodes.

yetiancn:
1. Add GetOperationInfo method for memo in order to print the best operator tree
2. Add GetInfo for some plan nodes
3. Shorten GetInfo for expressions
4. Add GetInfo for memo
fix #1369

### DIFF
--- a/src/include/optimizer/group.h
+++ b/src/include/optimizer/group.h
@@ -32,7 +32,7 @@ class ColumnStats;
 //===--------------------------------------------------------------------===//
 // Group
 //===--------------------------------------------------------------------===//
-class Group {
+class Group : public Printable {
  public:
   Group(GroupID id, std::unordered_set<std::string> table_alias);
 
@@ -90,10 +90,10 @@ class Group {
 
   int GetNumRows() { return num_rows_; }
 
-  inline GroupID GetID() { return id_; }
+  inline GroupID GetID() const { return id_; }
 
-  const std::string GetInfo(int num_indent);
-
+  const std::string GetInfo(int num_indent) const;
+  const std::string GetInfo() const override;
 
   // This is called in rewrite phase to erase the only logical expression in the
   // group
--- a/src/include/optimizer/group.h
+++ b/src/include/optimizer/group.h
@@ -92,6 +92,9 @@ class Group {
 
   inline GroupID GetID() { return id_; }
 
+  const std::string GetInfo(int num_indent);
+
+
   // This is called in rewrite phase to erase the only logical expression in the
   // group
   inline void EraseLogicalExpression() {
--- a/src/include/planner/csv_scan_plan.h
+++ b/src/include/planner/csv_scan_plan.h
@@ -61,7 +61,7 @@ class CSVScanPlan : public AbstractScan {
 
   const std::string &GetFileName() const { return file_name_; }
 
-  const std::string GetInfo() const override { return "CSVScan"; }
+  const std::string GetInfo() const override { return "CSVScanPlan"; }
 
   void GetAttributes(std::vector<const AttributeInfo *> &ais) const override;
 
--- a/src/include/planner/csv_scan_plan.h
+++ b/src/include/planner/csv_scan_plan.h
@@ -61,7 +61,7 @@ class CSVScanPlan : public AbstractScan {
 
   const std::string &GetFileName() const { return file_name_; }
 
-  const std::string GetInfo() const { return "CSVScan"; }
+  const std::string GetInfo() const override { return "CSVScan"; }
 
   void GetAttributes(std::vector<const AttributeInfo *> &ais) const override;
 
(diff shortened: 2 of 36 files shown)


src/optimizer/group.cpp
```diff
@@ -75,6 +75,74 @@ std::shared_ptr<ColumnStats> Group::GetStats(std::string column_name) {
...
 
+const std::string Group::GetInfo(int num_indent) {
+    std::ostringstream os;
+    os << StringUtil::Indent(num_indent)
+       << "GroupID: " << GetID() << std::endl;
+
+    if (logical_expressions_.size() > 0)
+        os << StringUtil::Indent(num_indent + 2)
+           << "logical_expressions_: \n";
+    for (size_t i = 0; i < logical_expressions_.size(); ++i) {
```
tli2: Is there a specific reason why these cannot be for-each loops?

src/include/optimizer/group.h
```diff
@@ -92,6 +92,9 @@ class Group {
 
   inline GroupID GetID() { return id_; }
 
+  const std::string GetInfo(int num_indent);
```
tli2: explicitly inherit from peloton::Printable if you intend this to be usable as one

src/include/planner/csv_scan_plan.h
```diff
@@ -61,6 +61,8 @@ class CSVScanPlan : public AbstractScan {
 
   const std::string &GetFileName() const { return file_name_; }
 
+  const std::string GetInfo() const override { return "CSVScan"; }
```
tli2: Maybe add "plan". Providing concrete information for each field would be even better (I realize the other plans don't do that either, so it's up to you)

src/optimizer/group.cpp
```diff
@@ -75,6 +75,74 @@ std::shared_ptr<ColumnStats> Group::GetStats(std::string column_name) {
...
+        os << StringUtil::Indent(num_indent + 4)
+           << logical_expressions_[i]->Op().GetName() << std::endl;
+        const std::vector<GroupID> ChildGroupIDs = logical_expressions_[i]->GetChildGroupIDs();
+        if (ChildGroupIDs.size() > 0) {        
+                os << StringUtil::Indent(num_indent + 6)
```
tli2: Check your indentation

src/optimizer/memo.cpp
```diff
@@ -69,6 +69,26 @@ std::vector<std::unique_ptr<Group>> &Memo::Groups() {
 
 Group *Memo::GetGroupByID(GroupID id) { return groups_[id].get(); }
 
+
+const std::string Memo::GetInfo(int num_indent) {
+    std::ostringstream os;
+    os << StringUtil::Indent(num_indent) << "Memo::\n";
+    os << StringUtil::Indent(num_indent + 1) 
+       << "rule_set_size_: " << rule_set_size_ << std::endl;
+    for (size_t i = 0; i < groups_.size(); ++i) {
```
tli2: Same here

apavlo:
@tli2 Is there a way for you to mark your requested changes as fixed? The only option I have is "Dismiss Review".

tli2:
@apavlo Yeah I think you just need to dismiss the review (or simply giving approval should overwrite that too)

PR #1473, opened 2018-08-17
Add first implementation of augmentedNN to predict selectivity

yetiancn:
The model is an initial implementation to predict selectivity for range predicates.
It can be applied to queries like:
  `SELECT * FROM table WHERE c >= l AND c <= u`.

I implement the model in augmentedNN.py and cpp wrapper code in augmentedNN.cpp, taking LSTM.py and LSTM.cpp as a reference. 
Hyperparameters, especially number of training epochs, need to be discussed based on real system experiments.
Test cases for the model are also added. The test cases include a uniform distribution dataset and a skewed distribution dataset.

There are two classes defined.
1. class `AugmentedNN` (in augmentedNN.cpp). This class is just like class `TimeSeriesLSTM`.
 - `Fit()`: applies backpropagation.
 - `Predict()`: returns the predictions for the input.
 - `TrainEpoch()`: trains for one epoch.
 - `ValidateEpoch()`: uses one epoch for validation.

2. class `TestingAugmentedNNUtil` (in testing_forecast_util.cpp)
 - `GetData()`: generates data for training and testing. Dataset is uniform or skewed distributed.
 - `Test()`: calls the APIs mentioned above to train and test the model.

Btw, in testing_forecast_util.cpp, the argument of `matrix_eig::bottomRows` was wrong. It should be the number of rows counted from the bottom of the `matrix_eig`. I've modified it. Please check if I am right.


test/brain/testing_forecast_util.cpp
```diff
@@ -110,5 +114,292 @@ matrix_eig TestingForecastUtil::GetWorkload(WorkloadType w, size_t num_samples,
...
+  matrix_eig data = all_data.topRows(all_data.rows() - num_tests*3);
+
+  brain::Normalizer n(normalize);
+  val_interval = std::min<size_t>(val_interval, model.GetEpochs());
+
+  // Determine the split point
+  size_t split_point =
+      data.rows() - static_cast<size_t>(data.rows() * val_split);
+
+  // Split into train/test data
```
malin1993ml: This is splitting into train/**validate** data, and the test data is separate, right?

test/brain/model_test.cpp
```diff
@@ -1,139 +1,184 @@
...
+  float VAL_THESH = 0.05;
+
+  TestingAugmentedNNUtil::Test(*model, DistributionType::UniformDistribution,
+                               LOG_INTERVAL, NUM_SAMPLES,
+                               VAL_SPLIT, NORMALIZE, VAL_THESH,
+                               brain::CommonWorkloadDefaults::ESTOP_PATIENCE,
+                               brain::CommonWorkloadDefaults::ESTOP_DELTA);
+}
+
+TEST_F(ModelTests, DISABLED_AugmentedNNSkewedTest) {
```
malin1993ml: I know we probably cannot enable them in the CI, but have you tested these tests locally?

src/include/brain/workload/augmentedNN.h
```diff
@@ -0,0 +1,108 @@
...
+using TfFloatIn = TfSessionEntityInput<float>;
+using TfFloatOut = TfSessionEntityOutput<float>;
+
+/**
+ * AugmentedNN is used to predict selectivity or cardinality
+ * for range predicates, e.g, 
+ *   SELECT * FROM table WHERE c1 >= l1 AND c1 <= u1
+ *                         AND c2 >= l2 AND c2 <= u2
+ *                         AND ... 
+ * Input is [l1, u1, l2, u2, ...]
```
malin1993ml: From the tests, it looks like right now we're doing only one pair of predicates, just `[l1, u1]`, right?

test/brain/testing_forecast_util.cpp
yetiancn: Right. I'll update the comment here.

test/brain/model_test.cpp
yetiancn: Yes, I've tested them.

src/include/brain/workload/augmentedNN.h
yetiancn: Right.

Issue #1467, opened 2018-07-25
Include optimizer estimates in resulting optimizer plan


GustavoAngulo:
We want to add the following statistics so @pmenon can use them at execution time:
* Operator cardinality estimation
* Selectivity for each predicate

The code should already be written for computing these, we just need to add them to the different objects.

PR #1475, opened 2018-08-31
Add cardinality estimate to AbstractPlan object

GustavoAngulo:
This PR adds the optimizer's cardinality estimate to the AbstractPlan object for use during plan execution. Includes test cases to verify it works. This is part of #1467 to add more optimizer estimates to plans.

Additionally, this PR adds an optimizer_test_util class with useful helper functions for creating test tables and generating plans. Future and possibly existing test cases should inherit from this class.


test/optimizer/cardinality_test.cpp
```diff
@@ -0,0 +1,45 @@
...
+  EXPECT_EQ(num_rows, plan->GetCardinality());
+}
+
+TEST_F(CardinalityTest, EstimatedCardinalityTestWithPredicate) {
+
+  const std::string test_table_name = "testtable";
+  const int num_rows = 10;
+  OptimizerTestUtil::CreateTable(test_table_name, num_rows);
+
+  auto plan = GeneratePlan("SELECT * from " + test_table_name + " WHERE " + test_table_name + "." + "a < 10;");
```
malin1993ml: Could you make it be something that is not selecting all the tuples?

src/optimizer/stats_calculator.cpp
```diff
@@ -64,18 +64,19 @@ void StatsCalculator::Visit(const LogicalGet *op) {
         AddBaseTableStats(col, table_stats, predicate_stats, false);
       }
     }
-    // Use predicates to update the stats accordingly
-    UpdateStatsForFilter(
-        table_stats->GetColumnCount() == 0 ? 0 : table_stats->num_rows,
-        predicate_stats, op->predicates);
+    // Use predicates to estimate cardinality
+    if (table_stats->GetColumnCount() == 0) {
+      root_group->SetNumRows(0);
```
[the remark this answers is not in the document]
GustavoAngulo: This would happen if [`ColumnStatsCatalog::GetTableStats`](https://github.com/cmu-db/peloton/blob/e738acb956306d6914a5a94e6773f410a3eae4a1/src/include/catalog/column_stats_catalog.h#L86-L90) returned no column stats. I believe this would happen if the catalog did not have stats for the table. This is assuming its not possible to have a table without columns.

test/optimizer/cardinality_test.cpp
GustavoAngulo: Done.

Issue #1478, opened 2018-09-19
packages.sh needs to be updated to change LLVM version in Travis


yetiancn:
Error log from Travis:
```
$ PATH=/usr/lib/llvm-3.7/bin:/usr/bin:$PATH cmake -DCOVERALLS=$COVERALLS -DCMAKE_PREFIX_PATH=`llvm-config-3.7 --prefix` -DCMAKE_BUILD_TYPE=$PELOTON_BUILD_TYPE -DUSE_SANITIZER=$SANITIZER ..
/Users/travis/.travis/job_stages: line 98: llvm-config-3.7: command not found
```
It seems that LLVM 3.7 was deprecated. We need to update `packages.sh` to fix this problem.

PR #1482, opened 2018-09-26
LLVM 3.9 for macOS

mbutrovich:
Fix for #1478, hopefully.


mbutrovich:
I dunno why Jenkins isn't running. I'll look again in the morning, but I don't think my changes would have any effect on Jenkins triggering.

crd477:
I noticed this in the other project as well.  Does it matter to Homebrew that the macOS builder VM is still using macOS 10.13 and XCode 9.1?

crd477:
Disregard my other stupid comment.  I believe I've fixed this in the VM image.  A permissions problem for the `jenkins` user when invoking brew prevented llvm from being installed

crd477:
For the record, it worked for me in the other project and I expect it will work here too.  I'll submit a PR in the other one shortly.  One other change I made was to reduce the number of executors on the macOS build VM from 4 to 1.  This will prevent those failures where two build are trying to run `brew` simultaneously.

PR #1490, opened 2018-11-30
 Fix CSV scanner for handling filter predicate on a VARCHAR column correctly

RonyMin:
CSV scan does not care about delimiter in a CSV file, but query
processor should care the end of each column value, i.e., ‘\0’.

Because query parser parses each selection predicate by adding ‘\0’ at
the end of the predicate value, query processor cannot handle the query
correctly if we ignore the existence of delimiter in a raw tuple in CSV
file.

This simple solution fix the case by replacing delimiter to ‘\0’, and
thus StringCompare in type/type_util.h works correctly.


apavlo:
Thanks for sending this. We're not actively developing Peloton anymore. We'll eventually port this over to the new system.

RonyMin:
@apavlo Hello, professor. It is my pleasure if this PR can improve the new system in any way.

PR #1494, opened 2019-04-09
Syntax-based query rewriter

newtoncx:
The code here uses templates across many optimizer files to leverage the old query rewriter and allow it to operate on expression trees. A very simple rewriting task is passing. Use of abstract interfaces may provide a cleaner way to generalize the rewriter, and the code is currently in progress in a separate branch of development.


src/include/optimizer/rule_rewrite.h
```diff
@@ -0,0 +1,45 @@
...
+namespace optimizer {
+
+/* Rules are applied from high to low priority */
+enum class RulePriority : int {
+  HIGH = 3,
+  MEDIUM = 2,
+  LOW = 1
+};
+
+class ComparatorElimination: public Rule<AbsExpr_Container,ExpressionType,AbsExpr_Expression> {
```
tanujnay112, line 30: It would be nice if you added some documentation to this, even if it's simplistic.

src/include/optimizer/memo.h
```diff
@@ -68,10 +73,10 @@ class Memo {
   //===--------------------------------------------------------------------===//
   // For rewrite phase: remove and add expression directly for the set
   //===--------------------------------------------------------------------===//
-  void RemoveParExpressionForRewirte(GroupExpression* gexpr) {
+  void RemoveParExpressionForRewirte(GroupExpression<Node,OperatorType,OperatorExpr>* gexpr) {
```
tanujnay112, line 76: Typo?

src/include/optimizer/absexpr_expression.h
```diff
@@ -0,0 +1,183 @@
...
+      case ExpressionType::COMPARE_EQUAL:
+      case ExpressionType::COMPARE_NOTEQUAL:
+      case ExpressionType::COMPARE_LESSTHAN:
+      case ExpressionType::COMPARE_GREATERTHAN:
+      case ExpressionType::COMPARE_LESSTHANOREQUALTO:
+      case ExpressionType::COMPARE_GREATERTHANOREQUALTO:
+      case ExpressionType::COMPARE_LIKE:
+      case ExpressionType::COMPARE_NOTLIKE:
+      case ExpressionType::COMPARE_IN:
+      case ExpressionType::COMPARE_DISTINCT_FROM: {
```
tanujnay112: I think fall-through from the above cases might mess with you here

src/include/optimizer/rewriter.h
```diff
@@ -0,0 +1,51 @@
...
+namespace peloton {
+namespace optimizer {
+
+class Rewriter {
+
+ public:
+  Rewriter(const Rewriter &) = delete;
+  Rewriter &operator=(const Rewriter &) = delete;
+  Rewriter(Rewriter &&) = delete;
+  Rewriter &operator=(Rewriter &&) = delete;
```
tanujnay112, line 31: Not sure if this is necessary as this is default behavior I think

src/optimizer/memo.cpp
```diff
@@ -14,34 +14,71 @@
 #include "optimizer/memo.h"
```
tanujnay112, line 14: It would be nice if you added more documentation to memo.h to give a better high-level idea of what these Memo objects are used for.

test/optimizer/rewriter_test.cpp
```diff
@@ -0,0 +1,206 @@
...
+  EXPECT_TRUE(casted != nullptr);
+  EXPECT_TRUE(type::ValuePeeker::PeekBoolean(casted->GetValue()) == true);
+
+  delete rewrote;
+}
+
+TEST_F(RewriterTests, SimpleEqualityTree) {
+  //                      [=]
+  //                  [=]     [=]
+  //                [4] [5] [3] [3]
```
lmwnshn, line 124: Maybe add a line to show what you expect it to be rewritten to, i.e. // false.

src/include/optimizer/absexpr_expression.h
```diff
@@ -0,0 +1,183 @@
...
+// correct interface that the rest of the system depends on.
+//
+// As an extreme level of simplification (sort of hacky), an AbsExpr_Container is
+// analogous to Operator and wraps a single AbstractExpression node. AbsExpr_Expression
+// is analogous to OperatorExpression.
+//
+// AbsExpr_Container does *not* handle memory correctly w.r.t internal instantiations
+// from Rule transformation. This is since Peloton itself mixes unique_ptrs and
+// hands out raw pointers which makes adding a shared_ptr here extremely problematic.
+// terrier uses only shared_ptr when dealing with AbstractExpression trees.
```
lmwnshn, line 40: (All Terrier parser behavior can be changed, just FYI. If anything would make it more convenient for you, make the case for it.)

src/include/optimizer/optimizer.h
```diff
@@ -85,16 +88,18 @@ class Optimizer : public AbstractOptimizer {
 
   void Reset() override;
 
-  OptimizerMetadata &GetMetadata() { return metadata_; }
+  OptimizerMetadata<Operator,OpType,OperatorExpression> &GetMetadata() { return metadata_; }
 
   /* For test purposes only */
```
lmwnshn, line 93: Kind of inherited from bad decisions before, but in terrier we try pretty hard to not have public test-only functions and use FRIEND_TEST instead.

src/include/optimizer/rewriter.h
lmwnshn: Consider DISALLOW_COPY_AND_MOVE macro.

src/include/optimizer/rule_rewrite.h
```diff
@@ -0,0 +1,45 @@
...
+#include "optimizer/rule.h"
+#include "optimizer/absexpr_expression.h"
+
+#include <memory>
+
+namespace peloton {
+namespace optimizer {
+
+/* Rules are applied from high to low priority */
+enum class RulePriority : int {
```
lmwnshn, line 24: (I know nothing about optimizers) Is priority a well-defined concept for optimization or is this a rough heuristic? What happens if you have two rules of the same priority that can both be applied, is it always arbitrary which should go first?

src/include/optimizer/absexpr_expression.h
```diff
@@ -0,0 +1,183 @@
...
+      // Probably not worth to create a "validator" since porting
+      // this to terrier anyways (?). == does not check Value
+      // so it's broken. ExactlyEqual requires precondition checking.
+      return false;
+    } else if (!IsDefined() && !r.IsDefined()) {
+      return true;
+    }
+    return false;
+  }
+
```
KatiaVi, line 112: This may not be a huge deal since, as you said, the == op doesn't affect correctness but do you intend to implement the == op for rewrites? If so, does there exist an == op for AbstractExpression you can use?

src/include/optimizer/absexpr_expression.h
```diff
@@ -0,0 +1,183 @@
...
+        return new expression::ConstantValueExpression(cve->GetValue());
+      }
+      default: {
+        int type = static_cast<int>(GetType());
+        LOG_ERROR("Unimplemented Rebuild() for %d found", type);
+        return nullptr;
+      }
+    }
+  }
+
```
KatiaVi, line 151: What does Rebuild do in context of its AbsExpr_Container? You could probably add some more documentation for this function.

src/include/optimizer/absexpr_expression.h
lmwnshn: I believe Terrier currently only has the notion of logical equality for abstract expressions.

src/include/optimizer/rule_rewrite.h
```diff
@@ -0,0 +1,45 @@
...
+// Copyright (c) 2015-16, Carnegie Mellon University Database Group
+//
+//===----------------------------------------------------------------------===//
+
+#pragma once
+
+#include "optimizer/rule.h"
+#include "optimizer/absexpr_expression.h"
+
+#include <memory>
```
tanujnay112, line 18: nitpick: clang-tidy will complain about the ordering of the imports so be aware of that in the future. I believe native libraries such as <memory> should go first here.

src/include/optimizer/absexpr_expression.h
```diff
@@ -0,0 +1,183 @@
...
+#pragma once
+
+// AbstractExpression Definition
+#include "expression/abstract_expression.h"
+#include "expression/conjunction_expression.h"
+#include "expression/comparison_expression.h"
+#include "expression/constant_value_expression.h"
+
+#include <memory>
```
tanujnay112, line 19: Same nitpick about ordering #include's.

src/optimizer/rewriter.cpp
```diff
@@ -0,0 +1,150 @@
...
+using OptimizerTaskStackTemplate = OptimizerTaskStack<AbsExpr_Container,ExpressionType,AbsExpr_Expression>;
+
+using TopDownRewriteTemplate = TopDownRewrite<AbsExpr_Container,ExpressionType,AbsExpr_Expression>;
+
+using BottomUpRewriteTemplate = BottomUpRewrite<AbsExpr_Container,ExpressionType,AbsExpr_Expression>;
+
+using GroupExpressionTemplate = GroupExpression<AbsExpr_Container,ExpressionType,AbsExpr_Expression>;
+
+using GroupTemplate = Group<AbsExpr_Container,ExpressionType,AbsExpr_Expression>;
+
```
KatiaVi, line 54: Would be nice to see more documentation of the functions in Rewriter as well explanations for the templates you use. Maybe a high level explanation of what the rewriter does at the top of the rewriter.h file would be helpful?

src/include/optimizer/optimize_context.h
```diff
@@ -22,18 +22,20 @@
...
-  OptimizeContext(OptimizerMetadata *metadata,
+  OptimizeContext(OptimizerMetadata<Operator,OperatorType,OperatorExpr> *metadata,
                   std::shared_ptr<PropertySet> required_prop,
                   double cost_upper_bound = std::numeric_limits<double>::max())
       : metadata(metadata),
         required_prop(required_prop),
         cost_upper_bound(cost_upper_bound) {}
 
-  OptimizerMetadata *metadata;
+  OptimizerMetadata<Operator,OperatorType,OperatorExpr> *metadata;
```
amlatyrngom, line 38: Just for style: the rest of the code names class members as metadata_ or required_prop_.
In addition, is there a particular reason why these members are public?

src/include/optimizer/optimizer_task_pool.h
```diff
@@ -24,32 +24,35 @@ namespace optimizer {
...
+  virtual std::unique_ptr<OptimizerTask<Node, OperatorType, OperatorExpr>> Pop() = 0;
+  virtual void Push(OptimizerTask<Node,OperatorType,OperatorExpr> *task) = 0;
   virtual bool Empty() = 0;
 };
 
 /**
  * @brief Stack implementation of the task pool
  */
-class OptimizerTaskStack : public OptimizerTaskPool {
+template <class Node, class OperatorType, class OperatorExpr>
```
amlatyrngom, line 39: Consider using the **final** keyword if this class is not be inherited.

test/optimizer/rewriter_test.cpp
amlatyrngom: Also add a comment similar to this one to the other tests so we can easily know what they are testing.

amlatyrngom (commented):
This may just be due to Peloton, but the code in some parts is fairly cryptic, so it would help to have more documentation to understand what it's doing. Also, I think terrier may reject PRs that do not contain enough documentation for functions.

